PR #2, opened 2021-09-15
Fix Vampyre pretending to have MP and be able to carry familiars

soolar:
I know the transition isn't actually complete yet, so feel free to wait to review this PR until it's finished. Just noticed an issue with Vampyre support that was inconveniencing me so I figured I'd fix it.

Also not sure if this is the most correct place to apply the fix. More than happy to listen to feedback on a better approach.

The branch name is no longer fully accurate because I was also annoyed by maximizer telling me to bjornify things as a Vampyre when you can't actually do that. Need to check other paths as listed in the commit message, unless others can confirm for me.

Worth noting this does not cover the case of Zombiecore, where you cannot bjornify/enthrone non-zombie familiars, since that case isn't handled by `canEquip()` presently (which probably warrants its own PR? I'd do it if I had any plans to run Zombiecore any time soon so I could test.)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -1921,6 +1921,24 @@ public static final long getBaseMaxHP()
...
+	public static final boolean usesMP()
+	{
+		if ( KoLCharacter.isVampyre() )
+		{
+			// Vampyre don't have MP, despite api.php saying otherwise
+			return false;
+		}
+
+		// Plumber notably not included because they actually can use MP with bird calendar
+
```
jaadams5, line 1938: Can Plumber use bird calendar?  Conceptually I might argue against a new method that is the same as isVampyre.  I'd be inclined to postpone usesMP until we had more cases it checked.  Ot you could also check isPlumber and has_bird_calendar here as well.  (Not real function names).

soolar: I feel like it's better practice to have `usesMP` now instead of having to search through every reference to `isVampyre` later and figure out which ones are about the fact that it doesn't have MP and which ones aren't once we have another case. Since that is, I imagine, only a matter of time.

Also plumber might be able to use MP with something else in the future too. Heck, it might be able to use it with Source Terminal now that that path isn't standard restricted, so maybe I should update that comment to remove the bird calendar reference.

jaadams5: From a maintainability point of view, I have some concerns/questions.  If this is only used in one place then it is a waste of time and introduces a potential bug when someone is doing something with Vampyre and using text searches to assist that.  Since this is basically an alias for isVampyre it will get overlooked unless/until someone realizes they need to look at usesMP as well.

I am all in favor of encapsulating something so that some future KoL change can be accommodated by one edit.  But this patch isn't there yet.

I might consider a more descriptive name.  Can a character in this path with these items use MP?  usesMP suggests that but canUseMP might be better as long as it is understood that the current MP is not a factor (because you can't use what you don't have).  If I had to pick something I might try pathCanUseMP and use a code comment to note that the availability of some items also effects the answer.  Having done that I would look at the entire code base and use this instead of a path comparison when the concern is MP.  If, and I haven't looked, there are other places then it would be appropriate to change those places now in the name of future proofing.  If there are no other such places then it is a judgement call whether my concerns about maintainability trump your hopes of futureproofing.

PR #3, opened 2021-09-16
Fix "always be equipping" maximizer bug.

heeheehee-kolmafia:
In particular, Maximizer would always prefer equipping something that
had no effect, over leaving nothing equipped.

This turned out to be two bugs:

1. We expected new CheckedItem( 0, ... ) to equal EquipmentRequest.UNEQUIP,
   which is populated by ItemPool.get( "(none)", 1 ). This function creates an
   object with item id -1 if no matches are found. Thus, we never considered
   equipping nothing as "simpler" than equipping something.

2. We use InventoryManager.getCount() and getAccessibleCount() to determine how
   many of the object we have on hand. This hurts us twice:

   a. First, we're fail the tiebreaker in MaximizerSpeculation.compareTo() -- we
      don't have any copies of (none) on-hand, so we prefer anything else.

   b. Second, Evaluator has some additional logic to make sure we don't try to
      equip more copies than we can acquire. Since we claim we can acquire 0
      copies...


heeheehee-kolmafia:
(Note that as it's still prior to the switchover date, I'll submit this via SVN if I get a thumbs up from another maintainer.)

The test failure seems like a gradle configuration problem.

src/net/sourceforge/kolmafia/maximizer/CheckedItem.java
```diff
@@ -62,16 +62,18 @@ public CheckedItem( int itemId, int equipScope, int maxPrice, int priceLevel )
 	{
 		super( itemId, 1, false );
 
+		this.inventory = InventoryManager.getCount( itemId );
+
+		this.initial = InventoryManager.getAccessibleCount( itemId );
+
 		// special case used to get a CheckItem that .equals( EquipmentRequest.UNEQUIP ).
```
jaadams5, line 69: It feels weird using itemId before checking it, but I recognize the idiom so no change needed.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -2039,7 +2039,7 @@ else if ( spec.compareTo( secondBest ) > 0 )
 			// If we currently have nothing equipped, also consider leaving nothing equipped
 			if ( EquipmentManager.getEquipment( Evaluator.toUseSlot( slot ) ) == EquipmentRequest.UNEQUIP )
 			{
-				ranked[ slot ].add( new CheckedItem( 0, equipScope, maxPrice, priceLevel ) );
+				ranked[ slot ].add( new CheckedItem( -1, equipScope, maxPrice, priceLevel ) );
```
jaadams5, line 2042: Since I am not viewing this in a rich environment, is this the only place CheckedItem is called?  If it isn't then it might be safer to replace the check in the above with something that captured all invalid itemIds and mapped them to NONE.  I don't know how the Mazimizer might get presented with a pseudo Item or Item 13 but if it can be ruled out....

test/net/sourceforge/kolmafia/MaximizerTest.java
```diff
@@ -120,7 +120,6 @@ public void freshCharacterShouldNotRecommendEverythingWithCurrentScore()
 
 	// Sample test for https://kolmafia.us/showthread.php?23648&p=151903#post151903.
 	@Test
-	@Ignore("Currently failing")
 	public void noTieCanLeaveSlotsEmpty()
```
jaadams5, line 123: Because I am still navigating the review process and tools, it is not clear to me whether this test passes with the changes or not.  If it is untested because of Gradle then perhaps ant could be used to run this test against the fixes?  I am not sure that "No tie can leave slots empty" is a correct description of the desired behavior.  Would it be reasonable to want -tie to leave a slot empty if it is currently empty?  If the Surprising Fist skills are perm'd sometimes it makes sense to use them instead of a weapon.

Rinn: I've always felt like -tie should leave slots empty.

heeheehee-kolmafia: This particular test now passes, as you can see by the checks below.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
heeheehee-kolmafia: There are a few other constructor invocations in Evaluator.

I missed another instance where itemId was specified as 0. I believe all other instances are receiving itemIds from actual items in inventory.

I wouldn't typically object to making this a catch-all for "hey this isn't a real item", although it may lead to unnecessary checks which may lead to Maximizer slowdowns, depending on how we do this.

The robust approach would be to check ItemDatabase to confirm that the itemId doesn't map to a real item. Each new CheckedItem would spawn a HashMap lookup.

The quick-and-dirty approach would be to check that itemId <= 0.

heeheehee-kolmafia:
Erm, hm. I apparently totally broke this history when I ran a git pull --rebase.

gausie:
> Erm, hm. I apparently totally broke this history when I ran a git pull --rebase.

Yeah that makes sense. Looking forward to being git native!

test/net/sourceforge/kolmafia/MaximizerTest.java
heeheehee-kolmafia: To better explain the test case naming: it should be parsed as:

"no tie" (i.e. -tie) can leave slots empty

A better name might be "noTiebreakerCan...", since it's implicit that because there is no tiebreaker expression, there is a tie, which should be broken in favor of making no changes.

The intended preference of -tie isn't actually to leave slots empty, but rather leave them unchanged. But in this case, empty == unchanged.

PR #4, opened 2021-09-16
Fix "always be equipping" maximizer bug, try #2

heeheehee-kolmafia:
Hopefully this time I won't ruin the PR. :/

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -2657,10 +2657,10 @@ public class Evaluator
 				}
 			}
 
-			// Blunt object fix for only having a foldable that might be neede elsewhere
+			// Blunt object fix for only having a foldable that might be needed elsewhere
 			if ( automatic[ slot ].size() == 1 && ItemDatabase.getFoldGroup( automatic[ slot ].get( 0 ).getName() ) != null )
 			{
-				automatic[ slot ].add( new CheckedItem( 0, equipScope, maxPrice, priceLevel ) );
+				automatic[ slot ].add( new CheckedItem( -1, equipScope, maxPrice, priceLevel ) );
 			}
 
 			if ( this.dump > 0 )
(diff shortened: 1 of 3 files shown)


heeheehee-kolmafia:
Note past history at https://github.com/kolmafia/kolmafia/pull/3.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -2657,10 +2657,10 @@ else if ( total < useful )
 				}
 			}
 
-			// Blunt object fix for only having a foldable that might be neede elsewhere
+			// Blunt object fix for only having a foldable that might be needed elsewhere
 			if ( automatic[ slot ].size() == 1 && ItemDatabase.getFoldGroup( automatic[ slot ].get( 0 ).getName() ) != null )
 			{
-				automatic[ slot ].add( new CheckedItem( 0, equipScope, maxPrice, priceLevel ) );
+				automatic[ slot ].add( new CheckedItem( -1, equipScope, maxPrice, priceLevel ) );
```
jaadams5, line 2663: I only found two instances of calling CheckedItem with an integer constant instead of a variable and it looks like you got them both.

The part of me that doesn't like Magic Numbers  will go and suppress itself :-)

jaadams5 (approved):
I'm not certain that the test is the right one but it works so we can deal with the other question elsewhere.

jaadams5 (approved):

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
heeheehee-kolmafia: I don't like magic numbers either! Sentinel values are evil, and all that.

heeheehee-kolmafia:
> I'm not certain that the test is the right one but it works so we can deal with the other question elsewhere.

Regarding this: the context is from the link right above the test: https://kolmafia.us/showthread.php?23648&p=151903#post151903

> I definitely think we can quash, eventually, the not preferring an unhelpful item over no item. If you have no item, and it doesn’t beat a tie with an item, then stick with no item. That is a bug I’ve tried to track down several times and over many hours.

Anyways, I'll close this PR now, then commit this change to SVN once my precommit hooks finish running. I think this was a useful exercise.

PR #6, opened 2021-09-17
Set revision when building jar with gradle

Rinn:
- Stop copying code to generated-src and writing revision into KolConstants.java
- Embed build info into jar when building with gradle
- Use the embedded build info to set the revision in mafia
- Override revision for tests since it is now unavailable without a jar
- Remove RELEASE code, comment out version update check until the migration is finished, then the code will be rewritten to check github for new releases
- Remove runtime check against svn to determine revision for locally edited mafia, as it is now always embedded in the jar when building with gradle.
- Add build info instead release date to log & about window
- Enable github releases


build.gradle
```diff
@@ -218,4 +190,13 @@ clover {
   }
 }
 
-compileJava.dependsOn setVersion
+jar.dependsOn gitRevList
```
heeheehee-kolmafia, line 193: Any reason why this shouldn't be compileJava?

Someone using ./gradlew run will no longer have revision set appropriately.

build.gradle
```diff
@@ -114,7 +120,8 @@ jar {
   }
   duplicatesStrategy = 'exclude'
   destinationDirectory = file('dist/')
-  archiveBaseName = ant.properties['version-name']
+  archiveBaseName = "KoLmafia"
+  classifier = "${versioning.info.dirty ? 'Modified' : ''}"
```
heeheehee-kolmafia, line 124: classifier is deprecated in favor of archiveClassifier.

How will the resulting file look? KoLmafia-r20999-Modified.jar?

(preference nit: make this lowercase?)

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -2149,11 +2149,7 @@ public static boolean isPlayerOnline( final String player )
 	{
 		public void run()
 		{
-			if ( KoLConstants.VERSION_NAME.startsWith( "KoLmafia r" ) )
-			{
-				return;
-			}
-
+			/* TODO: Check for new version on jenkins\github after migrated
```
heeheehee-kolmafia, line 2152: I personally view commenting out version-controlled code as an anti-pattern, since if you really need it, you can always grab it from a past commit.

Do you intend to merge this change in the next couple of days? If not, maybe just add the new implementation? If so, it's okay to just leave a TODO.

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -86,43 +83,48 @@ protected MafiaState initialValue()
...
+		// Get the revision from the jar manifest attributes
+		if ( KoLConstants.REVISION == null )
+		{
+			try
 			{
-				return StaticEntity.cachedSVNRevisionNumber;
+				ClassLoader classLoader = StaticEntity.class.getClassLoader();
+				if (classLoader != null)
+				{
+					Enumeration<URL> resources = classLoader.getResources("META-INF/MANIFEST.MF");
```
heeheehee-kolmafia, line 110: Formatting nit: missing spaces in method call.

(ditto below with manifest.getMainAttributes().getValue("Build-Revision"))

src/net/sourceforge/kolmafia/KoLConstants.java
```diff
@@ -73,7 +73,6 @@
 
 	String VERSION_NAME = "KoLmafia v20.7";
```
heeheehee-kolmafia, line 74: If we want to enforce that all accesses are done through StaticEntity.java's getVersion(), getRevision(), VERSION_NAME / REVISION (at the very least) should be private fields that live in StaticEntity.

heeheehee-kolmafia (commented):
I hope @gausie doesn't mind, but I took a pass at this change, since there hadn't been any movement for a week.

Holding off on approval pending changes to version-checking logic in KoLmafia.java, but otherwise I have no major concerns (just a bunch of nits).

build.gradle
Rinn: The revision is no longer embedded in the java code with this change, so adding it as a dependency there isn't necessary. It's only bundled into the jar now. But yeah it means running the code without a jar means the revision won't be available.

build.gradle
Rinn: I left it `KoLmafia` since that matches the current filename case from the ant build (locally modified for ant is KoLmafia-r###M.jar)

src/net/sourceforge/kolmafia/KoLmafia.java
Rinn: Yeah the intent is for this to get merged on the switch to github, then once git releases are working fine I'll rewrite this to find the latest jar from git or jenkins, but since we don't support auto-updating the "daily" builds at the moment I was going to do it in a separate commit and put the checking begind a global var. I'll rip the commented code out and leave the TODO.

build.gradle
heeheehee-kolmafia: Oh, I meant specifically to make "Modified" lowercase. But I'm fine with it either way.

BadHorseMonkey:
Why did we change from "M" as the modified build version number to "Modified"?  I don't think it makes any difference, but don't we strip the last character off the version number in a few places if it's "M" right now?  If it's just in StaticEntity, I see you took care of, but if anyone has scripts (ash or shell) that rely on it, we're changing it.

Not sure what the advantage of the longer name change is.

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -87,72 +90,115 @@ protected MafiaState initialValue()
...
+		catch ( IOException e )
 		{
 			// fall through
 		}
-		if ( KoLConstants.REVISION == null )
+
+		return 0;
+	}
+
+	public static final String getbuildInfo()
```
heeheehee-kolmafia, line 144: nit: getBuildInfo

Rinn:
All these are valid points, but I'm not going to be able to do anything else on this today.

heeheehee-kolmafia:
> All these are valid points, but I'm not going to be able to do anything else on this today.

(I don't expect you to! Take your time!)

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -87,72 +90,115 @@ protected MafiaState initialValue()
...
-				return StaticEntity.cachedSVNRevisionNumber;
-			}
-			if ( KoLConstants.REVISION == null && SVNWCUtil.isWorkingCopyRoot( KoLConstants.ROOT_LOCATION ) )
+			ClassLoader classLoader = StaticEntity.class.getClassLoader();
+			if ( classLoader != null )
 			{
-				SVNInfo info = SVNManager.doInfo( KoLConstants.ROOT_LOCATION );
-				StaticEntity.cachedSVNRevisionNumber = (int) info.getRevision().getNumber();
-				return StaticEntity.cachedSVNRevisionNumber;
+				Enumeration<URL> resources = classLoader.getResources( "META-INF/MANIFEST.MF" );
```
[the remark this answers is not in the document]
gausie: I'm changing this to `getResource` so this isn't applicable anymore. If we need to revisit I think you need to do `getResources(...).asIterator()`

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -87,72 +90,115 @@ protected MafiaState initialValue()
...
+				while ( resources.hasMoreElements() ) 
+				{
+					Manifest manifest = new Manifest( resources.nextElement().openStream() );
+					String buildRevision = manifest.getMainAttributes().getValue( "Build-Revision" );
+					if ( buildRevision != null && StringUtilities.isNumeric( buildRevision ) )
+					{
+						StaticEntity.cachedRevisionNumber = StringUtilities.parseInt( buildRevision );
```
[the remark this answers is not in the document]
gausie: The current impl would just return 0 in those instances anyway though right? We can fix this later, not going to fix in #17

PR #7, opened 2021-09-20
Language Server - parseAggregateLiteral arrays

fredg1:
(from https://kolmafia.us/threads/ash-language-server-features.26098/post-163764 )
Alters parseAggregateLiteral a little so that we try to read key-value pairs (components of a map literal) even once we identified that the user is supplying an array literal (which doesn't expect keys).


fredg1:
(sorry for the ~8 first commit. VScode didn't let me base the branch off of upstream (which it doesn't even seem to acknowledge the existence of...), so I had to base off of my "main". The first three commits zero-out each others, and the rest is just pulling from upstream. Squash-and-merge will leave no trace of those.)

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -1359,13 +1366,16 @@ else if ( !delim.equals( "}" ) )
...
 
 			// We are parsing a MapLiteral
-			if ( !delim.equals( ":" ) )
+			this.readToken(); // read :
+
+			if ( isArray )
 			{
-				throw this.parseException( ":", this.currentToken() );
+				// Not entirely correct, since, to get here, what we got so far must have matched
+				// the value's datatype, but we can't tell what they put after the :, so just
```
heeheehee-kolmafia, line 1374: > we can't tell what they put after the :

Why not? You've already read the colon, so why couldn't you use this.currentToken()?

tangent: This function would likely be cleaner / easier to understand if we split it into parseMapLiteral() and parseArrayLiteral(), which may be simpler to accomplish with the rewindTo() functionality you previously added.

heeheehee-kolmafia (commented):
Overall the diff looks reasonable, although it reads a bit awkwardly (I think that's a consequence of handling array literal and map literal parsing in the same function).

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: I think you may be misunderstanding that comment.

It doesn't claim that the exception it is replacing (the line that git shows was previously here) is incorrect
i.e.
`throw this.parseException( ":", this.currentToken() );`

What it refers to is the accuracy of the error message that follows (two lines below).
i.e.
`throw this.parseException( "Cannot include keys when making an array literal" );`


e.g.
supposing this array literal:
`string [int] {"a", "b", "c", "d":  <not yet read/known>`
Once we read the colon after the "d", we'll send the error "Cannot include keys when making an array literal".
However, that is not accurate (which is what the comment suggests). We *know* that this isn't a key-value pair, because for it to be a key, it would have needed to be an integer, yet "d" is a string, which matches the expected datatype of a "value" for this aggregate.

heeheehee-kolmafia: I recognize that.

Why is the comment necessary in the first place? We can in fact tell what's after the colon (because we've just read the colon, and the next token can be accessed via currentToken()). Either way, the colon is not valid in this position -- we could throw an unexpected token error, but are instead choosing to provide a more specific error message.

heeheehee-kolmafia: Might be simpler to just always throw "unexpected colon" so you don't have to worry about the sometimes-technically-incorrect error message.

fredg1: > Why is the comment necessary in the first place? We can in fact tell what's after the colon (because we've just read the colon, and the next token can be accessed via currentToken()).

This is a simple example, but what follows could be an expression, which takes more than currentToken() to parse.



> we could throw an unexpected token error, but are instead choosing to provide a more specific error message.

We could, but keep in mind: for someone not too well-versed in programming, they may not really get the difference between a map literal and an array literal, so they could not get why the colon is suddenly not accepted (maybe they just didn't notice that they forgot the colon at the first value/element)

An unexpected token error pretty much always highlights a syntax error. It's a token with which we don't know what to do.
Here, we *do* know what to do with it, but we're in a situation where it's forbidden.


If you still wish me to throw an unexpected colon error at this point, say so, and I'll do it. It was just a "does considering this change your mind?" thing.

heeheehee-kolmafia: I would prefer one of:

a) throw "Unexpected colon, expected comma"

b) check that the key is coercible to an int; only then throw "cannot include keys".

c) throw a more accurate error, like "Cannot mix map literal and array literal syntax" which would be unambiguous and wouldn't need a comment explaining why it's sort of wrong.

fredg1: let's go with a) and b)

heeheehee-kolmafia (approved):
Thanks, this looks good.

(I think technically I won't have permissions to approve changes until after the switchover, but we'll get those ironed out.)

PR #8, opened 2021-09-20
Language Server - parseVariable initializations

fredg1:
( from https://kolmafia.us/threads/ash-language-server-features.26098/post-163766 )
A bit more complex than the last few; this patch modifies parseVariable so that the signal indicating that we're parsing parameters is a boolean, rather than scope being null, and waits until the end to throw the error.
This, again, will allow us to still keep trying to parse a parameter initialization, and will give us a full location once we throw the error.

Also shuffles the parsing logic a little, for efficiency.


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -773,7 +773,7 @@ private Function parseFunction( final Type functionType, final Scope parentScope
 				vararg = true;
 			}
 
-			Variable param = this.parseVariable( paramType, null );
+			Variable param = this.parseVariable( paramType, parentScope, false );
```
heeheehee-kolmafia, line 776: Does this fix a bug (null -> parentScope)?

(If so, can you add a test case exercising that?)

test/net/sourceforge/kolmafia/textui/ParserTest.java
```diff
@@ -1268,6 +1268,18 @@
 				"Cannot initialize parameter x",
```
heeheehee-kolmafia, line 1268: When _are_ we supposed to be able to initialize parameters?

All the tests exercise failures, but not a single success.

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: No. This fixes a *future* bug.

When we will keep parsing even after seeing an error, if that parameter is null, we get a NPE.

test/net/sourceforge/kolmafia/textui/ParserTest.java
fredg1: Never. This is, again, preparation for the "keep parsing until the end of the file, and accumulate errors". We parse the initialization even though we know it's not allowed, so that we can tell when it *ends*, and the next parameter starts.

Additionally, this could lay the groundwork for allowing "default values" to parameters... But that's just a side-effect.

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -897,68 +897,81 @@ else if ( scope != null && scope.findVariable( variableName.content ) != null )
...
-		if ( this.currentToken().equals( "=" ) )
+		if ( postVariableToken.equals( "=" ) )
 		{
 			this.readToken(); // read =
+		}
 
-			if ( this.currentToken().equals( "{" ) )
+		if ( this.currentToken().equals( "{" ) )
+		{
+			// We allow two ways of initializing aggregates:
```
heeheehee-kolmafia, line 916: This comment seems like it would fit better in the "maybe read =" block.

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -874,7 +874,7 @@ private boolean parseVariables( final Type t, final BasicScope parentScope )
 		}
 	}
 
-	private Variable parseVariable( final Type t, final BasicScope scope )
+	private Variable parseVariable( final Type t, final BasicScope scope, final boolean allowInitialization )
 	{
```
heeheehee-kolmafia, line 878: Rereading this PR.

All of this seems unnecessarily complicated, since this function is handling two separate cases, controlled by the boolean parameter. It's made worse now that you have to track this state throughout the function, as opposed to in a single place.

What we should do instead is split this into parseDeclaration and parseDefinition (or repurpose parseAssignment to optionally only allow "="), then move the "cannot initialize parameter" complaint into something that actually deals with function parameters.

(or, if you want different names, parseVariable and parseInitialization.)

heeheehee-kolmafia: In particular, I might suggest as API:

Variable parseVariable(Type t) { ... }
Assignment parseInitialization(Variable v) {
  // parses " = ...;"
}

then add it to the scope in question parseVariables. That way, if you really want to parse initialization of a variable in a parameter declaration, you can do that without impacting the scope.

fredg1: (do you get notified if there are commits after you posted a comment, or do you need to be notified manually?)

heeheehee-kolmafia: (I do, I just haven't had time to look at this.)

test/net/sourceforge/kolmafia/textui/ParserTest.java
```diff
@@ -486,7 +486,7 @@
 			{
 				"Primitive assignment to aggregate",
 				"int[4] x = 1;",
-				"Cannot store int in x of type int [4]",
+				"Cannot store int in a variable of type int [4]",
```
heeheehee-kolmafia, line 489: Do you think it would be more helpful to say "Cannot store int in variable x of type int[4]"?

(Namely, call out _which_ variable is getting a bad assignment)

Maybe it's moot with line number + column info.

(Other than this, the PR looks reasonable.)

fredg1: The isolation of the initialization into a method means the name of the variable is now out of reach...

heeheehee-kolmafia: That seems like a situation where you'd catch the exception at the callsite, then rethrow it with more context.

It's basically the same rationale as extending "Unexpected = at line X" with " in file script.ash". (Not that Parser throws exceptions that look like that today... but just as an example.)

So, we could envision something like...

"Cannot store int in type int[4]" + " for variable x".

(It's not super important for this PR, as it will look weird initially with () representing the lack of a file. But you can fix that with the same pattern by shunting the file annotation into the corresponding parse() call.)

gausie:

1. Copy this branch
4. Hit "Reopen"!

PR #9, opened 2021-09-20
Language Server - parseFunction vararg

fredg1:
( from https://kolmafia.us/threads/ash-language-server-features.26098/post-163765 )
This modifies the logic of parseFunction a bit so that the vararg error checks happen after the variable name was read.
This will allow us to have the full variable + type's location, rather than only the type, once those are implemented.

Additionally, this allows the "Only one vararg parameter is allowed" error message to be reached.


heeheehee-kolmafia (approved):
This seems reasonable, thanks.

gausie:

1. Copy this branch
4. Hit "Reopen"!

PR #10, opened 2021-09-20
Language Server - parseType records

fredg1:
( from https://kolmafia.us/threads/ash-language-server-features.26098/post-163762 )
This changes the logic of parseType to attempt to parse records even when not allowed (again, with the continuous parsing, we need to try anyway).

This is possible because the method parseRecord starts by looking for the token "record", immediately returning null if it doesn't find it.
"record" is a reserved word exclusively used to declare new records, so nothing is hindered by checking for them right away.


heeheehee-kolmafia (approved):
This took me a bit of staring to decide this wasn't changing existing behavior (other than adding a more descriptive error message in one case), but it looks fine as well.

gausie:

1. Copy this branch
4. Hit "Reopen"!

PR #11, opened 2021-09-20
Language Server - parsetree abstracts

fredg1:
( from https://kolmafia.us/threads/ash-language-server-features.26098/post-163993 )
A few appetizers while we wait for the rest:

- The classes CompositeType, CompositeValue and AggregateValue should be made abstract.

- The method CompositeType.getDataType( Object ) can be made abstract.

- Rather than having Type.toString() and RecordType.toString() both be return this.name, just make Symbol.toString() { return this.name; } (this also incidentally adds the method to Variable and Function, which helps in the debugger)


src/net/sourceforge/kolmafia/textui/parsetree/CompositeValue.java
```diff
@@ -45,7 +45,7 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
-public class CompositeValue
+public abstract class CompositeValue
 	extends Value
```
heeheehee-kolmafia, line 49: It seems strange to have an abstract class without any abstract methods (even if it's technically legal). Are there any methods that subclasses always implement (and always need specialized logic for)?

Just skimming for methods that have useless stub implementations, clear / keys / aref / aset / remove / dumpValue?

src/net/sourceforge/kolmafia/textui/parsetree/AggregateLiteral.java
```diff
@@ -33,13 +33,44 @@
...
 {
 	protected AggregateValue aggr = null;
 
-        public AggregateLiteral( final AggregateType type )
+	public AggregateLiteral( final AggregateType type )
 	{
 		super( type );
 	}
+
+	@Override
```
heeheehee-kolmafia, line 48: These stub methods don't seem useful here -- they should presumably remain abstract.

src/net/sourceforge/kolmafia/textui/parsetree/CompositeValue.java
```diff
@@ -63,34 +63,23 @@ public Value aref( final Value key )
...
+	public abstract void aset( final Value key, final Value val, final AshRuntime interpreter );
 
-	public Value remove( final Value key, final AshRuntime interpreter )
-	{
-		return null;
-	}
+	public abstract Value remove( final Value key, final AshRuntime interpreter );
 
 	@Override
 	public void clear()
```
heeheehee-kolmafia, line 78: This feels like it should also be abstract, since it doesn't do anything useful or generic.

src/net/sourceforge/kolmafia/textui/parsetree/RecordType.java
```diff
@@ -90,6 +90,12 @@ public Type getIndexType()
 		return DataTypes.STRING_TYPE;
 	}
 
+	@Override
+	public Type getDataType()
```
heeheehee-kolmafia, line 94: The definition in the abstract class could use a comment describing why it's okay that it's null for a RecordType -- namely that it stores the value type, and records do not necessarily have a single value type for their fields.

(Truthfully, the same can probably be said for all the other abstract methods... but this one in particular stood out to me.)

src/net/sourceforge/kolmafia/textui/parsetree/Symbol.java
```diff
@@ -37,10 +37,11 @@
 	extends ParseTreeNode
 	implements Comparable<Symbol>
 {
-	public String name;
+	public final String name;
 
 	public Symbol()
```
heeheehee-kolmafia, line 42: Does anything even use this constructor? Can we get rid of it?

src/net/sourceforge/kolmafia/textui/parsetree/AggregateLiteral.java
fredg1: They're not stubs. Neither MapLiteral nor ArrayLiteral (both subclasses of AggregateLiteral) implement any of these methods, so they need a "default".

It's either that, or all of this goes in both MapLiteral and ArrayLiteral instead.

fredg1: They're not stubs. Neither MapLiteral nor ArrayLiteral (both subclasses of AggregateLiteral) implement any of these methods, so they need a "default".

It's either that, or all of this goes in both MapLiteral and ArrayLiteral instead.

fredg1 (commented):
damn big green button, I didn't want to click that...

src/net/sourceforge/kolmafia/textui/parsetree/AggregateLiteral.java
heeheehee-kolmafia: I get the aset / remove being implemented here, as an AggregateLiteral should be immutable. But MapLiteral and ArrayLiteral don't need to know how to access a value, or iterate over keys, or check for presence?

I'm probably misunderstanding what these methods are actually meant to signify.

src/net/sourceforge/kolmafia/textui/parsetree/Symbol.java
fredg1: Yeah, IDK why it's even there

src/net/sourceforge/kolmafia/textui/parsetree/AggregateLiteral.java
fredg1: Quickly looking at its two subclasses, it seems like they override a single method: Value.execute().

This method turns them into a *different* AggregateValue (either an ArrayValue or a MapValue), which are *then* acted on.

fredg1: (single, besides Value.count()...)

src/net/sourceforge/kolmafia/textui/parsetree/RecordType.java
fredg1: How's that? https://github.com/kolmafia/kolmafia/pull/11/commits/4c510d0f9e4a0f46130ae564951ff2d19d040998

src/net/sourceforge/kolmafia/textui/parsetree/AggregateLiteral.java
```diff
@@ -62,6 +62,11 @@ public Value remove( final Value key, final AshRuntime interpreter )
 		return null;
 	}
 
+	@Override
+	public void clear()
+	{
+	}
```
heeheehee-kolmafia, line 68: (Some of these invalid accesses might make sense as thrown exceptions, but this is fine either way.)

heeheehee-kolmafia (approved):
Thanks, this looks good now.

gausie:

1. Copy this branch
4. Hit "Reopen"!

PR #17, opened 2021-09-26
Set revision when building jar with gradle

gausie:
Replaces #6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,6 +4,7 @@ plugins {
   id 'application'
   id 'java'
   id 'com.bmuschko.clover' version '3.0.1'
+  id 'net.nemerosa.versioning' version '2.8.2'
 }
 
 ant.importBuild('build.xml') { String oldTargetName ->
@@ -13,7 +14,7 @@ ant.importBuild('build.xml') { String oldTargetName ->
 sourceSets {
   main {
     java {
-      srcDirs = [ 'build/generated-src' ]
+      srcDirs = [ 'src' ]
       destinationDirectory.set(file('build/main'))
     }
     resources {
@@ -105,7 +106,12 @@ test {
 jar {
   manifest {
     attributes(
-      'Main-Class': 'net.sourceforge.kolmafia.KoLmafia'
+      'Main-Class'      : 'net.sourceforge.kolmafia.KoLmafia',
+      'Build-Revision'  : new Object() { String toString() { getVersion() } },
+      'Build-Branch'    : versioning.info.branch,
+      'Build-Commit'    : versioning.info.commit,
+      'Build-Jdk'       : "${System.properties['java.version']} (${System.properties['java.vendor']} ${System.properties['java.vm.version']})",
+      'Build-OS'        : "${System.properties['os.name']} ${System.properties['os.arch']} ${System.properties['os.version']}"
     )
   }
 
@@ -114,7 +120,8 @@ jar {
   }
   duplicatesStrategy = 'exclude'
   destinationDirectory = file('dist/')
-  archiveBaseName = ant.properties['version-name']
+  archiveBaseName = "KoLmafia"
+  classifier = "${versioning.info.dirty ? 'Modified' : ''}"
 }
 
 task download(type: Copy) {
@@ -132,7 +139,9 @@ task gitRevList(type: Exec) {
   standardOutput = new ByteArrayOutputStream()
 
   doLast {
-    new File('build', 'revision.txt').text = standardOutput.toString().trim()
+    new File('build/revision.txt').text = standardOutput.toString().trim()
+    // Update the version to the new revision
+    project.setVersion(standardOutput.toString().trim())
   }
 }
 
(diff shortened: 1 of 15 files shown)


heeheehee-kolmafia:
@gausie Feel free to assign this PR to me if or when you want me to take a look

build.gradle
```diff
@@ -132,7 +139,9 @@ task gitRevList(type: Exec) {
   standardOutput = new ByteArrayOutputStream()
 
   doLast {
-    new File('build', 'revision.txt').text = standardOutput.toString().trim()
+    new File('build/revision.txt').text = standardOutput.toString().trim()
```
heeheehee-kolmafia, line 142: Ostensibly the two-arg constructor we were previously using was ever so slightly more portable, but I don't feel strongly about it.

We also weren't being consistent, which we now are.

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -87,72 +90,104 @@ protected MafiaState initialValue()
...
+			if ( classLoader != null )
 			{
-				SVNInfo info = SVNManager.doInfo( KoLConstants.ROOT_LOCATION );
-				StaticEntity.cachedSVNRevisionNumber = (int) info.getRevision().getNumber();
-				return StaticEntity.cachedSVNRevisionNumber;
+				URL resources = classLoader.getResource( "META-INF/MANIFEST.MF" );
+
+				Manifest manifest = new Manifest( resources.openStream() );
+				String buildRevision = manifest.getMainAttributes().getValue( "Build-Revision" );
+				if ( buildRevision != null && StringUtilities.isNumeric( buildRevision ) )
```
heeheehee-kolmafia, line 118: Optional: As noted in the previous PR, I think wrapping Integer.parseInt in a try/catch would be sufficient here and not be subject to the nonstandard parsing rules present in StringUtilities.isNumeric / parseInt.

heeheehee-kolmafia (approved):
Looks good, modulo some nits that I don't feel strongly about. Thanks!

gausie:
@heeheehee-kolmafia I'm having some issues whereby the revision always comes out at 0. Trying to work it out...

src/net/sourceforge/kolmafia/StaticEntity.java
heeheehee-kolmafia: (And as we both noted, it's not a meaningful change of behavior, in most cases. It just accepts and caches a slightly broader set of inputs.)

build.gradle
```diff
@@ -121,7 +121,7 @@ jar {
   duplicatesStrategy = 'exclude'
   destinationDirectory = file('dist/')
   archiveBaseName = "KoLmafia"
-  classifier = "${versioning.info.dirty ? 'Modified' : ''}"
+  classifier = "${versioning.info.dirty ? 'M' : ''}"
```
heeheehee-kolmafia, line 124: (Note this is subtly different from what we have today with `ant daily`, namely we'll now have "KoLmafia-r12345-M.jar" instead of "KoLmafia-r12345M.jar". I don't think this is a meaningful difference.)

heeheehee-kolmafia:
@gausie What situations are you seeing the revision as 0? I built a fresh jar, which was mostly named appropriately (marked as M even though my branch is up-to-date, although I do have some untracked files in the repo root).

And, both the get_revision() ASH function and the version string ("KoLmafia r25900" or so) looked good.

The ant build on the other hand had revision = 0, which is unsurprising given that it doesn't set the manifest appropriately.

heeheehee-kolmafia (approved):

heeheehee-kolmafia:
Some other misc comments, that I don't think are blocking:

- Build-Commit is the full commit hash. Apparently we could use versioning.info.build for the short hash.
- My Build-Jdk apparently contains "N/A" in place of ${System.properties['java.vendor']. I'm using OpenJDK; java.runtime.name is more informative for me ("OpenJDK Runtime Environment").

build.gradle
gausie: Yeah me and @BadHorseMonkey noticed this and thought that it might be alright given it allows you to split by - and grab [1]

gausie:
> @gausie What situations are you seeing the revision as 0? I built a fresh jar, which was mostly named appropriately (marked as M even though my branch is up-to-date, although I do have some untracked files in the repo root).
> 
> And, both the get_revision() ASH function and the version string ("KoLmafia r25900" or so) looked good.
> 
> The ant build on the other hand had revision = 0, which is unsurprising given that it doesn't set the manifest appropriately.

I had this issue when I built a jar though @BadHorseMonkey  didn't have the issue ejther. Me and @BadHorseMonkey found ourselves unable to change flatlaf themes, can you test that?

heeheehee-kolmafia:
I too can't change flatlaf themes. I don't see them bundled in the jar. Perhaps all the runtimeOnly deps should actually be implementation deps. Simple enough fix.

gausie:
I think I'd like to leave this to @Rinn when he's available!

gausie:
I get the feeling that my revision not being found in the jar could be a Linux compatibility issue?

BadHorseMonkey:
> I get the feeling that my revision not being found in the jar could be a Linux compatibility issue?

Could be, but we should be able to find/fix tha

> I think I'd like to leave this to @Rinn when he's available!

Before we do that, I have a two line fix that's working in my repo.

`plugins {
  id 'application'
  id 'java'
  id 'com.bmuschko.clover' version '3.0.1'
  id 'net.nemerosa.versioning' version '2.8.2'
  **id 'com.github.johnrengelman.shadow' version '6.0.0'**

}
**mainClassName = 'net.sourceforge.kolmafia.KoLmafia'**`

Then refresh your gradle tasks and in the new shadow subfolder, chose shadowJar.

There may be more cleanup, and it uses the default shadowJar directories (build/libs for the jar).

Also, I updated my gradle to the latest code that has the platform-agnostic files, and now I get revision 0 as well...

heeheehee-kolmafia:
By Linux compatibility issue, do you mean because you're using Linux, or that it only works on Linux? Because it works for me, on Linux.

I've been building and running off of this branch's head with no issues.

Can you unzip the resulting jar and see what's actually being written into your META-INF/MANIFEST.MF?

Do you need to set mainClassName outside of application { mainClass }? Hm.

heeheehee-kolmafia:
(Also, with changing runtimeOnly -> implementation, I was able to change flatlaf themes. Jar got a bit bigger, but that's expected given that we were omitting the flatlaf files previously...)

BadHorseMonkey:
> By Linux compatibility issue, do you mean because you're using Linux, or that it only works on Linux? Because it works for me, on Linux.
> 
> I've been building and running off of this branch's head with no issues.
> 
> Can you unzip the resulting jar and see what's actually being written into your META-INF/MANIFEST.MF?
> 
> Do you need to set mainClassName outside of application { mainClass }? Hm.

I'm pretty sure this is because shadowJar extends jar, but it's a separate plugin task.  There might be other ways to do it, but I tried adding the shadow plugin and then it complained.  I added the mainClassName and it stopped complaining.

It's how my other side project works, so I was pretty sure it would work here, too...

gausie:
@BadHorseMonkey could you add commits to that effect to this branch? You should have permission to do so

heeheehee-kolmafia:
With shadowJar I too get revision == 0, but that's because shadowJar doesn't depend on the version-setting logic (only jar does).

shadowJar.dependsOn gitRevList
shadowJar.dependsOn svnRevList

fixes that for me.

BadHorseMonkey:
> @BadHorseMonkey could you add commits to that effect to this branch? You should have permission to do so

I'll see if I can.

gausie:
Merging this, @Rinn it would be great if you could look over and throw in a supplementary PR if you deem it necessary. This is working great though, we just need to run `shadowJar` not `jar` as our task.

Also would be good to see if we can slim down the `build.gradle` file as I reckon we have a lot of redundant stuff in there.

PR #22, opened 2021-09-26
Language Server - parseAggregateLiteral arrays

fredg1:
reset of https://github.com/kolmafia/kolmafia/pull/7


heeheehee-kolmafia (approved):

PR #23, opened 2021-09-26
Rearranged StaticEntity check for null revision for developers runnin…

BadHorseMonkey:
So the "if" wasn't failing, it was succeeding if the revision was null.

Solved by initializing the cachedRevisionNumber to 0 prior to the if statement.
Tested with IJ.  works for me.


BadHorseMonkey:
wrong one.

PR #24, opened 2021-09-26
Rearranged StaticEntity check for null revision for developers runnin…

BadHorseMonkey:
fixed the revision 0 causing an NPE in the IDE.

initialized the cachedRevisionNumber prior to setting it.

Tested with IJ


BadHorseMonkey:
obsolete due to PR 26

PR #25, opened 2021-09-26
Update jenkinsfile

fewyn:
Update jenkinsfile to use --no-daemon and change the test build targets.


fewyn:
This should resolve any issues with Jenkins, starts gradle with no daemon support, moved testing to Java 16 (1.8 is causing major issues with Jenkins and hard crashing), also it should clean the workspace prior to build now.

PR #27, opened 2021-09-26
Language Server - parseVariable initializations

fredg1:
manual reset of https://github.com/kolmafia/kolmafia/pull/8

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/Parser.java
+++ b/src/net/sourceforge/kolmafia/textui/Parser.java
@@ -784,6 +784,11 @@ public class Parser
 				throw this.parseException( "Parameter " + param.getName() + " is already defined" );
 			}
 
+			if ( this.currentToken().equals( "=" ) )
+			{
+				throw this.parseException( "Cannot initialize parameter " + param.getName() );
+			}
+
 			if ( !this.currentToken().equals( ")" ) )
 			{
 				// The single vararg parameter must be the last one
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -864,68 +894,49 @@ else if ( scope != null && scope.findVariable( variableName.content ) != null )
...
-		// Otherwise, we must initialize the variable.
+		return result;
+	}
 
-		Value rhs;
+	/**
+	 * Parses the right-hand-side of a variable definition. It is assumed that the caller expects
+	 * an expression to be found, so this method never returns null.
+	 */
+	private Value parseInitialization( final Type t, final BasicScope scope )
```
heeheehee-kolmafia, line 906: Hm. If you really wanted access to the variable name, I suppose you could pass in the Variable (or VariableReference) here instead of just the Type. Doesn't make sense to initialize something other than a variable, anyhow.

heeheehee-kolmafia (approved):

(One optional suggestion, but feel free to skip if you don't want to make that change.)

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: I'm mostly reluctant to add a parameter for a single, edge-case purpose.
At the same time, I guess it would still make sense, since that Variable is tied to the assignment...

At at the same the same time time, one of the end-results of the next step is to add the exact location of the errors in the error message, so even without the name of the variable, people will still be able to tell which expression brought the error...

fredg1:
Need new approval after solving a merge conflict

src/net/sourceforge/kolmafia/textui/Parser.java
heeheehee-kolmafia: Oh, I meant in lieu of the Type, since the Variable(Reference) knows what its type is.

heeheehee-kolmafia (dismissed):

fredg1:
oh... yeah, that works

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: Oh, yeah, that works

fredg1:
Note: VariableReference.toString() => return this.target.getName();

heeheehee-kolmafia (approved):

PR #29, opened 2021-09-26
Language Server - parseType records

fredg1:
manual reset of https://github.com/kolmafia/kolmafia/pull/10


heeheehee-kolmafia (approved):
Re-approving, as this also looks about the same as it did before.

PR #35, opened 2021-09-27
Move over sourceforge references and remove unused donation section

gausie:
@Veracity0 I removed the donation buttons and references thereto. Have we ever had any money come in from that over the years? Not a huge worry either way. If we ever needed to go back to accepting donations, we could register with https://opencollective.com/

The only thing left is the code for running a buffbot. I don't know how it works and can't see why it needs to pull `buffbot.xsl` from our old website. Anyone have any ideas?


heeheehee-kolmafia:
I would just download the buffbot.xsl file and bundle it with the rest of our data depedencies. It looks like it's just some HTML templating.

heeheehee-kolmafia (commented):
Grep turned up a couple more Sourceforge references:

- textui/command/FortuneCommand.java (in a comment, it seemed)
- the aforementioned BuffBotManager logic.

Otherwise, this seems reasonable. I'll defer to Veracity on the donations front.

gausie:
>     * textui/command/FortuneCommand.java (in a comment, it seemed)

Weird... I can't see this at all!

>     * the aforementioned BuffBotManager logic.

Done

jaadams5:
I'm still getting used to the New World Order and didn't notice I was tagged.  I'm still getting used to the context but if this is grep driven removals of sourceforge references previously missed then they look fine.  I am surprised that there is mafia licensing/copyright info in spellcast since at one point there was a deliberate decision not to make any claims about third party source we used, but just include their license somewhere.  But mu surprise is just historical commentary.

I recall past discussion of donations and there weren't many donations, if at all.  We should consider reinstating it once we have a sense of who should be steward and how donations should be spent.

I'm going to Close with Comment which may mean this is an approving review (which is fine but I don't know for sure).

gausie:
Nope! This closes it without merging, you need to hit the green accept button on the top right!

heeheehee-kolmafia:
> Weird... I can't see this at all!

I think I just hadn't synced, and this was https://github.com/kolmafia/kolmafia/pull/32. You're good!

PR #36, opened 2021-09-27
Language Server - Type equals Type

fredg1:
Convert multiple
"(Type) == (Type)"
into
"(Type).equals( (Type) )"

The resultint logic is the same, as the only Types that were doing that are primitive types, and there's only one instance of each primitive type.


heeheehee-kolmafia:
Is there a guarantee that getType() != null? That's my main concern with a change like this. Otherwise looks reasonable.

fredg1:
> 
> 
> Is there a guarantee that getType() != null? That's my main concern with a change like this. Otherwise looks reasonable.

it is guaranteed, yes.

PR #37, opened 2021-09-27
feat: try multiple faxbots if the first fails in the CLI

midgleyc:
Reference: https://kolmafia.us/threads/trying-to-use-ash-to-faxbot-a-factory-overseer-fails.26271/

The bug occurs when, for example:

1. easyfax is tried first
2. easyfax offers the monster you want
3. easyfax has become trapped in a clan over rollover, so can't actually return the monster you want

The thread talks about several other options, but I think this change is still good.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
@@ -20,10 +20,10 @@ public class FaxbotCommand
 	{
 		this.usage = " [command] - send the command to faxbot";
 	}
-	
+
 	@Override
 	public void run( final String cmd, final String command )
-	{	
+	{
 		FaxBotDatabase.configure();
 
 		boolean tried = false;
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
```diff
@@ -59,9 +61,13 @@ public void run( final String cmd, final String command )
 			}
 
 			Monster monster = bot.getMonsterByCommand( (String)commands.get( 0 ) );
-			FaxRequestFrame.requestFax( botName, monster, false );
-			return;
+			tried = true;
+			if (FaxRequestFrame.requestFax( botName, monster, false )) {
```
gausie, line 65: This needs mafia-style formatting

gausie:
I can't remember, when does `requestFax` decide that a fax request has failed?

src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
midgleyc: do you have a styling file or lint I can use?

midgleyc:
Good point, lots and lots of reasons:
1. can't receive a fax
2. bot not online
3. used to check whether user is in chat, but that code ignored by the `|| true`
4. user refused to dump their fax
5. bot didn't respond
6. bot said the fax wasn't available

Of those, I guess 2, 5 and 6 should try another bot.

midgleyc:
Looks like ASH tries again for any failure.

src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
midgleyc: I found, and used, the one [posted by heeheehee](https://kolmafia.us/threads/mafia-source-code-formatting.26013/#post-161996).

jaadams5:
I'm still trying to figure out how to do things so forgive me for not commenting on a line of unchanged code.  It seems to me that if I try to get a fax from one bot and there are too many matches then I really ought to try another bot rather than bail immediately.

jaadams5 (approved):

PR #38, opened 2021-09-27
Language Server - BasicScope library function check

fredg1:
BasicScope.findFunction(String, List, MatchType)'s logic searches first in user functions, going up the parent scopes, then in library functions.

The thing is, the library functions are always *already included* as the topmost parent of the user functions, making the library function search redundant (as can be seen in coverage report, which shows that no match is ever found while looking in the library functions).

This removes that redundant logic.


heeheehee-kolmafia:
One edge case that comes to mind: function resolution via coercion. Not that I expect anyone to write this specific code in practice:

    string path_id_to_name(string x) { return "not a real path"; }
    // Outputs "not a real path".
    print(path_id_to_name(4.0));

heeheehee-kolmafia:
Really what ought to happen is that we shouldn't allow coercion to yield an ambiguous result (as is the case in basically every major language out there that deals with polymorphism in this way). But there's probably someone somewhere who's written code inadvertently using this.

heeheehee-kolmafia:
(You'll want to merge again, or at least cherry-pick 63ac9ad496a6318e913f0f803d0086d4ec978cd5, as we've updated the branch protection rules to run tests across multiple OSes.)

fredg1:
AscensionHistoryRequestTest > parseAscensionHistory() is still failing, aaaaaah :A:

test/root/scripts/test_function_coercion.ash
```diff
@@ -0,0 +1,38 @@
...
+str3 str3_date = str2_date;
+str3 str3_out_format = str2_out_format;
+
+print("> pre-overload; all coerce to original");
+print(format_date_time(str2_in_format, str2_date, str2_out_format));
+print(format_date_time(str3_in_format, str3_date, str3_out_format));
+print(format_date_time(str_in_format, str2_date, str3_out_format));
+print();
+
+string format_date_time(str2 inFormat, str2 dateTimeString, str2 outFormat) {
```
heeheehee-kolmafia, line 27: Are we sure we want to continue supporting this behavior, but only in this nonintuitive way?

Either we should want to allow script authors to override builtin functions, or we should disallow it entirely. But we shouldn't only allow it via a typedef backdoor.

src/net/sourceforge/kolmafia/textui/parsetree/BasicScope.java
```diff
@@ -122,84 +122,47 @@ public final Function findFunction(
     }
 
     // We will consider functions from this scope and from the RuntimeLibrary.
-    Function[] userFunctions = this.functions.findFunctions(name);
-    Function[] libraryFunctions = RuntimeLibrary.functions.findFunctions(name);
+    Function[] functions = this.functions.findFunctions(name);
```
heeheehee-kolmafia, line 125: One other possibility that's used by other languages like Python is to have a separate scope for builtins so if people _really_ want to override builtin functions, they can.

e.g. builtin scope -> global scope -> function-level scope

heeheehee-kolmafia (changes requested):
(Some food for thought.)

src/net/sourceforge/kolmafia/textui/parsetree/BasicScope.java
fredg1: well, isn't this already how ASH is doing it?

heeheehee-kolmafia: Not quite. You can shadow a user-defined function at a different scope, but you cannot shadow a builtin function no matter what you do.

For instance:
```
string a() {
	return "global";
}

print(a());

void main() {
  string a() {
    return "nested";
  }
  print(a());
}
```
will print:
```
global
nested
```

test/root/scripts/test_function_coercion.ash
fredg1: Disallowing it has the potential to break some scripts...

I'm thinking of the inconvenience it would put on people who learn that their function now needs a new name: we don't have a "rename this method" service, so they'd need to manually rename them all...

At the same time, it's not like we're expecting many people to have done that...

jaadams5:
OK.  I am going to make an arbitrary executive decision that a scripter is not allowed to override built-in functions, especially through a back door.  I would ask @fredg1 to change the code so that is true or point out to me that the code already does that.  I will presumably resolve the conversations and approve.

gausie:
> OK. I am going to make an arbitrary executive decision that a scripter is not allowed to override built-in functions, especially through a back door. I would ask @fredg1 to change the code so that is true or point out to me that the code already does that. I will presumably resolve the conversations and approve.

If this is the decision we should also have a test clearly confirming as such

jaadams5:
> > OK. I am going to make an arbitrary executive decision that a scripter is not allowed to override built-in functions, especially through a back door. I would ask @fredg1 to change the code so that is true or point out to me that the code already does that. I will presumably resolve the conversations and approve.
> 
> If this is the decision we should also have a test clearly confirming as such

Perfect.  @fred1 Please write a test.  Make it easy on me - write two new tests.  One can be a trivial script that defines a function.  The other can be the same trivial script except the name of the defined function is the name of a built-in.

Thanks.

fredg1:
Yup, working on it

fredg1:
Done.

jaadams5 (approved):
I hope.

PR #40, opened 2021-09-27
Prune old jars when building with gradle

Rinn:
* Delete any mafia jars except the current version from dist when building jars
* Add cleanDist task as dependency for clean, which will clean all mafia jars
* Force gradlew to always use unix line endings with .gitattributes


build.gradle
```diff
@@ -90,6 +90,36 @@ application {
...
+
+task pruneDist(type: Delete) {
+  onlyIf {
+    file('dist').exists()
+  }
+  inputs.dir('dist')
+  outputs.upToDateWhen { true }
+  doLast {
+    file('dist').eachFile(groovy.io.FileType.FILES) { File file ->
+      if (file.name.startsWith('KoLmafia-') && file.name.endsWith('.jar') && !file.name.contains(project.getVersion())) {
```
heeheehee-kolmafia, line 116: IMO we should prune KoLmafia-12345-M.jar too, which suggests we should check that !file.name.equals("KoLmafia-" + project.getVersion() + ".jar"), or !endsWith(getVersion() + ".jar").

(In addition to the startsWith / endsWith checks you have, to make sure we only nuke jars.)

Other than that, this change looks good.

heeheehee-kolmafia (approved):

build.gradle
Rinn: done

heeheehee-kolmafia:
Mmm, what I meant was "if it's dirty, always delete it, since there could be more changes." (even more dirty? That's not a phrase.)

Rinn:
> Mmm, what I meant was "if it's dirty, always delete it, since there could be more changes." (even more dirty? That's not a phrase.)

You don't really need to though because it will just get replaced when :jar or :shadowJar runs if the revision is the same.

build.gradle
```diff
@@ -90,6 +90,38 @@ application {
...
+
+task pruneDist(type: Delete) {
+  onlyIf {
+    file('dist').exists()
+  }
+  inputs.dir('dist')
+  outputs.upToDateWhen { true }
+  doLast {
+    file('dist').eachFile(groovy.io.FileType.FILES) { File file ->
+      if (file.name.startsWith('KoLmafia-') && file.name.endsWith('.jar')) {
```
BadHorseMonkey, line 116: Two questions: is it case sensitive?  I think that's ok if it is.

And also unimportantly, can we use .startsWith('KoLmafia') so that users who do ANT builds can get cleaned up, too?

heeheehee-kolmafia: It should be case-sensitive.

Don't Ant builds specify a dash, either `daily` or `jar`?

BadHorseMonkey: Nope.  It comes straight from sVN:
```

Michaels-MBP:kolmafia mcroft$ svnversion
20947M
```

I didn't fight it at the time because I didn't care that much and it gives me a visual clue as to which version we used.

heeheehee-kolmafia: That's the version, not the full jar. Right?

```
version-name=${ant.project.name}-${version}
jarfile=${dist}/${version-name}.jar
```

My jars that I built just now with Ant look like KoLmafia-25704.jar and KoLmafia-21.9.jar.

heeheehee-kolmafia: Oh, I see, this probably should be endsWith('M.jar') without the dash. That's what you're getting at, right?

heeheehee-kolmafia: Yeah. Ant is firmly *not supported* at this point, given that it can't even bundle revision numbers anymore.

BadHorseMonkey: Yeah, it's got a few features we don't have in gradle, but it's not something we want to support.

daily is probably the biggest missing item, which depends on ANT calling out to svn update.

We can switch that to grgit which relies on JGit, which I want to use anyway for git scripts, but it's a bigger task than I thought at first.

PR #42, opened 2021-09-28
Change some ints to longs in BasementRequest

Rinn:
This prevents truncating doubles to a narrower type, which could overflow and return negative values. Resolves CodeQL code analysis warning.


jaadams5:
Looks like Close with comment was not enough to autotrigger Squash and merge.  Maybe I was impatiend and needed to wait for tests to finish?

gausie:
No close doesn't merge, don't close the PR. Close is more like a "wontfix"

jaadams5:
OK.  When I am looking at a PR I am willing to approve and close how do I merge?  I don't think I have ever seen the Squash and merge option not marked with a red X.  Does making a comment change that?  Is there an Approve button I am just not seeing?  I don't think it is a permission issue because I am reported as member of Kolmafia and Kolmafia/staff.  Thanks.

PR #43, opened 2021-09-28
dev: add IntelliJ IDEA formatting file

midgleyc:
Originally from heeheehee's post:
https://kolmafia.us/threads/mafia-source-code-formatting.26013/#post-161996

Useful for contributors to have something like this somewhere accessible, and the repository is a good place. There was already a `util/format.xml` file for eclipse (probably outdated by the forum thread), so I added the IntelliJ one next to it.

### DIFF
--- a/util/format_intellij.xml
+++ b/util/format_intellij.xml
@@ -1,0 +1,77 @@
+<code_scheme name="KoLMafia Style" version="173">
+  <JavaCodeStyleSettings>
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+    <option name="PACKAGES_TO_USE_IMPORT_ON_DEMAND">
+      <value />
+    </option>
+  </JavaCodeStyleSettings>
+  <Properties>
+    <option name="SPACES_AROUND_KEY_VALUE_DELIMITER" value="true" />
+  </Properties>
+  <editorconfig>
+    <option name="ENABLED" value="false" />
+  </editorconfig>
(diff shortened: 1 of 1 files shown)


jaadams5:
@BadHorseMonkey - how does this compare to what you were using?

Anyone - Are we close enough to adopting a Google standard that can be easily applied across IDEs to make this obsolete real soon?  Or do we accept it, apply it and remove it when we do make a decision?

heeheehee-kolmafia:
@jaadams5 since you asked for it, I spent all of 10 minutes tinkering to create https://github.com/kolmafia/kolmafia/pull/44

Optionally, if we want, I can add spotless to build.gradle too so if people want to use that, they can.

jaadams5:
Does Add Google Java Format on push #44 make this overtaken by events or is it still useful for someone who prefers the old style and is willing to apply it on a case by case basis?

@heeheehee-kolmafia ?

midgleyc:
If you're adopting an external style with a known name, I think it's fine to not put that in the repository. Contributors can configure their editors to use that style.

Should probably make a note that that's the style used somewhere (`CONTRIBUTING.md`?), but having it be applied by action is fine too.

jaadams5:
I'm a little confused, but since the Google style will be enforced on commit, local formatting is up to the developer.  If the justification is that all formatting renegades using IntelliJ can use the same format, that's fine.  Shall I make this a review and approve?

heeheehee-kolmafia:
I think we're backing out of the github-action formatting model and instead opting to push the responsibility onto the developer.

In that case, IntelliJ users should instead use Google Java Style, rather than this one.

gausie:
Yeah I'm going to close this commit. We can add instructions on how to set up intellij to format on save, if you desire, to the eventual README.md/CONTRIBUTING.md

PR #44, opened 2021-09-28
Add Google Java Format on push

heeheehee-kolmafia:
This automatically formatted my local repo, so I think it must do something right.


heeheehee-kolmafia:
(I apologize for this changing literally every source file in the repo.)

midgleyc:
You might want to exclude the vendored code in `/lib` from the reformatting action.

Hopefully most of it can be replaced with jars to avoid that issue.

heeheehee-kolmafia:
Done. Unfortunately the vendored code isn't in jars because there are no jars -- a lot of it is code that someone found on forums somewhere. :(

midgleyc:
I think `files` should be specified separately (assuming it's not working as-is on this branch, while it was before):
```
- uses: axel-op/googlejavaformat-action@v3
  with:
    args: "--replace"
    files:
      - "src/**/*.java"
      - "test/**/*.java"
```
Not certain that'll work looking at the code, it might have to be `src/**/*.java\ntest/**/*.java`, but I hope that passing it as an array works.

heeheehee-kolmafia:
Well, I can try that too. Thanks, I'm basically clueless as far as what I'm doing here.

midgleyc:
It looks to be single string, but a limited number of constructs are supported -- specifically, those in the [`@actions/glob`](https://www.npmjs.com/package/@actions/glob) library.

Try `src/**/*.java` to confirm we have anything right at all.

If we do, then `[st]*/**/*.java` is my terrible attempt at a supported glob that matches src, test and nothing else.

gausie:
Is this then meant to make a commit that formats your code? I think that's a bad idea

I was looking at using spotless for gradle to apply java formatting and it worked really well. Maybe we can do that instead?

Also Google java formatting is different to our current styling. I'd be absolutely willing to move to that styling that though

gausie:
Yeah it's committing from GitHub. I'd really like to not do it that way

heeheehee-kolmafia:
If you prefer, I have a separate branch that uses spotless to format code. I'll close this PR out.

I don't feel strongly about it either way.

gausie:
I appreciate it. It's just a real pain from expertise to get commits added to your PRs all the time. It clutters the commit log and causes extra merge commits when you forget to pull before your next change etc.

We could have GitHub actions check that the changed files would not change if styled though and add a warning to the PR if so.

heeheehee-kolmafia:
Yeah, I agree. I didn't realize that's how it would work until I started tinkering.

I also agree that this could just be a check -- there's a --set-exit-if-changed flag, and also a skipCommit parameter that looks like it would do what we want. But I think applying formatting is orthogonal to checking that the formatting is correct.

Rinn:
You can add a commit hook that auto formats when committing to the local branch. I'd need to look this up because I don't remember the details but that was set up at a previous job I was at to enforce code style.

Rinn:
https://github.com/a1exsh/google-java-format-git-pre-commit-hook maybe

heeheehee-kolmafia:
I do have a pre-commit hook in place for me personally, but I'd prefer to enforce something for everyone.

PR #46, opened 2021-09-28
Implicit Narrowing part 2

jaadams5:
Address additional implicit narrowing warnings

### DIFF
--- a/src/net/sourceforge/kolmafia/request/HermitRequest.java
+++ b/src/net/sourceforge/kolmafia/request/HermitRequest.java
@@ -458,7 +458,7 @@ public class HermitRequest
 		if ( InventoryManager.canUseNPCStores() )
 		{
 			int cost = SewerRequest.currentWorthlessItemCost();
-			count += KoLCharacter.getAvailableMeat() / cost;
+			count += ((int)KoLCharacter.getAvailableMeat()) / cost;
 		}
 		return count;
 	}
--- a/src/net/sourceforge/kolmafia/session/ResultProcessor.java
+++ b/src/net/sourceforge/kolmafia/session/ResultProcessor.java
@@ -1185,7 +1185,7 @@ public class ResultProcessor
 				long diff = KoLCharacter.calculateBasePoints( stat + count ) -
 					    KoLCharacter.calculateBasePoints( stat );
 				AdventureResult.SESSION_SUBSTATS[ 0 ] += count;
-				AdventureResult.SESSION_FULLSTATS[ 0 ] += diff;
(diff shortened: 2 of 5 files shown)


src/net/sourceforge/kolmafia/request/HermitRequest.java
```diff
@@ -458,7 +458,7 @@ public static final int getAcquirableWorthlessItemCount()
 		if ( InventoryManager.canUseNPCStores() )
 		{
 			int cost = SewerRequest.currentWorthlessItemCost();
-			count += KoLCharacter.getAvailableMeat() / cost;
+			count += ((int)KoLCharacter.getAvailableMeat()) / cost;
```
jaadams5, line 461: Meat is stored as long but there are many places where it is truncated to an int.  Making count a long would effect several other pieces of code.

src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -1185,7 +1185,7 @@ else if ( resultName.equals( AdventureResult.SUBSTATS ) )
 				long diff = KoLCharacter.calculateBasePoints( stat + count ) -
 					    KoLCharacter.calculateBasePoints( stat );
 				AdventureResult.SESSION_SUBSTATS[ 0 ] += count;
-				AdventureResult.SESSION_FULLSTATS[ 0 ] += diff;
+				AdventureResult.SESSION_FULLSTATS[ 0 ] += (int) diff;
```
jaadams5, line 1188: AdventureResult.SESSION_FULLSTATS[  ] might perhaps be changed to long[] but that has ripple effects

src/net/sourceforge/kolmafia/request/HermitRequest.java
Rinn: Will this actually affect the narrowing warning? It's still going to narrowing a long to an int and possibly overflow if out of the int range.

jaadams5: I read the warning as being about an implicit conversion.  An explicit cast may be a bad thing but compilers and other tools can deal with explicit casts differently than implicit casts.  The explicit cast may be wrong but it was a deliberate.  That said, I don't know how to run the security check locally so I don't know for sure.  If this is submitted and the warning still occurs after test then I just make one more iteration and this time select the wider ranging changes.

gausie (approved):

src/net/sourceforge/kolmafia/request/HermitRequest.java
jaadams5: Missed one but the cast seems to have addressed the warning for the ones I did catch.

PR #47, opened 2021-09-28
Add spotless w/ google java style.

heeheehee-kolmafia:
This is a proof-of-concept, and does not actually apply any formatting
changes yet. I'd prefer to make that formatting change in a standalone
commit.

This also removes some comments in the middle of imports in
KoLConstants.java because `./gradlew spotlessCheck` was yelling at me,
because Google Java Format (gojf) doesn't like that.

To reformat the code, you'll run `./gradlew spotlessApply`.

This adds some flags so gojf continues to work for newer Java versions
(16+). But for that, requires that the Java compiler be new enough to
support exports, aka at least Java 9, which coincidentally is the
highest version supported by OpenClover.

We really should switch away from OpenClover to something like JaCoCo.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ plugins {
   id 'com.bmuschko.clover' version '3.0.1'
   id 'net.nemerosa.versioning' version '2.8.2'
   id 'com.github.johnrengelman.shadow' version '6.0.0'
+  id 'com.diffplug.spotless' version '5.15.1'
 }
 mainClassName = 'net.sourceforge.kolmafia.KoLmafia'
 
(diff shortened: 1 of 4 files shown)


heeheehee-kolmafia:
(If Java 8 is indeed subtly broken with our build, then I don't feel so bad about breaking it more explicitly.)

gausie (approved):

heeheehee-kolmafia:
If you don't mind, I'll merge this and post an announcement stating that Java 9 is now needed for Mafia (in part due to a previously reported Java 8 bug), but we strongly recommend upgrading to the latest LTS release (Java 17), as we intend to upgrade our codebase again in the near future.

heeheehee-kolmafia:
Hm, actually, the Java 9 thing is only needed to _build_ Mafia. But, I think it's still useful to enforce this requirement for various other reasons (subtle Java 8 breakages and whatnot). I'd bump this further, but then Clover would start breaking without additional flags, and I think it's simpler to do that with the JaCoCo migration if we want to do it then.

PR #48, opened 2021-09-28
Migrate from OpenClover to JaCoCo for coverage

heeheehee-kolmafia:
They have support for Java > 9...

I think I got all the references to Clover (outside of the ant build), but it's hard to be sure.

Also this may have broken the jenkinsfile, so please read this PR with extreme skepticism.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,10 +2,11 @@
 
 plugins {
   id 'application'
-  id 'java'
-  id 'net.nemerosa.versioning' version '2.8.2'
-  id 'com.github.johnrengelman.shadow' version '6.0.0'
   id 'jacoco'
+  id 'java'
+
+  id 'com.github.johnrengelman.shadow' version '6.0.0'
+  id 'net.nemerosa.versioning' version '2.8.2'
 }
 mainClassName = 'net.sourceforge.kolmafia.KoLmafia'
 
--- a/build.gradle
+++ b/build.gradle
@@ -3,9 +3,9 @@
 plugins {
   id 'application'
   id 'java'
-  id 'com.bmuschko.clover' version '3.0.1'
   id 'net.nemerosa.versioning' version '2.8.2'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -3,9 +3,9 @@
 plugins {
   id 'application'
   id 'java'
-  id 'com.bmuschko.clover' version '3.0.1'
   id 'net.nemerosa.versioning' version '2.8.2'
   id 'com.github.johnrengelman.shadow' version '6.0.0'
+  id 'jacoco'
```
gausie, line 8: Do we not want to pin a version?

gausie (commented):
Does this work with codecov also

build.gradle
heeheehee-kolmafia: Error resolving plugin [id: 'jacoco', version: '0.8.7']
> Plugin 'jacoco' is a core Gradle plugin, which cannot be specified with a version number. Such plugins are versioned as part of Gradle. Please remove the version number from the declaration.

heeheehee-kolmafia:
I think it must work with codecov, there's documentation about it that I didn't really read.

https://about.codecov.io/tool/jacoco/

https://github.com/codecov/example-gradle

build.gradle
heeheehee-kolmafia: Actually, if you want, I can add a commit to sort the plugins so all the core plugins are in one block.

Rinn (approved):
Seems fine, @fewyn said Jenkins has a JaCoCo plugin so I'd just check with them first.

heeheehee-kolmafia:
Yeah. I attempted to update the jenkinsfile to reference JaCoCo instead of Clover but I have no idea how to check if I did that right.

fewyn:
We can just merge this and I can update the jenkinsfile after some testing to make sure it works. I can just host and test one directly on the jenkins instance.

heeheehee-kolmafia:
Sounds good, thanks.

PR #52, opened 2021-09-29
Implicit narrowing

jaadams5:
Missed one in the security warnings.

Warning is about implicit casts.  Explicit casts aren't security concerns.

If I was supposed to format this to Google Java, would someone walk me through the process of updating the PR?  Otherwise does someone deal with that during a merge, depending upon the order things happen?


gausie:
@jaadams5 did you mean to close this without merging?

PR #53, opened 2021-09-29
Implicit Narrow

jaadams5:
Eliminate the last of the implicit narrowing by cast warnings.

The warning is about implicit casts.  Explicit casts are fine.


heeheehee-kolmafia (commented):
Not that I'd ever expect anyone to have 10^16 MP to spend on this, but it is technically a lossy operation.

Since Java doesn't have an integer exponentiation function, it's probably okay...

*I* would have probably written the loop body as mpCost = mpCost * 10 + 1;

Either way, this seems like something that's easy enough to add a test for. Can you do that?

jaadams5:
This was flagged as a security issue because a lossy implicit conversion is bad but a lossy explicit conversion (which is what making the cast did) is acceptable.  The coder knew lossy was a possibility.

I'll look at adding a test once I can run tests post-Ant on my machine.

heeheehee-kolmafia:
It's fine as a change either way, but you'll want to deal with the merge conflict (or drop + recreate the pull request if that's easier).

I'm just pointing out a way I would have done it without the lossy operation, and without performing pow() in a loop, which is unnecessarily expensive.

jaadams5:
I'll write a test and change.  While replacing Pow is trivial, giving me the opportunity to learn how to add code, and change code, to a pull request is worthwhile.

jaadams5:
@heeheehee-kolmafia 

Did you push a test that used Mockito and the associated dependencies?  If so I can't find it and would like an example.

If not how would you change a preference as part of a test?  Preferences.setInteger() is not changing anything in a test context.

heeheehee-kolmafia:
It's not been merged into the main branch yet. It's #56.

heeheehee-kolmafia:
(You'll want to ./gradlew spotlessApply, as this PR is failing formatting checks.)

src/net/sourceforge/kolmafia/persistence/SkillDatabase.java
```diff
@@ -916,11 +896,15 @@ public static final long birdSkillCasts(long availableMP) {
   }
 
   public static final long stackLumpsCost() {
+    return stackLumpsCost(Preferences.getInteger("_stackLumpsUses"));
```
jaadams5, line 899: I broke this up because I could not figure out how to Mock Preferences.getInteger and this was so much easier to test.

src/net/sourceforge/kolmafia/persistence/SkillDatabase.java
```diff
@@ -161,11 +93,63 @@ public static final int skillTypeNameToType(final String typeName) {
...
+  private static final AdventureResult SUPER_SKILL = EffectPool.get(EffectPool.SUPER_SKILL);
+  private static final ArrayList<String> skillNames = new ArrayList<>();
+  private static String[] canonicalNames = new String[0];
 
   static {
     SkillDatabase.reset();
   }
 
+  public static final String skillTypeToTypeName(final int type) {
+    return type == PASSIVE
```
heeheehee-kolmafia, line 106: (I would like to rip out this and the next function and replace them with switch statements and/or expressions, but best to deal with that separately.)

src/net/sourceforge/kolmafia/persistence/SkillDatabase.java
```diff
@@ -916,11 +896,15 @@ public static final long birdSkillCasts(long availableMP) {
...
+  }
+
+  public static final long stackLumpsCost(int casts) {
     long mpCost = 1;
-    int casts = Preferences.getInteger("_stackLumpsUses");
     if (casts < 0) return mpCost;
+    if (casts > 17) return Long.MAX_VALUE;
     for (int i = 0; i <= casts; i++) {
-      mpCost += 10 * Math.pow(10, i);
+      mpCost = (10 * mpCost) + 1;
```
jaadams5, line 907: The Pow formulation was suggested on the KoL wiki but this implementation actually gave an incorrect answer after 15 casts or so.  The different calculation is correct and overflows will be trapped.

src/net/sourceforge/kolmafia/persistence/SkillDatabase.java
heeheehee-kolmafia, line 899: Ah, nice workaround.

heeheehee-kolmafia (approved):

jaadams5:
I wrote a test so I could run it before my changes and after them and prove that the changes didn't make things worse.  I replaced the calculation as suggested and reformatted as noted.

heeheehee-kolmafia:
It's not surprising to hear that the pow() formula gives wrong results at 14, as double only has 53 bits of mantissa (which is just shy of 10^16, and calculateStacks(n) > 10^(n+1)).

I don't generally expect any character ever to have more than, like, a billion MP, much less a quadrillion, which is why I mentioned this is a purely theoretical concern.

PR #54, opened 2021-09-29
Updated to add JaCoCo test reports to Jenkins CI

fewyn:
Updated to add JaCoCo test reports to Jenkins CI.

This will also fail the build if coverage drops more than 5% which is adjustable via the jenkinsfile.


fewyn:
Already tested and running on Jenkins once it's merged I'll switch it back to using the jenkinsfile.

jenkinsfile
```diff
@@ -60,4 +47,4 @@ pipeline {
       }
     }
   }
-}
+}
```
heeheehee-kolmafia, line 50: nit: can you add a newline at the end of this file?

Otherwise this change looks fine to me.

fewyn: Done.

heeheehee-kolmafia (approved):
Thanks, looks good.

PR #56, opened 2021-09-30
Only increment skill usage for successes.

heeheehee-kolmafia:
NOTE: THIS BREAKS THE ANT TEST BECAUSE WE DON'T BUNDLE JUNIT5.

Thanks to phreddrickk @ kolmafia.us for reporting this issue.

This adds an example of how one might use Mockito to inject a custom
response. It's not super-well-documented, but I promise all the bits
are necessary.

In particular: we need to set session ID or else GenericRequest.run()
will short-circuit, and we need to set max / current MP and "learn"
the skill so UseSkillRequest.run() won't reject it before even passing
it on to GenericRequest.

These test cases follow the standard arrange / act / assert model.


gausie:
Can we make a kolmafia test base class that has a beforeall that sets the hash?

I imagine we'll want a few utilities like that to make setup/teardown easier

gausie (dismissed):
Awesome

gausie (approved):

PR #60, opened 2021-09-30
Use browser launching code available since Java 1.6

gausie:
We lose the ability to launch a browser other than the system default but we also get to delete a really old source dependency.


heeheehee-kolmafia (approved):
Seems reasonable. I think we'll definitely want to make it clear to users "hey this functionality will be broken, and yes, this is intentional."

gausie:
> Seems reasonable. I think we'll definitely want to make it clear to users "hey this functionality will be broken, and yes, this is intentional."

Agreed, I'll note on the forums

PR #61, opened 2021-09-30
Create CODEOWNERS file

gausie:
This will just get github to automatically request review from @kolmafia/staff for reviewing PRs. I don't really recommend we use it in any more fine-grained way than this.


jaadams5:
I like the concept.  Presumably if it generates too much notification email for folks we can back it out or teach ourselves how to fine tune which notifications generate email.

jaadams5 (approved):
OK

PR #64, opened 2021-09-30
Remove OSXAdapter and replace with Java built-ins

gausie:
Everything is replaced with its equivalent. There was a line that allowed frames to be made fullscreenable but I think OSX has changed to allow that by default, and Java indeed doesn't offer a replacement as a result.


util/KoLmafia.importorder
```diff
@@ -1,4 +1,3 @@
-0=apple.dts.samplecode.osxadapter
 1=ca.bcit.geekkit
```
heeheehee-kolmafia, line 1: Should this be zero-indexed, or does it not matter?

heeheehee-kolmafia (approved):

BadHorseMonkey:
I didn't see anything unusual when I checked out this branch and ran it.

Java broke taskbar badge support, but that was last year.

src/net/sourceforge/kolmafia/swingui/listener/LicenseDisplayListener.java
```diff
@@ -9,10 +9,8 @@
   private static final String[] LICENSE_FILENAME = {
     "kolmafia-license.txt",
```
BadHorseMonkey, line 10: kolmafia-license.txt is missing, but I think that predates your change.  Should be in src/licenses, but it's no longer there and the page in About is blank.

BadHorseMonkey (approved):
Works.  Doesn't need to fix broken licenses.txt file, but it could.

util/KoLmafia.importorder
gausie: Nah it just traverses in order I think

gausie:
> I didn't see anything unusual when I checked out this branch and ran it.
> 
> Java broke taskbar badge support, but that was last year.

All my badges work perfectly by the way!

PR #66, opened 2021-10-01
Remove Ant build

heeheehee-kolmafia:
We add grgit integration (thereby removing the explicit need for the git CLI tool), but not dmg / exe building, or the documentation task.

Since that's now done, we can then remove the Ant build files, as well as the various bundled jars.


build.gradle
```diff
@@ -188,19 +185,27 @@ task download(type: Copy) {
...
-    project.setVersion(standardOutput.toString().trim())
+    project.setVersion(revision.toString().trim())
+  }
+}
+
+task gitUpdate() {
+  doLast {
+  grgit.pull {
+    remote= 'origin'
+    branch= 'main'
```
BadHorseMonkey, line 207: Would prefer if this updated the current branch instead.

Maybe check for dirty state, and stash/pull/stash pop

Would a rebase command be useful or overwhelming?

heeheehee-kolmafia:
Regarding packaging, if we still want to distribute binaries, it looks like the modern way to do so is via jpackage (as of JDK16), which can generate exe / deb / dmg and several other formats.

I haven't found a gradle plugin to just do that, although I imagine we could create an exec task that runs it directly.

build.gradle
```diff
@@ -188,19 +185,27 @@ task download(type: Copy) {
...
+  }
   }
 }
```
BadHorseMonkey, line 211: Should we keep svnRevList?

BadHorseMonkey (commented):
I don't see anything in here for detecting a modified repository and adding a 'M' to the version string.  Did that concept go away?  I found it useful.

BadHorseMonkey:
> Regarding packaging, if we still want to distribute binaries, it looks like the modern way to do so is via jpackage (as of JDK16), which can generate exe / deb / dmg and several other formats.
> 
> I haven't found a gradle plugin to just do that, although I imagine we could create an exec task that runs it directly.

It's actually already there:

Distribution tasks
------------------
assembleDist - Assembles the main distributions
assembleShadowDist - Assembles the shadow distributions
distTar - Bundles the project as a distribution.
distZip - Bundles the project as a distribution.
installDist - Installs the project as a distribution as-is.
installShadowDist - Installs the project as a distribution as-is.
shadowDistTar - Bundles the project as a distribution.
shadowDistZip - Bundles the project as a distribution.

Michaels-MBP:kolmafia mcroft$ ./gradlew assembleDist

BUILD SUCCESSFUL in 5s
9 actionable tasks: 4 executed, 5 up-to-date

Michaels-MBP:kolmafia mcroft$ ls -l build/distributions/
total 210312
-rw-r--r--  1 mcroft  staff  35215360 Sep 30 22:29 KoLmafia-25724.tar
-rw-r--r--  1 mcroft  staff  31822443 Sep 30 22:29 KoLmafia-25724.zip
drwxr-xr-x  5 mcroft  staff       160 Sep 30 22:27 KoLmafia-shadow-25724
-rw-r--r--  1 mcroft  staff  20541440 Sep 30 22:26 KoLmafia-shadow-25724.tar
-rw-r--r--@ 1 mcroft  staff  19106309 Sep 30 22:26 KoLmafia-shadow-25724.zip

build.gradle
heeheehee-kolmafia: So, the pull command also has a boolean for rebase which we could use as well.

build.gradle
heeheehee-kolmafia: > I don't see anything in here for detecting a modified repository and adding a 'M' to the version string. Did that concept go away? I found it useful.

I didn't do anything about that! It was always handled by the nemerosa versioning package.

heeheehee-kolmafia: Deleted in the latest commit.

heeheehee-kolmafia:
>  I don't see anything in here for detecting a modified repository and adding a 'M' to the version string. Did that concept go away? I found it useful.

I didn't do anything about that! It was always handled by the nemerosa versioning package, and should still be. ./gradlew shadowJar created KoLmafia-25724-M.jar for me.

heeheehee-kolmafia:
Regarding distribution tasks: does that mean we should remove util/{debian,macosx,windows}?

build.gradle
heeheehee-kolmafia: Sadly, grgit does not have stash support.

Also, spotless isn't formatting this config well, clearly. Hm.

BadHorseMonkey:
> > I don't see anything in here for detecting a modified repository and adding a 'M' to the version string. Did that concept go away? I found it useful.
> 
> I didn't do anything about that! It was always handled by the nemerosa versioning package, and should still be. ./gradlew shadowJar created KoLmafia-25724-M.jar for me.

I don't know how much change you've made.  I commented out a line in GenericRequest.java, so it was definitely changed.

I was wondering if we could use grgit to do this GIT command and if it's the same:

> Michaels-MBP:kolmafia mcroft$ git describe --tags --always --first-parent --dirty=M --abbrev=0
> r25724M

I got the idea from [gradle-git-version](https://github.com/palantir/gradle-git-version), but they can't configure the --dirty flag yet.

build.gradle
BadHorseMonkey: Is there an option to print some sort of status?  Right now it acts silently, which isn't super great for knowing if everything is up to date.

heeheehee-kolmafia:
(I probably want to cherry-pick the gradle formatting changes into a separate PR, so this one is easier to review. Not now.)

BadHorseMonkey:
> > I don't see anything in here for detecting a modified repository and adding a 'M' to the version string. Did that concept go away? I found it useful.
> 
> I didn't do anything about that! It was always handled by the nemerosa versioning package, and should still be. ./gradlew shadowJar created KoLmafia-25724-M.jar for me.

whatever was wrong is fixed now.  My dirty repo works.

`Michaels-MBP:kolmafia mcroft$ ls dist
KoLmafia-25729-M.jar
`

BadHorseMonkey:
> Regarding distribution tasks: does that mean we should remove util/{debian,macosx,windows}?

Probably.  And if we go to JPackage, definitely, but the zip/tar commands just make a jar and also a windows .bat file.  It doesn't have to be in scope of this PR to fix that, unless you want to.

build.gradle
BadHorseMonkey: seeing that now, thank you.

BadHorseMonkey (dismissed):
This is working well enough for this PR, which is to say nothing is broken.

I'm good with this as-is, but will also look at any further tweaks you want to do with it

Thank you!  It was interesting going through the review process.

build.gradle
```diff
@@ -191,15 +191,19 @@ task download(type: Copy) {
...
+task getRevision() {
 	onlyIf {
 		file('.git').exists()
 	}
 	inputs.dir('.git')
 	outputs.files file('build/revision.txt')
 
 	doLast {
-		def revision = grgit.log(includes:['origin/main']).size()
+		def remote = System.getenv('HEAD_COMMIT_HASH')
```
heeheehee-kolmafia, line 202: @Rinn is this sufficient for your use case? I'm not sure how else to do this.

Rinn: I'll prep a PR separately for this after you commit this, there's some gradle stuff to consider the environment variable as a dependency so it invalidates the task correctly.

Rinn (approved):

build.gradle
Rinn: #72 should do it, but I'll wait for you to commit this first.

PR #70, opened 2021-10-02
Run checks against latest mac/windows.

heeheehee-kolmafia:
This also opts to use ubuntu-latest so we don't have to manually
configure it in the future.

Note that there are technically newer OSes than windows-latest and
macos-latest at this time, but I anticipate that these labels will be
automatically migrated over time (unlike a fixed label that will
require periodic maintenance).

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,8 +10,7 @@ jobs:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
 
-    # This needs to be called Run Tests to satisfy our github setup.
-    name: Run Tests
+    name: Run Checks
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia:
Hrm. The Run Tests thingy is unhappy with the matrix.

heeheehee-kolmafia:
(In particular, I believe this is based on how our branch protection is set up -- the matrix modifies the check names to "Run Tests (ubuntu-latest)" etc. We'd either have to edit that, or put this into a different check. I'd prefer to change branch protection.)

BadHorseMonkey:
I'll hold off on reviewing until it's passing tests...

heeheehee-kolmafia:
(It is passing tests! It's just not matching the expected test target by name.)

BadHorseMonkey:
> (It is passing tests! It's just not matching the expected test target by name.)

I stand corrected.  Do you want me to review or wait for more changes?

heeheehee-kolmafia:
So, this is the entire change I wanted to make, but I wonder if I could set up a task that just expects the sub-tests to succeed, so we don't need to modify our branch protection at all. I'll stare some more at github docs.

heeheehee-kolmafia:
(I should add comments to make it clearer what this placeholder action is trying to accomplish.)

gausie (approved):

PR #71, opened 2021-10-02
Update SVNManager.java

jaadams5:
Change error message as requested.

### DIFF
--- a/src/net/sourceforge/kolmafia/svn/SVNManager.java
+++ b/src/net/sourceforge/kolmafia/svn/SVNManager.java
@@ -515,8 +515,7 @@ public class SVNManager {
 
     if (failed && !quiet) {
       KoLmafia.updateDisplay(
-          MafiaState.ERROR,
-          "The requested repo failed validation.  Complain to the script's author.");
+          MafiaState.ERROR, "The requested repo (" + repo.getPath() + " failed validation.");
     } else {
       if (!quiet) RequestLogger.printLine("Repo validated.");
     }


src/net/sourceforge/kolmafia/svn/SVNManager.java
```diff
@@ -515,8 +515,7 @@ private static boolean validateRepo(SVNURL repo, boolean quiet) {
```
heeheehee-kolmafia, line 515: hm, it looks like failed is potentially being cleared depending on loop order, right before this (in the dependencies.txt check; should probably read failed |= ...).

(I was curious whether we could provide a better error message for _why_ it failed validation, in addition to what as you have here.)

heeheehee-kolmafia (approved):

src/net/sourceforge/kolmafia/svn/SVNManager.java
jaadams5: I couldn't really get a useful debugger run without actually changing a repository and that was more work than just making the change.   I wondered about a better message but I wasn't sure I wanted to understand the SVN code well enough to actually make an intentional change.  Worth revisiting if it is not made obsolete by adding git support to KoLmafia or refactoring how scripts get installed and updated.

src/net/sourceforge/kolmafia/svn/SVNManager.java
```diff
@@ -515,8 +515,7 @@ private static boolean validateRepo(SVNURL repo, boolean quiet) {
 
     if (failed && !quiet) {
       KoLmafia.updateDisplay(
-          MafiaState.ERROR,
-          "The requested repo failed validation.  Complain to the script's author.");
+          MafiaState.ERROR, "The requested repo (" + repo.getPath() + " failed validation.");
```
gausie, line 518: @jaadams5  This is missing a closing bracket

jaadams5: @gausie 
Would you recheck, please?  On my system the only diff is in the string being passed to updateDisplay.

If failed has a an opening bracket, the else has the closing and then an opening and then there is a close before the return.

BadHorseMonkey: I think he means in the error text.

+          MafiaState.ERROR, "The requested repo (" + repo.getPath() + "**)** failed validation.");

jaadams5: See https://github.com/kolmafia/kolmafia/pull/73

I tend to use "bracket" for { or [ and "parenthesis" for ( so I was looking for the wrong thing.

PR #72, opened 2021-10-02
Fix release build not using correct commit

Rinn:


heeheehee-kolmafia:
(Per the removal of the ant build and the use of grgit, this will need some changes.)

build.gradle
```diff
@@ -195,15 +195,13 @@ task getRevision() {
 	onlyIf {
 		file('.git').exists()
 	}
+	def commit = System.getProperty("commit") ?: 'origin/main'
```
BadHorseMonkey, line 198: will this fail if origin is not a remote repository for this clone?

build.gradle
```diff
@@ -195,15 +195,13 @@ task getRevision() {
...
 	doLast {
-		def remote = System.getenv('HEAD_COMMIT_HASH')
-		if (remote == null) {
-			remote = 'origin/main'
-		}
-		def revision = grgit.log(includes:[remote]).size()
+		def revision = grgit.log(includes:[commit]).size()
 		file('build/revision.txt').text = revision.toString().trim()
 		// Update the version to the new revision
 		project.setVersion(revision.toString().trim())
```
BadHorseMonkey, line 207: would like to see this in the ./gradlew getRevision output.
Maybe add
    println 'Version :' + project.getVersion()

BadHorseMonkey (changes requested):
please make the log/output show the version. 

We should document the flag in the build output or help

Can you give me an example of how to test this?

I'm again not seeing the dirty bit set.  I was sure we'd corrected that, is it on a different PR?

Rinn:
> please make the log/output show the version.
> 
> We should document the flag in the build output or help
> 
> Can you give me an example of how to test this?
> 
> I'm again not seeing the dirty bit set. I was sure we'd corrected that, is it on a different PR?

It's already there. It's been there the whole time https://github.com/kolmafia/kolmafia/blob/f6d1103c18f14bb0048af0ef847bb2c353ba79c6/build.gradle#L180

build.gradle
Rinn: It was already using origin/main I didn't change that. If you think it's an issue please address it in a different pr.

Rinn: Set either the ORG_GRADLE_PROJECT_commit environment variable to a commit hash, or pass -Pcommit=HASH on the command line

heeheehee-kolmafia: To add a little more context:

We need some way of determining what the canonical branch is. We can't just use the tracked branch -- otherwise, say, this feature branch with 6 non-mainline commits would be r+6.

I experienced issues with this when @Rinn initially pushed the change to use git rev-list, but it was a simple enough change to rename the corresponding remote.

Rinn: yeah idk i'd need to think about it more, maybe it can just check if the current commit is tagged with a revision and if it's not consider it dirty.

BadHorseMonkey:
> > please make the log/output show the version.
> > We should document the flag in the build output or help
> > Can you give me an example of how to test this?
> > I'm again not seeing the dirty bit set. I was sure we'd corrected that, is it on a different PR?
> 
> It's already there. It's been there the whole time

I think I see the problem.

ANT tasks set the revision that the app uses to r12345M and also made the jar named KoLmafia-12345M.jar

gradle does not (for me) set the revision in the app uses for dirty versions but does make the jar named KoLmafia-12345-M.jar

I want to know in the console, the debug log, the about box, and at the top of the window if I'm on a modified version or not, which was functionality we used to have.

BadHorseMonkey:
> > > please make the log/output show the version.
> > > We should document the flag in the build output or help
> > > Can you give me an example of how to test this?
> > > I'm again not seeing the dirty bit set. I was sure we'd corrected that, is it on a different PR?
> > 
> > 
> > It's already there. It's been there the whole time
btw, if you want to go ahead with this PR and one of us can fix the other thing in another one, I'm fine with that.  I'm not trying to block progress with my request for changes, and I don't really know the etiquette of PR approval...

Rinn:
> > > It's already there. It's been there the whole time
> 
> btw, if you want to go ahead with this PR and one of us can fix the other thing in another one, I'm fine with that. I'm not trying to block progress with my request for changes, and I don't really know the etiquette of PR approval...

Yeah I'd rather deal with it separately since i need to think about it more.

BadHorseMonkey:
> > btw, if you want to go ahead with this PR and one of us can fix the other thing in another one, I'm fine with that. I'm not trying to block progress with my request for changes, and I don't really know the etiquette of PR approval...
> 
> Yeah I'd rather deal with it separately since i need to think about it more.

Fair.  I think it's bigger, since I think it's going to require a change to StaticEntity

BadHorseMonkey (approved):
approved, but please look at the M version as used by the code.

PR #77, opened 2021-10-03
Add -M modified flag to version number inside KoLmafia code (for title bar, logs, about, CLI version, etc)

BadHorseMonkey:
This needs to wait on @Rinn's PR #72 , and can certainly get cleaned up.

Tested with both dirty and clean git repos from the command line and IJ


BadHorseMonkey:
oh, good, there's the first bug.  It's double-adding the -M to the jar.

build.gradle
```diff
@@ -204,9 +204,15 @@ task getRevision() {
 			remote = 'origin/main'
 		}
 		def revision = grgit.log(includes:[remote]).size()
-		file('build/revision.txt').text = revision.toString().trim()
+		def revString = revision.toString().trim()
+		def dirty = versioning.info.dirty
+		if (dirty) {
+			// adding the dirty flag to the version
+			revString += "-M"
```
heeheehee-kolmafia, line 211: Can we instead add a new field to the manifest? That seems like it would be a lot cleaner, and express intent better than `isModified := getAttributes().getValue("Build-Revision").endsWith("-M")`.

BadHorseMonkey: sure.  I was following the SVN convention, but we can make better choices.

BadHorseMonkey:
Apologies for not fixing the description of the pushes.

Added the changes we discussed.

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -81,17 +81,33 @@ public static final String getVersion() {
     if (StaticEntity.cachedVersionName == null) {
       StringBuilder versionName =
           new StringBuilder(PRODUCT_NAME).append(" r").append(StaticEntity.getRevision());
+
+      if (isCodeModified()) {
+
```
heeheehee-kolmafia, line 86: Extraneous newline?

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -81,17 +81,33 @@ public static final String getVersion() {
...
+        versionName.append("-M");
+      }
+
       StaticEntity.cachedVersionName = versionName.toString();
     }
 
     return StaticEntity.cachedVersionName;
   }
 
+  public static final boolean isCodeModified() {
```
heeheehee-kolmafia, line 96: Does this need to be public? (Do we want it to be for some future use case?)

build.gradle
```diff
@@ -204,9 +205,11 @@ task getRevision() {
 			remote = 'origin/main'
 		}
 		def revision = grgit.log(includes:[remote]).size()
-		file('build/revision.txt').text = revision.toString().trim()
+		def revString = revision.toString().trim()
+		file('build/revision.txt').text = revString
 		// Update the version to the new revision
-		project.setVersion(revision.toString().trim())
+		project.setVersion(revString)
+		println("KoLmafia version :" + project.getVersion())
```
heeheehee-kolmafia, line 212: This prints as:

    KoLmafia version :12345-M

Is that intended?

heeheehee-kolmafia (dismissed):
Looks fine other than some small nits.

build.gradle
BadHorseMonkey: Hmm.  Probably could be another PR, but I generally want to be able to run "./gradlew getRevision and check my revision.  Right now you can call the function and get nothing.  This is also a carryover from ANT functionality.  Will put it back in my sandbox.

heeheehee-kolmafia:
Also note that this will mark a revision as clean if git thinks it's clean, not necessarily if it's clean compared to origin/main.

That may not be quite what we want (perhaps: "is clean and is also tagged with a revision").

build.gradle
heeheehee-kolmafia: Oh, my comment was more that it prints "version :12345-M" and not "version: 12345-M"

BadHorseMonkey: ah.  will fix.

src/net/sourceforge/kolmafia/StaticEntity.java
BadHorseMonkey: yep!  probably because I hacked the function up between commits.   will clean up.  Spotless didn't find that, btw.

BadHorseMonkey:
Yeah, I thought about that, mostly in the context of "how could we use Update4J?" and mostly there around how can we tell you that you're outdated but on a fork, so if we trigger an update, we'd want the user to know they were getting the main branch again.

That's all future stuff.  This is to make the app version match the jar version.

src/net/sourceforge/kolmafia/StaticEntity.java
BadHorseMonkey: nope, maybe, but it can change then.

heeheehee-kolmafia (dismissed):

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -81,17 +81,29 @@ public static final String getVersion() {
...
 
+  private static boolean isCodeModified() {
+    Attributes attributes = getAttributes();
+    if (attributes != null) {
+      String buildRevision = attributes.getValue("Build-Dirty");
+      return buildRevision.equals("true");
+    }
+    return false;
+  }
+
```
gausie, line 100: This is obviously completely optional but I would always write this like:

```suggestion
  private static boolean isCodeModified() {
    Attributes attributes = getAttributes();
    if (attributes == null) {
      return false;
    }
  
    return attributes.getValue("Build-Dirty").equals("true");
  }
```

in the spirit of return early and indent less

gausie (approved):

BadHorseMonkey:
Interesting!  My experiment to commit @gausie ’s suggestion from GitHub’s review page worked, but failed because it need to be spotless-scrubbed.

BadHorseMonkey:
ok, now we're in sync. 

I can tell I don't have the git model well internalized, because it took me about 12 hours to actually get the upstream merge to my fork, and I didn't do it in a very "clean" way.  And now I have no idea why it wants to upload test_charsheet.html, when it says "the files are identical".

gausie (approved):

PR #78, opened 2021-10-03
Migrate JUnit4 tests to JUnit5

heeheehee-kolmafia:
We have a number of JUnit 4 tests. We should move them to JUnit 5, which has improved assertions and better support for parameterized tests, among many other things.

Note that the number of commits looks much bigger / scarier than it should be, since this was based off the ant removal PR. b8aca2335677d8a90a9dfa898fbcfc3be19ae002 is the main change in question.


heeheehee-kolmafia:
There are apparently still some vestiges of JUnit 4 in the imports. Will want to clean that up, and resolve the failing checks...

heeheehee-kolmafia:
Hm. I have no idea why test_location_id.ash is failing, but _only for Mac_. Based on the test duration, I suspect what happened is that it actually timed out. Dunno why.

heeheehee-kolmafia:
And... converting CustomScriptTest mysteriously fixed the Mac failures. Either that or it was transient and resolved itself by the time I made that last change.

gausie (approved):
This is great!

PR #79, opened 2021-10-03
Upgrade swingx to 1.6.

heeheehee-kolmafia:
~This disables the auto-sorting tables that SwingX enables, since it's
incompatible with LockableListModel, I guess.~

~(This seems to work for me, at least with what was previously a
consistent reproducing example no longer reproduces with this change.)~

This removes the optional "force update" parameter of
LockableListModel.updateFilter, as it is error-prone and can cause listeners
to desynchronize from the actual list. This is especially evident due to
changes in SwingX 1.0 -> 1.6 implementation details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
 	runtimeOnly 'net.java.dev.jna:jna-platform:5.9.0'
 	runtimeOnly 'org.antlr:antlr-runtime:3.5.2'
 
-	libImplementation 'org.swinglabs:swingx:1.6'
+	libImplementation 'org.swinglabs:swingx:1.6.1'
 
 	implementation files('build/lib') {
 		builtBy 'compileLibJava'
@@ -80,7 +80,7 @@ dependencies {
 	implementation 'org.fusesource.jansi:jansi:2.3.4'
 	implementation 'org.json:json:20210307'
 	implementation 'org.mozilla:rhino:1.7.13'
-	implementation 'org.swinglabs:swingx:1.6'
+	implementation 'org.swinglabs:swingx:1.6.1'
 	implementation 'org.tmatesoft.svnkit:svnkit:1.10.3'
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -70,7 +70,7 @@ dependencies {
 	runtimeOnly 'net.java.dev.jna:jna-platform:5.9.0'
 	runtimeOnly 'org.antlr:antlr-runtime:3.5.2'
 
-	libImplementation 'org.swinglabs:swingx:1.0'
+	libImplementation 'org.swinglabs:swingx:1.6'
 
 	implementation files('build/lib') {
 		builtBy 'compileLibJava'
@@ -80,10 +80,8 @@ dependencies {
 	implementation 'org.fusesource.jansi:jansi:2.3.4'
 	implementation 'org.json:json:20210307'
 	implementation 'org.mozilla:rhino:1.7.13'
-	implementation 'org.swinglabs:swingx:1.0'
+	implementation 'org.swinglabs:swingx:1.6'
 	implementation 'org.tmatesoft.svnkit:svnkit:1.10.3'
-
-	toCopy 'org.swinglabs:swingx:1.0'
 }
 
 application {
@@ -184,11 +182,6 @@ shadowJar {
 	duplicatesStrategy = 'exclude'
 	destinationDirectory = file('dist/')
 	archiveClassifier.set("${versioning.info.dirty ? 'M' : ''}")
-}
-
-task download(type: Copy) {
-	from configurations.toCopy
-	into 'src/jar'
 }
 
 task getRevision() {
--- a/build.gradle
+++ b/build.gradle
@@ -44,11 +44,6 @@ repositories {
 	mavenCentral()
 }
 
-configurations {
-	toCopy
-	implementation.canBeResolved = true
-}
-
 dependencies {
 	// Use JUnit Jupiter for running JUnit5 tests.
 	testRuntimeOnly 'org.junit.platform:junit-platform-launcher:1.8.0'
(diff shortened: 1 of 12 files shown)


heeheehee-kolmafia:
My consistent reproducing test case prior to this change: log in, note that Item Manager frame isn't present.

I also checked that it still reproduces if I revert the changes in ShowDescriptionTable.

heeheehee-kolmafia:
Hm. Looks like that actually broke table sorting entirely.

I wonder what else needs fixing.

heeheehee-kolmafia:
Hm. So, I still think the problem is a startup race. We use AbstractTableModel with a RowSorter just fine with CandyTable, and also I don't reproduce the issue if I load the frames before I log in (which populates inventory).

heeheehee-kolmafia:
More experimenting indicates that changing AutoFilterTextField to use LockableListModel.updateFilter(true) instead of false (which I suspect was an optimization of sorts) fixes the initially observed panel load failure.

There are many other instances of LockableListModel.updateFilter(false), and I'm not sure which if any may trigger this same style of issue.

build.gradle
```diff
@@ -70,7 +70,7 @@ dependencies {
 	runtimeOnly 'net.java.dev.jna:jna-platform:5.9.0'
 	runtimeOnly 'org.antlr:antlr-runtime:3.5.2'
 
-	libImplementation 'org.swinglabs:swingx:1.0'
+	libImplementation 'org.swinglabs:swingx:1.6'
```
BadHorseMonkey, line 74: why not 1.6.1?

build.gradle
```diff
@@ -80,10 +80,8 @@ dependencies {
 	implementation 'org.fusesource.jansi:jansi:2.3.4'
 	implementation 'org.json:json:20210307'
 	implementation 'org.mozilla:rhino:1.7.13'
-	implementation 'org.swinglabs:swingx:1.0'
+	implementation 'org.swinglabs:swingx:1.6'
 	implementation 'org.tmatesoft.svnkit:svnkit:1.10.3'
```
BadHorseMonkey, line 84: also why not 1.6.1?

build.gradle
```diff
@@ -80,10 +80,8 @@ dependencies {
...
-
-	toCopy 'org.swinglabs:swingx:1.0'
```
BadHorseMonkey, line 86: Do we no longer need to copy it?

BadHorseMonkey (approved):
works well for me on MacOSX Monterrey.  no crashes.  
up to you if you want to wait for a reviewer on another platform.

build.gradle
heeheehee-kolmafia: This target was always intended for downloading jars into the src/jar directory.

Since Gradle is bundling our jars for us (as opposed to Ant grabbing all the jars in a directory), we no longer need this target.

build.gradle
heeheehee-kolmafia: Sure, why not. Done.

heeheehee-kolmafia:
@jaadams5 Mind taking this for a spin on Windows? I don't expect any problems, but confirmation would be good to have.

jaadams5:
Missed this but I will try and check on Windows.  The practice of working with someone else's branch will do me good.  I assume a local build and then trying to interact with a lot of menus will be sufficent?

heeheehee-kolmafia:
Yes, that should be good. It's a separate branch on the same remote repo, so I'd expect it to be well-supported by GUIs.

jaadams5:
I think I tested on Windows 11.  I convinced some GUI tool (and I didn't take notes) to fetch this branch.  I built via shadowjar and my build.gradle referenced 1.6 and 1.6.1.  I ran the code bringing up windows and sorting some columns and nothing unexpected occurred.

src/net/sourceforge/kolmafia/swingui/widget/AutoFilterTextField.java
```diff
@@ -223,73 +223,66 @@ public static final int getResultQuantity(final Object element) {
...
-          && AutoFilterTextField.this.text.charAt(0) == "-".charAt(0)) {
-        AutoFilterTextField.this.notChecked = true;
-        AutoFilterTextField.this.text = AutoFilterTextField.this.text.substring(1).trim();
-      }
+    AutoFilterTextField.this.qtyChecked = false;
+    AutoFilterTextField.this.asChecked = false;
+    AutoFilterTextField.this.notChecked = false;
+    AutoFilterTextField.this.text = AutoFilterTextField.this.getText().toLowerCase();
+
+    if (AutoFilterTextField.this.text != null
```
jaadams5, line 231: IntelliJ says AutoFilterTextField.this.text != null is always true.

But if I am reading right AutoFilterTextField.this.getText() can return null.  So it is either always true or we have previously had some kind of error caused by .toLowerCase() on a null value, right?

jaadams5 (approved):
I'm fine with the changes.  Someone could look at my comment on AutoFilterTextField but I am not sure enough that I am actually correct to insist on that.

src/net/sourceforge/kolmafia/swingui/widget/AutoFilterTextField.java
heeheehee-kolmafia: If getText() returns null, then toLowerCase() will result in a NPE, yeah.

src/net/sourceforge/kolmafia/swingui/widget/AutoFilterTextField.java
```diff
@@ -223,73 +223,66 @@ public static final int getResultQuantity(final Object element) {
...
 
-        String op = mas.group(1);
+      String op = mas.group(1);
 
-        AutoFilterTextField.this.asEQ = op.indexOf("=") != -1;
-        AutoFilterTextField.this.asLT = op.indexOf("<") != -1;
-        AutoFilterTextField.this.asGT = op.indexOf(">") != -1;
-        AutoFilterTextField.this.text = mas.replaceFirst("");
-      }
+      AutoFilterTextField.this.asEQ = op.indexOf("=") != -1;
```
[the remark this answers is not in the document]
heeheehee-kolmafia: Ah, I see you already committed the String.contains change. Thanks!

heeheehee-kolmafia:
I'll merge this as-is. We can do other cleanups in a followup PR, since I'd rather not conflate too many things, especially since this has the potential to cause another breakage, so it might need to be rolled back again. I don't think it will, but I've been wrong before.

PR #80, opened 2021-10-03
Experimental: Check github for update on launch once a day (set checkRssUpdate=true)

Rinn:
Queries the GitHub api for the current release and prompts if a newer release is available. Basically the same style update check as the sourceforge point release update check, but behind a setting now. Only prompts to update on launch once a day.

Possible improvements for additional PRs
- Add checkRssUpdate option to preferences window
- Add manual update check that bypasses settings \ daily limit
- Query and show change log in app
- Download jar directly \ auto-relaunch?
- Disable update check if mafia is built locally

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -17,6 +17,7 @@ global	chatBeep	false
 global	chatFontSize
 global	chatLinksUseRelay	false
 global	chatStyle	0
+global	checkRssUpdate	false
 global	cloverProtectActive	true
 global	coinMasterIndex	1
 global	combatHotkey0
--- a/src/net/sourceforge/kolmafia/KoLmafia.java
+++ b/src/net/sourceforge/kolmafia/KoLmafia.java
@@ -6,6 +6,7 @@ import static net.sourceforge.kolmafia.KoLGUIConstants.FLATMAP_LIGHT_LOOKS;
 import java.awt.Color;
 import java.awt.Frame;
 import java.awt.Taskbar;
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintStream;
@@ -15,6 +16,7 @@ import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
 import java.util.List;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
 import javax.swing.JEditorPane;
 import javax.swing.JFrame;
 import javax.swing.SwingUtilities;
@@ -105,11 +107,16 @@ import net.sourceforge.kolmafia.swingui.SystemTrayFrame;
 import net.sourceforge.kolmafia.swingui.listener.LicenseDisplayListener;
 import net.sourceforge.kolmafia.swingui.panel.GenericPanel;
 import net.sourceforge.kolmafia.textui.AshRuntime;
+import net.sourceforge.kolmafia.utilities.FileUtilities;
+import net.sourceforge.kolmafia.utilities.InputFieldUtilities;
 import net.sourceforge.kolmafia.utilities.LockableListFactory;
 import net.sourceforge.kolmafia.utilities.LogStream;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
 import net.sourceforge.kolmafia.utilities.SwinglessUIUtils;
+import net.sourceforge.kolmafia.webui.RelayLoader;
 import net.sourceforge.kolmafia.webui.RelayServer;
+import org.json.JSONException;
+import org.json.JSONObject;
 
 public abstract class KoLmafia {
   private static boolean isRefreshing = false;
@@ -259,7 +266,8 @@ public abstract class KoLmafia {
       KoLConstants.saveStateNames.add(actualName);
     }
 
-    // Set a user agent preemptively.  Workaround to allow https support for file_to_map and price
+    // Set a user agent preemptively. Workaround to allow https support for
+    // file_to_map and price
     // updates to coexist.
     GenericRequest.setUserAgent();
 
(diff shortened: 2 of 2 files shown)


src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -660,7 +670,8 @@ public static void refreshSession() {
 
     boolean shouldResetCounters = false;
     boolean shouldResetGlobalCounters = false;
-    // Assume if rollover has changed by an hour, it is a new rollover. Time varies slightly between
+    // Assume if rollover has changed by an hour, it is a new rollover. Time varies
+    // slightly between
     // servers by a few seconds.
```
heeheehee-kolmafia, line 675: Can you reflow this comment? Now you have two short lines next to each other.

ditto below. Do you have your column limit set to 80 chars? 100 is the standard in Google Java Style.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1861,15 +1874,79 @@ public static String whoisPlayer(final String player) {
...
+      if (now - lastUpdate < 86400000L) {
+        return;
+      }
+
+      String contents = null;
+
+      try {
+        BufferedReader reader =
+            FileUtilities.getReader(
+                "https://api.github.com/repos/kolmafia/kolmafia/releases/latest");
```
heeheehee-kolmafia, line 1901: Consider reframing this as a try-with-resources statement.

This has the potential to leak resources if it throws an exception before getting to reader.close().

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1861,15 +1874,79 @@ public static String whoisPlayer(final String player) {
...
+      Preferences.setString("lastRssVersion", currentVersion);
+      Preferences.setString("lastRssUpdate", Long.toString(now));
+
+      if (currentVersionInt <= StaticEntity.getRevision() || currentVersion.equals(lastVersion)) {
+        return;
+      }
+
+      if (InputFieldUtilities.confirm(
+          String.format(
+              "A new version KoLmafia r%s is now available. Would you like to download it now?",
```
heeheehee-kolmafia, line 1946: formatting nit: perhaps "new version of KoLmafia (r%s) is now available"?  This reads a bit awkwardly to me as it is right now.

(Maybe that's just me)

src/data/defaults.txt
```diff
@@ -17,6 +17,7 @@ global	chatBeep	false
 global	chatFontSize
 global	chatLinksUseRelay	false
 global	chatStyle	0
+global	checkRssUpdate	false
```
heeheehee-kolmafia, line 20: Do you also need to set a default value for lastRssUpdate?

I'm worried that Long.parseLong("") will throw an exception.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1861,15 +1874,79 @@ public static String whoisPlayer(final String player) {
...
   private static class UpdateCheckRunnable implements Runnable {
     public void run() {
-      // TODO: Check for new version on jenkins\github after migration is complete. See revision
-      // history for old release update check.
+      boolean checkUpdate = Preferences.getBoolean("checkRssUpdate");
+      if (!checkUpdate) {
+        return;
+      }
+
+      long lastUpdate = Long.parseLong(Preferences.getString("lastRssUpdate"));
```
heeheehee-kolmafia, line 1890: Why not just Preferences.getLong?

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1861,15 +1874,79 @@ public static String whoisPlayer(final String player) {
...
+      try {
+        currentVersionInt = Integer.parseInt(currentVersion) + 1;
+      } catch (NumberFormatException e) {
+        System.out.println(e.getMessage() + " while parsing version number");
+        return;
+      }
+
+      String lastVersion = Preferences.getString("lastRssVersion");
+      Preferences.setString("lastRssVersion", currentVersion);
+      Preferences.setString("lastRssUpdate", Long.toString(now));
```
heeheehee-kolmafia, line 1938: similarly, Preferences.setLong?

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1861,15 +1874,79 @@ public static String whoisPlayer(final String player) {
...
+
+      if (currentVersion == null) {
+        return;
+      }
+
+      int currentVersionInt;
+      try {
+        currentVersionInt = Integer.parseInt(currentVersion) + 1;
+      } catch (NumberFormatException e) {
+        System.out.println(e.getMessage() + " while parsing version number");
```
heeheehee-kolmafia, line 1932: Perhaps also print the current version in the error message?

e.g. System.out.println(e.getMessage() + " while parsing version number ' " + currentVersion + " ' ");

(extraneous spaces added only to make the single-quotes easier to see)

BadHorseMonkey:
If this wasn’t small and a return to pre-watershed parity, I’d say “just implement update4J”. 

But it isn’t a lot of code and it may even lead the way to that or a similar tool.

src/net/sourceforge/kolmafia/KoLmafia.java
Rinn: oh i think this was an auto format in my editor, i'll revert them

src/net/sourceforge/kolmafia/KoLmafia.java
Rinn: this is an existing preference, I didn't want to change it from a string

src/net/sourceforge/kolmafia/KoLmafia.java
Rinn: existing preference, didn't want to change the type

src/data/defaults.txt
Rinn: It already does, it's an existing preference and it defaults to 0

Rinn:
> If this wasn’t small and a return to pre-watershed parity, I’d say “just implement update4J”.
> 
> But it isn’t a lot of code and it may even lead the way to that or a similar tool.

Oh okay, let's just do that then.

PR #81, opened 2021-10-03
Numeric cast

jaadams5:
Remove cast flagged by security scanning.   Some code cleanup suggested by inspection.  spotlessApply

### DIFF
--- a/src/net/sourceforge/kolmafia/request/BasementRequest.java
+++ b/src/net/sourceforge/kolmafia/request/BasementRequest.java
@@ -1192,8 +1192,8 @@ public class BasementRequest extends AdventureRequest {
     return (long) BasementRequest.basementTestCurrent;
   }
 
-  public static long getBasementTestValue() {
-    return (long) BasementRequest.basementTestValue;
+  public static  double getBasementTestValue() {
+    return BasementRequest.basementTestValue;
   }
 
   public static int getActualStatNeeded() {
--- a/src/net/sourceforge/kolmafia/request/BasementRequest.java
+++ b/src/net/sourceforge/kolmafia/request/BasementRequest.java
@@ -1184,8 +1184,8 @@ public class BasementRequest extends AdventureRequest {
     return (long) BasementRequest.basementTestCurrent;
   }
 
-  public static double getBasementTestValue() {
-    return BasementRequest.basementTestValue;
+  public static long getBasementTestValue() {
+    return Double.valueOf(BasementRequest.basementTestValue).longValue();
   }
 
   public static int getActualStatNeeded() {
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/request/BasementRequest.java
```diff
@@ -1192,8 +1184,8 @@ public static long getBasementTestCurrent() {
     return (long) BasementRequest.basementTestCurrent;
   }
 
-  public static long getBasementTestValue() {
-    return (long) BasementRequest.basementTestValue;
```
jaadams5, line 1196: This is the cast that the security scanning didn't like.  This function is only called in BasementDecorator and the result is only used in a String so I am not expecting the change to create an issue.

heeheehee-kolmafia: Is the security scanning happy with instead using Double.longValue()? Or Math.round()?

I mention this because printing a double can be much more verbose.

jaadams5: I don't know how to run the scanning locally.  If it is easy I will try.  Otherwise I'm fine with trying Double.longValue().

jaadams5: And the change was made.

src/net/sourceforge/kolmafia/request/BasementRequest.java
```diff
@@ -526,7 +520,7 @@ private static boolean checkForElementalTest(boolean autoSwitch, final String re
     }
 
     BasementRequest.changeBasementOutfit("element");
-    BasementRequest.canHandleElementTest(autoSwitch, true);
```
heeheehee-kolmafia, line 529: Was this supposed to change?

heeheehee-kolmafia (approved):

src/net/sourceforge/kolmafia/request/BasementRequest.java
jaadams5: Yes, as suggested by lint.  

In context

    if (!autoSwitch) {
      return true;
    }

    BasementRequest.changeBasementOutfit("element");
    BasementRequest.canHandleElementTest(true, true);
    return true;
    
 so the changed line will never be executed unless autoSwitch is true.  I couldn't make a case that the change reduced comprehension so I accepted it.

heeheehee-kolmafia:
(If I had any more requests, it would be for tests, which should be pretty straightforward with the RequestTestBase I added. Just inject a response that matches a stat check at a certain floor, and then check getBasementTestValue() and related fields. Bonus points for making it parameterized to check a bunch of different stats.)

jaadams5:
I thought about tests but have not really wrapped my mind around RequestBase and how to use it, yet.  But coverage++ remains my default task so.... :-)

heeheehee-kolmafia:
I started tinkering a bit with some tests in https://github.com/heeheehee-kolmafia/kolmafia/tree/basement but still need to figure out why they're so slow (5s per stat test, yuck).

PR #82, opened 2021-10-04
Warn desktops who cannot launch a browser

gausie:
This really just applies to very old Linux installations... but mafia users tend to do things like use very old Linux installations


BadHorseMonkey:
Haven’t tried that, but that doesn’t pop up for headless users, does it?

Rinn (approved):

PR #84, opened 2021-10-04
Add tests for BasementRequest.

heeheehee-kolmafia:
Another proof-of-concept of parameterized JUnit5 tests with injecting response data.

This also incrementally bumps the required version of Java to 11 (the next LTS release) because I wanted to use var. This should be a non-issue for users who had to upgrade earlier, since Java 9 is unsupported, and we strongly recommended that users switch to Java 17. (But also, if this is a dealbreaker, I can spell out a few redundant types...)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,8 +58,9 @@ dependencies {
 
 	testImplementation 'org.hamcrest:hamcrest:2.2'
 	testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.0'
+	testImplementation 'org.junit.jupiter:junit-jupiter-params:5.8.0'
 	testImplementation 'junit:junit:4.13.2'
-	testImplementation 'org.mockito:mockito-core:3.+'
+	testImplementation 'org.mockito:mockito-inline:3.+'
 
 	implementation 'com.formdev:flatlaf:1.6'
 	implementation 'com.formdev:flatlaf-intellij-themes:1.6'
--- a/gradle.properties
+++ b/gradle.properties
@@ -6,5 +6,5 @@ org.gradle.jvmargs=--add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAME
   --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED \
   --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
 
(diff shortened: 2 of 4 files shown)


heeheehee-kolmafia:
@jaadams5 as a fyi

(this switches the dep of mockito to inline so I could mock some static methods.)

heeheehee-kolmafia:
Hm. I thought this fixed the performance issue, but apparently not.

heeheehee-kolmafia:
So, it turns out the performance issue is ~entirely due to mockStatic(Preferences.class). I guess I can do something wherein I inject a Preferences object, and then override it...

gausie:
Full support of the version bump. I think we should consider removing references to minimum version, it's confusing and a mixed message for users

jaadams5:
I am perfectly fine with mandating a Java version for kolmafia/staff, staff wannabees and wildcards who build their own.  If they are smart enough to do that then they should be smart enough to ask for help if they can't figure something out.   Shall we just float the idea of going tom 17?

I would prefer to have a less stringent requirement for "normal" users and if the only thing driving 11 is testing we should be able to avoid inflicting that on them.

This may be the incentive we need to bundle the JRE we want users to use, with our jar.  The issues would be an installer or runner that did not any Java on the target system and a process that didn't need escalated privs.

heeheehee-kolmafia:
I'm okay with holding off on this change until we have bundling / native binaries in place. I figured 11 was a reasonable stepping stone, since it's at least been out for a few years.

At this time, in my opinion, the primary driver of a move to a newer version of Java is cleaner, safer code (in the form of more modern features that make it easier to write clear code). Java 11 also brings TLS 1.3 support and a new HttpClient, but between then and Java 17, I don't know if there are many library changes that I really want (as was the case with java.awt.Taskbar in Java 9).

BadHorseMonkey:
I don’t think 11 has switch expressions, which will make it possible to replace some tortured  code with something more readable.  But it’s not a must-have.

jaadams5 (approved):
OK by me.

heeheehee-kolmafia:
Are we holding off on this until bundling is in place? I can poke at that over the next couple of days.

jaadams5:
Not my call.  I'm always in favor of more test examples but I might be at the point (in git learning) where I could just juggle branches locally.

PR #85, opened 2021-10-04
Delete svnrepo.json if it's bad.

heeheehee-kolmafia:
If the file throws a JSONException when we try to parse it, it's most
likely not valid JSON. So, let's delete it.

Anything else that we should do here? At the very least, this should get users out of a bad state.

Maybe we should also make sure that the Content-Type isn't text/html?

We already reject file downloads that aren't 200s.


jaadams5 (approved):
Looks fine to me.

PR #86, opened 2021-10-05
Create OS-specific builds via jpackage

heeheehee-kolmafia:
This generates Linux, macOS, and Windows builds. It's admittedly hard to show this in detail, but you can look at the historical intermediate actions that I triggered when developing this.

For instance: https://github.com/heeheehee-kolmafia/kolmafia/runs/3799238426?check_suite_focus=true

(Publish artifacts -> List all downloaded artifacts)

These are all bundled as zipballs as the lowest common denominator, since we need to bundle the jars alongside the Java runtime and potentially any other dependencies.


build.gradle
```diff
@@ -252,6 +252,33 @@ tasks.withType(JavaCompile) {
...
+	}
+	delete 'build/releases/KoLmafia'
+}
+
+task bin(type: Exec) {
+	dependsOn shadowJar
+	outputs.files file('build/releases/')
+
+	executable 'jpackage'
+	doFirst {
```
BadHorseMonkey, line 267: Would recommend a gradle plugin instead of using exec (even if it is just calling the exec).
e.g. : https://github.com/petr-panteleyev/jpackage-gradle-plugin or https://github.com/xcporter/jpkg

heeheehee-kolmafia: Done.

BadHorseMonkey:
I added the mac/windows/linux sections and appVersion which is required on MacOS at least.  Also the mac icns.  We probably need a windows icon for it as well.

heeheehee-kolmafia:
Any need for signing on the Mac side?

I'm also not sure what the resulting files / zipballs will look like on other distributions if we're not using the APP_IMAGE approach. We might have to update the github actions (although if they really are standalone files, we may be able to upload just the binaries as opposed to a zip that users then have to extract).

Neither the deb / rpm targets wanted to build for me, despite being on Linux. Probably because I'm not on a distro that uses either of those for packaging (and I don't have the corresponding tools installed).

BadHorseMonkey:
It runs for me without signing, but I'm not running a typical environment.  We probably need to have this vetted on a standard user account.  I may need to borrow my wife's mac. :)

[AppImage](https://en.wikipedia.org/wiki/AppImage) is a linux only thing.  It created an unusable folder for me.

The mac can make a DMG (disk image, which was how we did it in the Java 6 days) or a pkg, which is an installer.  both are single files.

The windows options are .exe (just the raw executable) or .msi, which is the windows MicroSoft Installer format.  both are also single files.

BadHorseMonkey (approved):
works correctly for me on my mac, and it looks like it should work for windows (and you've tested it for your distro of unix).  I can't test the build.yml, but I didn't see anything that looked crazy there.

If this is experimental and publishes *beside* or existing builds, than we can recruit testers after it's on the site.

PR #87, opened 2021-10-05
Add initial pass of CONTRIBUTING.md.

heeheehee-kolmafia:
There's a lot more room for expansion, but I think this is a good
start.

Draft can be seen [here](https://github.com/kolmafia/kolmafia/blob/c3bb2af48b86a5c92b50a1957f6688c738a1d901/CONTRIBUTING.md).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -65,6 +65,9 @@ Some tasks of interest:
 We currently require at least Java 9, but strongly recommend building
 and running via the latest LTS release (Java 17 as of October 2021).
 
+See [this thread](https://kolmafia.us/threads/rocking-the-gradle.26583/) for
+some additional notes on Gradle usage.
+
 ### Testing
 
 We use [JUnit 5](https://junit.org/junit5/docs/current/user-guide/) for all of
@@ -80,6 +83,9 @@ We provide CustomScriptTest.java if you wish to write ASH / CLI / JS scripts in
 test/root/scripts. This wrapper compares the output of these scripts against
 golden files in test/root/expected. Tests of this form must clean up after
 themselves, to avoid polluting the shared environment.
+
+See [this thread](https://kolmafia.us/threads/scripts-in-tests.26380/) for a
+discussion on setting up the environment for these custom script tests.
 
 ### Code style
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,13 +31,13 @@ members of the dev team tend to have their proposed patches ignored.
 The ability to determine the appropriateness of a patch and the level of support
 for it is extremely useful. A patch that violates the spirit of KoLmafia (for
 example, that allows access to non-KoLmafia directories in the user's file
-system) will be rejected. A patch that changes an existing feature would need
+system) will be rejected. A patch that changes an existing feature will need
 some discussion. A patch that changes a lot of code to implement a feature of at
 best niche interest is unlikely to be accepted.
 
 The technical skills are the ability to write and test Java code that works in
 the context of KoLmafia. A prerequisite for success is the ability to read and
-understand KoLmafia's code and infer its archetecture since those are not well
+understand KoLmafia's code and infer its architecture since those are not well
 documented in the traditional sense.
 
 Individuals with a good track record of creating and submitting patches can ask
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,15 +3,17 @@
 First off, thank you for contributing your time and knowledge. This project
 would not exist if not for volunteers like you!
 
-## Some useful prerequisites (lifted from [Github docs](https://github.com/github/docs/blob/main/CONTRIBUTING.md))
+## Workflow Overview
+
+Some suggested reading (lifted from [Github docs](https://github.com/github/docs/blob/main/CONTRIBUTING.md)):
 
 - [Set up Git](https://docs.github.com/en/get-started/quickstart/set-up-git)
 - [GitHub flow](https://docs.github.com/en/get-started/quickstart/github-flow)
 - [Collaborating with pull requests](https://docs.github.com/en/github/collaborating-with-pull-requests)
 
-All work must be done in feature branches (typically within your fork of the
-project), and reviewed by at least one of the core developers through Github's
-Pull Request feature.
+In the spirit of Github flow, all work must be done in feature branches
+(typically within your fork of the project), and reviewed by at least one of the
+core developers through Github's Pull Request (PR) feature.
 
 We strongly encourage that you provide tests as part of your pull request, as
 this helps us understand the problem at hand and avoid breaking your change in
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+this helps us understand the problem at hand and avoid breaking your change in
+the future.
+
+## Guidelines (adapted from [our previous guide](https://kolmafia.us/threads/25364))
+
+(Editor's note: you can mentally replace all instances of "patch" with
+"pull request".)
+
+A successful submit requires social and technical skills.
+
```
jaadams5, line 26: I would not complain if this were tweaked.  It was written at a time when the social aspects were prominent - at least one dev (me) was ignoring patches from people who were being rude, annoying and acting as if the fact that they did a little work entitled them to have that work incorporated into the repository.  I think the GitHub workflow with reviews is a good step towards reducing the issues caused by personalities.  I would like to encourage discussions "should we do this?  is this a good idea?" before someone writes code and submits a pull request but we could probably rework this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+some discussion. A patch that changes a lot of code to implement a feature of at
+best niche interest is unlikely to be accepted.
+
+The technical skills are the ability to write and test Java code that works in
+the context of KoLmafia. A prerequisite for success is the ability to read and
+understand KoLmafia's code and infer its architecture since those are not well
+documented in the traditional sense.
+
+Individuals with a good track record of creating and submitting patches can ask
+to be made a dev. The active devs who can do this are MCroft and gausie.
```
jaadams5, line 44: Is this still true and what we want or how we expect to do things?  Can non-staff contribute to a review even if they cannot approve the PR?  I'd like the answer to be yes and part of the audition to become staff would include constructive participation in reviews.

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
+# Contributing to KoLmafia
+
+First off, thank you for contributing your time and knowledge. This project
+would not exist if not for volunteers like you!
+
+## Some useful prerequisites (lifted from [Github docs](https://github.com/github/docs/blob/main/CONTRIBUTING.md))
+
```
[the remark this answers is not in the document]
heeheehee-kolmafia: Done. I renamed the section, and added a comment regarding "in the spirit of github flow, we do things this way" Let me know if that's any better!

CONTRIBUTING.md
heeheehee-kolmafia: I've rewritten that initial paragraph to incorporate a lot of the things you mentioned. Thoughts?

CONTRIBUTING.md
heeheehee-kolmafia: > Can non-staff contribute to a review even if they cannot approve the PR? I'd like the answer to be yes and part of the audition to become staff would include constructive participation in reviews.

Yes, agreed on both fronts. I also added a vague comment about how we may initiate contact, since IIRC that's how I became a dev.

What I don't want is to encourage a system of basically mini-mods where non-devs go around nitpicking others' PRs. For now "constructively engaging" seems good enough, but we can revisit if it becomes an issue.

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+the future.
+
+## Guidelines (adapted from [our previous guide](https://kolmafia.us/threads/25364))
+
+Before creating a PR, we strongly recommend creating a bug in the [KoLmafia
+forums](https://kolmafia.us/forums/bug-reports.24/), in order to provide broader
+context regarding the change, and to discuss whether the change is a good
+idea. We'd hate for you to waste effort and be discouraged from submitting
+future PRs because some fundamental concerns arise during code review.
+
```
BadHorseMonkey, line 29: Can we link to the “how to write a good bug report?” In this paragraph?

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+:----------------:|----------------------------------------------------
+ tasks            | List all Gradle tasks.
+ runShadow        | Build and run the application.
+ shadowJar        | Create a new jar in dist/.
+ spotlessApply    | Apply formatting changes.
+ check            | Run all checks, including formatting and tests.
+ jacocoTestReport | Generate a coverage report in build/reports/jacoco.
+
+We currently require at least Java 9, but strongly recommend building
+and running via the latest LTS release (Java 17 as of October 2021).
```
BadHorseMonkey, line 66: Some tasks, like jpackage, require Java14.   Should we mention that or just wait until that lands and update this doc?

CONTRIBUTING.md
```diff
@@ -0,0 +1,99 @@
...
+best niche interest is unlikely to be accepted.
+
+The technical skills are the ability to write and test Java code that works in
+the context of KoLmafia. A prerequisite for success is the ability to read and
+understand KoLmafia's code and infer its architecture since those are not well
+documented in the traditional sense.
+
+Individuals with a good track record of creating PRs and constructively engaging
+in other pull requests can ask to be made a dev. The active devs who can do this
+are MCroft and gausie. We may also reach out to you.
```
BadHorseMonkey, line 44: Is it still just us two?

CONTRIBUTING.md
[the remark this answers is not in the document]
BadHorseMonkey: Do we want to link to the (not currently up to date) build and IDE setup guides on the wiki?  Should there be general links to the wiki in the suggested reading section?

BadHorseMonkey (approved):

BadHorseMonkey:
I approved it, but if you want to tweak it, I’ll re-review.

CONTRIBUTING.md
heeheehee-kolmafia: Depends on the definition. For expanding commit access, I think it's "whoever's listed as an owner in the github organization" which also includes Rinn and Veracity.

CONTRIBUTING.md
heeheehee-kolmafia: Well, it's not necessary to run the jpackage task for general contributors, unless you're working on that feature.

CONTRIBUTING.md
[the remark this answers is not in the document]
heeheehee-kolmafia: I guess my question on this front is "what do we see as the purpose of the wiki?" I view it primarily as a user guide, not as a developer guide. It could be both, but I think it makes sense to put dev-only content where devs are already looking.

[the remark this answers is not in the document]
jaadams5: If someone needs help in setting things up and we want to preserve the interaction so we don't have to repeat ourselves for some visitor from the future, I think there should be some minimal guidance on KoLmafia.us.  We do have at least one person who wants to setup and build locally but has no desire to share what they are doing so that conversation might fit better there then here.  If we can clearly identify something of benefit/interest only to devs then it belongs here.

CONTRIBUTING.md
jaadams5: Do we want to link to that here?  WE aren't reporting bugs on GitHub and reports at KoLmafia are driving activities.

jaadams5:
We can probably tweak this some more but I'm fine with it as it stands.  We can wait until someone else reads it and asks questions :-)

jaadams5 (approved):
OK by me, at least for round 1.

CONTRIBUTING.md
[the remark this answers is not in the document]
heeheehee-kolmafia: Hey, if you've already downloaded the repo, you'll have a copy of CONTRIBUTING.md locally :)

heeheehee-kolmafia:
I'll commit this; we can discuss further changes in a followup.

PR #90, opened 2021-10-06
Replace line endings in test/ file with LF.

heeheehee-kolmafia:
I guess this file snuck in before we finally added coverage for the
directory in .gitattributes.

It bugs me every now and then when I switch branches and find
this file needs attention.


BadHorseMonkey (approved):
thank you!

PR #91, opened 2021-10-06
Reject risky character names.

heeheehee-kolmafia:
We don't expect to ever receive character names containing '.', '/',
or '\', but if we do, we shouldn't accept them, as this introduces a
risk of directory traversal which can allow attackers to break out of
$KOLMAFIA_ROOT.

(This attempts to address some/most of the identified security risks.)


gausie (approved):
Nice

PR #93, opened 2021-10-06
Directly upload Mac / Windows executables.

heeheehee-kolmafia:
Currently, we're generating macOS.zip, Windows.zip which each only contain a single file. We can upload them directly in each release.


.github/workflows/build.yml
```diff
@@ -105,6 +109,8 @@ jobs:
           files: |
             dist/*.jar
             releases/*.zip
+            releases/*.exe
+            releases/*.pkg
           tag_name: r${{ env.KOLMAFIA_VERSION }}
           name: ${{ env.KOLMAFIA_VERSION }}
```
gausie, line 115: Can we add an "r" prefix here too? Or even "Revision " or "Release "

build.gradle
```diff
@@ -265,14 +265,15 @@ tasks.jpackage {
...
+		icon = 'util/linux/KoLmafia.ico'
 	}
 	mac {
 		type = 'pkg'
 		icon = 'util/macosx/limeglass.icns'
 	}
 	windows{
 		type = 'exe'
-		icon = 'util/windows/KoLmelion.ico'
+		icon = 'util/windows/KoLmafia.ico'
```
gausie, line 276: rip

gausie (approved):
Looks good. We should be able to make an `.AppImage` for the linux release too though

gausie:
Ah I see what's going on with Linux. Look at https://bugs.openjdk.java.net/browse/JDK-8225428 - "APP_IMAGE" doesnt exist, so we're actually just seeing fallthrough behaviour. Does that mean "deb" works?

build.gradle
```diff
@@ -265,14 +265,15 @@ tasks.jpackage {
 	appVersion = new Date().format("yy.MM")
```
gausie, line 265: Don't we want to use our rev number here?

heeheehee-kolmafia:
APP_IMAGE works for me locally, as well as when I download and unzip the latest release. I just have to manually run ./KoLmafia/bin/KoLmafia. It's not a real app-image....

Using deb means that we're foregoing support for Fedora, RHEL, CentOS (all rpm-based), and basically everyone else who's not on a deb-based distro.

build.gradle
heeheehee-kolmafia: Can we instead use yy.MM.dd? Or 'yy.MM.' + version or something that gives us finer granularity?

heeheehee-kolmafia:
Yeah, I saw. The problem is that if I run with deb or rpm without having the corresponding tools installed, jpackage fails entirely.

BadHorseMonkey:
@fewyn and I figured out the issue with windows (by default it puts things in C:\Program Files, by default it wants to write config files to the Current Working Directory if it doesn't find them, by default Windows says `Hell No` when someone wants to write to C:\Program Files and that said the JDK failed to launch.

Otherwise it works perfectly. :/
We added the dir chooser option and maybe also a default directory, but users may need to be steered towards using something like `%CSIDL_PROFILE%\KoLmafia`

And I'm about to change and test the mac DMG solution, expect changes incoming...

gausie:
I still think we shouldn't use "APP_IMAGE". It's not doing anything, we'd get the same result with "BANANA"

heeheehee-kolmafia:
"KoLmafia-r25762-21.10.25762.dmg"

I feel like we could drop the r25762 part, but it's also okay as it is...

PR #95, opened 2021-10-06
Check for null in UseItemEnqueuePanel.

heeheehee-kolmafia:
We apparently need to do this every time we want to access a method of
item. Worth refactoring on a later date so this is less dangerous.

This change also resolves a mismatch between the item name (1950
Vampire Vintner wine) and its presence in inebriety.txt,
modiiers.txt, and statuseffects.txt.

For reasons, we only expanded DataFileConsistencyTest to cover
inebriety.txt (since it best fit the model used by the existing check
of equipment.txt).

### DIFF
--- a/src/data/inebriety.txt
+++ b/src/data/inebriety.txt
@@ -4,6 +4,7 @@
 
 # Drink	Inebriety	Level Req	quality	Adv	Musc	Myst	Moxie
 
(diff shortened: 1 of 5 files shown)


BadHorseMonkey (approved):
tested fine on my system, but I don't know how to trigger the problem and I don't have the IotM...

PR #13, opened 2021-09-25
Create jenkinsfile

fewyn:
First version of jenkinsfile for [ci.kolmafia.us](ci.kolmafia.us)

PR #96, opened 2021-10-07
Coerce get_player_name argument to int

heeheehee-kolmafia:
Coerce first argument of get_player_name to int.

This caused problems when invoked with a type that's coercible to int,
namely a float. While it was most visible in JS (which only has a
Number type, equivalent to ASH's float), it could also be reproduced
in ASH.

For a more systematic fix, we might want to investigate always
coercing arguments of library functions, rather than doing it on a
one-off basis in each method.

This also fixes issues with CustomScriptTest that were introduced when
migrating it to JUnit5 parameterized tests.

Associated bug report: https://kolmafia.us/threads/javascript-bugs.25638/page-5#post-164821


heeheehee-kolmafia:
Hm. I don't know why this is failing Windows tests. It's passing locally on Linux, and the test report isn't being published. :/

BadHorseMonkey:
js has no such thing as an int.  It just has doubles.

https://stackoverflow.com/questions/23376880/parseint-returns-double-in-rhino

test/net/sourceforge/kolmafia/textui/javascript/AshInteropTest.java
```diff
@@ -0,0 +1,34 @@
...
+    assertEquals("", KoLmafia.lastMessage);
+    assertEquals("354981", js.execute(null, null, true).toString());
+  }
+
+  @Test
+  void getPlayerIdInvertsGetPlayerName() {
+    ContactManager.registerPlayerId("heeheehee", "354981");
+
+    var js = new JavascriptRuntime("getPlayerName(parseInt(getPlayerId('heeheehee')))");
+    assertEquals("", KoLmafia.lastMessage);
```
BadHorseMonkey, line 31: let's simplify this.  You're already sure you're getting an Int from getPlayerIdReturnsInt() above.  Don't chain it here, so we don't test three functions.

heeheehee-kolmafia: Actually, that test name lies. We're getting a string from getPlayerId.

heeheehee-kolmafia: Either way, simplified to just testing a single function in each case, and replaced comparisons against v.toString() with Value comparisons. It's not much better, since equals does coercions, but at least it won't coerce numbers to strings.

Maybe I should instead be using rawValue... although that doesn't perform any type checking either.

heeheehee-kolmafia: Thing is, _both_ test cases were failing under Windows. So even the simple one that was just doing getPlayerId('heeheehee') was returning null.

BadHorseMonkey:
```
AshInteropTest > getPlayerIdReturnsString() FAILED

    org.opentest4j.AssertionFailedError at AshInteropTest.java:24


AshInteropTest > getPlayerIdNameCoercesArgumentToInt() FAILED

    org.opentest4j.AssertionFailedError at AshInteropTest.java:32
```
the first one was passing before, IIRC.

the second one is weird, isn't it?  What's in LastMessage on that line that's breaking the 2nd test.

heeheehee-kolmafia: > the first one was passing before, IIRC.

We can actually see historical executions in the Actions tab.

https://github.com/kolmafia/kolmafia/actions?query=branch%3Ajs-coercion

https://github.com/kolmafia/kolmafia/runs/3826425362

```
AshInteropTest > getPlayerIdReturnsInt() FAILED
    java.lang.NullPointerException at AshInteropTest.java:15

AshInteropTest > getPlayerIdInvertsGetPlayerName() FAILED
    java.lang.NullPointerException at AshInteropTest.java:23
```

I have a second branch whose sole purpose is to publish reports on commit (since cross-repo reports don't get published due to security concerns) so I can see more report details.

Results of latest check: https://github.com/heeheehee-kolmafia/kolmafia/runs/3834392716

jaadams5: My git-fu is not advanced enough to work with this branch.  I think it is because I can make my tools work for branches of kolmafia but not branches of heeheehee-kolmafia.  I did clone h3-k but then could not get the branch.  In any event, I extracted the one test file and put it in my own local branch.  There's an educational opportunity :-) 

I know the point of this PR is that JS doesn't have strong typing and what it does have needs to behave with Ash.  The only test failure I had was in what had been the last test and that was because 354981 was not equal to 354981.0 .  I tracked that down to js_parseInt which took 354981 as input and returned 354981.0  My conclusion was that either js_parseInt is buggy or we are using it inappropriately.  

I used the debugger a lot and unchained some things so that I could inspect an intermediate object or two.  Everything I looked at but the one conversion behaved as expected.  Breakpoints in the Ash functions showed the expected inputs and outputs.  The new test I added worked.  Basically we are composing two functions where f(x) = y, g(y) = x and thus f(g(y)) = g(f(x)) and my new test made sure both "singletons" were tested separately.

I was thinking I was looking at a test that had failed on Windows but passed in other environment.  It does sound like there are various failures that may not be the same across OSes.

I've got a couple of (time) windows later today and will try and see if I can get the null pointer and other reported errors.

jaadams5: Missed there were changes there...  Will fetch

BadHorseMonkey:
So, getPlayerId() is an interesting function.
It responds slightly differently to a string, an int, a string of an int ("13"), an invalid player, a fraction (1/2), an invalid name ("..", "\'", "", broken shite (! or ** or a+b ) , and array (["13", "1"]), etc.

Sometimes it shows the result of /whois $string.  Not always.  I wonder if that's tripping up the lastMessage check and other things.

Also, there are a lot of fun negative cases for a test case for getPlayerId.  

> js getPlayerId('Mr Skullhead')
Returned: 13

> js getPlayerId(13)
**Mr Skullhead (#13), Father, Son, and Wholly Gross. (In Ronin)**
Returned: 13

> js getPlayerId('Darwinlet')
Returned: 263717
> js getPlayerId('263717')

**Darwinlet (#263717), the Level 9 Basillusionist (In Ronin)
This player is currently online in channel normal and listening to clan, games, hobopolis, slimetube and wizards.**

Returned: 263717

> js getPlayerId(263717)

**Darwinlet (#263717), the Level 9 Basillusionist (In Ronin)
This player is currently online in channel normal and listening to clan, games, hobopolis, slimetube and wizards.**

Returned: 263717

> js getPlayerId('Darwinlet')

Returned: 263717
> js getPlayerId('BANANANANANANANAAAAANA')
**Unknown Player: BANANANANANANANAAAAANA**
Returned: BANANANANANANANAAAAANA

heeheehee-kolmafia:
Hm, I might have broken it when I migrated it to be a parameterized test.

heeheehee-kolmafia:
Yeah, it's broken because the test method is private, which is invalid.

heeheehee-kolmafia:
Here's a fun one... now it's only failing under macOS.

BadHorseMonkey:
> Here's a fun one... now it's only failing under macOS.

It is, but only when run from the test target!  If I run the test in isolation from the IDE, it's fine.

!  If I put something in the try or the catch block, it works:
  // Looks for the file "test/root/expected/" + script + ".out".
  private static String getExpectedOutput(String script)  {
```
    try {
      //System.out.println("lookingfor out"+script);
      return new String(
          Files.readAllBytes(new File(EXPECTED_LOCATION, script + ".out").toPath()),
          StandardCharsets.UTF_8);
    } catch (IOException e) {
      System.out.println("Error: "+ e.toString());
      return e.toString();
    }
  }
```

When I broke it out, it only failed if I didn't initialize my returnValue:
```  private static String getExpectedOutput(String script)  {
    String returnValue="undefined!";

    try {
      //System.out.println("lookingfor out"+script);
      returnValue =  new String(
          Files.readAllBytes(new File(EXPECTED_LOCATION, script + ".out").toPath()),
          StandardCharsets.UTF_8);
    } catch (IOException e) {
      System.out.println("Error: "+ e.toString());
      returnValue = e.toString();
    }
    return returnValue;
  }```

heeheehee-kolmafia:
Hm. My impression from the action results was that the issue was with initializing output, not expectedOutput. Is that not the case for you?

https://github.com/heeheehee-kolmafia/kolmafia/runs/3844932557

That was why I experimented with explicitly calling out.close(), although it didn't seem to change anything.

heeheehee-kolmafia:
I've updated the PR title / description, squashed these commits together, and force-pushed, since the meandering path I took is ultimately uninteresting from the perspective of the actual change.

It's now passing (and running!) tests, so I think it's ready for review.

heeheehee-kolmafia:
I... did not mean to close this PR. Whoops!

heeheehee-kolmafia:
I'm sure nobody relied on passing 123.0 to get_player_name (since that never worked). This doesn't impact the output of get_player_id, since for whatever reason that returns a string (even though IMO it should return an int).

BadHorseMonkey (approved):
So, I ran a bunch of tests and I only see one oddity, which is the first test pair and AFAICT does no harm (now).

It occurs to me that it would be useful to have a parameterized test which checked that the output of a single js command was the same as the output of the identical ash command, but that is not this PR.

> js getPlayerId(1)
Jick (#1), Jick-Or-Treat! (In Ronin)
Returned: 1.0

> ash get_player_id(1)
Jick (#1), Jick-Or-Treat! (In Ronin)
Returned: 1

> js getPlayerId('jick')
Returned: 1

> ash get_player_id('jick')
Returned: 1

> js getPlayerName(1)
Returned: Jick

> ash get_player_name(1)
Returned: Jick

> ash get_player_name(13)
Mr Skullhead (#13), Father, Son, and Wholly Gross. (In Ronin)
Returned: Mr Skullhead

> js getPlayerName(13)
Returned: Mr Skullhead

> ash get_player_name(to_int(get_player_id(1)))
Jick (#1), Jick-Or-Treat! (In Ronin)
Returned: Jick

> js getPlayerName(toInt(getPlayerId(1)))
Jick (#1), Jick-Or-Treat! (In Ronin)
Returned: Jick

> js getPlayerName(1.0)
Returned: Jick

> ash get_player_name(1.0)
Returned: Jick

> js getPlayerName(parseInt(getPlayerId("WoRtHaWhOlEbEaN")))
worthawholebean (#1972588), the Level 59 Pastamancer
This player is currently online.
Returned: worthawholebean

heeheehee-kolmafia:
Yeah. Nothing I can do about JS Numbers, and agreed that it would be nice to have an ASH / JS comparison to make sure they generate the same Values (and that it's out of scope for this PR).

PR #97, opened 2021-10-07
Make $outfit and $tattoo a type in ASH

gausie:
This is an breaking change for ASH. 

```
get_outfits() is now $outfits[]
all_normal_outfits() is now $outfits[]
outfit_treats(string) is now $outfit[].treats
outfit_tattoo(string) is now $outfit[].tattoo
```

Still in draft and we might not want to do it, but it addresses a long standing deficiency (in my opinion) in the constants available to our scripting runtimes.

This still allows custom outfits to be referenced with a string.

In the future we can add functions to check if you have a specific tattoo unlocked and if you current have it on your profile

### DIFF
--- a/src/data/tattoos.txt
+++ b/src/data/tattoos.txt
@@ -1,0 +1,306 @@
+1
+# image	requirement type	requirement
+aerotat.gif	outfit	Aeroutfit
+animelftat.gif	outfit	Animelf Apparel
+aol1.gif	other
+aol2.gif	other
+aol3.gif	other
+aol4.gif	other
+aol5.gif	other
+arbortat.gif	outfit	Arrrbor Day Apparrrrrel
+arborween.gif	outfit	Vestments of the Treeslayer
+armortat.gif	outfit	Antique Arms And Armor
+asc01.gif	ascension	1
+asc02.gif	ascension	2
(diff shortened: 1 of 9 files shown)


src/data/tattoos.txt
```diff
@@ -0,0 +1,306 @@
+1
```
jaadams5, line 1: New file, versioned according to KoLmafia idiom, and tab delimited?  Can't really verify the latter.

gausie:
@jaadams5 I meant to make this a draft but still. Yes! We should talk about it here and give some warnings. These are the breaking changes:

```
get_outfits() is now $outfits[]
all_normal_outfits() is now $outfits[]
outfit_treats(string) is now $outfit[].treats
outfit_tattoo(string) is now $outfit[].tattoo
```

heeheehee-kolmafia:
The breaking change that worries me the most is that all variables called "outfit" or "tattoo" will now cause problems with the Parser.

    // This won't generate an error...
    string outfit = "";
    // nor this
    if (outfit == "test") { }
    // but this will
    outfit = "test";

Typically this is made more explicit by marking new datatypes as reserved words.

gausie:
Going to close this for now, branch will still exist though

PR #98, opened 2021-10-07
Fix out of bounds error in Robocore

gausie:
Relevant issue: https://kolmafia.us/threads/you-robot-spring-challenge-path.25870/page-4#post-164832

Many guesses were made as to what the issue was. None of them were correct.


taltamir:
> Relevant issue: https://kolmafia.us/threads/you-robot-spring-challenge-path.25870/page-4#post-164832
> 
> Many guesses were made as to what the issue was. None of them were correct.

thanks for the fix. sorry for the wrong guesses

PR #101, opened 2021-10-07
Little mess up in ParserTest.java

fredg1:
When making the transition to Junit5, one of the parenthesis was added *in a comment*!
This effectively corrupted half of the tests!


heeheehee-kolmafia (approved):
Thanks for the catch.

PR #102, opened 2021-10-08
Interop

jaadams5:
Draft.  Just trying to make my code visible since my git-fu isn't finding a better way to share.


heeheehee-kolmafia:
Frono -- I think this PR is obsolete now. Mind if I close + delete the associated branch?

jaadams5:
Well I like AshInteropTest :-) but we can withdraw this (since it was only published for collaboration) and if I retain my fondness for AshInteropTest I can make a new PR.

PR #103, opened 2021-10-08
Remove java 11 check, as that is the default now

Rinn:
No need for a separate step, the default check is set to java 11 compatibility now.


heeheehee-kolmafia (approved):
Ah, I see you updated branch protection. Thanks!

PR #106, opened 2021-10-10
Disable DebugDatabaseTest.checkZapGroups().

heeheehee-kolmafia:
This test relies on the wiki, which seems to be down for me right
now.

In general, tests that rely on external data (especially ones that
rely on the network) should be avoided to ensure future
reproducibility.

I don't know if the test is hanging just for me or for everyone else.
Maybe they've IP-banned me for running too many tests :(

### DIFF
--- a/test/net/sourceforge/kolmafia/persistence/DebugDatabaseTest.java
+++ b/test/net/sourceforge/kolmafia/persistence/DebugDatabaseTest.java
@@ -363,6 +363,7 @@ public class DebugDatabaseTest {
   }
 
   @Test
+  @Disabled("Relies on external resources (wiki)")
   public void checkZapGroups() {
     String expectedOutput = "Checking zap groups...\n";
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia:
Ha. setSocketImplFactory is also deprecated as of Java 17. Likely a slightly more involved refactor.

BadHorseMonkey:
While I like the idea of checks for data integrity, it would be nice to have another place for them. 

How would we test the ash wiki command or SVN if we couldn’t use a socket connection?

jaadams5:
We had some conversations about separating tests that were actually confirming data files were correctly formatted and contents were internally consistent and tests that exercised code.  Perhaps we should continue those discussions?

I agree that tests that need network access need to be written carefully and intentionally.  If access is a "side effect" we can presumably use Forbid to detect a test needs it and the mock to not need it any more?

PR #109, opened 2021-10-11
You robot fixes

gausie:
Multiple fixes for You, Robot

### DIFF
--- a/test/net/sourceforge/kolmafia/request/AscensionHistoryRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/AscensionHistoryRequestTest.java
@@ -38,7 +38,7 @@ public class AscensionHistoryRequestTest extends RequestTestBase {
     assertEquals(43, Preferences.getInteger("borisPoints"));
     assertEquals(31, Preferences.getInteger("zombiePoints"));
     assertEquals(15, Preferences.getInteger("jarlsbergPoints"));
-    assertEquals(0, Preferences.getInteger("petePoints"));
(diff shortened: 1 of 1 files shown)


gausie:
> > This is only failing on OSX :( @heeheehee-kolmafia is it possible that OSX can't persist prefs in memory when `saveSettingsOnSet` is false?
> 
> Those tests fail on my local mac when run as part of the tests target but work when run on their own.
> 
> I don't yet know why. Maybe some state is persisting?

This will be because when you run them on their own presumably they don't get the `@BeforeAll` run and so they're writing to a file

heeheehee-kolmafia:
> This is only failing on OSX :( @heeheehee-kolmafia is it possible that OSX can't persist prefs in memory when `saveSettingsOnSet` is false?

I wouldn't expect this to be the case, since that would indicate that either TreeMap or Collections.synchronizedSortedMap is horribly broken but only on macOS. (Aside: we're probably using that synchronized map incorrectly.)

I haven't looked into the specific failures.

gausie:
> Should you be clearing preference state between test cases?

Maybe, but it shouldn't cause any issues. Might be nice to have a sort of checkpoint thing for preferences to use in these cases.

heeheehee-kolmafia:
(I merged main into this branch so we can hopefully see the report failures.)

jaadams5:
Perhaps related, I grabbed this test locally.  8 custom scripts tests failed.  I did dos2unix on several of the expected files and all tests passed including ScrapHeapRequestTest.  I may have a local line ending issue but things are behaving on Windows.

PR #112, opened 2021-10-11
Move legacy dailies

gausie:
Tidying up where we keep track of what preferences are daily - we have old prefs that aren't preceded with a `_`, it's neater to keep that in Preferences.


jaadams5:
And our comments crossed.

Effectively all this does now is define isDaily() and use it instead of checking for a leading underscore, correct?  If so we can ignore my comment and I will approve.

PR #114, opened 2021-10-12
Parser roadmap

fredg1:
A comment on top of the main class showing how the various classes of parseTreeNode get parsed.


fredg1:
> 
> 
> given that Java demands (primary) single inheritance.

Well, that wouldn't matter here, since this is actually a *call* hierarchy, rather than a *type* hierarchy.
There's a "parse" implicitly preceding every single individual word/identifier in the roadmap.

jaadams5 (approved):
Thank You

PR #117, opened 2021-10-13
revert_lib_removal_temporarily

BadHorseMonkey:
IntelliJ is falling all over itself failing because srcDirs doesn't include 'lib'.  While I am all in favor of pulling it out, it's breaking things.  We may need a more tailored srcDirs, or some excludes, or to move the 3rd party stuff into src.  But that's a follow-on fix.


BadHorseMonkey:
I'm good with the goal, I just don't know how to express it in Gradle.

PR #119, opened 2021-10-13
Language Server - Evaluable

fredg1:


fredg1:
Actually, I *just* noticed (aaand you just merged...)

Since I added the "TypedNode", I no longer need the changes I made to everything that ended up calling Parser.undefinedFunctionMessage

PR #122, opened 2021-10-15
add function matching tests

fredg1:
Let's try to avoid getting another https://github.com/kolmafia/kolmafia/pull/121, a'ight?


gausie:
Can you add any tests that you'd expect to throw?

gausie:
Write code that you would expect to throw an exception and test that those exist. If there aren't any that you can do thats fine.

PR #124, opened 2021-10-15
Disable writing settings more directly in tests.

heeheehee-kolmafia:
We think there may be some settings leaking across tests, which
is... not ideal.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -3175,6 +3175,7 @@ public abstract class KoLCharacter {
         break;
       case GLOVER:
         Preferences.increment("gloverPoints", points, 11, false);
+        Preferences.increment("garlandUpgrades", 1, 10, false);
         break;
       case DISGUISES_DELIMIT:
         Preferences.increment("masksUnlocked", points, 25, false);


heeheehee-kolmafia:
Why is this the fix, I wonder? Is macOS not running the BeforeAll- / BeforeEach-decorated methods properly / consistently?

heeheehee-kolmafia:
Okay. It looks like the root cause per the @BadHorseMonkey's latest commits is that we were using private methods for these annotations. But why did it even work in the first place outside of Mac??

heeheehee-kolmafia:
Actually, hm. It's failing again not just on Mac, but also Windows, now that you've removed the inline KoLCharacter.reset call.

But, they're for different tests. This feels like case of flaky tests caused by a race condition, but I'm not sure what could be racing.

heeheehee-kolmafia:
(This latest commit is just to validate my guess that it's a race condition. I don't know that we want it long-term.)

BadHorseMonkey:
> (This latest commit is just to validate my guess that it's a race condition. I don't know that we want it long-term.)

I put a sleep(9000) on either side of it.  I am not convinced it's a race condition.

heeheehee-kolmafia:
Fascinatingly, I set `forkEvery = 1` in build.gradle in https://github.com/heeheehee-kolmafia/kolmafia/commit/042775a40cefea6315773c66bce49890b81d33ac, and all tests passed. Um, what?

heeheehee-kolmafia:
My approach of gauging test ordering from timestamps in xml logs was actually imprecise, since they only have second-level granularity.

But, I think it's very likely that some test is leaving things in a weird state.

heeheehee-kolmafia:
Heck yes. Tests are green now.

gausie:
@heeheehee-kolmafia just saw someone mention their prefs weren't persisting between mafia sessions. They could be wrong, but this could possibly have a defect.

heeheehee-kolmafia:
>  If the only way Preferences.saveSettingsToFile will be changed is via a test then there are some annotation options or documentation that might be helpful. So the next person to read the code without an IDE search doesn't jump to the wrong conclusion.

Is this comment on the variable insufficient?
```
// If false, blocks saving of all preferences. Do not modify outside of tests.
```

> @heeheehee-kolmafia just saw someone mention their prefs weren't persisting between mafia sessions. They could be wrong, but this could possibly have a defect.

Hm. I don't _see_ anything that this specific commit should have changed on that front. There are three non-test files that changed. KoLCharacter and AscensionPath both deal with setting gloverPoints / garlandUpgrades and are irrelevant.

The change to Preferences.java would be the most likely culprit, except we initialize saveSettingsToFalse = true, and only bail from saveToFile() if that setting is false. All other changes are tied to Test files.

It's possible that preference-saving is broken from some other commit. Other than GLOBAL_prefs, I see that none of my preferences have been updated since my Oct 15 cronjob run (which predates this change; I won't rule out me running from off the main branch though).

PR #125, opened 2021-10-16
Tweak a test

jaadams5:
Fix some line and adjust a test that was previously ignored.

### DIFF
--- a/test/net/sourceforge/kolmafia/utilities/CharacterEntitiesTest.java
+++ b/test/net/sourceforge/kolmafia/utilities/CharacterEntitiesTest.java
@@ -29,12 +29,12 @@ public class CharacterEntitiesTest {
     String eua = CharacterEntities.escape(ua);
     assertEquals(a, eua);
     String b = "This is mixed and matched \u00a3 \u00a4";
-    String ub = CharacterEntities.unescape(b);  //nothing to unescape so should be equal
+    String ub = CharacterEntities.unescape(b); // nothing to unescape so should be equal
     assertEquals(b, ub);
-    String eub = CharacterEntities.escape(ub);  //there should be a replacement
-    assertNotEquals(ub, eub); //so not equal
-    assertNotEquals(b, eub); //and not inverse
-    assertEquals(a, eub); //true because of data choice.
+    String eub = CharacterEntities.escape(ub); // there should be a replacement
+    assertNotEquals(ub, eub); // so not equal
+    assertNotEquals(b, eub); // and not inverse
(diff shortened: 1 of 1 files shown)


jaadams5:
I got distracted by Real Life.  My intent was to post this and solicit feedback,

My test for b = escape(unescape(b)) passed when running in the debugger on Windows but failed when running gradlew test.  I think the problem is that the JRE is pretty aggressive about replacing "escape sequences" such as \u00a3 at the earliest possible moment.  So when running in the debugger the substitution happens at a different time compared to the test for equality in gradle.  So if we insist that a = escape(unescape(a)) and a = unescape(escape(a)) for all values of a I think we have an implementation error.  Does anyone agree with me or did I just write a "bad" test and make it worse?

Oh yeah, the test failures on GitHub that had nothing to do with my changes contributed to the distraction :-)

I'll call @heeheehee-kolmafia out to discuss before I squash and merge.

heeheehee-kolmafia:
My take is that fundamentally, escape takes arbitrary unicode and transforms non-ascii (and some other special characters like &) characters into HTML entities; unescape transforms those entities into unicode.

If your string contains no entities (as is the case with b), then unescape will do nothing.

In general, unescape(escape(x)) should always equal x, since even entities will undergo some escaping (`&quot;` -> `&amp;quot;`). But escape(unescape(x)) as you've found is not always the same as x (namely, when x contains unicode characters).

```
    String b = "This is mixed and matched \u00a3 \u00a4";
    String eb = CharacterEntities.escape(b);
    String ueb = CharacterEntities.unescape(eb);
    assertEquals(b, ueb);
```
passed for me in `./gradlew test`, for what it's worth. Maybe there's some secondary issue in Windows where that's not the case? (encoding mismatch, once again?)

heeheehee-kolmafia:
https://github.com/heeheehee-kolmafia/kolmafia/commit/5c8de803522b7d304eaf54cdd23ec288384eb0ed passes tests for me, curious what you tried that failed via ./gradlew test.

heeheehee-kolmafia:
Anyways. In short: IMO, insisting that unescape and escape are inverses is incorrect. It might be more accurate to say that unescape "undoes" the action of escape, or to describe unescape as a left inverse (as opposed to a two-sided inverse) of escape. If you want to look at it in math-y terms:

```
unescape o escape == id != escape o unescape
```
using `o` for function composition, and `id` for the identity function `id(x) -> x`.

Equivalently:
```
unescape(escape(x)) == x != escape(unescape(x))
```
for all non-null Unicode strings x.

PR #126, opened 2021-10-17
Fix taskbar detection for setAdventuresLeft.

heeheehee-kolmafia:
Right now, useDockIconBadge is overriding our careful taskbar feature detection. This is generating debug logs for users who have this preference set to true (e.g. if they previously used these settings on a Mac once upon a time).


gausie:
Thanks for fixing this!

PR #131, opened 2021-10-18
Simplify new ascension pref resets

gausie:
There has been talk about this refactor for years... and here it is! Fixed a few bugs/incongruencies in our prefs. Some were misspelt/used incorrect case and (e.g. "bondDR" was typoed to "bondDA", "twinPeakProgress" was formatted as "twinpeakprogress")


BadHorseMonkey:
I wasn't sure where resetDailies() was called, but I found it inside resetCounters().

PR #133, opened 2021-10-19
Js test

jaadams5:
All tests pass in IntelliJ.  Current code passes in Gradle.  If there was no operator error then enabling one of the disabled tests fails in Gradle.  Seems to be related to the number of enabled tests and not which ones.  Don't think there is a spill over between tests and wonder if it is the Gradle threading?


jaadams5:
I copied JavaScriptRuntime from 135 and am running locally.  I added a test for Kolmafia,getLastmessage() to be null and it is Not enough mana to cast Experience Safari.  Now what I am expecting.  Pondering.

PR #136, opened 2021-10-20
Unit Tests for Preferences.java

BadHorseMonkey:
This PR includes unit tests for each method in Prefs.


jaadams5:
I have had tests fail that had nothing to do with any changes I have made.  I can often trigger that by doing gradlew clean before the tests.

gradlew clean
gradelw test

has often generated failures on "my changes"

but immediately doing a second

gradlew test

shows everything passing.

If I want to maximize my certainty about tests working I run them in IntelliJ and then in the context of clean and then run every test.

If all I do is a single test I get surprised with a PR had failing tests.

PR #137, opened 2021-10-20
Implement CLI command for tuning moon sign

midgleyc:
Implement a CLI command for changing moon sign, currently named "spoon". Unequips the spoon, runs the command for the appropriate sign, re-equips it.

Refactored part of zodiac signs in `KoLCharacter` into their own enum. Left `ZODIACS` as it was: could generate it from a stream of the enum values, but I didn't see the point in that.

No unit test for the command because I don't know how I'd test it: I couldn't find any existing tests for any commands. I tested it manually.


gausie:
After looking at all the usages of `String[] ZODIACS` we should be fine to refactor each usage to use this new enum, but that can be part of a later PR if you'd like

PR #148, opened 2021-10-21
DisplayCaseManager raw types

fredg1:
from https://kolmafia.us/threads/reducing-number-of-warnings-raw-types-displaycasemanager.26004/

Now, the discussion over there actually ended with an unfinished talk about adding a test to this class.

The only non-solely-generics-related change of this PR is a change from
```java
move(<List>.toArray(), [...]);

void move(Object[], [...]) {
  <turn the array into a List>
  [...]
}
```
to
```java
move(<List>, [...]);

void move(List, [...]) {
  [...]
}
```

I would like to push the fact that this change should be straightforward/simple enough to not warrant a test.
N.B. this fact was not mentioned in the discussion so far.

If the test is still mandated, I would ask for help in adding it, because I don't really know where to start...

### DIFF
--- a/test/net/sourceforge/kolmafia/session/DisplayCaseManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/DisplayCaseManagerTest.java
@@ -36,7 +36,7 @@ class DisplayCaseManagerTest {
   // The disabled status should cause someone to revisit this eventually.
 
   @Test
-  @Disabled("DisplayCaseManager.update is not be the right injection")
+  @Disabled("DisplayCaseManager.update is not the right injection")
   public void itShouldHaveSomeContents() {
     // This file was generated from CafeBabe's Display Case which had no shelves at the time.
     String displayCase = null;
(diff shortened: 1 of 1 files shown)


jaadams5:
I noticed a @SuppressWarnings.  My philosophy at the moment is that we should not do that because we are much more tolerant of PRs that deal with fixing compiler complaints.  If there is a warning we cannot address, perhaps a comment in the code about the warning and why it remains to be suppressed?

I'm not happy with the test situation, or lack there of but since you asked for help I will either get you some or let things pass as is.

fredg1:
> 
> 
> I noticed a @SuppressWarnings. My philosophy at the moment is that we should not do that because we are much more tolerant of PRs that deal with fixing compiler complaints. If there is a warning we cannot address, perhaps a comment in the code about the warning and why it remains to be suppressed?
> 
> I'm not happy with the test situation, or lack there of but since you asked for help I will either get you some or let things pass as is.

It's because of an array of generics.
I'm generally surprised that the compiler even says anything about it, since there is, AFAIK, literally no way to make it "safe"
```java
new ArrayList[]; // <<< Allowed, but whines about raw types
new ArrayList<Object>[]; // <<< Not allowed
```

jaadams5:
I added a simple test.  You could add to it and make a better one.  Grab the html from a display case with shelves, put it in a file in test\root\data.  Copy and rename itShouldHaveSomeContents(), change it to read your file and then add some assertions to deal with the shelves defined in your file.  For extra credit look at the test report and see if there is something you can do that covers uncovered code.

@heeheehee-kolmafia might look at the test but I think we're almost there.

PR #149, opened 2021-10-21
feat: add fax send/receiver ash commands

oxc:
Add RuntimeLibrary commands for sending and receiving faxes. They are just wrappers around the CLI command, but are less awkward to use from ASH/JS scripts.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -933,6 +933,12 @@ public abstract class RuntimeLibrary {
     params = new Type[] {DataTypes.INT_TYPE, DataTypes.ITEM_TYPE};
     functions.add(new LibraryFunction("retrieve_item", DataTypes.BOOLEAN_TYPE, params));
 
+    params = new Type[] {};
+    functions.add(new LibraryFunction("receive_fax", DataTypes.VOID_TYPE, params));
+
+    params = new Type[] {};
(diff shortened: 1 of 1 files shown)


oxc:
Of course in the end, I would rather argue than there should not be two separate scripting engines anyway, but CLI should be migrated to be just another RuntimeLibrary frontend. Probably not gonna happen, but that would be nice.

jaadams5 (changes requested):
**Might I suggest something like mocking RequestThread.postRequest, and verifying that it's called with an appropriately constructed ClanLoungeRequest?**

I understand that as a request for a test.  That is why I am not approving but if someone else so chooses....

oxc:
I will try to add some tests, but I'm not sure when I will get around to. I'm not expecting you to approve this until then, even though I find these changes a bit small to expect them to lay the ground for library function testing.

gausie:
> I will try to add some tests, but I'm not sure when I will get around to. I'm not expecting you to approve this until then, even though I find these changes a bit small to expect them to lay the ground for library function testing.

I think that's a fair point. Maybe it should be a kick for us to lay that ground

jaadams5:
OK.  My first response is a script that exercised the commands would be better than nothing but setting up that as a test that didn't need external resources is non-trivial.  The current coverage of RuntimeLibrary is minimal, to the point that I see no point in delaying this to make it the first test.  So I am going to unrequest changes and will push.

jaadams5 (approved):

PR #150, opened 2021-10-22
simple raw types

fredg1:
from https://kolmafia.us/threads/reducing-number-of-warnings-raw-types-easy-straightforward-simple-only-batch-n%C2%B01.26003/

There's many, but none of them depend on something else.


jaadams5:
Since the original patch dates to April, has this been updated to reflect the change in Java version?

PR #152, opened 2021-10-22
Try to auto-merge before running checks in PRs.

heeheehee-kolmafia:
If it can't merge cleanly, then it should abort the change.

This might work on Windows -- it seems that Powershell v7 added
pipeline-chain operators:

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pipeline_chain_operators?view=powershell-7.1


heeheehee-kolmafia:
Hm, so, it does look like it is merging properly, per the number of tests that are being run. I _think_ I still want to add kolmafia/kolmafia as a remote explicitly so this works for external contributors.

PR #153, opened 2021-10-23
KoLmafiaCLI.findScriptFile(File, String, List, boolean)

fredg1:
In KoLmafiaCLI.findScriptFile(String, List), when we don't want to recursively search in a directory, instead of replicating the logic of KoLmafiaCLI.findScriptFile(File, String, List), we can instead pass to it whether or not we want it to recurse, simplifying the logic.


gausie (dismissed):
This seems reasonable to me but I feel the same as @BadHorseMonkey with regards to seeing red

jaadams5:
I am a little concerned about a possible behavior change because the subdirectory flag doesn't seem to be set the way I expect.  I won't block this but I hope someone has actually run this against a real user setup full of files that don't have unique names until the entire path is considered.

fredg1:
What do you mean, "not the way I expect"?

jaadams5:
It was presented as a refactoring.  So if it is truly a refactor I would expect the new parameter to always be true.  The second step would be to decide that subdirectories did not need to be searched and flip the parameter.  I do not have enough confidence in my understanding of the old code to decide whether the behavior has changed or not when the parameter is flipped.

jaadams5:
I've looked at this enough that I can't find a case where behaviour changes so my reservations on that point are gone.  I still believe it is an unnecessary change and does nothing to advance any project metric that is of interest to me - rearranging the deck chairs on the Titanic before encountering the iceberg - but since @gausie seems to disagree I will let them press Squash and merge and I acknowledge my own complicity :-)

fredg1:
@jaadams5 well, now I'm intrigued/interested. All three of you pointed out hatred for the current system, labelled as "flat namespace".

I am not well educated on this side of things; what's the alternative? Is it really preferable?
Is it something that *could* be achieved with/in mafia, or would it carry changes that we can't impose because it would break other people's workflows/scripts?

jaadams5:
A flat namespace in this context is a situation where all of the files are assumed to be in the same directory.  That means names must be unique.  The advantage of a flat namespace is that it, since it maps to a single directory, it is easy to enforce file operations so that no access is allowed outside of the namespace.  

Because it supports user written scripts KoLmafia has some obligation to prevent a hostile script from accessing the  filesystem outside of the "advertised" namespace.

IIRC in the beginning scripts could only be run from /scripts and /relay.  They could only read and write files in /data and scripts in /scripts could only reference other scripts in /scripts.  A decade later we have subdirectories and scripts in /planting, /scripts and /relay can reference each other.  We have a local repository of remotely published scripts and a requirement that they must be copied elsewhere before they can be run.  Filenames alone no longer determine uniqueness.  "Kludge" and "fragile" are some of the more polite descriptions for the current situation.  No one has agreed on a new design, let alone a migration plan.  But no one really wants to mess with things now either because a) the work will almost certainly be jettisoned when there is a new scheme; and b) every change has to be inspected for possible exploits which might allow a malicious scripts access to "forbidden" parts of the file system.

So this change is neutral in terms of taking us towards the as yet undefined Goal and negative because it creates additional work to make sure it does not create or expose an exploit.

See PR181 as an example of a case where a simple change of using equals to compare two strings broke code because it turns out one of the strings was allowed to be null and that just happened to be the one used to call equals.

fredg1:
WHOOPS, wrong branch :|

gausie (approved):
Scuttlebutt aside, this looks good to me

PR #160, opened 2021-10-27
Preferences

jaadams5:
Lint, typos and cleanup so root\settings remains empty after tests.  Cleanup should happen automatically but tests that create multiple preferences files may have to do additional work.  A human who wants to examine a preference file after a test is probably not going to be happy.  Draft status as I ponder these things.

### DIFF
--- a/test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
+++ b/test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
@@ -16,7 +16,7 @@ public class ClearSharedState implements AfterAllCallback {
   @Override
   public void afterAll(ExtensionContext context) {
     // Clean up user files
-    deleteEm(KoLCharacter.baseUserName());
+    deleteUserPrefsAndMoodsFiles(KoLCharacter.baseUserName());
     // and GLOBALS
     deleteGlobals();
     // Among other things, this sets KoLCharacter.username.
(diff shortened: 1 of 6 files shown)


jaadams5:
root\settings remains empty after tests

added tests for more coverage

commented about ways to make tests leak, but these don't seem to.

when tests pass on github, will change draft status

jaadams5:
and nothing has been done about the security scanner item that was why I went down this rabbit hole.  Next one.

jaadams5:
Ans maybe I have made the scanner happier so one more commit

jaadams5:
Probably easier ways to resolve conflicts but...

test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
```diff
@@ -29,4 +34,24 @@ public void afterAll(ExtensionContext context) {
     KoLmafia.lastMessage = "";
     KoLmafia.forceContinue();
   }
+
+  public static void deleteEm(String user) {
```
gausie, line 38: Why is this called `deleteEm`?

gausie: Ohhh that was silly of me

gausie (approved):
Approved. What should live inside `extensions`? If this is purely test related, shouldn't it live in the test package?

jaadams5:
I was following he3s lead and just put code that dealt with preferences leaking in tests in the place where there was already code to do that.  Worthy discussion topic.

PR #166, opened 2021-10-29
maximizer raw types

fredg1:
the `instanceof Boost` was only done because the Iterator was a raw type.
The LockableListModel was already enforcing that the Iterator would only contain Boosts.


jaadams5 (approved):
Looks like partial coverage by a test already exists.

PR #168, opened 2021-10-30
Add CLI command for upgrading the cosplay saber

midgleyc:
Implement a CLI command for upgrading the cosplay saber. Unlike spoon, this one is easy enough to do manually, but convenient to have through Mafia anyway.


jaadams5 (approved):
Added some tests to increase coverage.  Couldn't find an easy way to equip the saber and test, which would raise coverage tom 100%.

The presence of a test expedited my review process.  Thank you.

PR #169, opened 2021-10-30
Extend what it means for a build to be "dirty".

heeheehee-kolmafia:
~In particular, consider it to be dirty if it's not tagged with the
corresponding revision.~ Consider a build to be dirty if it has
any changes not in origin/main. 

This should be more flexible than just checking the HEAD of
origin/main, since it'll continue to work if you build an old
revision.

~We also need to make sure that we're tagging revisions before building
jars / binaries for distribution, so they're not all marked as
"dirty".~

This PR also fixes revision detection for old builds, which previously just
used "whatever origin/main is at". Now we count the total number of
revisions, minus those that are not present upstream. (It would be nice if
grgit supported git merge-base, but this works well enough.)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,22 +6,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  tag:
-    if: github.repository == 'kolmafia/kolmafia'
-    name: Publish Artifacts
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
-
-      - name: Retrieve version and tag
-        run: |
-          KOLMAFIA_VERSION=$(git rev-list --count $GITHUB_SHA)
-          git tag "r$KOLMAFIA_VERSION"
   bin:
-    needs: [tag]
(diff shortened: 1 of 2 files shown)


heeheehee-kolmafia:
I... don't know how this interacts with ci.kolmafia.us, or at what stage the Jenkins actions run. Could lead to a race if we're not careful.

heeheehee-kolmafia:
Hm. Our revision detection is also wrong for feature branches. Updating origin/main should not result in a revision bump unless we're incorporating those changes.

jaadams5:
Thank you.

heeheehee-kolmafia:
Case in point for our revision detection being wrong:
```
$ ./gradlew runShadow
... 
KoLmafia r25823
Build HEAD-67f7730
...
$ git show 67f7730
commit 67f77308473d2b49561010ea01e6bcc5b465dc38 (tag: r25820)
...
```

jaadams5 (approved):
Makes sense to me.  Built locally as -M but can't really test.  I will approve trusting your judgement.

heeheehee-kolmafia:
I did also confirm that merging main back into the branch incremented the revision number as expected, for what it's worth.

PR #170, opened 2021-10-30
Refactor AbstractCommand: avoid raw types

midgleyc:
`AbstractCommand` had two instances of imprecise types: one in `AbstractCommand.lookup`, where `Object` was used instead of `AbstractCommand`, and one in `AbstractCommand.substringLookup`, which was a `List` that alternated `String`s and `AbstractCommand`s.

The former was generified and `AbstractCommand` used. The latter was replaced by a `Map` of `String` to `AbstractCommand`: the Map has the additional restriction that no key can appear twice, but looking at the implementation that is desired anyway (in fact, an even stricter requirement is desired: no key can strictly contain another).

Tests have been added. I renamed the existing `AbstractCommandTest` because it was used as an abstract base class for the other tests, instead of a test for the `AbstractCommand` methods.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafiaCLI.java
+++ b/src/net/sourceforge/kolmafia/KoLmafiaCLI.java
@@ -497,7 +497,7 @@ public class KoLmafiaCLI {
     return this.elseRuns;
   }
 
-  static {
+  public static void registerCommands() {
     new AbortCommand().register("abort");
     new AbsorbCommand().register("absorb");
     new AccordionsCommand().register("accordions");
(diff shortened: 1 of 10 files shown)


fredg1:
...infinite loop? That test's been going for more than an hour...

midgleyc:
I don't know exactly what the issue was, but moving the cleanup into the after test callback has fixed it.

jaadams5 (approved):
I wish I could figure out how test without modifying source but since the mods are basically about visibility I can live with them.

PR #171, opened 2021-10-30
Modifiers-raw-types

fredg1:


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2675,19 +2675,20 @@ public final void applyFloristModifiers() {
   public void applySynergies() {
     int synergetic = this.getRawBitmap(Modifiers.SYNERGETIC);
     if (synergetic == 0) return; // nothing possible
-    Iterator i = Modifiers.synergies.iterator();
+    Iterator<Entry<String, Integer>> i = Modifiers.synergies.entrySet().iterator();
```
gausie, line 2678: Why not just change this section to `Modifiers.synergies.forEach((name, mask) -> {`

test/net/sourceforge/kolmafia/ModifiersTest.java
```diff
@@ -2,11 +2,13 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.util.Iterator;
+import java.util.Map.Entry;
 import org.junit.jupiter.api.Test;
 
 public class ModifiersTest {
   @Test
-  public void patriotSheildClassModifiers() {
+  public void patriotShiedClassModifiers() {
```
gausie, line 11: Typo here

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2675,19 +2675,20 @@ public final void applyFloristModifiers() {
...
       if ((synergetic & mask) == mask) {
         this.add(Modifiers.getModifiers("Synergy", name));
       }
     }
   }
 
-  // Returned iterator yields alternating names / bitmaps
-  public static Iterator getSynergies() {
-    return Modifiers.synergies.iterator();
+  // Returned iterator yields bitmaps keyed by names
```
gausie, line 2689: Could then make this just return the map

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -1002,10 +1003,11 @@ void enumerateEquipment(int equipScope, int maxPrice, int priceLevel)
     }
 
     int usefulSynergies = 0;
-    Iterator syn = Modifiers.getSynergies();
+    Iterator<Entry<String, Integer>> syn = Modifiers.getSynergies();
```
gausie, line 1006: Then if this returned a map, just use forEach here (and other similiar places) too

test/net/sourceforge/kolmafia/ModifiersTest.java
fredg1: fix a typo, make a typo... Le sigh...

src/net/sourceforge/kolmafia/Modifiers.java
fredg1: This, as well as multiple other getters, return iterators to make sure the underlying map doesn't get mutated (as far as I can tell)

src/net/sourceforge/kolmafia/Modifiers.java
fredg1: Consistency. Iterators are used in various places in KoLmafia (even when the map/array itself is accessible), so this is just made to match this.

If someone then wants to change those to forEaches, they could just do it for all of them at once.

gausie: Nah it makes more sense to fix things as we go

src/net/sourceforge/kolmafia/Modifiers.java
gausie: Then we can just make it an unmodifiable map

fredg1: but we WANT it to be modifiable. Just not from outside Modifiers.java...

gausie: Yeah so you make it private and then use collections.unmodifiablemap in the getter

src/net/sourceforge/kolmafia/Modifiers.java
fredg1: what do I do if there are *break*s in the loops? I know *continue*s can be replaced with *return*s, but what about those?

fredg1: @gausie this doesn't seem to work. Lambdas require external variables to be final (or effectively final). However, some of these iterations modify variables from outside the loop.

Map.forEach() for when you want to *apply* something to each element. But here, we're looking at each of them, and *tallying* something about the map.

fredg1: (Here, this is because of the line just 4 lines below:
```java
if (delta > 0.0) usefulSynergies |= value;
```

gausie: Ah ok never mind then. Can we foreach through them instead? I think this iterator typing isn't amazing

PR #172, opened 2021-10-31
ItemDatabase FoldGroup raw types

fredg1:
Turns ItemDatabase's foldGroups from List<Comparable> to FoldGroup (new class).

The list would always have the damage you receive upon folding the item (an Integer) as the list's first element, and the rest after (Strings).
This now places them in their own fields.

This means I had to make sure to alter every point which assumed the item names started at index 1.

The class that would need testing the most is FoldItemCommand, but I'm not sure how we'd do that...


jaadams5:
There is an unenforced convention to have test scripts start with test_.  You could rename things.

My preferred way to review things is to make a local branch with nothing but new tests and changes to existing tests.  I run them and try to understand which test changes, if any, are driven by the new code.  I run the jacocotestreport and see what the coverage looks like for the classes that will be changed in the complete PR.  This just increases my confidence in the ability of the tests to detect a problem.  If the code that will be eventually changed is not covered by a test then I will spend a little time trying to write a test that does cover where the changes will be.  

This is still a draft but addressing testing would give it a higher priority in my queue when it becomes reviewable :-)

fredg1:
> 
> 
> This is still a draft but addressing testing would give it a higher priority in my queue when it becomes reviewable :-)

Actually, this PR is only a draft BECAUSE OF the lacking test. In other words, the wanted changes are all there, but I "know" that a test for FoldItemCommand (at the very least) will be required for it to be accepted.
(I, myself, cannot even be 100% sure that I didn't make a mistake/miss anything, because of the nature of shifting that 1-indexed list to 0-indexed)

gausie (approved):

PR #175, opened 2021-11-01
RequestTestBase upgrade

fredg1:


test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +82,105 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+  /** Set {@link #factory} to generate the given response if the path matches {@code pattern}. */
+  protected ArgumentMatcher<URL> respondIfMatches(final Pattern pattern, final String response) {
+    return respondIfMatches(Collections.singletonList(pattern), response);
+  }
+
+  /**
+   * Set {@link #factory} to generate the given response if the path matches every element of {@code
+   * patterns}.
+   */
+  protected ArgumentMatcher<URL> respondIfMatches(
```
heeheehee-kolmafia, line 118: I'm not sure I understand the use case for all of these overloads.

You want to check that the path matches multiple regex patterns? and/or multiple strings? Are you trying to match multiple query parameters that may show up in an uncertain order?

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -1,22 +1,72 @@
...
+ *   <li>For {@link GenericRequest}s made locally, gives access to {@link
+ *       RequestTestBase#expectSuccess}, allowing the simulation of running a {@link
+ *       org.mockito.Mockito#spy spy} of that request.
+ *   <li>For the rest, gives access to {@link #respondIfContains}/{@link #respondIfMatches}, causing
+ *       future executed requests to succeed with the given response if the path of the URL
+ *       submitted contains the submitted text(s)/matches the submitted pattern(s)/matches the
+ *       submitted {@link ArgumentMatcher}. These methods all return the {@link ArgumentMatcher}
+ *       submitted for the stubbing. This allows them to be re-used to {@link
+ *       org.mockito.Mockito#verify verify} the access of the method (using the {@link
+ *       #verifyRequest} shortcut method).
```
heeheehee-kolmafia, line 43: You shouldn't normally need to directly verify() your requests. Instead, prefer to use strict stubs (which are planned to be the default in Mockito 4).

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -1,22 +1,72 @@
...
+    HttpUtilities.setOpen(factory);
+  }
+
   @AfterAll
   static final void cleanSessionId() {
     GenericRequest.sessionId = null;
   }
 
+  @AfterAll
+  static final void reBlockNetwork() {
```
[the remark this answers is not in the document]
fredg1: doesn't "BeforeAll" mean it will only run once (talking about ForbidNetworkAccess.beforeAll() )? Otherwise, it would be "BeforeEach"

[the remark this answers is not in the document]
fredg1: Also, no, `openNetwork` needs to be `BeforeEach`, as the factory needs to be refreshed between the tests (otherwise the stubs will carry between tests)

test/net/sourceforge/kolmafia/request/RequestTestBase.java
fredg1: The plan is to expect that people will want to match various parameters in URLs.

We don't want to force an "order" to those parameters.
For example,
```java
{"inventory.php"},
{"which=2"}
```
can match "inventory.php?which=2&pwd=\<pwd\>"
as well as "inventory.php?pwd=\<pwd\>&which=2"

heeheehee-kolmafia: I would prefer matchers for query parameters in that case.

For instance, something like:

    argThat(allOf(hasPath("inventory.php"), hasParam("which", 2)))

Although, even that is incorrect if you've got POST parameters that aren't part of the query string but are instead in the POST body.

heeheehee-kolmafia: In particular, checking for the substring "inventory.php" will incorrectly match a page "subinventory.php", and checking for "which=2" will incorrectly match "which=20" and "itemwhich=2".

fredg1: Mh... That seems possible + an interesting challenge.

This might take the rest of the day, at least. Is it possible to mark these as "changes requested" so that I can notify when done?

heeheehee-kolmafia (changes requested):
requesting changes as requested

jaadams5 (dismissed):
I am inclined to defer to @heeheehee-kolmafia for approval but if the requested changes have been resolved then I can merge.

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,175 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+    @Override
+    public boolean usingProxy() {
+      return false;
+    }
+
+    @Override
+    public void connect() throws IOException {}
+  }
+
+  private class SectionedUrl {
```
heeheehee-kolmafia, line 193: I'd prefer to call this something like "ParsedUrl" personally.

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,175 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+      final int formSplitIndex = fullPath.indexOf("?");
+
+      if (formSplitIndex == -1) {
+        this.path = fullPath;
+      } else {
+        this.path = GenericRequest.decodePath(fullPath.substring(0, formSplitIndex));
+
+        final String[] paramsString = fullPath.substring(formSplitIndex + 1).split("&");
+        for (final String param : paramsString) {
+          if (param.length() == 0) {
```
heeheehee-kolmafia, line 208: nit: param.isEmpty()

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,175 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+    private SectionedUrl(final URL url) {
+      final String fullPath = url.getPath();
+      final int formSplitIndex = fullPath.indexOf("?");
+
+      if (formSplitIndex == -1) {
+        this.path = fullPath;
+      } else {
+        this.path = GenericRequest.decodePath(fullPath.substring(0, formSplitIndex));
+
+        final String[] paramsString = fullPath.substring(formSplitIndex + 1).split("&");
```
heeheehee-kolmafia, line 206: Instead of fullPath.substring(...), use fullPath.getQuery(). Similarly above, check that getQuery() != null instead of using indexOf("?").

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,175 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+    @Override
+    public void connect() throws IOException {}
+  }
+
+  private class SectionedUrl {
+    private final String path;
+    private final Map<String, String> params = new HashMap<>();
+
+    private SectionedUrl(final URL url) {
+      final String fullPath = url.getPath();
```
heeheehee-kolmafia, line 198: This won't work.

getPath() only returns the part of the path prior to the query and/or fragment.

heeheehee-kolmafia (changes requested):

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,171 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+   * Set {@link #factory} to generate the given response if the path matches every one of {@code
+   * matchers}.
+   *
+   * <p>Use {@link #hasPath} and/or {@link #hasParam} to generate them.
+   */
+  protected void respondIf(final String response, final Iterable<UrlMatcher> matchers) {
+    prepareResponse(
+        response,
+        url -> {
+          final ParsedUrl sectionedUrl = new ParsedUrl(url);
```
heeheehee-kolmafia, line 110: nit: rename the variable name to parsedUrl?

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,171 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
...
+    boolean matches(final ParsedUrl url) {
+      return this.matcher.matches(url.path);
+    }
+  }
+
+  private class ParamMatcher extends UrlMatcher {
+    final ArgumentMatcher<Entry<String, String>> matcher;
+
+    private ParamMatcher(final String param, final String value) {
+      this.matcher =
```
heeheehee-kolmafia, line 235: Why not store the String param and an ArgumentMatcher\<String> matcher?

That way instead of iterating through all the entries of params, you can use something like `matcher.matches(params.get(param))`.

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -32,4 +86,171 @@ protected void expectSuccess(GenericRequest spy, String responseText) {
         .when(spy)
         .externalExecute();
   }
+
+  /**
+   * Set {@link #factory} to generate the given response if the path matches every one of {@code
```
heeheehee-kolmafia, line 91: nit: "Sets ..."

"This is a fragment—a noun phrase or verb phrase, not a complete sentence. It does not begin with `A {@code Foo} is a...`, or `This method returns...`, nor does it form a complete imperative sentence like `Save the record.`. However, the fragment is capitalized and punctuated as if it were a complete sentence."

https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment

Ditto elsewhere.

test/net/sourceforge/kolmafia/request/RequestTestBase.java
```diff
@@ -1,17 +1,71 @@
...
 import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.mockito.ArgumentMatcher;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoSession;
+import org.mockito.quality.Strictness;
 
+/**
+ * Extend this class when making tests involving {@link GenericRequest} and/or its descendants.
```
heeheehee-kolmafia, line 33: nit: this should be in the body of the javadoc describing how to use it, not the description of what it is.

heeheehee-kolmafia (dismissed):
I don't have much more input on this front.

I encourage anyone who's actually actively playing the game and/or engaged in active development of new features to weigh in regarding how they would want to use such a library.

@gausie @jaadams5

test/net/sourceforge/kolmafia/request/RequestTestBase.java
fredg1: I think I was still thinking about users maybe wanting to also submit a Pattern for the key.
I changed my mind mid-way, as the keys are pretty much specific and predictable enough that the user will know what to expect.

gausie (approved):
This is cool and a well-needed base for better testing in the project. It all looks good but this will be properly tested by an implementation. Thank you for writing it.

Will wait for heeheehee to approve your changes before merging

test/net/sourceforge/kolmafia/request/RequestTestBase.java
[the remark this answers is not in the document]
jaadams5: I think I agree with BeforeEach so am marking resolved.

jaadams5 (approved):
I'm going to approve with one comment.  @fredg1 please do what you can to make sure that this plays nicely with #220.  My life would be easier not having a merge  conflict or tests that stop playing nicely together.  I'll push tomorrow unless some other dev beats me to it.

fredg1:
@jaadams5 do you want me to pull #220 into it, or do you think it may take longer?
There are indeed merge conflicts with it.

jaadams5:
> @jaadams5 do you want me to pull #220 into it, or do you think it may take longer? There are indeed merge conflicts with it.

I am lazy.  If this gets committed first the I'm going to have to resolve merge conflicts in #220.  If this gets committed second I'm going to have to resolve merge conflicts because I press the button.

If there is an easy way for you to resolve the conflicts instead of me then I am in favor of it.  But if this one goes first there are at most two files to deal with so maybe I'm making a mountain out of a molehill?

fredg1:
You could tell me when #220 is planned to be merged, and I could resolve the conflicts myself right after.

jaadams5:
> You could tell me when #220 is planned to be merged, and I could resolve the conflicts myself right after.

That will work.  I will ping you when I close 220 and be prepared to reapprove and merge this.

jaadams5:
@fredg1 OK - please resolve conflicts for me.  Thank you.

jaadams5 (approved):

PR #177, opened 2021-11-02
Universal Seasonings can be used multiple times, retire old bool pref and add new int one

gausie:


gausie:
This is actually dependent on inventory so I'll need to make a change to this when I can

jaadams5:
> This is actually dependent on inventory so I'll need to make a change to this when I can

Does this mean it is Draft at the moment?

Will the questions about versioning and breaking scripts be rendered moot?

gausie:
> > This is actually dependent on inventory so I'll need to make a change to this when I can
> 
> Does this mean it is Draft at the moment?
> 
> Will the questions about versioning and breaking scripts be rendered moot?

They won't be rendered moot per se, but they won't be urgent and seriously affect any scripts (and I doubt universal seasoning is widely scripted anyway). I'll make an announcement on kolmafia.us once this is merged

gausie:
Ok I think the checking for this is actually fine, but I should add an `int _universalSeasoningsActive` pref in the future.

PR #178, opened 2021-11-02
No-effect-assignments

fredg1:
Removal of assignments doing nothing.

The source of the existence of these useless assignments is documented in the commits.

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -233,8 +233,6 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
         RequestLogger.printLine(
             "Effect database error: id = " + this.id + " name = \"" + this.name + "\"");
       }
-    } else {
-      this.name = this.name;
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/net/sourceforge/kolmafia/request/BatFellowRequest.java
```diff
@@ -59,7 +59,7 @@ else if (action.equals("batman_industrial_car")) {
       return false;
     }
 
-    String message = message = "{" + BatManager.getTimeLeftString() + "} " + location;
+    String message = "{" + BatManager.getTimeLeftString() + "} " + location;
```
jaadams5, line 62: Weird.  That's been like that since 2016.

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -233,8 +233,6 @@ public void normalizeEffectName() {
         RequestLogger.printLine(
             "Effect database error: id = " + this.id + " name = \"" + this.name + "\"");
       }
-    } else {
```
jaadams5, line 236: It took too much work to confirm the else did nothing but this is ok.

jaadams5 (approved):

PR #180, opened 2021-11-02
formatting errors caused by end-of-line comments

fredg1:
I can't assert that I definitely got everything; VScode was *really* not cooperative with this...

### DIFF
--- a/src/net/sourceforge/kolmafia/AreaCombatData.java
+++ b/src/net/sourceforge/kolmafia/AreaCombatData.java
@@ -887,9 +887,7 @@ public class AreaCombatData {
     } else if (weighting == -4) {
       buffer.append("0%");
     } else if (weighting == 0) {
-      {
-        buffer.append("special");
-      }
+      buffer.append("special");
     } else {
       buffer.append(
           this.format(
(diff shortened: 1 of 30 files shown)


jaadams5:
What does this do that gradlew spotlessapply didn't?

heeheehee-kolmafia:
From a quick skim, spotless doesn't want to move comments relative to code. So, for instance,

```
if (someCondition)
// here's a comment
{
...
}
```

will not get reformatted as
```
if (someCondition) {
  // here's a comment
  ...
}
```

jaadams5 (commented):
Have you built and then run a few turns with a character with these changes?  I'll approve when the answer is yes but the Merge will happen sooner of someone else makes that check.

fredg1:
I haven't...

jaadams5:
> I haven't...

Would you consider doing so or does this wait until someone else gets around to it?

:-)

fredg1:
I can give it a go. What exactly am I supposed to be looking for, other than a sudden debug log?

jaadams5:
> I can give it a go. What exactly am I supposed to be looking for, other than a sudden debug log?

That and just something that does not work as you expect it to.  I have found changes that stopped the app from starting, stopped windows from updating or displaying, scripts that didn't run, etc.  I'm not expecting any issues but with 30 files changed and two of us thinking nothing should change, actually building and burning turns increases confidence in the patch.

fredg1:
Lot of errors all right... but this branch was based off of 25831, i.e. right before https://github.com/kolmafia/kolmafia/pull/181...
Let's pull from upstream and retry...

fredg1:
Way better now~~
Nothing to report.

jaadams5 (approved):

PR #181, opened 2021-11-02
Fix hiccup

jaadams5:
Lint saw a comparison between what was declared as two strings and suggested it be replaced by .equals

I did so but it turns out the first string can be legitimately null.  Fixed that.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/RelayRequest.java
+++ b/src/net/sourceforge/kolmafia/request/RelayRequest.java
@@ -3370,7 +3370,7 @@ public class RelayRequest extends PasswordHashRequest {
     String limitmode = KoLCharacter.getLimitmode();
 
     // If we are playing Spelunky, a specialized set of warnings are relevant
-    if (limitmode.equals(Limitmode.SPELUNKY)) {
+    if ((limitmode != null) && limitmode.equals(Limitmode.SPELUNKY)) {
       return this.sendSpelunkyWarning(adventure);
     }
 


src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3370,7 +3370,7 @@ private boolean sendWarnings(KoLAdventure adventure, String adventureName, Strin
     String limitmode = KoLCharacter.getLimitmode();
 
     // If we are playing Spelunky, a specialized set of warnings are relevant
-    if (limitmode.equals(Limitmode.SPELUNKY)) {
+    if ((limitmode != null) && limitmode.equals(Limitmode.SPELUNKY)) {
```
heeheehee-kolmafia, line 3373: I don't think you need these parentheses.

As an alternative, you could switch the order of comparison:

SPELUNKY.equals(limitmode)

heeheehee-kolmafia (approved):
your call if you want to make the change I suggested; I'll reapprove if needed

src/net/sourceforge/kolmafia/request/RelayRequest.java
jaadams5: As a defensive measure I tend to use extra parens instead of trying to recall the evaluation hierarchy.  

limitmode is declared as a String but null also is expected and has meaning.  This formulation makes that slightly more obvious, at least until limitmode becomes an Enum.

I'll leave it as is since that gets this fix out faster.  Thank you.

PR #182, opened 2021-11-03
WeakReferences raw type

fredg1:
They are for listeners!

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/panel/GenericPanel.java
+++ b/src/net/sourceforge/kolmafia/swingui/panel/GenericPanel.java
@@ -35,7 +35,7 @@ public abstract class GenericPanel extends ActionVerifyPanel {
   protected ConfirmedListener CONFIRM_LISTENER = new ConfirmedListener();
   protected CancelledListener CANCEL_LISTENER = new CancelledListener();
 
-  protected HashMap listenerMap;
+  protected HashMap<Object, WeakReference<ActionConfirmListener>> listenerMap;
 
   public JPanel southContainer;
   public JPanel actionStatusPanel;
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/swingui/panel/GenericPanel.java
```diff
@@ -171,26 +171,26 @@ public void addListeners() {
...
           (JTextComponent) ((AutoFilterComboBox) component).getEditor().getEditorComponent();
 
       editor.addKeyListener(listener);
-      this.listenerMap.put(editor, new WeakReference(listener));
-    } else if (component instanceof JComboBox && ((JComboBox) component).isEditable()) {
+      this.listenerMap.put(editor, new WeakReference<>(listener));
+    } else if (component instanceof JComboBox && ((JComboBox<?>) component).isEditable()) {
       JTextComponent editor =
-          (JTextComponent) ((JComboBox) component).getEditor().getEditorComponent();
+          (JTextComponent) ((JComboBox<?>) component).getEditor().getEditorComponent();
```
heeheehee-kolmafia, line 190: (I would like to weaken this to just Component editor = ..getEditorComponent() since addKeyListener doesn't depend on those concrete details, but that's irrelevant to this change.)

src/net/sourceforge/kolmafia/swingui/panel/GenericPanel.java
```diff
@@ -35,7 +35,7 @@
   protected ConfirmedListener CONFIRM_LISTENER = new ConfirmedListener();
   protected CancelledListener CANCEL_LISTENER = new CancelledListener();
 
-  protected HashMap listenerMap;
+  protected HashMap<Object, WeakReference<ActionConfirmListener>> listenerMap;
```
heeheehee-kolmafia, line 38: Should this be JTextComponent instead of Object?

heeheehee-kolmafia (changes requested):

src/net/sourceforge/kolmafia/swingui/panel/GenericPanel.java
fredg1: I didn't notice that indeed! Nice catch

heeheehee-kolmafia (approved):

PR #183, opened 2021-11-03
Revisit Security

jaadams5:
Previous attempt to make security scanner happy didn't.  Try this.

New Preferences tests kept breaking old ones.  Decided to move the setup and cleanup into Before/After Each instead of All.  Seems to work much better.  Might be willing and able to deprecate ClearSharedState but for the moment just duplicated code from there instead.

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -530,17 +530,14 @@ public class Preferences {
     }
   }
 
-  private static void loadUserPreferences(final String username) {
-    String lusername;
+  private static void loadUserPreferences(String username) {
     // Apparently the CodeQL security scan requires this as a fix...
     if ((username != null) && (username.contains(".."))) {
-      lusername = null;
-    } else {
-      lusername = username;
+      username = null;
     }
     File file =
         new File(
-            KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(lusername) + "_prefs.txt");
+            KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.txt");
     Preferences.userPropertiesFile = file;
 
     Properties p = Preferences.loadPreferences(file);
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -536,8 +536,7 @@ public class Preferences {
       username = null;
     }
     File file =
-        new File(
-            KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.txt");
+        new File(KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.txt");
     Preferences.userPropertiesFile = file;
 
     Properties p = Preferences.loadPreferences(file);
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -531,12 +531,16 @@ public class Preferences {
   }
 
   private static void loadUserPreferences(final String username) {
+    String lusername;
     // Apparently the CodeQL security scan requires this as a fix...
     if ((username != null) && (username.contains(".."))) {
-      return;
+      lusername = null;
+    } else {
+      lusername = username;
     }
     File file =
-        new File(KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.txt");
+        new File(
+            KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(lusername) + "_prefs.txt");
     Preferences.userPropertiesFile = file;
 
     Properties p = Preferences.loadPreferences(file);
--- a/test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
+++ b/test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
@@ -551,10 +551,50 @@ class PreferencesTest {
   }
 
   @Test
-  public void exerciseReset() {
+  public void exerciseResetNull() {
+    // Allow files to be written
+    Preferences.saveSettingsToFile = true;
+    // Global preferences name
+    String globalName = "settings/" + "GLOBAL" + "_prefs.txt";
+    File globalfile = new File(globalName);
+    if (globalfile.exists()) {
+      globalfile.delete();
+    }
+    assertFalse(globalfile.exists());
+    // Reset should save global.
     Preferences.reset(null);
+    assertTrue(globalfile.exists());
+  }
+
+  @Test
+  public void exerciseResetEmpty() {
+    // Allow files to be written
+    Preferences.saveSettingsToFile = true;
+    // Global preferences name
+    String globalName = "settings/" + "GLOBAL" + "_prefs.txt";
+    File globalfile = new File(globalName);
+    if (globalfile.exists()) {
+      globalfile.delete();
+    }
+    assertFalse(globalfile.exists());
+    // Reset should save global.
     Preferences.reset("");
+    assertTrue(globalfile.exists());
+  }
+
+  @Test
+  public void exerciseResetDots() {
+    // Allow files to be written
+    Preferences.saveSettingsToFile = true;
+    // Global preferences name
+    String globalName = "settings/" + "GLOBAL" + "_prefs.txt";
+    File globalfile = new File(globalName);
+    if (globalfile.exists()) {
+      globalfile.delete();
+    }
+    assertFalse(globalfile.exists());
+    // Reset should save global.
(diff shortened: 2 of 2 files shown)


src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -531,12 +531,16 @@ private static void loadGlobalPreferences() {
   }
 
   private static void loadUserPreferences(final String username) {
+    String lusername;
```
heeheehee-kolmafia, line 534: Instead of introducing a new variable, why not just remove the final qualifier from username?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -2,25 +2,59 @@
...
+    KoLmafiaCLI.registerCommands();
+    KoLmafia.lastMessage = "";
+    KoLmafia.forceContinue();
+  }
+
+  public static void deleteUserPrefsAndMoodsFiles(String user) {
+    String begin = "settings/" + user;
+    File file = new File(begin + "_prefs.txt");
+    if (file.exists()) {
+      file.deleteOnExit();
```
heeheehee-kolmafia, line 41: Do you want to delete this only when the VM terminates? This reads like it has the potential to leak state between tests.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -488,12 +522,12 @@ public void exerciseIsPerUserGlobalProperty() {
   @Test
   public void actuallySaveFileToIncreaseCoverage() {
```
heeheehee-kolmafia, line 523: In general, I would prefer if tests actually assert something about the resulting state, rather than just calling a bunch of methods with side effects.

Coverage that doesn't check anything about the code only tells you that the happy path doesn't cause crashes, not that the code is correct.

(ditto below with exerciseReset)

heeheehee-kolmafia (commented):
I don't think we should deprecate ClearSharedState, as the entire point of it is to serve as a catch-all in case a test author accidentally leaks something (e.g. a preference).

src/net/sourceforge/kolmafia/preferences/Preferences.java
jaadams5: I had no idea why someone had declared it static so it was easier to leave it that way.  Habit from pre-Java where call by reference and call by value were considerations.  I'll change.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
jaadams5: I watched this and no tests fail and test\root\sessions was empty at the end so it did what I wanted.  file.deleteOnExit() is a defensive move from a multi-threaded environment where file.delete() executed immediately and caused errors on other threads.  I'd like to understand more about how JUnit is threading before I change to file.delete() since everything I can observe at the OS and debugger level says this is working and there are no leaks.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
jaadams5: I agree in general.  But a test that covers a change is better than a change with no coverage.  It is not clear but this test did pass before the change to Preferences as well as after.  I'm open to suggestions as to how to test something that is not visible, produces only side effects and verifying the side effect is irrelevant to the code that was changed.  I can ponder this a little harder.

jaadams5: Not static.  No new variable.

jaadams5:
> I don't think we should deprecate ClearSharedState, as the entire point of it is to serve as a catch-all in case a test author accidentally leaks something (e.g. a preference).

I haven't seen evidence that it is preventing any leakage but deprecation was a flip comment based upon an observation that it didn't seem to be working to prevent leaks.  A lot more study before that is a serious recommendation.

midgleyc:
`ClearSharedState` fixed an infinite loop that only happened on MacOS in #170. I think it's convenient.

jaadams5:
> `ClearSharedState` fixed an infinite loop that only happened on MacOS in #170. I think it's convenient.

Any chance you could try running this on a Mac and see whether tests behave?

I'm not going to rip out ClearSharedState any time soon but it did not fix leaking between tests contained in PreferenceTest and at the moment there are no other tests that are actually using preferences and leaking.  

ClearSharedState is currently wired to that it runs after all tests are completed and the leakage is addressed after each test completes.  There may be a performance trade off based upon what ClearSharedState does and how often it does it.  

If there is no performance hit then I think the end state should be a ClearSharedState that runs after every test guarantees a known state for each test.  But that is not this PR :-)

midgleyc:
> Any chance you could try running this on a Mac and see whether tests behave?

Unfortunately I don't have a Mac to test with. I could only tell by using the GitHub actions runner.

jaadams5:
OK.  I guess I need to learn about the actions runner.  Thank you.

jaadams5:
Tweaked tests to be less about coverage :-)

BadHorseMonkey:
> > Any chance you could try running this on a Mac and see whether tests behave?
> 
> Unfortunately I don't have a Mac to test with. I could only tell by using the GitHub actions runner.

I can test on a Mac, but probably not until Sunday.

jaadams5:
> I can test on a Mac, but probably not until Sunday.

I'll try and be patient, then.

heeheehee-kolmafia:
> I haven't seen evidence that it is preventing any leakage but deprecation was a flip comment based upon an observation that it didn't seem to be working to prevent leaks. A lot more study before that is a serious recommendation.

You're evidently forgetting the context in which it was introduced, where we were adding reset() and the like to clean up after other tests.

ClearSharedState was never meant to catch 100% of shared state leakages across tests, but rather to be a last line of defense to avoid known issues that have bitten developers in the past.

If you really want to insist that there is no leakage of shared state across test classes whatsoever, then set [forkEvery = 1](https://docs.gradle.org/current/userguide/java_testing.html#sec:test_execution). ClearSharedState is a compromise between increasing test execution time by 2x or worse, versus having flaky test failures.

jaadams5:
Maybe there is a philosophy issue at odds here?

I set forkEvery = 1 and have three failing tests without any of the changes in this PR.

I removed the Before/After each from this version of PreferencesTest, included it and got 19 more failing tests.

I restored the Before/After each and was back to the three failing tests.

So there is leakage to be tracked down.

Once that is done then there is certainly code embedded in tests that ideally should be in ClearSharedState.

I wonder if, when the errors discovered by forkEvery =1 are resolved and committed, whether the tests run on GitHub as part of the commit checks should also use forkEvery =1.

We don't have to resolve ClearSharedState and forkEvery =1 here because there will be another PR once this one is committed.

heeheehee-kolmafia:
> Maybe there is a philosophy issue at odds here?
> 
> I set forkEvery = 1 and have three failing tests without any of the changes in this PR.

If there are failing tests _with_ forkEvery=1, that means they're relying on teardown logic that should instead be setup logic. That said, I agree that we should resolve those issues, as those are test reordering races waiting to happen.

Two of the failures are in UseItemRequestTest. Those are tied to not setting saveSettingsToFile, and tests therefore persisting state across individual methods by way of writing preference files when calling KoLCharacter.reset().

The third failure is in AshInteropTest, which expects KoLmafia.lastMessage to be empty, when it's initialized to a single space.

BadHorseMonkey:
> > I haven't seen evidence that it is preventing any leakage but deprecation was a flip comment based upon an observation that it didn't seem to be working to prevent leaks. A lot more study before that is a serious recommendation.
> 
> You're evidently forgetting the context in which it was introduced, where we were adding reset() and the like to clean up after other tests.
> 
> ClearSharedState was never meant to catch 100% of shared state leakages across tests, but rather to be a last line of defense to avoid known issues that have bitten developers in the past.
> 
> If you really want to insist that there is no leakage of shared state across test classes whatsoever, then set [forkEvery = 1](https://docs.gradle.org/current/userguide/java_testing.html#sec:test_execution). ClearSharedState is a compromise between increasing test execution time by 2x or worse, versus having flaky test failures.

I was thinking that perhaps the tests that do want to write to the user folder might want to set an individual (and thus isolated) location for them.

BadHorseMonkey:
> > I can test on a Mac, but probably not until Sunday.
> 
> I'll try and be patient, then.

Or I could've underestimated my business!  Current results:
ran `./gradlew clean test --rerun-tasks` twice, to see if tests failed after first run.


641 tests | 0 failures | 7 ignored | 17.798s duration

jaadams5:
OK.  I'll take a review here and start working on forkEvery=1 failures.  When that is resolved I will start looking at tests that have their own before/after and consider which ones should be replaced by an improved ClearSharedState.  I personally am not prepared to try and build an infrastructure so sets up and tears down directories on a test by test basis although that might change after I finish forkEvery=1 excursions.

jaadams5:
I'm going to self approve...

PR #184, opened 2021-11-03
JComboBoxes' raw type

fredg1:
This PR *ONLY* adds Generics, and *ONLY* to JComboBoxes (was hard to restrain myself from even just adding generics to other nearby types).

\<?\> for when we can't tell (it's a parameter, and we're just passing it on)
\<String\> for when it's only fed strings (most of them)
\<Object\> for when it's a mix of strings and something else (sadly...)

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
+++ b/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
@@ -1629,15 +1629,14 @@ public class OptionsFrame extends GenericFrame {
                   {"allowCloseableDesktopTabs", "Allow tabs on main window to be closed"},
                 };
 
-    private final JComboBox<Object> looks;
-    private final JComboBox<String> toolbars, toolIcons, scripts;
+    private final JComboBox<String> looks, toolbars, toolIcons, scripts;
 
     public UserInterfacePanel() {
       super(new Dimension(80, 22), new Dimension(280, 22));
       PreferenceListenerRegistry.registerPreferenceListener("swingLookAndFeel", this);
 
       UIManager.LookAndFeelInfo[] installed = UIManager.getInstalledLookAndFeels();
-      Object[] installedLooks = new Object[installed.length + 1];
+      String[] installedLooks = new String[installed.length + 1];
       String CurrentLook = getLookAndFeel().getClass().getName();
 
       installedLooks[0] = "Always use OS default look and feel";
--- a/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
+++ b/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
@@ -1628,7 +1628,8 @@ public class OptionsFrame extends GenericFrame {
                   {"allowCloseableDesktopTabs", "Allow tabs on main window to be closed"},
                 };
 
-    private final JComboBox looks, toolbars, toolIcons, scripts;
+    private final JComboBox<Object> looks;
+    private final JComboBox<String> toolbars, toolIcons, scripts;
 
     public UserInterfacePanel() {
       super(new Dimension(80, 22), new Dimension(280, 22));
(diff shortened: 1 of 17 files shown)


gausie (commented):
Is there a way we can narrow `<?>` at all? No is an acceptable answer

fredg1:
Nope.
Even if we could, these are only in methods for which it doesn't matter, meaning it would only be detrimental for the future (if someone else wanted to use that method, they'd have to match the generic we narrowed it down to, even though they shouldn't have to)

gausie:
So there is no requirement whatsoever from the generic. Hmm.. I wonder if that would be clearer then if we made it `<? extends Object>` explicitly or something. That way someone looking at it would not assume it was just lazy, as `<?>` might otherwise look. Just musing on it.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1628,7 +1628,8 @@ public void update() {
                   {"allowCloseableDesktopTabs", "Allow tabs on main window to be closed"},
                 };
 
-    private final JComboBox looks, toolbars, toolIcons, scripts;
+    private final JComboBox<Object> looks;
```
heeheehee-kolmafia, line 1631: Can this be JComboBox<String>? It looks like we either add the constant string "Always use OS default look and feel" or we add installed[i].getName().

heeheehee-kolmafia (changes requested):
Mostly looks reasonable, just one potential instance of a bare Object we might be able to remove.

src/net/sourceforge/kolmafia/swingui/panel/AdventureSelectPanel.java
```diff
@@ -535,12 +535,12 @@ public static JPanel getAdventureSummary(final String property) {
     private final String property;
     private final CardLayout resultCards;
     private final JPanel resultPanel;
-    private final JComboBox resultSelect;
+    private final JComboBox<?> resultSelect;
```
heeheehee-kolmafia, line 538: This instance looks like it's only populated above in this file with a JComboBox<String>, right?

src/net/sourceforge/kolmafia/swingui/panel/RestoreOptionsPanel.java
```diff
@@ -148,7 +148,7 @@ private String getSettingString(final JCheckBox[] restoreCheckbox) {
     return restoreSetting.toString();
   }
 
-  private float getPercentage(final JComboBox option) {
+  private float getPercentage(final JComboBox<?> option) {
```
heeheehee-kolmafia, line 151: Similarly, can this be replaced with a JComboBox<String>? (ditto below in setSelectedIndex)

heeheehee-kolmafia (commented):
(optional things to consider, as these are cases of <?> in parameters when we _don't care_ and are just referencing the selected index or something like that.)

src/net/sourceforge/kolmafia/swingui/panel/RestoreOptionsPanel.java
fredg1: sure, I guess, since they are private methods... not like someone is gonna pop in with a JComboBox of a different generic and will be sad they can't use that method...

heeheehee-kolmafia (approved):
This looks good to me, although I'll let @gausie make the decision on merging it, since I totally missed that he had started the review process earlier.

fredg1:
> 
> 
> This looks good to me, although I'll let @gausie make the decision on merging it, since I totally missed that he had started the review process earlier.

Well, to be fair, the only two instance of <?> that are left are of this exact scenario:
```java


  protected boolean isValidEvent(final ActionEvent e) {
    if (e == null || e.getSource() == null) {
      return true;
    }

    if (e.getSource() instanceof JComboBox) {
      return ((JComboBox<?>) e.getSource()).isPopupVisible();
    }

    return true;
  }```

Since we're casting from a non-generic type, trying to turn them into
```java


  protected boolean isValidEvent(final ActionEvent e) {
    if (e == null || e.getSource() == null) {
      return true;
    }

    if (e.getSource() instanceof JComboBox) {
      return ((JComboBox<? extends Object>) e.getSource()).isPopupVisible();
    }

    return true;
  }```
would have the compiler complain about an unchecked cast...

PR #191, opened 2021-11-05
forkEvery = 1

jaadams5:
Pushing for eventual collaboration.  First, set forkEvery = 1 and fix failing tests.


jaadams5:
Test failures are expected :-)

jaadams5:
@heeheehee-kolmafia @BadHorseMonkey @gausie 

Help?  Test passes on my machine but not here.  How do I force GitHub to run the Windows and Mac tests even if the Linux test fails?  The failure is a CustomScriptTest show_skills.txt and it looks like a whitespace issue.

Also I'd be much happier if someone else approved #183 or continued the discussion in light of what is happening here.  My intention is to have everything working with 1 fork and then try and consolidate what works to isolate tests in  ClearSharedState.

heeheehee-kolmafia:
I had a different set of changes in mind:

- initialize `KoLmafia.lastUpdate = "";` instead of `" "`
- set Preferences.saveSettingsToFile in milkSetup().

(as in https://github.com/heeheehee-kolmafia/kolmafia/commit/429adac32fbb4681b9b369cfad5ecacf902b2f06)

I guess you could also just change ClearSharedState to be a BeforeAllCallback instead of an AfterAllCallback.

jaadams5:
> * initialize `KoLmafia.lastUpdate = "";` instead of `" "`

Yeah.  I think I am going to introduce a NO_MESSAGE constant and use it everywhere.  The core issue is that some non-test portions of the code think no message is "" but some think it is " ".

> * set Preferences.saveSettingsToFile in milkSetup().

That did not change any failing tests to passing tests :-(


> I guess you could also just change ClearSharedState to be a BeforeAllCallback instead of an AfterAllCallback.

I feel like we are talking past each other on ClearSharedState.  You focus on Before/After but my experience with preferences has been that All/Each is the root of the solution.  I suspect you know something I don't.  In any event Step 1 for me is to get all known tests working with numFork=1.  When that is the case then I will try and consolidate the setup and cleanup needed to make the tests work in one place, i.e. ClearSharedState.

No love for #183?  :-)

BadHorseMonkey:
I have *a* solution, but it may not be *a good* solution.  I do not know why in one case we get a leading CR and in the other two we do not.  Or, indeed if it's consistent.

jaadams5:
> I have _a_ solution, but it may not be _a good_ solution. I do not know why in one case we get a leading CR and in the other two we do not. Or, indeed if it's consistent.

I think I would like to make this non-Draft, commit to main after review/discussion or time and then start a new effort to consolidate all of the state preservation and clearing in ClearSharedState.

test/net/sourceforge/kolmafia/utilities/FileUtilitiesTest.java
```diff
@@ -42,5 +42,6 @@ public void exerciseSomeMethodsForCoverage() {
     FileUtilities.copyFile(source, destination);
     assertTrue(destination.exists());
     destination.delete();
+    source.delete();
```
BadHorseMonkey, line 45: I was a little concerned by this, but it doesn't seem to be changing my source tree, so I'm OK with it.

BadHorseMonkey (dismissed):

test/net/sourceforge/kolmafia/utilities/FileUtilitiesTest.java
jaadams5: In a test environment, everything is under test\root and AFAIK this test only deletes what it created.  That said this test looks confusing to me now and I'd like more confidence that nothing is coming from "outside".  Maybe I'll shelve this for now.

jaadams5:
I'm going to commit and then circle back to clarify the file utilities test and start consolidating things into CkearSharedState.  Not having this committed is an issue because the state of test\root does not clear when changing branches.

PR #193, opened 2021-11-06
cli: add ability to autoconfirm sending spading data

midgleyc:
If you have a lot of spading data, confirming each time is a bit annoying. This adds a command to confirm everything pre-emptively.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/LoginManager.java
+++ b/src/net/sourceforge/kolmafia/session/LoginManager.java
@@ -216,8 +216,9 @@ public class LoginManager {
     if (Preferences.getString("spadingData").length() > 10) {
       KoLmafia.updateDisplay(
           "Some data has been collected that may be of interest to others. "
-              + "Please type `spade' to examine and submit or delete this data. If you have a lot, "
-              + "typing `spade autoconfirm' will send all of it without prompting you for each bit.");
+              + "Please type `spade' to examine and optionally submit the data or `spade autoconfirm'"
+              + " to submit all of the spaded data. Either way the data will be deleted whether shared"
+              + " or not.");
     }
 
     // Rebuild Scripts menu if needed
--- a/src/net/sourceforge/kolmafia/session/LoginManager.java
+++ b/src/net/sourceforge/kolmafia/session/LoginManager.java
@@ -215,8 +215,9 @@ public class LoginManager {
 
     if (Preferences.getString("spadingData").length() > 10) {
       KoLmafia.updateDisplay(
-          "Some data has been collected that may be of interest "
-              + "to others.  Please type `spade' to examine and submit or delete this data.");
+          "Some data has been collected that may be of interest to others. "
+              + "Please type `spade' to examine and submit or delete this data. If you have a lot, "
+              + "typing `spade autoconfirm' will send all of it without prompting you for each bit.");
     }
 
     // Rebuild Scripts menu if needed
--- a/src/net/sourceforge/kolmafia/textui/command/SubmitSpadeDataCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/SubmitSpadeDataCommand.java
@@ -9,7 +9,7 @@ import net.sourceforge.kolmafia.utilities.InputFieldUtilities;
 
 public class SubmitSpadeDataCommand extends AbstractCommand {
   public SubmitSpadeDataCommand() {
-    this.usage = " [prices <URL>] - submit automatically gathered data.";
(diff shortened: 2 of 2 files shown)


jaadams5:
Not about the patch but the use case.  I don't see anything in mafia that is actually collecting spading data so this change is presumably to make a script easier to use?  @gausie 's Excavator has code that bypasses the need for confirmation.  Not sure it is still being maintained, though.

midgleyc:
It's for excavator's 70s volcano quest collection, which isn't automatically submitted. Might be an excavator bug if it's supposed to be.

src/net/sourceforge/kolmafia/textui/command/SubmitSpadeDataCommand.java
```diff
@@ -29,13 +31,14 @@ public void run(final String cmd, final String parameters) {
...
+          || InputFieldUtilities.confirm(
+              "Would you like to send the data \""
+                  + contents
+                  + "\" to "
+                  + recipient
+                  + "?\nThis information will be used "
+                  + explanation)) {
         RequestThread.postRequest(new SendMailRequest(recipient, contents));
       }
     }
```
jaadams5, line 44: The previous behavior was the spade command always cleared data whether it had been sent or not.  Since we are here is that still what we want?

src/net/sourceforge/kolmafia/session/LoginManager.java
```diff
@@ -215,8 +215,9 @@ public static void initialize(final String username) {
 
     if (Preferences.getString("spadingData").length() > 10) {
       KoLmafia.updateDisplay(
-          "Some data has been collected that may be of interest "
-              + "to others.  Please type `spade' to examine and submit or delete this data.");
+          "Some data has been collected that may be of interest to others. "
+              + "Please type `spade' to examine and submit or delete this data. If you have a lot, "
```
jaadams5, line 219: Please type `spade' to examine and optionally submit the data or `spade autoconfirm` to submit all of the spaded data.  Either way the data will be deleted whether shared or not.

Not sure that is really better but it is clearer.

jaadams5 (dismissed):
Consider the always delete behaviour and the wording of the "nag" but those are suggestions not requirements.

src/net/sourceforge/kolmafia/textui/command/SubmitSpadeDataCommand.java
midgleyc: I also thought that was weird but it's outside the scope of this change, in my opinion: I just left it as it was.

src/net/sourceforge/kolmafia/session/LoginManager.java
midgleyc: updated, thanks

midgleyc:
Looks like `Dismiss stale pull request approvals when new commits are pushed` is enabled on this repository: I've changed the wording as you requested and it's dismissed your review (:o).

jaadams5:
> Looks like `Dismiss stale pull request approvals when new commits are pushed` is enabled on this repository: I've changed the wording as you requested and it's dismissed your review (:o).

That is the expected behavior.  :-)

jaadams5:
> It's for excavator's 70s volcano quest collection, which isn't automatically submitted. Might be an excavator bug if it's supposed to be.

My experience is that the automatic submission never worked and I had to add something to my logout file.  See https://kolmafia.us/threads/excavator-gausies-spading-script.25076/

Maybe the best solution is to fix Excavator?


I also note that deleting all data, sent or not, via space is at least documented, if not necessarily desirable, behaviour.

jaadams5 (approved):
I'm OK with the changes.  Discussion about changing mafia or excavator should be resolved before merging :-)

jaadams5:
@midgleyc 

if (get_property("spadingData") != ""){
	cli_execute("try; ash import excavator; flush_spading_data()");
}

seems to be an existing solution.

Does it make sense to close this without committing?

midgleyc:
Possibly. There's a bit of a halfway house in terms of support at the moment, with Mafia having the "spade" command, but it only really being used by excavator.

There's a workaround for the problem I was having (as you've posted, thanks), but it's not very discoverable, and the problem would re-arise if a non-excavator spading script came into use.

jaadams5:
> Possibly. There's a bit of a halfway house in terms of support at the moment, with Mafia having the "spade" command, but it only really being used by excavator.
> 
> There's a workaround for the problem I was having (as you've posted, thanks), but it's not very discoverable, and the problem would re-arise if a non-excavator spading script came into use.

Bingo.  Either this gets approved or the existing spade command gets deprecated, IMO.  Since only one of those is non-controversial...

PR #195, opened 2021-11-07
Updating references to Ed the Undying to fix my_class()

fewyn:
This is a partial fix for https://kolmafia.us/threads/my_class-returning-void-for-actually-ed-the-undying.26728/

It now returns:
```
> ash my_class()

Returned: Ed the Undying
primestat => Mysticality
```

This still would break compatibility for all other scripts that still refer to the path as Ed. I'm not sure how to alias it or anything otherwise.

### DIFF
--- a/test/root/scripts/test_class.ash
+++ b/test/root/scripts/test_class.ash
@@ -3,3 +3,4 @@ ok = ($class[Avatar of Sneaky Pete] == $class[Avatar of Sneaky Pete]);
 print("comparison " + to_string(ok));
 ok = ($class[Gelatinous Noob].primestat == $stat[Moxie]);
 print("primestat " + to_string(ok));
+print($class[Ed] == $class[Ed the Undying]);
--- a/src/net/sourceforge/kolmafia/AscensionClass.java
+++ b/src/net/sourceforge/kolmafia/AscensionClass.java
@@ -44,7 +44,7 @@ public enum AscensionClass {
 
   public static AscensionClass nameToClass(String name) {
     for (AscensionClass ascensionClass : AscensionClass.values()) {
-      if (name.equalsIgnoreCase(ascensionClass.getName())) {
+      if (ascensionClass.getName().toLowerCase().contains(name.toLowerCase())) {
         return ascensionClass;
       }
     }
--- a/src/net/sourceforge/kolmafia/AscensionClass.java
+++ b/src/net/sourceforge/kolmafia/AscensionClass.java
@@ -19,7 +19,7 @@ public enum AscensionClass {
   ZOMBIE_MASTER("Zombie Master", 12, "tombstone", 0, "Corpse Pile"),
   AVATAR_OF_JARLSBERG("Avatar of Jarlsberg", 14, "path12icon", 1, "Blend"),
   AVATAR_OF_SNEAKY_PETE("Avatar of Sneaky Pete", 15, "bigglasses", 2, "Snap Fingers"),
-  ED("Ed", 17, "thoth", 1, "Curse of Indecision"),
+  ED("Ed the Undying", 17, "thoth", 1, "Curse of Indecision"),
   COWPUNCHER("Cow Puncher", 18, "darkcow", 0),
   BEANSLINGER("Beanslinger", 19, "beancan", 1),
   SNAKE_OILER("Snake Oiler", 20, "tinysnake", 2),
--- a/src/net/sourceforge/kolmafia/AscensionClass.java
+++ b/src/net/sourceforge/kolmafia/AscensionClass.java
@@ -43,6 +43,10 @@ public enum AscensionClass {
   private final String stun;
 
   public static AscensionClass nameToClass(String name) {
+    if (name.equals("")) {
+      return null;
+    }
+
     for (AscensionClass ascensionClass : AscensionClass.values()) {
       if (ascensionClass.getName().toLowerCase().contains(name.toLowerCase())) {
         return ascensionClass;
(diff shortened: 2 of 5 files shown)


heeheehee-kolmafia:
Maybe we should instead explicitly fix up "Ed the Undying" -> "Ed" right before turning it into an AscensionClass? I'd prefer not to break all existing scripts if we can help it.

Also, how can we test this? Is it just a matter of checking class after parsing json originally from api.php?

gausie:
> Maybe we should instead explicitly fix up "Ed the Undying" -> "Ed" right before turning it into an AscensionClass? I'd prefer not to break all existing scripts if we can help it.
> 
> Also, how can we test this? Is it just a matter of checking class after parsing json originally from api.php?

This sounds right. I wonder what scripts it would actually break though, `$class[Ed]` will still resolve to `$class[Ed the Undying]` (I think? We can certainly make sure it does). In that instance users will just get a warning (which is good in my opinion).

It would break if people are doing `my_class().to_string() == "Ed"`, but I wonder if anyone is doing that sort of thing.

gausie:
> > Maybe we should instead explicitly fix up "Ed the Undying" -> "Ed" right before turning it into an AscensionClass? I'd prefer not to break all existing scripts if we can help it.
> > Also, how can we test this? Is it just a matter of checking class after parsing json originally from api.php?
> 
> This sounds right. I wonder what scripts it would actually break though, `$class[Ed]` will still resolve to `$class[Ed the Undying]` (I think? We can certainly make sure it does). In that instance users will just get a warning (which is good in my opinion).
> 
> It would break if people are doing `my_class().to_string() == "Ed"`, but I wonder if anyone is doing that sort of thing.

I put a commit onto this branch that does this. Users will now get a `Changing "Ed" to "Ed the Undying" would get rid of this message. (test_class.ash, line 6)` sort of message but `$class[Ed]` will still work. I'll drop a note on the forums about it and to scoop up the one or two scripts that do something silly with strings.

heeheehee-kolmafia:
> > > Maybe we should instead explicitly fix up "Ed the Undying" -> "Ed" right before turning it into an AscensionClass? I'd prefer not to break all existing scripts if we can help it.
> > > Also, how can we test this? Is it just a matter of checking class after parsing json originally from api.php?
> > 
> > 
> > This sounds right. I wonder what scripts it would actually break though, `$class[Ed]` will still resolve to `$class[Ed the Undying]` (I think? We can certainly make sure it does). In that instance users will just get a warning (which is good in my opinion).
> > It would break if people are doing `my_class().to_string() == "Ed"`, but I wonder if anyone is doing that sort of thing.
> 
> I put a commit onto this branch that does this. Users will now get a `Changing "Ed" to "Ed the Undying" would get rid of this message. (test_class.ash, line 6)` sort of message but `$class[Ed]` will still work. I'll drop a note on the forums about it and to scoop up the one or two scripts that do something silly with strings.

This is fine by me.

test/root/scripts/test_class.ash
```diff
@@ -2,4 +2,5 @@ boolean ok;
 ok = ($class[Avatar of Sneaky Pete] == $class[Avatar of Sneaky Pete]);
 print("comparison " + to_string(ok));
 ok = ($class[Gelatinous Noob].primestat == $stat[Moxie]);
-print("primestat " + to_string(ok));
\ No newline at end of file
+print("primestat " + to_string(ok));
+print($class[Ed] == $class[Ed the Undying]);
```
heeheehee-kolmafia, line 6: nit: missing newline (already existing, so I don't care that much if you fix it)

heeheehee-kolmafia (approved):

gausie:
Can't quite work out why this PR is making MaximizerTests fail. Stepping through it in the debugger didn't help after 20 minutes, but I'll try again this evening.

jaadams5:
> Can't quite work out why this PR is making MaximizerTests fail. Stepping through it in the debugger didn't help after 20 minutes, but I'll try again this evening.



> Can't quite work out why this PR is making MaximizerTests fail. Stepping through it in the debugger didn't help after 20 minutes, but I'll try again this evening.

Failed locally for me on Windows as well.  Merging main into my local didn't make the failure go away.

gausie:
Would appreciate some help from anyone who can work out how this is failing the maximizer tests

heeheehee-kolmafia:
hm.

`if (ascensionClass.getName().toLowerCase().contains(name.toLowerCase())) {`

what if name is ""

heeheehee-kolmafia:
This breaks modifier checking for all items that don't have a Class: modifier, per Modifiers.add(final Modifiers).

It also breaks smithsness detection for non-class-specific items, since they're registering as AscensionClass.ASTRAL_SPIRIT (first matching entry).

This hilarious patch fixes MaximizerTest, but I recommend not applying it (and instead fixing the code to not force unknown classes to be ASTRAL_SPIRIT).

```
index e7faa71f5f..d05b23208b 100644
diff --git a/src/net/sourceforge/kolmafia/Modifiers.java b/src/net/sourceforge/kolmafia/Modifiers.java
index 12f15556ad..32817e72f4 100644
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -1722,7 +1722,7 @@ public class Modifiers {
 
     // Make sure the modifiers apply to current class
     AscensionClass ascensionClass = AscensionClass.nameToClass(mods.strings[Modifiers.CLASS]);
-    if (ascensionClass != null && ascensionClass != KoLCharacter.getAscensionClass()) {
+    if (ascensionClass != AscensionClass.ASTRAL_SPIRIT && ascensionClass != KoLCharacter.getAscensionClass()) {
       return;
     }
```

gausie:
Thanks 3h. I'll make that check return null on a blank string!!

PR #196, opened 2021-11-07
Language Server - expect aggregates when seeing {

fredg1:
Move forward in the logic when seeing {, instead of also checking for if the correcponding type is an AggregateType.
An error will still be thrown when seeing that the type isn't an AggregateType, but this will result in a better error message.

What this replaces:
line 478: the alternative was to throw an error, so this really just means a "better error message".

line 1072, 1151 and 2304: the alternative is always to try to call parseExpression.
Since no expression may simply start with a "{", this would result in the parser thinking that there is "nothing", leaving the user wondering why their aggregate literal wasn't parsed (considering that it couldn't have been anything else)


heeheehee-kolmafia (changes requested):
I assume you'll need to resolve merge conflicts for the split ParserTest change first?

Core of this commit looks fine, otherwise.

heeheehee-kolmafia (approved):

PR #197, opened 2021-11-08
Allow zap groups to contain items with commas

libraryaddict:
The item "tiny plastic Hank North, Photojournalist" contains a comma which cannot be split properly.

This PR aims to allow zappable groups to quote items which means trailing spaces, and extra commas will be correctly handled and permitted.


libraryaddict:
Yes. Realized it doesn't handle the case.

```"Hello there", the greeting item!```

Better to make something where it escapes a comma instead.

```"Hello there"\, the greeting item!```

Just a matter of figuring out the correct regex / giving up and looking for an old project where I had the regex sorted.

PR #198, opened 2021-11-08
Language Server - No longer need Stream<Arguments>

fredg1:
Arguments is no longer needed for ParserTest.data() since there is a single, consistent class for each element.


gausie (approved):
LGTM

PR #199, opened 2021-11-08
Allow zap groups to contain items with commas

libraryaddict:
The item "tiny plastic Hank North, Photojournalist" contains a comma which cannot be split properly.

This PR aims to allow zappable groups to gain the ability to escape commas with the usage of backslashes \


gausie:
I hated that regex enough to write a string tokenizer. Also added a test

gausie (approved):

PR #201, opened 2021-11-09
Language Server - split ParserTest

fredg1:
Splitting Parser into parsetree may not be recommended right now, but we can at least split ParserTest, which will make it easier to add tests for individual classes afterwards.

No test was altered in the making of this patch.


heeheehee-kolmafia (approved):
sure, why not

PR #202, opened 2021-11-09
Some lint discovered while exploring Concoctions.

jaadams5:
Tweaked Concoction Test.  Faster but still a candidate for moving to DebugDatabase.  Feedback welcomed.


jaadams5:
So it seems that Windows is getting more heap space than Ubuntu?

And not even "my" test...

Caused by: java.lang.OutOfMemoryError: Java heap space
	at java.base/java.util.regex.Matcher.<init>(Matcher.java:247)
	at java.base/java.util.regex.Pattern.matcher(Pattern.java:1134)
	at net.sourceforge.kolmafia.Modifiers.parseModifiers(Modifiers.java:1923)
	at net.sourceforge.kolmafia.Modifiers.getModifiers(Modifiers.java:1895)
	at net.sourceforge.kolmafia.DebugModifiers.allModifiers(DebugModifiers.java:127)
	at net.sourceforge.kolmafia.textui.command.ModifierListCommand.run(ModifierListCommand.java:23)
	at net.sourceforge.kolmafia.KoLmafiaCLI.doExecuteCommand(KoLmafiaCLI.java:458)
	at net.sourceforge.kolmafia.KoLmafiaCLI.executeCommand(KoLmafiaCLI.java:419)
	at net.sourceforge.kolmafia.KoLmafiaCLI.executeLine(KoLmafiaCLI.java:338)
	at net.sourceforge.kolmafia.KoLmafiaCLI.executeLine(KoLmafiaCLI.java:225)
	at net.sourceforge.kolmafia.KoLmafiaCLI$CommandProcessorThread.run(KoLmafiaCLI.java:160)
	at net.sourceforge.kolmafia.KoLmafiaCLI.listenForCommands(KoLmafiaCLI.java:220)
	at net.sourceforge.kolmafia.textui.command.CallScriptCommand.call(CallScriptCommand.java:217)
	at net.sourceforge.kolmafia.textui.command.CallScriptCommand.run(CallScriptCommand.java:34)
	at net.sourceforge.kolmafia.CustomScriptTest.testScript(CustomScriptTest.java:55)

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -30,61 +52,51 @@ public void itShouldSortUsables() {
...
-        int x = acon.compareTo(bcon);
-        int y = bcon.compareTo(acon);
-        String msg = acon.toString() + " * " + bcon.toString();
-        assertEquals(sgn(x), -sgn(y), msg);
-        if (x == 0) assertEquals(acon, bcon, msg);
+  public void compareToShouldBeQuasiSymmetric() {
+    for (int i = 0; i < maxIndex; i++) {
+      for (int j = 0; j < maxIndex; j++) {
+        msg = "comparing (quasi symmetry)" + idsAndMaps.get(i) + "and " + idsAndMaps.get(j);
+        assertEquals(sgn(result[i][j]), -sgn(result[j][i]), msg);
```
heeheehee-kolmafia, line 61: sgn is redundant here, since you've already invoked it in populating this array.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -30,61 +52,51 @@ public void itShouldSortUsables() {
...
+        if (result[i][j] > 0) {
+          for (int k = 1; k < maxIndex; k++) {
+            if (result[j][k] > 0) {
+              msg =
+                  "comparing (transitive)"
+                      + idsAndMaps.get(i)
+                      + "and "
+                      + idsAndMaps.get(j)
+                      + "and "
+                      + idsAndMaps.get(k);
```
heeheehee-kolmafia, line 97: You're missing several spaces.

This will print as: "comparing (transitive)seal-clubbing cluband seal toothand bar skin"

You're also not using msg anywhere in this function.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -2,24 +2,46 @@
...
-dynamic data, i.e. something that changed because a character left Ronin, but it can't hurt to verify
-the comparison against static data.
+said "Comparison method violates its general contract!"
  */
 
 public class ConcoctionTest {
 
+  static final Map<Integer, Concoction> idsAndMaps = new HashMap<>();
+  static int maxIndex;
+  static String msg;
```
heeheehee-kolmafia, line 20: I don't see why msg / maxIndex have to be static.

You're not saving any allocations given that you're constructing this string from scratch every time, and maxIndex is basically a glorified idsAndMaps.size().

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -2,24 +2,46 @@
...
 /* This test was triggered by a runtime error traced back to sorting usable concoctions that
-said "Comparison method violates its general contract!"  It is likely that the error was caused by
-dynamic data, i.e. something that changed because a character left Ronin, but it can't hurt to verify
-the comparison against static data.
+said "Comparison method violates its general contract!"
  */
 
 public class ConcoctionTest {
 
+  static final Map<Integer, Concoction> idsAndMaps = new HashMap<>();
```
heeheehee-kolmafia, line 18: Why not store this in an ArrayList or an array with true constant (and much lower overhead) random access?

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -30,61 +52,51 @@ public void itShouldSortUsables() {
...
+          for (int k = 1; k < maxIndex; k++) {
+            if (result[j][k] > 0) {
+              msg =
+                  "comparing (transitive)"
+                      + idsAndMaps.get(i)
+                      + "and "
+                      + idsAndMaps.get(j)
+                      + "and "
+                      + idsAndMaps.get(k);
+              assertTrue(result[i][k] > 0);
```
heeheehee-kolmafia, line 98: I would prefer using Hamcrest matchers here, as they'll provide more useful error messages by default.

```
assertThat(result[i][k], greaterThan(0))
```

will tell you what the actual value was, rather than just "false does not equal true" or whatever.

heeheehee-kolmafia (changes requested):

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -2,24 +2,46 @@
...
   }
 
+  @BeforeAll
+  private static void buildIndexAndData() {
+    LockableListModel<Concoction> usables = ConcoctionDatabase.getUsables();
+    maxIndex = usables.getSize();
+    int i = 0;
+    for (Concoction con : usables) {
+      idsAndMaps.put(i, con);
+      i++;
```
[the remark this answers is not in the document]
jaadams5: The prefix version is just unnatural to me.  If it is a standard or maves a measurable difference I will gladly change but I am almost never going to write ++i without prompting.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
jaadams5: msg doesn't need to be static.  maxIndex is used in buildIndexAndData which has to be static to be @BeforeAll and code didn't compile without it.  In Ye Olden Days idsAndMaps.size() was an object reference and a function call which was more expensive than accessing a field.  I don't know whether that is true for 21st Century Java but it is habit from a time and place where it was less efficient.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
jaadams5: It seemed like a good idea at the time when there was also a corresponding Map<Concoction, Integer> so I could go back and forth between a Concoction and an integer.  I thought I was going to need Concoction as an index.  It doesn't look like I do.  Will revisit.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
jaadams5: Hamcrest matchers - not in my vocabulary or tool box.  In that case I was trying to mirror the contract and I'm not convinced that test does so as well as I thought it did a couple hours ago.  I'll look at Hamcrest.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
jaadams5: Fixed

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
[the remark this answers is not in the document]
jaadams5: And done.

jaadams5:
After much thought I moved the compareTo contract testing to a checkconcoctions cli command that can be invoked on demand.  It doesn't do much for test coverage but doing so reduces test running time.  It also makes sense that someone changing concoctions.txt might need the check but it was a lot of effort for a test being run all of the time.

PR with the new command hasn't been pushed yet but there were some things worth keeping that crept into this PR so...

PR #203, opened 2021-11-09
AutoFilterComboBox-raw-type

fredg1:
AutoFilterComboBox was split into two "modes": editable, and not.

This PR isolates the editable mode into its own class, "EditableAutoFilterComboBox", whose generic is locked as String.

This allows AutoFilterComboBox to have its own Generic.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/widget/EditableAutoFilterComboBox.java
+++ b/src/net/sourceforge/kolmafia/swingui/widget/EditableAutoFilterComboBox.java
@@ -8,6 +8,20 @@ public class EditableAutoFilterComboBox extends AutoFilterComboBox<String> {
     super(model);
   }
 
+  public String getText() {
+    return (String) (this.getSelectedItem() != null ? this.getSelectedItem() : this.currentMatch);
+  }
+
+  public void setText(final String text) {
+    if (this.model.contains(text)) {
+      this.setSelectedItem(text);
+    } else {
+      this.setSelectedItem(null);
+      this.currentMatch = text;
+      this.editor.setText(text);
+    }
+  }
+
   @Override
   public void forceAddition() {
     if (this.currentName == null || this.currentName.length() == 0) {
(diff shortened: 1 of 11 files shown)


src/net/sourceforge/kolmafia/swingui/widget/EditableAutoFilterComboBox.java
```diff
@@ -0,0 +1,77 @@
...
+import java.awt.event.KeyEvent;
+import net.java.dev.spellcast.utilities.LockableListModel;
+
+public class EditableAutoFilterComboBox extends AutoFilterComboBox<String> {
+  public EditableAutoFilterComboBox(final LockableListModel<String> model) {
+    super(model);
+  }
+
+  public String getText() {
+    return (String) (this.getSelectedItem() != null ? this.getSelectedItem() : this.currentMatch);
```
heeheehee-kolmafia, line 12: (I hate that LcokableListModel.getSelectedItem returns an Object instead of a genericized type...)

src/net/sourceforge/kolmafia/swingui/widget/EditableAutoFilterComboBox.java
```diff
@@ -0,0 +1,77 @@
...
+  @Override
+  public boolean isVisible(final Object element) {
+    if (!this.isActive()) {
+      return true;
+    }
+
+    // If it's not a result, then check to see if you need to
+    // filter based on its string form.
+
+    if (this.matchString == null || this.matchString.length() == 0) {
```
heeheehee-kolmafia, line 70: as elsewhere, this.matchString.empty() instead of length() == 0

heeheehee-kolmafia (changes requested):
Looks fine modulo a tiny nitpick.

src/net/sourceforge/kolmafia/swingui/widget/EditableAutoFilterComboBox.java
fredg1: (why are we even still using that? Wasn't it used for the lockable aspect, which got removed?)

heeheehee-kolmafia: At this point, LockableListModel is a (Swing) ListModel that has a notion of visible / hidden elements based on a custom filter.

heeheehee-kolmafia (approved):

PR #205, opened 2021-11-11
Fix scan take 3-1

jaadams5:
Trying to fix security scan by running scan manually before commit


jaadams5:
If I did things right this silenced the scanner.  Some lint mixed in, but the meat is in LoginRequest

heeheehee-kolmafia (approved):

PR #207, opened 2021-11-12
Concoctions Contract

jaadams5:
Added a new cli command checkconcoctions.

Modified compareTo and equals for concoctions so that concoctions that make the same item are not considered the same concoction.

Needs some real world testing.


jaadams5:
Curiously the checkconcoctions works when not logged in but gets an array index out of bounds error on two logged in characters.  Pondering.

jaadams5:
Index was a size issue.  LockableListModel size() is the size of the list.  getSize() is the size of the visible elements in the list.

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2583,6 +2577,92 @@ private static void checkPower(final int itemId, final boolean force) {
...
+                + getIString(ids[i])
+                + " and "
+                + getIString(ids[j]);
+        if (result[i][j] != -result[j][i]) {
+          KoLmafia.updateDisplay(msg);
+        }
+      }
+    }
+    // tests the portion of the contract that says (x.compareTo(y)==0) == (x.equals(y))
+    for (i = 0; i < maxIndex; ++i) {
```
jaadams5, line 2626: A case could be made for combining the two i, j loops into one but I haven't done that yet.

PR #213, opened 2021-11-14
Language Server - Locations - Operator

fredg1:


src/net/sourceforge/kolmafia/textui/parsetree/Operator.java
```diff
@@ -6,15 +6,17 @@
...
 
 public class Operator extends Command {
-  String operator;
+  final String operator;
 
   // For runtime error messages
-  String fileName;
-  int lineNumber;
+  private final String fileName;
+  private final int lineNumber;
```
gausie, line 16: Do we need to keep tracking lineNumber/fileName separately if we now have a Location?

fredg1: That's something that I'll eventually look into. There are a few classes who do this, and I'd prefer to address all of them in the same patch.

fredg1: (after verifying that the information they use is really the same as the one provided by the Location)

gausie: Ok. Do Locations know about what file name they're referring to? Surely they must, but I can't see it in the API for them.

fredg1: It's in the URI. However, we'd need to trim it first, since it's the full path up to the partition...

fredg1: (or root, for linux users)

PR #219, opened 2021-11-16
Protonic ghosts are nocopy and fix adv range for tin cup of mulligan stew

Malibu-Stacey:
As title.
[spading data for 100+ tin cups of mulligan stew](https://github.com/kolmafia/kolmafia/files/7544138/mulligans.txt) consumed in the past 2 weeks. Range is most definitely 15-18.

Changes have been neither built nor tested as I don't have a Java dev environment set up to do so however they are simply data file changes not actual code changes.


gausie (approved):
Thank you for the contribution!

PR #223, opened 2021-11-16
Language Server - Locations - Conditional

fredg1:
Because we refer to If for conditional chains, it is the one to have the location of the whole chain.

I'd rather prefer if we had another way to group them, because it means we can't access the location of the first If anymore...

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/parsetree/ConditionalTest.java
+++ b/test/net/sourceforge/kolmafia/textui/parsetree/ConditionalTest.java
@@ -2,8 +2,13 @@ package net.sourceforge.kolmafia.textui.parsetree;
 
 import static net.sourceforge.kolmafia.textui.ScriptData.invalid;
 import static net.sourceforge.kolmafia.textui.ScriptData.valid;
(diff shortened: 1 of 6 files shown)


test/net/sourceforge/kolmafia/textui/parsetree/ConditionalTest.java
```diff
@@ -22,7 +27,57 @@
...
+            scope -> {
+              List<Command> commands = scope.getCommandList();
+
+              // Conditional location test - whole
+              Conditional conditional = assertInstanceOf(If.class, commands.get(0));
+              // From the "if" up to the end of the last conditional's scope
+              ParserTest.assertLocationEquals(1, 1, 1, 60, conditional.getLocation());
+
+              // Conditional location test - If
+              // TODO save conditional chains as their own class so that we can access the first
```
gausie, line 40: This TODO should live in the Parser probably, not here

jaadams5 (approved):
> This TODO should live in the Parser probably, not here

I agree but won't require a change.

PR #238, opened 2021-11-17
Language Server - Locations - FunctionReturn

fredg1:


gausie:
Does that mean this is the last Command to which you need to add Locations?

fredg1:
This is the last *direct* descendant of Command that needed a Location.
The rest is now all descendants of Evaluable.

jaadams5 (approved):

PR #241, opened 2021-11-20
RequestTestBase redirections

fredg1:
Adds utilities to make redirections with RequestTestBase

Also makes RequestTestBase public, because request handling is most likely to be need more or less everywhere, and it shouldn't all go in the same folder...


gausie (approved):
Thanks

PR #251, opened 2021-11-24
Library function classes

gausie:


src/net/sourceforge/kolmafia/textui/function/LibraryClassFunction.java
```diff
@@ -0,0 +1,61 @@
...
+    String returns() default "void";
+  }
+
+  @Retention(RetentionPolicy.RUNTIME)
+  @Target({ElementType.PARAMETER})
+  public @interface LibraryFunctionParameter {
+    String value();
+  }
+
+  public final List<Function> getFunctions() {
```
gausie, line 35: Perhaps this should return a `FunctionList`? I'm not sure what the benefit would be.

fredg1: Well, I plan to make it so that function references are saved in TypeList, VariableList and FunctionList, so there's that... no sure if it's relevant here, though...

Veracity0:
I will point out that we can consider two approaches:

- Library Functions in multiple Class files where the methods are static methods. Fast, compile time method lookup. Still quite testable, to my eye - depending on what you have in mind for testing.
- Library Functions in Objects, where the methods are object methods. Method lookup on am object at Runtime is no where near as fast. Advantage is that the classes can extend a base class which provides ... stuff.

With the 2nd, you can still have SIngleton objects:

```
private static VisitUrl INSTANCE = null;
public static VisitUrl getInstance() {
    if (INSTANCE == null) INSTANCE = new VisitUrl();
    return INSTANCE;
}
```

And all the calls to "new VisitUrl()" become "VisitURL.getInstance()".

I want to understand exactly how you feel that the non-static Object approach makes testing easier.

We already can mock static methods.
session/MallPriceManagerTest mocks MallPriceManager.getSystemClock()

We can already mock actual class objects
session/MallPriceManagerTest has "MockMallSearchRequest extends MallSearchRequest"
It also mocks the static method MallPriceManager.newMallSearchRequest to return a mocked request.
The code for that was not obvious. It took a lot of googling to figure out, but now I have a model. :)

But, if you feel that making Runtime Library classes for functions will make them easier to test - or use in tests - I would be very interested in seeing some examples of how that plays out.

Show me some tests, please!

gausie:
The ultimate goal here is to be able to generate documentation for the runtime library and enforce its documentation as part of code review. This might not be the way to do it but I was looking for:

- Meaningful argument names for LibraryFunctions, with the understanding that it varies per overload
- An opportunity for describing what each function does and/or what it returns in Javadoc format (probably)

This can all be static, I will move this PR to using statics when I start working on it again, which will be soon

gausie:
Having said that, I think I might just start again. I'm going to close this PR. Thanks for your review @Veracity0

PR #252, opened 2021-11-24
Update monsters.txt

jaadams5:
These were items where Monster Manual and KoLmafia disagreed.

Changing Protector Spectre (Wildfire) will cause some scripts to complain because Protector Spectre is now ambiguous.

Local test showed this change eliminated the disagreements.


gausie:
What's the rationale for renaming the Spectre?

As far as I'm aware, its not a requirement that our monsters.txt matches Monster Manual. We have monsters like:

- Ninja Snowman (Chopsticks)
- Ninja Snowman (Hilt)
- Gremlin (tool)
- factory worker (woman)

For the same reason, I added the " (Wildfire)" on purpose.

jaadams5:
> What's the rationale for renaming the Spectre?

My recollection is that when the Manual came out a decision was made to use the Manual name as the canonical name of the monster.  I think one of the drivers for that was the ability to automatically recognize new monsters.  In any event the code that flagged the name mismatch was KoLmafia code so someone before me thought it should be fixed :-)

Run checkmanual.

That said there is a wiki entry for https://kol.coldfront.net/thekolwiki/index.php/Protector_Spectre_(Wildfire) so deleting "(Wildfire)" does not seem to be the correct solution.  I won't commit this until I figure out whether it is the data, the wiki or the checking code that needs to be adjusted.

jaadams5:
The Manuel entry changes the font color for "(Wildfire)".

<span style="font-weight:normal;color:#777777">(Wildfire)</span>

The problem is that KoLmafia isn't seeing the match.  My regex is not up to a quick fix so I reverted this portion.

PR #253, opened 2021-11-25
(slightly) safe(r) cloning

fredg1:
For types without a generic: use covariant return types to save some type castings.
(source: http://jtechies.blogspot.com/2012/07/item-11-override-clone-judiciously.html )

For ArrayList: use the copy constructor instead of ArrayList.clone()
(source: https://stackoverflow.com/questions/3410035/how-to-make-a-copy-of-arraylist-object-which-is-type-of-list/3410133#3410133 )

For LockableListModel... ugh... just ignore the warning, I guess...

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/MuseumFrame.java
+++ b/src/net/sourceforge/kolmafia/swingui/MuseumFrame.java
@@ -252,6 +252,8 @@ public class MuseumFrame extends GenericFrame {
 
   public class OrderingPanel extends ItemListManagePanel<String> {
     @SuppressWarnings("unchecked")
+    // LockableListModel (the type of getHeaders())'s clone() method still claims to return an
+    // Object, but we know its internal list has the same elements, so their types definitely match
     public OrderingPanel() {
       super((LockableListModel<String>) DisplayCaseManager.getHeaders().clone());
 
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/swingui/MuseumFrame.java
```diff
@@ -251,6 +251,7 @@ public void updateDisplay(final PanelList list, final Object value, final int in
   }
 
   public class OrderingPanel extends ItemListManagePanel<String> {
+    @SuppressWarnings("unchecked")
```
gausie, line 254: Why are you adding this

fredg1: See the PR's comment:

> For LockableListModel... ugh... just ignore the warning, I guess...

This clone is a LockableListModel, which is one of the files we have in /lib
Its `clone` method still returns an `Object`, so we can't have the type-safety be asserted earlier.

gausie: Please add a short but clear comment on the previous (or next - does Java allow that?) line explaining why you've added it. This is general advice for all `SuppressWarnings` annotations.

gausie (approved):

PR #255, opened 2021-11-25
Raw arrays - ChatManager

fredg1:
I've been thinking about https://github.com/kolmafia/kolmafia/pull/148#issuecomment-949555223 for a while, and decided to finally apply it (split between the next few PRs)

They replace arrays of classes with a type parameter with ArrayLists.

### DIFF
--- a/src/net/sourceforge/kolmafia/chat/ChatManager.java
+++ b/src/net/sourceforge/kolmafia/chat/ChatManager.java
@@ -50,7 +50,7 @@ public abstract class ChatManager {
 
   private static final TreeMap<String, StyledChatBuffer> instantMessageBuffers =
       new TreeMap<String, StyledChatBuffer>();
-  private static Entry<String, StyledChatBuffer>[] bufferEntries = new Entry[0];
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/chat/ChatManager.java
```diff
@@ -236,8 +236,8 @@ public static final StyledChatBuffer getBuffer(final String bufferKey) {
 
     synchronized (ChatManager.bufferEntries) {
       ChatManager.instantMessageBuffers.put(bufferKey, buffer);
-      ChatManager.bufferEntries = new Entry[ChatManager.instantMessageBuffers.size()];
-      ChatManager.instantMessageBuffers.entrySet().toArray(ChatManager.bufferEntries);
```
gausie, line 240: I'm sort of confused about what this line was even doing before, so I'm worried about removing it!

fredg1: same D:

gausie: hmm ok im willing to go ahead

gausie (approved):

PR #262, opened 2021-11-25
Raw arrays - ConsumablesDatabase

fredg1:
I was ready for this to take time, but then I realized that it was just the equivalent of a bitmap...


gausie:
Ideally:

- Make some sort of static function that takes `boolean perUnit`, `boolean gainEffect1`, `boolean gainEffect2`, `boolean gainEffect3`, `boolean gainEffect4` and spits out our integer key.
- This is a lot less complicated than nested loops and does the same thing with this design.
```java
for (int i=0; i<Math.pow(2, 5); i++) {
  ConsumablesDatabase.advsByName.put(i, new HashMap<>());
}
```

gausie (approved):

PR #264, opened 2021-11-25
ScrollablePanel raw types

fredg1:
passes the type of ScrollablePanel.scrollComponent as a type parameter.


src/net/sourceforge/kolmafia/swingui/panel/ItemListManagePanel.java
```diff
@@ -19,11 +19,11 @@ public ItemListManagePanel(
         addFilterField,
         addRefreshButton);
 
-    ShowDescriptionList<E> elementList = (ShowDescriptionList<E>) this.elementList;
+    ShowDescriptionList<E> elementList = this.scrollComponent;
```
gausie, line 22: Is this correct? I was surprised to see a change of property name referenced.

fredg1: This is because `elementList` was only made to "save" a type casting

`JComponent elementList = this.scrollComponent;`

(see ItemManagePanel)

gausie (approved):

PR #267, opened 2021-11-27
Raw arrays - Evaluator

fredg1:


gausie (approved):
We clearly need a class for mapping slots to *things* - we do it all the time!

fredg1:
> We clearly need a class for mapping slots to _things_ - we do it all the time!

I think we need a dedicated class of *many* things 😓

PR #291, opened 2021-11-28
Language Server - error locations - EscapeSequence

fredg1:


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -3192,9 +3192,10 @@ private Evaluable parseString(final BasicScope scope) {
   }
 
   private int parseEscapeSequence(final StringBuilder resultString, int i) {
+    final int backslashIndex = i++;
     final String line = this.restOfLine();
 
-    if (++i == line.length()) {
+    if (i == line.length()) {
       resultString.append('\n');
       this.currentLine.makeToken(i);
```
jaadams5, line 3200: I think this is wrong.  

In the old code, if i and line.length() are both 2 at the conditional, i will be incremented and the true branch executed.  So i would be 3 in the call to  this.currentLine.makeToken(i) whereas it would be 2 in the changed code

jaadams5 (commented):
Correct the prefix increment or explain the error of my thinking. please.

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: ...no, because `i` is still incremented in this code, two lines above?

fredg1: `final int backslashIndex = i++;`

fredg1: previous operations done to/with `i`:
1- increment
2- check against line.length()

current operations done to-with `i`:
1- save its value as `backslashIndex`
2- increment
3- check against line.length()

Nothing downhill is affected in any way

jaadams5: Ok.  final int backslashIndex = i++; didn't jump out at me.  I can make a case that there is a more comprehensible implementation possible but that can be another conversation.

jaadams5 (approved):

PR #293, opened 2021-11-30
Language Server - error locations - Variable

fredg1:


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -3720,7 +3721,7 @@ private Evaluable parseVariableReference(final BasicScope scope) {
     Variable var = scope.findVariable(name.content, true);
```
gausie, line 3721: I know this wasn't you but it would be good not to use `var` as a variable name

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -784,9 +784,10 @@ private Variable parseVariable(final Type t, final BasicScope scope) {
     Variable result;
 
     if (Parser.isReservedWord(variableName.content)) {
-      throw this.parseException("Reserved word '" + variableName + "' cannot be a variable name");
+      throw this.parseException(
+          variableName, "Reserved word '" + variableName + "' cannot be a variable name");
     } else if (scope != null && scope.findVariable(variableName.content) != null) {
-      throw this.parseException("Variable " + variableName + " is already defined");
+      throw this.parseException(variableName, "Variable " + variableName + " is already defined");
```
gausie, line 790: Can we not add a proper location here and above? Or do they get added by parseException

fredg1: parseException has multiple overloads that build the Location if it was submitted a Range or a Position, yes

fredg1: (and Token extends Range)

gausie (approved):

PR #295, opened 2021-11-30
Automatically generate release notes.

heeheehee-kolmafia:
Per documentation for action-gh-release, this should automatically
generate a name and body if not present. If body is specified, then
it'll append to the specified body.


jaadams5 (approved):
That was easy.  I'll trust that it works :-)

PR #297, opened 2021-12-01
Force html to wrap, add support for the html attribute "title" in CLI & Chat

libraryaddict:
This adds a custom html editor kit or so, which basically forces text to wrap in the html views.

Two common examples of html views are mafia chat, and the CLI.

This means text that is normally unable to wrap properly (Long words & links being an example) are now forcibly wrapped as is normal expected behavior.

Alongside this, it finds usages of the "title" attribute, mostly known for use in the chat effects. And utilizes the swing api to register them as something the user can hover over and show as if the "title" attribute worked normally.

https://stackoverflow.com/a/26583365


gausie:
Please update the description saying what this does

src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
```diff
@@ -1,25 +1,123 @@
...
+      }
+
+      private class WrapLabelView extends LabelView {
+        private String title;
+
+        public WrapLabelView(Element elem) {
+          super(elem);
+
+          Enumeration<?> itel = elem.getAttributes().getAttributeNames();
+
```
jaadams5, line 55: What is "itel" supposed to suggest to the reader?

src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
```diff
@@ -1,25 +1,123 @@
...
+      this.viewFactory = new WrappedHtmlFactory();
+    }
+
+    @Override
+    public ViewFactory getViewFactory() {
+      return this.viewFactory;
+    }
+
+    private class WrappedHtmlFactory extends HTMLEditorKit.HTMLFactory {
+      @Override
```
jaadams5, line 31: I think this can also be static.

src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
libraryaddict: Renamed to iterator. Old habit of mine.

src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
libraryaddict: Both inner classes changed to static

jaadams5 (approved):

SoftwareEthicsResearcher:
Hello, Sir, 

I am doing research in software engineering ethics. I found the commit code reference https://stackoverflow.com/questions/17533451/jeditorpane-linewrap-in-java7/26583365#26583365. 
It would be good if acknowledgment or credit to the author in code. I would like to highlight this as an attribution concern. 

Regards, Software Ethics Researcher.

PR #301, opened 2021-12-02
Serialize build.yml executions.

heeheehee-kolmafia:
I think this might eliminate the case where multiple releases end up
with the same summary data if they're merged in quick succession,
although it's possible the problem is deeper than that.

(I'm really not sure on this one, caveat lector.)


jaadams5 (approved):
Can't figure out how the change would hurt...

PR #302, opened 2021-12-02
Language Server - error locations - Expression

fredg1:
1- adds a type check after the '-' unary operator, which we were previously lacking (resulting in a runtime error)

2- tweak where the "Bad 'remove' statement" error is triggered (it would previously always be overriden by an "unknown variable REMOVE" (case may vary) exception.


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -2967,18 +2966,33 @@ private Evaluable parseExpression(final BasicScope scope, final Operator previou
...
+            && !lhs.getType().equals(DataTypes.FLOAT_TYPE)) {
+          throw this.parseException(
+              lhs.getLocation(), "\"-\" operator requires an integer or float value");
+        }
       }
-    } else if (operator.equals("remove")) {
+    } else if (operator.equalsIgnoreCase("remove")) {
+      // Due to the workings of Operator, we need 'remove' to be case-sensitive
+      if (!operator.equals("remove")) {
+        throw this.parseException(operator, "Bad 'remove' statement");
```
gausie, line 2984: This will only ever happen if `remove` is not lowercase. We could say as much in the exception text.

gausie: Though could we not just overwrite `operator` with lowercase "remove" here?

fredg1: we could, and we can!

gausie (approved):

PR #303, opened 2021-12-02
Use new cloudfront url for KoL images.

heeheehee-kolmafia:
Apparently this changed recently and broke the CAB.

https://kolmafia.us/threads/26911


gausie:
Hmmm this looks like something that CloudFront could choose to change often.

PR #306, opened 2021-12-02
Revert "Use new cloudfront url for KoL images."

heeheehee-kolmafia:
Reverts kolmafia/kolmafia#303

latest reports say this breaks other things, sounds like we need to dig into this further and/or invest in a more robust solution.


jaadams5 (approved):

gausie:
@heeheehee-kolmafia chatted with cdm and he confirmed that it shouldn't change frequently

PR #309, opened 2021-12-03
fix regex for text of cabinet while it's on cooldown

C2Talon:


gausie (approved):
Whoops, that was a typo from me. Thanks

PR #311, opened 2021-12-03
Language Server - Parse all - preparations - error methods and "bad" classes

fredg1:


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -4461,6 +4475,44 @@ public static Location mergeLocations(final Command start, final Command end) {
...
+      }
+    }
+
+    public String toString() {
+      StringBuilder result = new StringBuilder();
+
+      result.append(this.message);
+      result.append(" (");
+
+      result.append(Parser.getFileAndRange(Parser.this.shortFileName, this.location.getRange()));
```
gausie, line 4504: This means that if you toString an ashdaognostic after the fact, you may get the wrong filename because the parser is looking at a different file

fredg1: This is actually the same as line 4802
```java
  private ScriptException parseException(Location location, final String message) {
    if (location == null) {
      location = this.makeZeroWidthLocation();
    }

    return new ScriptException(
        message + " (" + Parser.getFileAndRange(this.shortFileName, location.getRange()) + ")");
  }
```

This is *intentional*.
If "the parser is looking at a different file", then it means that the AshDiagnostic was put in the wrong Parser.
A Parser should only report errors happening in the file it parsed.

I could add a check that the submitted location's URI matches shortFileName, but I'm worried about false positives, due to not knowing how File.getUri() works with every OS...

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -4461,6 +4475,44 @@ public static Location mergeLocations(final Command start, final Command end) {
...
+
+      result.append(")");
+
+      for (final String additionalMessage : this.additionalMessages) {
+        result.append(" " + additionalMessage);
```
heeheehee-kolmafia, line 4509: This ends up copying additionalMessage twice: once to construct " " + additionalMessage, and a second time to append it to result.

Since you're already using a StringBuilder, instead consider:

    result.append(" ");
    result.append(additionalMessage);

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -1,5 +1,7 @@
 package net.sourceforge.kolmafia.textui;
 
+import static org.eclipse.lsp4j.DiagnosticSeverity.*;
```
heeheehee-kolmafia, line 3: Can we just write out DiagnosticSeverity.Error and DiagnosticSeverity.Warning in the two places that you're using this static import?

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -4570,10 +4664,53 @@ private void enforceSince(String revision, final Range directiveRange) {
     }
   }
 
+  public final void error(final String msg, final String... otherInfo) {
+    this.error(this.getCurrentPosition(), msg, otherInfo);
+  }
+
+  public final void error(final Position start, final String msg, final String... otherInfo) {
```
heeheehee-kolmafia, line 4671: Do you actually use all these overloads? (other than as helper functions)

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -4570,10 +4664,53 @@ private void enforceSince(String revision, final Range directiveRange) {
...
+    this.diagnostics.add(
+        new AshDiagnostic(
+            location != null ? location : this.makeZeroWidthLocation(), Error, msg, otherInfo));
+  }
+
   public final void warning(final String msg) {
     RequestLogger.printLine("WARNING: " + msg + " " + this.getLineAndFile());
   }
 
+  /*
```
heeheehee-kolmafia, line 4694: Delete instead of commenting out code.

src/net/sourceforge/kolmafia/textui/parsetree/Value.java
```diff
@@ -25,6 +25,8 @@
  * as some sort of... hazmat suit..?
  */
 public class Value implements TypedNode, Comparable<Value> {
+  public static final Value BAD_VALUE = new Value(new Type.BadType(null, null));
```
heeheehee-kolmafia, line 28: Do we want BAD_VALUE.isBad() to be true? That doesn't seem to be the case.

heeheehee-kolmafia (changes requested):

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: This isn't something to be deleted, it is just "not added" yet because it would conflict with the existing one

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: yes, we do

src/net/sourceforge/kolmafia/textui/parsetree/Value.java
fredg1: BAD_VALUE.isBad() doesn't exist.
isBad() is for Symbol
to check foe BAD_VALUE, we use Evaluable.evaluatesTo(BAD_VALUE)

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: ...sure (._. )

src/net/sourceforge/kolmafia/textui/Parser.java
heeheehee-kolmafia: ... then add it when you're ready to delete the existing one?

fredg1: based on what's coming, I preferred to at least show that it was coming rather than fitting something unexpected

fredg1: (is it still ok if it stays there?)

fredg1: poke poke pokedy prod pokey?

jaadams5: >poke poke pokedy prod pokey?

I see a PR marked as "Changes Requested" by another reviewer so I am going to ignore it until there is evidence the requested changes have been made.  So the ball is in your court.  Make and commit the changes or make it clear why you do not intend to do so.  If you think you have already done so then it was not clear to me.

fredg1: > I see a PR marked as "Changes Requested" by another reviewer so I am going to ignore it until there is evidence the requested changes have been made. So the ball is in your court. Make and commit the changes or make it clear why you do not intend to do so. If you think you have already done so then it was not clear to me.

There were 5 "conversations" from heeheehee.
I successfully addressed and resolved two of them.
Two others were questions/concerns which I responded to (didn't mark them as resolved in case there's a follow-up, but until then, I can assume they are)

This leaves only this one, in which I marked a reason to not address the change. Was it insufficient/unclear?

jaadams5: >Was it insufficient/unclear?

Obviously or I would have responded differently.

Can you flag something (a conversation) as resolved?  If you can then it would help if you did so.  If you tell me No then I will look harder and resolve the items I am comfortable resolving.  Then I'll reassess status.

fredg1: I can mark them as resolved, yes. However, I prefer to not assume that a conversation is resolved when I'm not "resolving" it as the other person expected (i.e. answering why I would/should not)

fredg1: when doing this, I prefer getting a confirmation from the other party that the reason is satisfactory

jaadams5: In that case the nudge needs to be more specific :-)

src/net/sourceforge/kolmafia/textui/Parser.java
gausie: Someone could take a copy of the AshDiagnostic produced and expect it to be valid later on in execution though. You should change this to ensure that we don't need to rely on your having made sure it never happens.

src/net/sourceforge/kolmafia/textui/Parser.java
gausie: @fredg1 Do not commit any commented code. Remove it and then you can resolve the conversation.

Also I don't know how many times I have to tell you but stop prodding people if they don't respond to you for periods of less than 24 hours or over weekens. Stop doing it.

fredg1: ...noted.
For now, I'll guess I'll just comply and delete the comment, since the intent, which was to just make you (collectively) aware that this method was coming, was achieved either way...

src/net/sourceforge/kolmafia/textui/Parser.java
fredg1: What exactly do you mean by taking a copy of it?
You mean adding it to a Parser's `diagnostics` field?

gausie: In the future code could be written that holds an AshDiagnostic and does something with it later.

fredg1: That was... shamefully easier than expected (o///o)

gausie:
@fredg1 I just dont see what stops you running `Parser.getFileAndRange(Parser.this.shortFileName, this.location.getRange());` in the AshDiagnostic constructor instead of what you did

fredg1:
> @fredg1 I just dont see what stops you running `Parser.getFileAndRange(Parser.this.shortFileName, this.location.getRange());` in the AshDiagnostic constructor instead of what you did

What's stopping us is the fact that these AshDiagnostics will also need to be available to the language server, so no, we can't prepare all the information in a single string straight from the constructor

fredg1:
Is there... still... a problem..?

gausie:
Yes the problem is that heeheehee hasn't been able to approve your changes yet.

fredg1:
Seems like heeheehee is MIA again; is it really not possible for someone else to do/finish the review instead?

jaadams5:
> Seems like heeheehee is MIA again; is it really not possible for someone else to do/finish the review instead?

I sort of offered to do that but you were not comfortable with unilaterally resolving issues that @heeheehee-kolmafia had raised.  If that has changed, say so, and I'll try and review and approve in the next day or so.

fredg1:
I did end up complying.

> ...noted. For now, I'll guess I'll just comply and delete the comment, since the intent, which was to just make you (collectively) aware that this method was coming, was achieved either way...

https://github.com/kolmafia/kolmafia/pull/311#discussion_r762611987

jaadams5:
> I did end up complying.
> 
> > ...noted. For now, I'll guess I'll just comply and delete the comment, since the intent, which was to just make you (collectively) aware that this method was coming, was achieved either way...
> 
> [#311 (comment)](https://github.com/kolmafia/kolmafia/pull/311#discussion_r762611987)

I'll look things over.  But if you are saying you complied with he3's suggestion you could resolve the conversation yourself.

jaadams5 (approved):
Trusting you on this, but what I understand seems OK.

fredg1:
> I'll look things over. But if you are saying you complied with he3's suggestion you could resolve the conversation yourself.

I... did?

PR #312, opened 2021-12-04
Add my script "AccountVal"

libraryaddict:
https://github.com/libraryaddict/KolAccountVal/


jaadams5:
So I just copied the file locally, put it in data. ran mafia and there was nothing in my Install scripts menu.

Debug log was in included "org.json.JSONException: JSONObject["forumThread" not found."

Not sure where to point fingers but this should get resolved before approval :-)

jaadams5 (approved):

PR #314, opened 2021-12-04
Use cloudfront for images, take two.

heeheehee-kolmafia:
KoL is currently serving a mix of cloudfront / S3 traffic, so we should be
strict about requesting from cloudfront if we originate a request, but lenient
about accepting responses that contain non-cloudfront URLs.

Further, downloadImage() was constrained such that it would look for the last
occurrence of .com, which naturally stopped matching once we moved to
cloudfront.net URLs.

Also also, there were a bunch of regexes that expected image URLs to contain
"images.kingdomofloathing.com", which was true under S3 but is no longer true
with Cloudfront. Many of these are underspecified out of laziness, but I can fix
that if needed.

### DIFF
--- a/src/net/sourceforge/kolmafia/utilities/FileUtilities.java
+++ b/src/net/sourceforge/kolmafia/utilities/FileUtilities.java
@@ -355,6 +355,9 @@ public class FileUtilities {
     }
     String images = ".com";
     int index = filename.lastIndexOf(images);
+    if (index == -1) {
+      index = filename.lastIndexOf(".net");
+    }
     int offset = index == -1 ? 0 : (index + images.length() + 1);
     String localname = offset > 0 ? filename.substring(offset) : filename;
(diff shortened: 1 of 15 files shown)


src/net/sourceforge/kolmafia/utilities/FileUtilities.java
```diff
@@ -355,6 +355,9 @@ private static String localImageName(final String filename) {
     }
     String images = ".com";
     int index = filename.lastIndexOf(images);
+    if (index == -1) {
```
jaadams5, line 358: This is fine as is.  But it could be futureproofed by using the image server enumeration to derive candidate strings rather than hardwire com and net.  Of course if there are ever different images on different servers and it matters which one is returned then this, and the derivation, would need to be documented or reimplemented :-)

jaadams5 (approved):

PR #318, opened 2021-12-06
Introduces commerceGhostCombats pref

Thoth19:
This will allow us to determine how long we think it will be before the
commerce ghost triggers.

This is useful because it lets scripts like CHiT tell the player how many more combats they need to enter in order to trigger.
Note that it is combats entered rather than combats won or turn costing. I just checked using tatters.

I ran gradlew test and ran this once or two by hand on my own account. Is there a particular set of things I am meant to do other than that?


gausie:
Ideally you could add some tests to confirm that the value changes. Adding tests is a little complicated but there are some existing FightRequest tests you could look at

Also I would have this pref count up from 0 instead of down from 10

fredg1:
> Ideally you could add some tests to confirm that the value changes. Adding tests is a little complicated but there are some existing FightRequest tests you could look at
> 
> Also I would have this pref count up from 0 instead of down from 10

If it counts up, how does the user know at which possibly arbitrary number does the count end?

gausie:
It's not arbitrary, you know it by looking at the wiki. It would be nice
eventually to be able to document and annotate preferences and that would
be the place to include that max in mafia source. I agree that counting
down is nicer in many ways but this way results in 10 being the value even
if you don't have your ghost out. It's also the standard we use across the
existing codebase

Thoth19:
That makes sense. I chose to count down for that exact reason -- to make it immediately obvious how far away we are. But the consumer of the data will end up being someone/something with access to the right info anyway.
I will note the information in the source where I use it. Is there another place to document it?

I see that there are some FightReq tests in FightRequestTest.java. It doesn't look like there are many. Am I looking in the correct place?

Thoth19:
I have made the change for going from 0->10 rather than 10->0.
I am having some difficulty getting the tests to work the way I wanted. I asked a question over in discord, but not sure if this is a better place to ask those questions.
Posting the incomplete tests so that my question(s) are easier to answer. I figure they will get fixed in a followup commit and squashed together before we would merge anyway.

jaadams5:
handleGhostOfCommerce needs to be non-private before tests will compile on my system.  Tests are also failing.  I ran gradlew spotlessapply and will push momentarily.  More feedback later.

jaadams5:
I did some things so that tests would pass.  I reverted making handleGhosts.. private because I think a good test is possible without it.  I think the test should be using  FightRequest.parseFamBattleHTML instead of FightRequest.updateCombatData but I don't have the ghost and don't have sample html to try.  I understand the wip status so feel free to revert or undo anything.  I will check in later.

Thoth19:
So in order to use parseFamBattleHTML, what HTML do I need to provide? Is this the HTML from the whole fight? Or just a round of it?

gausie:
I've written a few tests for FightRequest [here](https://github.com/kolmafia/kolmafia/pull/323), so that might be something to base yours off once 323 is merged.

jaadams5:
I've merged.  We need html and this looks to be done.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -2323,6 +2323,17 @@ public static final void updateCombatData(
...
 
+      FamiliarData familiar = KoLCharacter.getEffectiveFamiliar();
+      int familiarId = familiar.getId();
+      // If other familiars also end up getting charged at start of fight rather than end we can put
+      // them here
+      switch (familiarId) {
+        case FamiliarPool.GHOST_COMMERCE:
+          {
+            Preferences.increment("commerceGhostCombats");
+          }
```
gausie, line 2334: I'd like a break here ideally, just to avoid future issues

jaadams5: Done.  Good catch.  Thank you.

gausie:
I think we can merge this as is, @jaadams5. Unless you want to convert to the pattern I used.

jaadams5:
> I think we can merge this as is, @jaadams5. Unless you want to convert to the pattern I used.

Since PRs with tests is an aspiration and not a requirement I'm OK with moving forward and hoping we revisit in the future.

Thoth19:
Sorry. I meant to work on this last week, but I was kinda busy with traveling and work things. Thanks for taking the time to fix this up. :)

jaadams5:
> Sorry. I meant to work on this last week, but I was kinda busy with traveling and work things. Thanks for taking the time to fix this up. :)

No problem.  If you do grab the HTML let me know and between us we can upgrade the tests.  Stick it in root/request and make a PR.

PR #319, opened 2021-12-06
Updating 2CRS seeds with new Potions and DB Wallaby reshuffle

chunkinaround:
This is to update all seeds to the new 2CRS potions and also to update DB wallaby after its effect shuffle. I believe it to have shuffled due to the fact that it was the only seed that had Brined Liver available.


jaadams5:
I have no particular interest in 54 files :-) If you will tell me what you did to generate them and what you did to verify at least one then I am willing to trust this ;-)

gausie (approved):
Looks good to me, thanks as always @chunkinaround!

chunkinaround:
@jaadams5  I have a multi in every class/moonsign combination and just re-derive the files using tcrs reset, tcrs derive, and tcrs save so I can keep the seeds up to date in mafia. Im also threebullethamburgler in game, and thus the same 3BH from the last update!

PR #320, opened 2021-12-06
TCRS derive now runs in parallel

gausie:
As far as I can see runtime changed from 2-3 hours to 5 minutes per seed


gausie:
Both I and @chunkinaround have tested this and it speeds up this run significantly!

PR #324, opened 2021-12-10
Rhino v1.7.14

gausie:
v1.7.14 released, ready to merge!

-----

I've been running RC1 since it was released without an issue. This branch is for people to try out the same if they like.

Once 1.7.14 final is released, I'll update this branch and we can merge.

Milestone for reference: https://github.com/mozilla/rhino/milestone/14


jaadams5 (approved):
Might as well push and see what breaks

PR #326, opened 2021-12-10
Keep appearance rate calculation DRY

gausie:
This is primarily in advance of implementing Be Gregarious to our queue management. We have our queue management logic in multiple different places, so this combines them and makes it easier to support more queuey things in future.

I did this tests-first, so it shouldn't have an effect on any behaviour


gausie:
This has given me the idea to include clover and semirare adventures in this sort of stuff - might look into it in the future.

jaadams5 (approved):

PR #328, opened 2021-12-11
Add CLI command for the guzzlr tablet

midgleyc:
Another CLI command that's useful to me personally. Includes a test.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
@@ -43,22 +43,26 @@ public class GuzzlrCommand extends AbstractCommand {
       }
       parameters = parameters.substring(7);
       Integer option = null;
-      if (parameters.equals("bronze")) {
-        option = 2;
-      } else if (parameters.equals("gold")) {
-        if (Preferences.getInteger("guzzlrBronzeDeliveries") < 5) {
-          KoLmafia.updateDisplay(
-              MafiaState.ERROR, "You need to make 5 bronze deliveries to serve gold clients.");
-          return;
-        }
-        option = 3;
-      } else if (parameters.equals("platinum")) {
-        if (Preferences.getInteger("guzzlrGoldDeliveries") < 5) {
-          KoLmafia.updateDisplay(
-              MafiaState.ERROR, "You need to make 5 gold deliveries to serve platinum clients.");
-          return;
-        }
-        option = 4;
+      switch (parameters) {
+        case "bronze":
+          option = 2;
+          break;
+        case "gold":
+          if (Preferences.getInteger("guzzlrBronzeDeliveries") < 5) {
+            KoLmafia.updateDisplay(
+                MafiaState.ERROR, "You need to make 5 bronze deliveries to serve gold clients.");
+            return;
+          }
+          option = 3;
+          break;
+        case "platinum":
+          if (Preferences.getInteger("guzzlrGoldDeliveries") < 5) {
+            KoLmafia.updateDisplay(
+                MafiaState.ERROR, "You need to make 5 gold deliveries to serve platinum clients.");
+            return;
+          }
+          option = 4;
+          break;
       }
       if (option == null) {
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
```diff
@@ -0,0 +1,90 @@
...
+public class GuzzlrCommand extends AbstractCommand {
+  public GuzzlrCommand() {
+    this.usage = " [abandon | accept <bronze|gold|platinum>] - Use the Guzzlr tablet";
+  }
+
+  @Override
+  public void run(final String cmd, String parameters) {
+    if (!InventoryManager.hasItem(ItemPool.GUZZLR_TABLET)
+        && !KoLCharacter.hasEquipped(ItemPool.GUZZLR_TABLET)) {
+      KoLmafia.updateDisplay(MafiaState.ERROR, "You don't have a Guzzlr tablet.");
```
jaadams5, line 21: "You don't have a Guzzlr tablet." or "You don't have a Guzzlr tablet equipped"

It is a nit but a message that tells the user exactly what to fix is more helpful than a suggestion.

src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
```diff
@@ -0,0 +1,90 @@
...
+        case "platinum":
+          if (Preferences.getInteger("guzzlrGoldDeliveries") < 5) {
+            KoLmafia.updateDisplay(
+                MafiaState.ERROR, "You need to make 5 gold deliveries to serve platinum clients.");
+            return;
+          }
+          option = 4;
+          break;
+      }
+      if (option == null) {
```
jaadams5, line 67: if the command parameter is bad is the right message for the user?  It would be nice if 

"guzzler <anything except abandon or accept> printed the usage string.

test/net/sourceforge/kolmafia/textui/command/GuzzlrCommandTest.java
```diff
@@ -0,0 +1,143 @@
...
+    // Stop requests from actually running
+    GenericRequest.sessionId = null;
+  }
+
+  public GuzzlrCommandTest() {
+    this.command = "guzzlr";
+  }
+
+  private static void hasGuzzlrTablet() {
+    AdventureResult.addResultToList(KoLConstants.inventory, ItemPool.get(ItemPool.GUZZLR_TABLET));
```
jaadams5, line 34: hasGuzzlrTablet sounds like it shout return a boolean.  maybe a more descriptive name?

jaadams5 (changes requested):
I'm OK with asking got the nits to be addressed.  I might also rethink some of the test names, knowing that there is no consistent convention.

src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
midgleyc: It can be in your inventory or equipped: either works.

src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
midgleyc: changed, but this is a common pattern across commands -- e.g. DaycareCommand has "Choice not recognised".

test/net/sourceforge/kolmafia/textui/command/GuzzlrCommandTest.java
midgleyc: changed to `playerHasGuzzlrTablet`.

src/net/sourceforge/kolmafia/textui/command/GuzzlrCommand.java
jaadams5: Well the right solution would be to fix DaycareCommand etc. but I didn't even see this as giving "Choice not recognised" so maybe I made an error.

jaadams5 (approved):

PR #330, opened 2021-12-11
typo

fredg1:


jaadams5:
where are telescope6 and 7 actually used in the code?  I can't find them.  They aren't in any of my settings so if if you have them it is presumably because of a script or code that has not been committed yet.

fredg1:
> where are telescope6 and 7 actually used in the code? I can't find them. They aren't in any of my settings so if if you have them it is presumably because of a script or code that has not been committed yet.

They're used indirectly, by concatenation

TelescopeRequest.java, line 128
`      String setting1 = "telescope" + upgrades;`

gausie:
Nah just remove 'em. It's impossible for them to have a value, right? So the default return of "" is correct for them (in case someone is still using them erroneously).

As long as `telescopeUpgrades` is not affected, I think we're good.

jaadams5 (approved):
Seriously?  ;-)

gausie:
Oh @fredg1 I would have removed the abiltiy for them to be set too in that loop!

PR #331, opened 2021-12-12
add a generic to ShowDescriptionTable

fredg1:
The generic still needs to be applied to its `getSelectedValues` method, but this will be done in a separate step, as it looks like the changes needed will be rather massive (lots of transitions from arrays to lists)


gausie (approved):
LGTM

PR #333, opened 2021-12-12
Fix html title hover not working as intended

libraryaddict:
The invalidation of the component is not needed, especially as it is called fairly often.

No need to unregister the component, as the component is not saved in a global object that needs to be removed later as originally thought.


jaadams5 (approved):
Trusting you on this :-)

PR #339, opened 2021-12-14
Language Server - Parse all - Transition from Exceptions to Diagnostics

fredg1:


gausie:
If this allows multiple errors to be collected per script (rather than throwing out at the first one), I'd like to see some tests that describe that behaviour

fredg1:
> If this allows multiple errors to be collected per script (rather than throwing out at the first one), I'd like to see some tests that describe that behaviour

Would it be okay if it was only a single test?
Currently, the main issue with these errors is that they aren't filtered at all (which is the next step).
This means that if we submit a script with syntax errors, the diagnostics get clogged with a couple of unwanted errors.

So, I'd only add a test with simple errors, like a break outside of a loop, unknown import, invalid aggregate key, etc.

gausie (approved):
Amazing!

PR #340, opened 2021-12-14
Fix text filtering in SkillBuffFrame.

heeheehee-kolmafia:
Some time back, we refactored AutoFilterComboBox. In doing so, we
changed its update() method from private to protected to expose it to
a subclass. What we didn't realize was that we were subsequently
shadowing this method in SkillSelectComboBox.

Thanks Veracity for [reporting this](https://kolmafia.us/threads/26961).


fredg1:
Wait... if this is the issue, then how was it working for the past 8 years??

gausie (approved):

PR #341, opened 2021-12-14
"enum constant needs a corresponding case label"

fredg1:
I remember seeing that having an empty "default" case was disliked, so instead, this only focuses on default cases which can immediately be inferred.
For example, if every case ends with a "return", this moves what's outside of the switch into the default case.

This cuts the "missing enum case" warnings by roughly half (~400 -> ~200)


gausie (approved):
These are all very reasonable and many are improvements on the existing code, though I do think that having a default case vs behaviour following a switch statement is more of a stylistic choice than a place for a warning.

PR #342, opened 2021-12-14
Cannot use `this` before it is defined (i.e. before super is called)

gausie:
Minor issue with the `reference` functions @fredg1 added to Type and RecordType. Unless I'm missing something, these were broken before,.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/parsetree/RecordType.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/RecordType.java
@@ -184,16 +184,16 @@ public class RecordType extends CompositeType {
 
   @Override
   public RecordType reference(final Location location) {
-    return new RecordTypeReference(location);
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):

gausie:
I was going to wait for @fredg1 to confirm this was right

fredg1:
Code_Wz1QbsDbPd
I don't really get what the problem was, what exactly was broken??

fredg1:
You only touched Type and RecordType, but *all* implementations of Type.reference() used `<super>.this.<field>` in their `super()` call.

What was the problem?

gausie:
> You only touched Type and RecordType, but _all_ implementations of Type.reference() used `<super>.this.<field>` in their `super()` call.
> 
> What was the problem?

I haven't looked at those other ones

PR #344, opened 2021-12-14
Make sure that all of the decorations after an Effect in the charpane do not wrap

gausie:
This fixes a weird error on safari where the size of the different decorated elements are recalculated but not repainted


Veracity0:
This worked for me. I see that it puts the (duration) and the uparrow into a no-wrap-on-space span, so it is no longer possible to get the uparrow on a line by itself. I approve.

Veracity0 (approved):
Tested, works fine on Safari.

PR #345, opened 2021-12-14
Cannot use `this` before it is defined (i.e. before super is called) 2

fredg1:


jaadams5 (approved):
Since I screwed up before @gausie needs to merge this.  I won't.

PR #347, opened 2021-12-15
Enforce @Override annotations

gausie:
Add `@Override` annotations where missing (thanks @heeheehee-kolmafia in [this thread](https://kolmafia.us/threads/text-filtering-no-longer-works-on-skill-name-field-in-skill-casting.26961/#post-165908))


gausie:
@heeheehee-kolmafia @BadHorseMonkey to get this to work I had to remove the `lib` directory from the `main` source sets. It's something I've been meaning to do for a while anyway. Can either of you explain why it is in there? I could be missing something but this change seems correct (and fixes some weirdness in my IDE while we're there)

jaadams5:
Historically lib contained source code KoLmafia needed.  It was not distributed by the authors as a jar.  There was a deliberate intention to use the code unmodified, primarily so that changes made by the authors could be assessed and perhaps incorporated.  As KoLmafia has grown and the infrastructure evolved, the requirement to compile and package lib has become problematic as show by lib's inclusion in test coverage statistics and code cleanup.  Perhaps it is a time to break the linkage?  My first thought would be to make the source in lib a separate project on Git, have that project generate a jar, manually include the jar in KoLmafia and then just remove the lib tree (and references thereto) from KoLmafia.

Maybe we shelve the annotation for the moment and do that instead?

gausie:
> Historically lib contained source code KoLmafia needed. It was not distributed by the authors as a jar. There was a deliberate intention to use the code unmodified, primarily so that changes made by the authors could be assessed and perhaps incorporated. As KoLmafia has grown and the infrastructure evolved, the requirement to compile and package lib has become problematic as show by lib's inclusion in test coverage statistics and code cleanup. Perhaps it is a time to break the linkage? My first thought would be to make the source in lib a separate project on Git, have that project generate a jar, manually include the jar in KoLmafia and then just remove the lib tree (and references thereto) from KoLmafia.
> 
> Maybe we shelve the annotation for the moment and do that instead?

Are you saying that doing it this way breaks jar building?

jaadams5:
> > Historically lib contained source code KoLmafia needed. It was not distributed by the authors as a jar. There was a deliberate intention to use the code unmodified, primarily so that changes made by the authors could be assessed and perhaps incorporated. As KoLmafia has grown and the infrastructure evolved, the requirement to compile and package lib has become problematic as show by lib's inclusion in test coverage statistics and code cleanup. Perhaps it is a time to break the linkage? My first thought would be to make the source in lib a separate project on Git, have that project generate a jar, manually include the jar in KoLmafia and then just remove the lib tree (and references thereto) from KoLmafia.
> > Maybe we shelve the annotation for the moment and do that instead?
> 
> Are you saying that doing it this way breaks jar building?

I'm not sure why you are asking that.  I was giving history.  The code in lib was never packaged in jars by the "owners" so we just built it.  Now that adding a jar to KoLmafia is trivial we could build a jar and remove lib from the Kolmafia source tree and this code we don't maintain would be excluded from lint type actions and test coverage.  Sorry I confused you.

gausie:
That's ok! I just didn't know if I was misunderstanding. I think this change would happen either way in that instance, so sounds good to me.

heeheehee-kolmafia:
@gausie see https://github.com/kolmafia/kolmafia/pull/117 for more context

gausie:
Ugh. There has to be a better solution.

src/net/sourceforge/kolmafia/utilities/PHPMTRandom.java
```diff
@@ -59,15 +60,17 @@ public int next(int bits) {
...
 
+  @Override
   public double nextDouble() {
     return nextInt() / (Integer.MAX_VALUE + 1.0);
   }
 
+  @Override
   public int nextInt(final int max) {
-    return nextInt(0, max);
+    return nextBoundedInt(0, max);
```
heeheehee-kolmafia, line 70: Would you mind making the non-override changes in a separate PR?

(This file, PHPRandom.java, and VotingBoothManager.java)

heeheehee-kolmafia (changes requested):
Other than a request to break out the non-annotation changes to a separate PR (and whatever you decide to do about lib/), looks reasonable.

src/net/sourceforge/kolmafia/utilities/PHPMTRandom.java
gausie: This is sort of part of the override changes - pmd says that `int nextInt(int int)` is an override and needs annotation and fails the build. It's not an override, so if I add that annotation my IDE highlights it as an issue. I think it's a pmd issue but this seemed like the easiest way to solve it.

heeheehee-kolmafia: Are you sure about that? I ran pmd locally and for PHPRandom, I see `int nextInt(int)` is an override (correct), but nothing about `int nextInt(int, int)`.

gausie: I am sure at least that it raised a false positive on my end. I'll revert the change and if it doesn't happen on the CI I'll commit it and look into why it's doing so for me only

gausie:
@heeheehee-kolmafia see now we get a PMD false positive

BadHorseMonkey:
I don't recall all the things that compile and stop working, but I know that we lose all of the 3rd party license text that we're supposed to display.  We probably need to hunt down the places those libraries exist and remove them if we can.

The licenses one should be easy to fix by moving the license text to data, but I know there were functional errors as well.  This would a bunch of testing.  

The last time we tried it, we reverted it instead of fixing it and rolling forward.

gausie:
Ok I've changed this PR to only add @Override annotations where they were missing. We don't actually enforce PMD and we don't make any changes to the build at all.

PR #348, opened 2021-12-15
Add Crimbo 21 Items and Effects

matt-chugg:
First pull request, apologies if it isn't right

### DIFF
--- a/src/data/fullness.txt
+++ b/src/data/fullness.txt
@@ -108,6 +108,7 @@ brains casserole	5	1	decent	6-14	0	8-10	0	Zombie Slayer
 bran muffin	1	1	awesome	4-5	150	150	150	25 All Branned Up (+40% Item Drop, 5-7 MP Regen)
 bread line	2	1	good	5-7	10-20	0	0	40 Bread-Lined (+50% Muscle, +25 Weapon damage)
 bread man	1	1	crappy	0	0	0	0	Unspaded
+bread pie	3	1	decent	0	0	0	0	Unspaded
 bread roll	2	1	awesome	8-10	0	0	10-20	leftovers, 20 Thanksgetting (Variable +meat, +item, +familiar weight, +all res)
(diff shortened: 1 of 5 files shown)


gausie:
@matt-chugg thanks for the PR! i would expect to see some changes to other files here:

* modifiers.txt will tell us what the items and effects do
* i'd expect to see the new booze in inebriety.txt
* and the new food in fullness.txt

We also need to add the new shop, but that doesn't need to come in this PR

matt-chugg:
Fair enough, understood, 
Was posting here as a PR when I would normally post on kolmafia.us, someone suggested data updates would be better in pull requests, and that's normally the extent of the data which comes from debug logs, I'll have a dig back through historic commits  for examples and submit a new PR (if I get time before someone else does it!)

matt-chugg:
> and that's normally the extent of the data which comes from debug logs

Apologies, that isn't a fair statement, there is of course all the data I need!

jaadams5:
I added some Advent calendar items in https://github.com/kolmafia/kolmafia/pull/316

It was a PITA for me because whatever I did to get from the forum post to the text files trashed the tabs even though quoting the session log was supposed to preserve them.  I think the culprit was my IDE but I have not confirmed and fixed.

jaadams5:
I also note failing tests.  My recollection is that I wrote the data file consistency tests to catch at least some of the tabs vs. spaces issues.  So that failure suggests there might be some editing to do.

matt-chugg:
- Have added fullness and inebriety
- Have added effect in modifiers

How do I add the source of Sauna-Fresh ? Its choice 145 option 1

gausie (approved):
Looks good to me!

gausie:
> How do I add the source of Sauna-Fresh ? Its choice 145 option 1

I'm not sure its possible or necessary (and I assume that's a typo for 1457 or something)

PR #349, opened 2021-12-15
KoLmafia.makePurchases accounts for PurchaseRequests with store limits

Veracity0:
KoLmafia.makePurchases is the method to purchase items from the mall. Both the "buy" command and InventoryManager.retrieveItem (used by "acquire" and retrieve_item()) use this.

Every time an item is successfully purchased, it will update the cached PurchaseRequests for that item so that subsequent purchases will have an idea of the currently state of the mall (modulo the fact that the entire rest of the kingdom is also interacting with the mall).

If you purchase from a store which sets daily limits on the item you just purchased, fewer of that item are available from that store and subsequent purchases may have to come from a different store - possibly with vastly higher prices.

This commit modifies the cached PurchaseRequest that was used to buy the item to account for the amount of the daily limit you have consumed. This will also recalculate the public "mall price" for the item - which is the price you'd pay for the 5th most expensive copy of the item.

This allows, for example, Sweet Synthesis to choose a new pair of candies before being forced into spending way more than expected to synthesize an effect.


gausie (approved):
Looks good to me! I assume this solves the reported mall limit bug

Veracity0:
I'm new at this. I added a few data file updates - new adventure zone and monsters - and didn't realize it was going into this same pull request. Please re-review; nothing was changed in the previously approved actual code.

src/data/items.txt
```diff
@@ -10864,9 +10864,7 @@
 10836	anti-creep cream	776583266	creamtube.gif	potion, usable	t,d	10
 10837	anti-pain cream	646241186	creamtube.gif	potion, usable	t,d	10
 10838	Doc's Special Reserve Wine	510170168	rxwine.gif	drink	t,d	10
-10839	bread pie	734136895	pie.gif	food	t,d	5
```
gausie, line 10867: Is this right?

src/data/monsters.txt
```diff
@@ -2175,13 +2175,17 @@ stumbling-drunk congoer (female)	1972	congoerf1.gif,congoerf2.gif,congoerf3.gif
 stumbling-drunk congoer (male)	1969	congoer1.gif,congoer2.gif,congoer3.gif	Scale: 5 Init: -10000 P: dude
 swimming pool monster	1970	poolmonster.gif	Scale: 5 Init: -10000 P: elemental
 
-# Crimbo 2018
+# Crimbo 2021
```
gausie, line 2178: I think you've accidentally switched 2018 and 2021

src/data/combats.txt
```diff
@@ -211,6 +211,7 @@ Shivering Timbers	100	deadwood tree	fur tree	hangman's tree	pumpkin tree	toilet-
 Shop Class	100	out-of-control bandsaw	deadly vise	demonic jigsaw	screw golem	X-fingered Shop Teacher: 0
 Simple Tool-Making Cave	100	flint-scraping cave elf	hunter-gatherer cave elf	rock-banging cave elf	sinew-stretching cave elf
 Sinister Dodecahedron	100	death ray in a pear tree: 1	turtle mech: 2	Swiss hen: 3	killing bird: 4	golden ring: 5	goose a-laying: 6	swarm a-swarming: 7	blade a-spinning: 8	laser lancing: 9	borg a-beeping: 10	strangler strangling: 11	stomper stomping: 12
+Site Alpha Dormitory	100	gooified elf-thing	gooified dog-thing
```
gausie, line 214: The zone definitely has NCs but I'm not sure anyone knows the %age (or if it's just timed or something)

gausie:
@Veracity0 no worries, next time you just have to checkout `main` again and make a new branch. It's possible to cherry pick these unrelated commits over to a new branch right now, but I don't think its necessary.

src/data/items.txt
Veracity0: oops. No.

src/data/monsters.txt
Veracity0: Sigh.

src/data/combats.txt
Veracity0: I've run a bit on multiple characters and it seems timed - but supposedly Clara's bell can force it.
I also think elf-things are more common than dog-things - but this all needs spading.
We can expect updates/corrections. :)

gausie (approved):

PR #351, opened 2021-12-15
Language Server - error filter

fredg1:
Look for BadType & co before making new errors. Since bad nodes are only generated from errors, it means we already made one.


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -2843,9 +2870,23 @@ private Evaluable parseCall(final BasicScope scope, final Evaluable firstParam)
...
       params = this.autoCoerceParameters(target, params, scope);
     } else {
-      this.undefinedFunctionError(name, params);
-
       target = new BadFunction(name.content);
+
+      // Don't make an error if the function couldn't be found
+      // because the user messed up one of the parameters
+      // (since that means we already made one earlier)
+      boolean error = true;
```
gausie, line 2878: You could write this like

`boolean alreadyErrored = params.stream().anyMatch(param -> param != null && param.getType().isBad());`

gausie:
This all looks good but I think this in particular needs a lot of tests. There are so .any situations to test that we don't double error and while you might have written it well now, tests ensure that it's written well when others work with it in the future.

src/net/sourceforge/kolmafia/textui/Parser.java
gausie: You could even do a parallel stream!

gausie:
I don't really understand why you wrote the tests like this - I was expecting to see more tests like `testErrorFilter`. This may test the functionality but I'm not sure how clearly it defines the behaviour. What do you think?

Also presumably we can get rid of "we currently can't read past the..." comments now?

fredg1:
> Also presumably we can get rid of "we currently can't read past the..." comments now?

Incorrect. We indeed still can't read past the \<x\> _before making the error_.
Currently, we still can't because doing so could create other errors before the one we wanted to make.
Later on, we'll be able to read the full aggregate before making the error, while still making sure that said error will always have priority over any thrown during the parsing of said aggregates.

fredg1:
> I don't really understand why you wrote the tests like this - I was expecting to see more tests like `testErrorFilter`. This may test the functionality but I'm not sure how clearly it defines the behaviour. What do you think?

The first test clearly shows a script throwing an error.
The second one is a modification to the script that results in another error being thrown earlier. Additionally, the earlier that happened in the first script does _not_ show up in any of the errors thrown in the second script, showing that it was effectively filtered.


The main issue I faced was that it was hard to convey how the modified script "should" have made the error, and that the filter is the only reason it didn't do so.

gausie:
> > I don't really understand why you wrote the tests like this - I was expecting to see more tests like `testErrorFilter`. This may test the functionality but I'm not sure how clearly it defines the behaviour. What do you think?
> 
> The first test clearly shows a script throwing an error. The second one is a modification to the script that results in another error being thrown earlier. Additionally, the earlier that happened in the first script does _not_ show up in any of the errors thrown in the second script, showing that it was effectively filtered.
> 
> The main issue I faced was that it was hard to convey how the modified script "should" have made the error, and that the filter is the only reason it didn't do so.

So its just running two different tests, but linking them because they're conceptually similar? Sounds like those should be split up for more unity unit tests. Why do you not just have one test per instance added in this PR that checks if a child isBad and thus doesn't throw?

fredg1:
> Why do you not just have one test per instance added in this PR that checks if a child isBad and thus doesn't throw?

But how would I check that said child "isBad", and how can I then prove that the lack of an error is specifically due to that?

gausie:
> > Why do you not just have one test per instance added in this PR that checks if a child isBad and thus doesn't throw?
> 
> But how would I check that said child "isBad", and how can I then prove that the lack of an error is specifically due to that?

You don't, you just write code that would result in that and assert the expected behaviour.

fredg1:
> you just write code that would result in that and assert the expected behaviour.

And... how is that different from what I did (the expected behaviour being "not throwing one particular error")

fredg1:
> So its just running two different tests, but linking them because they're conceptually similar? Sounds like those should be split up for more unity unit tests. Why do you not just have one test per instance added in this PR that checks if a child isBad and thus doesn't throw?

I can't do that because if I was to do it, these tests wouldn't be unit tests. They would need to test more than what they are supposed to, only to prove the absence of something

`testErrorFilter` was really the simplest example I could make (and even then, that test was stuck having to imitate `ScriptData`), but they all aren't that simple (in that they all don't make a single error), and these additional errors are unrelated to what these 'unit tests' are meant to test.
This is why the assertion made by the test needs to be limited to "none of the errors made by this script is 'that one'".

However, if those would then become individual tests, the issue I then thought of was that it would be too easy to submit to those tests an error message that just can't happen at all.
For example, what if someone changes the message of an error, and only fixes tests based on which fail?
They would miss those tests, because they wouldn't fail.
This is because even if the error we thought was filtered out was still thrown, it still wouldn't match what we thought its message was (since it was changed).

_This is why I paired the tests I made with another_. The tests need to prove that the error they assert isn't being thrown is an error that _can_ be thrown, which here is done by pairing them with a script that just so happens to throw them. Otherwise, you could just make a typo in the error message, and it would pass.

gausie:
I've found with you Fred that when you post a wall of text I only need to read the last paragraph. It makes interacting with you much easier :)

gausie:
Can we get a little bit more documentation of `invalid`, perhaps even a JSDoc with each parameter describes, and then I will merge

gausie (approved):

PR #354, opened 2021-12-16
Property forbiddenStores also excludes a store from appearing in mall searches

Veracity0:
We have an existing property "forbiddenStores". The contains a list of shopIds separated by commas.

Currently, it is only checked in MallPurchaseRequest.run() to abort the request if you try to purchase from such a store.

This submit also uses that property to exclude such stores from appearing in the results of MallSearchRequests, as well, so its items do not figure into mall price calculations; since we will not buy from the store, its inventory is irrelevant to finding the "5th least expensive price".


gausie:
@Veracity0 ideally you need to check out a new branch before making a PR, rather than committing to your own main branch. That'll allow you to submit multiple unrelated PRs

gausie (approved):

PR #355, opened 2021-12-16
Add raisin meringue pizza item

taltamir:
adds the item:
```
10847 raisin meringue pizza 268310979 pizza.gif food q,d 11
raisin meringue pizza 3 1 EPIC 0 0 0 0 Unspaded
Item raisin meringue pizza Effect: "Preemptive Medicine", Effect Duration: 13
```


gausie (changes requested):
Thank you! It looks like this item is randomly generated under a single item id so you should implement it like diabolic pizza

taltamir:
> Thank you! It looks like this item is randomly generated under a single item id so you should implement it like diabolic pizza

does the buff vary?

taltamir:
since this is randomly generated different items under the same ID this is too complicated for me. I will leave it to someone more experienced to implement

PR #356, opened 2021-12-16
More Crimbo 2021 support

gausie:
- NPC Stores
- New food item
- Decrementing items you put on the christmas tree
- Items that are in items.txt but not in npcstores.txt will still generate "learned a new npc store item" gcli output


jaadams5 (approved):
OK within the scope of my ignorance...

PR #357, opened 2021-12-16
Clear state leaked by CharPaneDecoratorTest.

heeheehee-kolmafia:
These fields are typically set by an ApiRequest shortly after login, but we're
not actually getting the full response text.

While I'm at it, this also changes ClearSharedState to a BeforeAllCallback (so
the same initialization will also apply to the first test suite that's
executed).

### DIFF
--- a/test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
+++ b/test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
@@ -12,15 +12,15 @@ import net.sourceforge.kolmafia.*;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.GenericRequest;
 import net.sourceforge.kolmafia.textui.command.AbstractCommand;
-import org.junit.jupiter.api.extension.AfterAllCallback;
+import org.junit.jupiter.api.extension.BeforeAllCallback;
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
```diff
@@ -12,15 +12,15 @@
...
 import org.junit.jupiter.api.extension.ExtensionContext;
 
-public class ClearSharedState implements AfterAllCallback {
+public class ClearSharedState implements BeforeAllCallback {
 
   // Prevents leakage of shared state across test classes.
 
   @Override
-  public void afterAll(ExtensionContext context) {
+  public void beforeAll(ExtensionContext context) {
```
jaadams5, line 23: I went for afterAll because I could imagine a case where a test prepopulated a directory with files before tests and beforeAll could break that.  If there is a test specific before/after that can still be done and, realistically, it's not highly probable so...

jaadams5 (approved):

test/net/sourceforge/kolmafia/extensions/ClearSharedState.java
gausie: BeforeAll is before a whole test class though right? We shouldn't be encouraging any statefulness between test classes

PR #358, opened 2021-12-16
MoodManager test and delinting

jaadams5:
coverage++

I wanted to use a real, prexisting moods file so stuck it in root/data but if there is a better place and/or a convention I missed then let me know.

I had much confusion getting triggers since the triggers for a particular mood are not always the same as MoodManager's list of triggers for display.  I think I fixed all that but more eyeballs won't hurt.

I handled leaks as I found them but I feel like there is a more uniform/consistent way to do so than what I found.

With all that I will still make this a PR and not a Draft :-)


gausie:
You could add an overload to `MoodManager.loadSettings()` that takes a File or something and uses that instead of the results of `getFile()`, then you can send your mock mood in via that. That's what I'd probably do anyway.

jaadams5:
> You could add an overload to `MoodManager.loadSettings()` that takes a File or something and uses that instead of the results of `getFile()`, then you can send your mock mood in via that. That's what I'd probably do anyway.

In a previous life changing code to support tests was A Bad Thing but that was then.  I'll try and figure something out.  Thanks.

gausie:
> > You could add an overload to `MoodManager.loadSettings()` that takes a File or something and uses that instead of the results of `getFile()`, then you can send your mock mood in via that. That's what I'd probably do anyway.
> 
> In a previous life changing code to support tests was A Bad Thing but that was then. I'll try and figure something out. Thanks.

My counter argument is that good code with good architecture is easily testable, so you're just fixing poor design choices. What it really says is that we need to split the function that reads the file and then parses it into one function that reads the file and another that parses it. Then we can test the latter.

jaadams5:
> > > You could add an overload to `MoodManager.loadSettings()` that takes a File or something and uses that instead of the results of `getFile()`, then you can send your mock mood in via that. That's what I'd probably do anyway.
> > 
> > 
> > In a previous life changing code to support tests was A Bad Thing but that was then. I'll try and figure something out. Thanks.
> 
> My counter argument is that good code with good architecture is easily testable, so you're just fixing poor design choices. What it really says is that we need to split the function that reads the file and then parses it into one function that reads the file and another that parses it. Then we can test the latter.

While researching I came across a quote, pulled out of context, "Don't Mock objects that you don't own".

But I don't need the file anymore...

gausie (approved):
Awesome!

PR #361, opened 2021-12-16
Stop automated adventuring in Site Alpha Dormitory when need more cold resistance.

Veracity0:
"It's getting colder! Better bundle up."

It also tells you how much you need. We should probably parse that and set a property or something, but this commit stops automation, at least.


gausie (approved):

jaadams5:
FWIW - Automated adventuring stopped for me with a nothing more to do message before this change.  I still had to visit in the relay browser to see what the Cold Resistance requirement was.

PR #363, opened 2021-12-16
Sweet synthesis iteration

Veracity0:
Although the SweetSynthesisRequest can be called with a "count", which will consume that many of each candy to get that many (*30) turns of the effect, it is slightly less request efficient but possibly vastly more cost (Meat) effective to call it 1 request at a time.

The ASH function - sweet_synthesis( count, effect ) used try to make a single request using count. Now it makes count requests, recalculating the best candy pair each time.

(I used to have code to do this in my VMF script, because I knew it was better. Now it's in the ASH runtime, so every ASH coder will get the benefit without needing special code. Although existing special code will still work.)


gausie (approved):
Removing lines of code seems good to me

PR #364, opened 2021-12-16
Track cold resistance required by the site alpha dormitory

gausie:
Also mention it in CLI failures when they show

### DIFF
--- a/src/net/sourceforge/kolmafia/request/AdventureRequest.java
+++ b/src/net/sourceforge/kolmafia/request/AdventureRequest.java
@@ -235,6 +235,16 @@ public class AdventureRequest extends GenericRequest {
     if (index >= 0) {
       String failure = KoLAdventure.adventureFailureMessage(index);
       MafiaState severity = KoLAdventure.adventureFailureSeverity(index);
+
+      // Add more details to the failure message when adventuring in the 2021 crimbo cold resistance
+      // zone.
+      if (index == 107) {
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -235,6 +235,16 @@ public void processResults() {
...
+
+      // Add more details to the failure message when adventuring in the 2021 crimbo cold resistance
+      // zone.
+      if (index == 107) {
+        int required = Preferences.getInteger("_crimbo21ColdResistance");
+        int current = KoLCharacter.getElementalResistanceLevels(MonsterDatabase.Element.COLD);
+        failure +=
+            " You need " + required + " (" + (required - current) + " more than you have now).";
+      }
+
```
gausie, line 247: @Veracity0 do you think failure-specific code here is a bad idea?

Veracity0: 107 is a magic number. It is the index in the array of failures?

I would prefer if it checked for a failure in the Site Alpha Dormitory, which is snarfblat #554. Something like is done later in the method: 

   if (this.formSource.equals("adventure.php") && this.adventureId.equals(AdventurePool.SITE_ALPHA_DORMITORY_ID) {
          ...code...
    }

Other than that, I like this idea.

gausie: Great!

PR #370, opened 2021-12-17
Clean up disk state

jaadams5:
A test was leaking files into root\settings.  I could not find it so went for the nuclear option of cleaning out the directory after all tests.  An argument could be made that cleaning the directories before and after tests is inefficient but it provides a stronger guarantee of state.


gausie (approved):
I like the change. I would probably do it before not after but still, lgtm

jaadams5:
> 

It was being done before and something slipped through.  It was easier to do this than find it.  I think we have a fair amount of experience plugging leaks and the time may come when we consolidate all that.  But after Christmas and Crimbo :-)

PR #371, opened 2021-12-17
Stop indents being applied to selected text

libraryaddict:
HTMLWriter has a default maximum line limit of 80, although I think lines with no spaces bypasses this.

What this commit does is adds a check if we're using an HTML Document, and applies a large maximum line limit.

This means lines are no longer indented.
This commit could perhaps be improved further with a proper class declaration, instead of an anonymous class.
setLineLength is a protected method you see.

Quick example of the indention follows, a straight copy/paste from the Mafia CLI display. Just a bunch of "test" with a space in between.

```
test test test test test test test test test test test test test test test     test test test test test test test test test test test test test test test     test test test test test test test test test test test test test test test     test test test test test test test test test test
```

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
+++ b/src/net/sourceforge/kolmafia/swingui/widget/RequestPane.java
@@ -127,18 +127,20 @@ public class RequestPane extends JEditorPane {
     StringWriter sw = new StringWriter();
 
     try {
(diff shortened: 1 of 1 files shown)


gausie:
I'm not really aware of the issue, but writing a test would both solve that problem and be good in general!

jaadams5 (approved):
I'd really like to see a test as well but since testing is an aspiration and a requirement I'll give this a pass.

PR #374, opened 2021-12-18
<string> == <string>

fredg1:
I've run through and removed some string comparisons that use == instead of String.equals


gausie:
> I've run through and removed some string comparisons that use == instead of String.equals

Thank you

gausie (approved):

PR #380, opened 2021-12-19
Fix up forbidden store stuff

libraryaddict:
* Slight cleanup of code
* Don't use string.valueof when its an int hashset
* Make sure it's a number before we parse the number


BadHorseMonkey:
Can you add tests for the empty property?

gausie:
> Can you add tests for the empty property?

Gonna merge this anyway just because its absence is breaking `main`, but I agree.

PR #381, opened 2021-12-19
Language Server - ErrorManager

fredg1:
First of all: while I did mark this PR as not-draft, unless you guys are sold on the idea, this PR is more of a proof-of-concept than anything, and the implementation is potentially subject to changes.

# What is it?
We currently have two general classes/severity of errors:
"Nuh-uh, you can't do that"
and
"Wait, WTF?? What are you doing???"

In both cases, we're currently stuck having to guess "why" these undesirable things are in our way, and to try to keep parsing based on that guess (at the very least to exit the current method).

If, however, we guess wrong, it can result in multiple redundant errors being made.
This feature is an attempt at "cushioning" the impact of that guess.

## So, what does it do, exactly?
Basically, the ErrorManager is something that keeps track of the worst level of error it has seen (currently only two, "default" and "syntax").
If it's submitted an error, it will only add it to the diagnostics if this is the worst level of error it has seen so far.

## Tests! Tests! Tests!
Adding tests for this behaviour... is potentially a bad idea.

This is because:
1- The severity at which each and every error has been assigned is rather arbitrary. Moreover, there's currently only two severity, so adding more is a possibility (changing which error is or isn't blocked by a previous one)
2- This feature and its behaviour can easily change. Currently, for example, ErrorManagers all only span a single method each, but in the future, could reach across methods (parsePluralConstant comes to mind)


gausie:
> ## Tests! Tests! Tests!
> 
> Adding tests for this behaviour... is potentially a bad idea.

I bet it isn't!

> This is because: 1- The severity at which each and every error has been assigned is rather arbitrary. Moreover, there's currently only two severity, so adding more is a possibility (changing which error is or isn't blocked by a previous one) 2- This feature and its behaviour can easily change. Currently, for example, ErrorManagers all only span a single method each, but in the future, could reach across methods (parsePluralConstant comes to mind)

I recognise that aspects of this might change in the future. That's not a reason to ignore testing. Please add comprehensive testing.

fredg1:
How's the concept, for starters?

jaadams5:
A mentor of mine was one of many people whom observed that it was easier to make a working program fast than to get a fast program working.  In context it was a focus mechanism.  I feel very much like this idea is before its time.  I do not understand how the LSP support with this will be any better or more robust than LSP support without it.  I do hope that LSP support might arrive sooner without it and that if I am wrong about the utility it could be added later.

To the extent that I understand what is being proposed we still have a fundamental philosophy difference that remains unresolved.  Unless the parser knows exactly what is wrong and is confident enough to fix the problem and continue without user intervention then the severity of an error is not relevant to me as the user.  Regardless of how severe an error is I still have to edit something and reparse.  I spent decades writing code in environments where I saw cascading errors with a single cause and so I am quite willing to accept that if resolving it impedes progress.

Note that I am at best lukewarm about the LSP idea but am trying to be supportive of the devs who support it.  I freely admit that there are many nuances I don't understand.

I agree with @gausie that tests are needed.  I will review and request those as changes.  Maybe that will help close this, one way or another :-)

jaadams5 (changes requested):
As per @gausie, more and better tests, please.  If that has been done then please walk me through the new work.  Thanks.

jaadams5:
The coverage metric for Parser seems pretty good to me so I will resolve the request for more tests.  However @fredg1 can resolve the merge conflicts before I approve and commit, please.

jaadams5 (approved):

jaadams5 (approved):

PR #382, opened 2021-12-19
Add some tests to the maximizer

midgleyc:
The maximizer is very big and complicated, and a good target for additional tests before refactoring (or in general).

I refactored the tests a bit to prefer using names to ids; I feel it reads better.

I added tests for max and min (inside modifiers, or starting for max), stat restrictions and clownosity. The documentation says that max and min at the start of an expression should apply to the entire expression, but I tried this and I don't think it works that way for min. It looks like "min" fails the maximization if the _current_ character items / effects don't meet the minimum.

### DIFF
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
@@ -29,6 +29,15 @@ public class MaximizerTest {
   }
 
   @Test
+  public void equipsItemsOnlyIfHasStats() {
+    addItem("helmet turtle");
+    addItem("wreath of laurels");
+    assertTrue(maximize("mus"));
+    assertEquals(1, modFor("Buffed Muscle"), 0.01);
+    recommendedSlotIs(EquipmentManager.HAT, "helmet turtle");
+  }
+
+  @Test
   public void nothingBetterThanSomething() {
     addItem("helmet turtle");
     assertTrue(maximize("-mus"));
@@ -77,9 +86,8 @@ public class MaximizerTest {
         modFor("Cold Resistance") - modFor("Combat Rate"),
         0.01,
         "Maximizing one slot should reach 27");
-    Optional<AdventureResult> acc1 = getSlot(EquipmentManager.ACCESSORY1);
-    assertTrue(acc1.isPresent());
-    assertEquals(acc1.get(), AdventureResult.parseResult("Krampus horn"));
+
+    recommendedSlotIs(EquipmentManager.ACCESSORY1, "Krampus horn");
   }
 
   @Test
@@ -180,9 +188,7 @@ public class MaximizerTest {
     assertEquals(3, modFor("Cold Resistance"), 0.01);
     assertEquals(20, modFor("Item Drop"), 0.01);
 
-    Optional<AdventureResult> hat = getSlot(EquipmentManager.HAT);
-    assertTrue(hat.isPresent());
-    assertEquals(hat.get(), AdventureResult.parseResult("bounty-hunting helmet"));
+    recommendedSlotIs(EquipmentManager.HAT, "bounty-hunting helmet");
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -175,27 +177,113 @@ public void keepsBonusBjornUnchanged() {
...
+  private Optional<AdventureResult> getSlot(int slot) {
+    return Maximizer.boosts.stream()
+        .filter(Boost::isEquipment)
+        .filter(b -> b.getSlot() == slot)
+        .map(Boost::getItem)
+        .findAny();
+  }
+
+  private void recommendedSlotIs(int slot, String item) {
+    Optional<AdventureResult> hat = getSlot(slot);
```
heeheehee-kolmafia, line 285: This variable probably shouldn't be called hat :)

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -120,7 +122,7 @@ public void freshCharacterShouldNotRecommendEverythingWithCurrentScore() {
   @Test
   public void noTieCanLeaveSlotsEmpty() {
     // 1 helmet turtle.
```
heeheehee-kolmafia, line 124: nit: this comment is no longer useful now that you've removed loadInventory.

(ditto above in a few other situations)

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -25,30 +23,39 @@ public void init() {
 
   @Test
   public void changesGear() {
-    // 1 helmet turtle.
-    loadInventory("{\"3\": \"1\"}");
+    addItem("helmet turtle");
+    assertTrue(maximize("mus"));
+    assertEquals(1, modFor("Buffed Muscle"), 0.01);
```
heeheehee-kolmafia, line 28: Probably doesn't need to change in this PR, but it's probably better to assert on expected gear rather than observed modifiers (which is more indirect and harder to understand).

(I recognize this is existing code. Up to you either way.)

heeheehee-kolmafia (changes requested):
Thanks! This is already looking a lot better than the hodgepodge of approaches we took historically for populating inventory / state.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
midgleyc: In general, I half agree, but I think there are cases where `modFor` is easier: e.g. weapon / off-hand that could be in either order, accessories that again could be in any order.

I think it's nicer to leave this as it is for this PR. I might change my mind after I think about this more / figure out a clean way of doing it.

BadHorseMonkey: [both.gif](https://tenor.com/I0sW.gif)

BadHorseMonkey: which is to say, testing modFor tests that the mod is applied, it's the expected amount, there are no other buffs from the maximizer, etc.

It doesn't test that the helmet turtle is selected, but that some item that makes the mod go to 1.  

Practically, that has to be the helmet turtle, because it's the only item in inventory, but these are two different things and could fail in different ways.  Which would both need to be checked the same way, so this is totally a nit...

heeheehee-kolmafia: Yeah. If your goal is specifically to check the modifier (as is the case with the noncombat test), then I agree that it's better to assert on the modifier's value.

If your goal is to check that a particular item is or isn't equipped but don't care which slot it's in, that's a place for improving the testing ergonomics (e.g. adding `void assertRecommended(String itemName) { ... }` that's slot-agnostic).

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -175,27 +173,113 @@ public void keepsBonusBjornUnchanged() {
...
   }
 
-  private void loadInventory(String jsonInventory) {
-    try {
-      InventoryManager.parseInventory(new JSONObject(jsonInventory));
-    } catch (JSONException e) {
-      fail("Inventory parsing failed.");
+  private void addItem(String item, int count) {
+    for (int i = 0; i < count; i++) {
+      AdventureResult.addResultToList(KoLConstants.inventory, AdventureResult.parseResult(item));
```
heeheehee-kolmafia, line 241: optional nit: you probably don't need to call AdventureResult.parseResult() in a loop (either store the value outside the loop, or figure out a way to set AdventureResult.count, e.g. by parsing `count + " " + item`).

That said, in this context you're unlikely to ever call it with a value of more than 4, so I don't care _that_ much.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -175,27 +173,113 @@ public void keepsBonusBjornUnchanged() {
...
 
+  private void addEffect(String effect) {
+    KoLConstants.activeEffects.add(EffectPool.get(EffectDatabase.getEffectId(effect)));
+  }
+
+  private void addSkill(String skill) {
+    KoLCharacter.addAvailableSkill(skill);
+  }
+
+  private void setStats(int muscle, int mysticality, int moxie) {
```
heeheehee-kolmafia, line 253: (I generally dislike functions with multiple parameters of the same type, but in this case, there's a canonical + well-known order.)

heeheehee-kolmafia (approved):
This looks fine now.

PR #384, opened 2021-12-19
When we attempt to search mall, always ensure the "Force Sort" option…

libraryaddict:
… is properly read.

This fixes searching mall from chat, without having searched through the Purchases tab not reading the checkbox.

Aka, checking the box, then attempting to visit someone's store through mafia's chat. Previously, the box would be ignored.


jaadams5 (approved):
Trusting you on this one.  chat is evil :-) so I tend to avoid it.

PR #385, opened 2021-12-19
Default value of a filter type is not true, but is checkbox.selected()

libraryaddict:
This fixes the exclusive setting of maximizer having checkboxes ticked wrongly


gausie:
Can you make a regression test for this?

jaadams5 (approved):

PR #386, opened 2021-12-19
Add chat frame width/height restoring

libraryaddict:
When it restored the size of other frames outside of tabbed chat, I found it both disorientating and annoying.
It'll still remember the sizes for now however, incase we plan to change this in the future.


jaadams5:
I'm lazy.  What happens if an old version of the preference does not have the width and height set and a tabbed chat is opened?

libraryaddict:
> I'm lazy. What happens if an old version of the preference does not have the width and height set and a tabbed chat is opened?

"location.length > 3" it already accounts for that :P

jaadams5 (approved):

PR #389, opened 2021-12-19
Update statuseffects.txt

Blunderbuss:
Added "Celestial Sheen", "Scaredy Dog", and "Furrier Than Thou" as wish-able effects (I think)


jaadams5:
Your "(I think)" concerns me :-)

You seem to be defining a new syntax for the file - instead of a single default action you are introducing a way to have multiple default actions.  I am guessing what you want to do but am not convinced it will work.  Could you elaborate on what you are trying to accomplish and why you think this accomplishes that?  Are there other genie wishes in the game that should be here if the file format supports that? Have you built this locally and seen mafia do what you hope for?

Blunderbuss:
Yeah you can reject this PR. I already confirmed on discord that my mafia
and someone else's are behaving differently.

jaadams5:
OK.

PR #390, opened 2021-12-20
Increment gold guzzlr quests by 1 on completion

s-k-z:
Prior to #374 this always incremented by 1 because Java matchers return a String. String == string is generally false because they do not exist at the same memory location. This PR is intended to revert to the original behavior that is now expected by users and simplify logic.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ChoiceManager.java
+++ b/src/net/sourceforge/kolmafia/session/ChoiceManager.java
@@ -13262,8 +13262,7 @@ public abstract class ChoiceManager {
                 // Increment the number of gold or platinum deliveres STARTED today
                 if (!tier.equals("bronze")) {
                   Preferences.increment(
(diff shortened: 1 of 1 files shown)


jaadams5:
I'm not sure about this.  The pre 12/18 code looks to me like the intent was to increment gold by three and platinum by one.  This change increments both by one.  

I'm also confused by your assertions concerning "String" and "string" and wonder if you are confusing Java with some other language?  "string" is not a primitive type in Java.

s-k-z:
I might have misunderstood the cause of why but this never actually incremented gold by 3. I pushed this after myself and another person observed that _guzzlrGoldDeliveries was always incremented by 1 until recently.

jaadams5:
So your assertion is that the gold and platinum preferences should both be incremented by one?

Makes sense, from looking at the code, but might as well confirm.

s-k-z:
Yep.

jaadams5 (approved):

PR #391, opened 2021-12-20
Add test for forbidden stores

libraryaddict:


libraryaddict:
Hmm, adding a super large number (eg 999,999,999,999) would fail the test as it'd pass the regex but fail to be parsed into an Integer.

And of course, the easiest way to "fix" that is to wrap every attempt at parsing the store id in a try / catch.
Uglier though.

Not sure if it's worth bothering given that you'd need to have a script going wild, or a user playing god to replicate it.

jaadams5:
Looks close.  I wouldn't worry as much as you did about invalid store ids and malformed preferences but it certainly cannot hurt.

I would consider two things.  There is a toggleForbiddenStore(int shopId) that could be exercised - pick a store check state, toggle and check for changed state.  Do this for something known to be in the list as well as something that is not there.  I would also consider a test that set the preference for one character, changed character and confirmed it was not set.

Thank you.

jaadams5 (approved):

PR #396, opened 2021-12-22
Show limits even if they're equal to or higher than quantity

libraryaddict:
This is stored in a separate field and the original "limit" field left mostly untouched.

To do otherwise would result in a large refactor over tens of classes. Either for renaming, for in correcting logic.

This allows limits to display even when the quantity in store is equal to, or lower than the store's limit.

image

### DIFF
--- a/src/net/sourceforge/kolmafia/request/MallPurchaseRequest.java
+++ b/src/net/sourceforge/kolmafia/request/MallPurchaseRequest.java
@@ -152,7 +152,8 @@ public class MallPurchaseRequest extends PurchaseRequest {
 
     this.quantity = quantity;
     this.price = price;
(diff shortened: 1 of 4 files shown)


gausie:
Why do you want this?

Also, I'd be tempted to do the full refactor - this solution will be really confusing in the future.

jaadams5:
> Why do you want this?
> 
> Also, I'd be tempted to do the full refactor - this solution will be really confusing in the future.

Agreed.  I'm not sure I see the benefit of the change in general and particularly if the cost is avoiding a refactor.

libraryaddict:
Fair enough, this change is mostly beneficial towards /games, and keeping inline with KOL itself.

PR #398, opened 2021-12-22
Extract EquipmentRequirement, more maximizer tests

midgleyc:
Extracted `EquipmentRequirement` (the equipment requiring parsing code) from `EquipmentManager` because I wanted to re-use it in the tests to set stats appropriately.

Refactored, reordered and added a few more maximizer tests.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/EquipmentRequirement.java
+++ b/src/net/sourceforge/kolmafia/session/EquipmentRequirement.java
@@ -1,0 +1,46 @@
+package net.sourceforge.kolmafia.session;
+
+import net.sourceforge.kolmafia.KoLConstants.Stat;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
+
+public class EquipmentRequirement {
+    private Stat stat = Stat.NONE;
+    private int amount;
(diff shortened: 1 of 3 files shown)


gausie (approved):
Love it, thank you

src/net/sourceforge/kolmafia/session/EquipmentRequirement.java
```diff
@@ -0,0 +1,46 @@
...
+
+    if (requirement.startsWith("Mys:")) {
+      stat = Stat.MYSTICALITY;
+      amount = StringUtilities.parseInt(requirement.substring(5));
+    }
+
+    if (requirement.startsWith("Mox:")) {
+      stat = Stat.MOXIE;
+      amount = StringUtilities.parseInt(requirement.substring(5));
+    }
```
gausie, line 24: Should we throw at the bottom here?

midgleyc: no, it's possible for the requirement to be "none" (in which case stat is Stat.NONE and amount is 0)

PR #399, opened 2021-12-22
Remove simple custom arrays

midgleyc:
Remove:
* AdventureResultArray, which just passed its input straight through to its internal list
* SortedListModelArray, which had some custom behaviour (if an invalid id was passed, return null) which was perfectly replicable with the standard HashMap.

The replacement saves on adding nulls in item id gaps, and also removes some custom code.

In newer versions of Java, `new AdventureResult[0]` is generally preferred to `new AdventureResult[this.internalList.size()]`: I understand it was poorly optimised in earlier versions, but this is no longer the case.

Remaining are `BooleanArray`, `IntegerArray` and `StringArray`. I think these could be swapped out with a custom HashMapWithDefault implementation (if not found, they return `false`, `0` and `""` respectively), like Python's `defaultdict`.


jaadams5 (changes requested):
Could we hold off on this for a bit?  I went through an effort to eliminate some of the custom arrays that could theoretically be replaced with modern Java structures.  I found out there was some custom behavior that I did not find easily replicable.  I'd like to see some tests that explicitly exercise the custom behavior and seem them pass before and after this change.  I feel obligated to write the tests myself (or say Never Mind) but that won't happen for several days.

gausie:
> Could we hold off on this for a bit? I went through an effort to eliminate some of the custom arrays that could theoretically be replaced with modern Java structures. I found out there was some custom behavior that I did not find easily replicable with a simple replacement. I'd like to see some tests that explicitly exercise the custom behavior and see them pass before and after this change. I feel obligated to write the tests myself (or say Never Mind) but that won't happen for several days.

What was the custom behaviour - what would you want @midgleyc to test here?

midgleyc:
Of the two I've replaced:
* AdventureResultArray doesn't have any custom behaviour
* SortedListModelArray has the behaviour that if you request an index that is not present, you get null. This is also true for hash maps.

The other ones have the behaviour that if you request an index that isn't present, you get a particular default: 0, empty string, or false. I've not replaced those because they were trickier.

They also have the behaviour that indexes up to the one requested are populated with whatever the default is, but this isn't externally visible and so isn't necessary to replicate (that is, the behaviour that getting an invalid key returns the default covers that).

gausie:
> The other ones have the behaviour that if you request an index that isn't present, you get a particular default: 0, empty string, or false. I've not replaced those because they were trickier.

Sounds like you could just use `getOrDefault` in those cases

midgleyc:
You could, but there are hundreds of usages, and I can see it being very easy to miss.

The other difference between the custom arrays and maps is that in some places, the custom arrays are iterated over, and that is currently in key order, but I'm not sure that it has to be. This maps to `map.values().sort(by key)` in pseudocode, so it's manageable, but again something to be considered.

jaadams5:
You have touched on some of my concerns but basically I was trying to buy time as a church musician looking at Christmas weekend who doesn't have time for much else for the next few days.  Specifically if my concerns can be addressed by tests, before and after then I feel an obligation to write those tests.  But doing so is a few days in the future.

jaadams5 (approved):

jaadams5:
Rehearsal was short and so I had some time.  The special behavior has been noted in comments above.  The one thing I am not sure about is size.  In the old code accessing an entry "after the end" padded the array so that subsequent calls to size were different.  I don't think that was relied on anywhere but...  I would prefer future changes (and perhaps this one could be redone) so that only one utility structure was replaced per PR.  That will make it easier for me to review and convince myself the special behavior isn't being relied upon.

midgleyc:
Filed #412 and #413.

They were done as two separate commits, so it's easy to split them out.

If you click on the individual commits (e.g. [the first](https://github.com/kolmafia/kolmafia/pull/399/commits/83fcba92067e70f4f86faf5b1bee9b712b7c69f0)), you can view them individually and use next / previous to navigate between them. But I'm also happy to file separately, as I have done.

PR #400, opened 2021-12-22
When checking underwater locations, check environment instead of list

midgleyc:
Several places in the codebase checked only for The Sea and the Party Yacht instead of all underwater locations. This PR changes that, and adds a test for the Maximizer.


gausie:
What are the other test changes?

midgleyc:
Not written, because no test existed to begin with. I can look into creating them.

gausie (approved):

PR #402, opened 2021-12-23
On purchase, include player id, and total cost in the CLI message

libraryaddict:
Player run store

> Purchasing BACON (5 @ 100 = 500) from #3272033...

NPC store

> Purchasing chewing gum on a string (2 @ 47 = 94)...

### DIFF
--- a/src/net/sourceforge/kolmafia/request/PurchaseRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PurchaseRequest.java
@@ -240,7 +240,7 @@ public abstract class PurchaseRequest extends GenericRequest
             + " @ "
             + this.getPriceString()
             + (this.limit > 1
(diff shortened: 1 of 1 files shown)


gausie:
I'm neutral on this change, learning towards reject. I'll let someone else weigh in.

jaadams5:
The change does break some session log parsing.

I wouldn't have spent the time to do it but I can lean towards approval.

libraryaddict:
> The change does break some session log parsing.

To be clear, you mean session log parsing by third parties?
This shouldn't effect the sessions file I believe, and the CLI has a max line limit regardless.

jaadams5:
> > The change does break some session log parsing.
> 
> To be clear, you mean session log parsing by third parties? This shouldn't effect the sessions file I believe, and the CLI has a max line limit regardless.

Definitely third party parsers.

jaadams5 (approved):

PR #405, opened 2021-12-23
Propagate negative weights, unchanged.

heeheehee-kolmafia:
Certain negative weights have special meaning (-2 == IS_BANISHED, -1
== IS_ULTRA_RARE). Users apparently rely on this.

Thanks Malibu Stacey @ kolmafia.us for [reporting this bug](https://kolmafia.us/threads/27039/).


gausie (approved):
Sorry for the bug, thanks for the fix

PR #410, opened 2021-12-23
Don't send empty macros to KoL

buttercookie42:
Sometimes Kolmafia might generate a combat macro which, after optimisation,
constists solely of the placeholders

#mafiaround
#mafiamp
#mafiaheader

handleMacroAction() then further optimises that kind of macro down into a bunch
of newlines, which then cause an "Invalid macro" error when sent to KoL.

With this change, we also check that the macro isn't empty *after* the pro-
cessing done by handleMacroAction(), and only send it on to KoL if it's truly
non-empty.

See also https://kolmafia.us/threads/consult-script-ccs-that-gets-optimised-out-to-nothing-leads-to-invalid-macro-abort.26620/

### DIFF
--- a/src/net/sourceforge/kolmafia/request/FightRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FightRequest.java
@@ -1040,10 +1040,6 @@ public class FightRequest extends GenericRequest {
   }
 
   private void handleMacroAction(String macro) {
-    FightRequest.nextAction = "macro";
-
-    this.addFormField("action", "macro");
-
     // In case the player continues the script from the relay browser,
     // insert a jump to the next restart point.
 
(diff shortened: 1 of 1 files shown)


gausie (commented):
Do you think it would be wise to log a message here saying that we're choosing not to submit a blank macro? When does this tend to happen.

buttercookie42:
Yes, probably not a bad idea, will add a log message.

Like I explained in the forum thread, at least for me this seems to happen when I have a custom combat script consisting of a single conditional action (like olfaction or the generic "special action") plus a consult script. When the conditional action doesn't take place (e.g. in the case of olfaction because I've already olfacted something, so Kolmafia seems to directly skip generating an olfaction action in the macro, or in the case of the "special action" because there's no special action scheduled in that round) Kolmafia seems to generate this sort of

#mafiaround
#mafiamp
#mafiaheader

de facto (but not technically) empty macro, which the `MACRO_COMPACT_PATTERN` replace step then turns into a bunch of newlines. This still used to work last year, but it no longer does now, so probably something changed on the KoL-side, because I think I also tried a build from last year and this time it no longer worked, either.

gausie:
> Yes, probably not a bad idea, will add a log message.
> 
> Like I explained in the forum thread, at least for me this seems to happen when I have a custom combat script consisting of a single conditional action (like olfaction or the generic "special action") plus a consult script. When the conditional action doesn't take place (e.g. in the case of olfaction because I've already olfacted something, so Kolmafia seems to directly skip generating an olfaction action in the macro, or in the case of the "special action" because there's no special action scheduled in that round) Kolmafia seems to generate this sort of
> 
> #mafiaround #mafiamp #mafiaheader
> 
> de facto (but not technically) empty macro, which the `MACRO_COMPACT_PATTERN` replace step then turns into a bunch of newlines. This still used to work last year, but it no longer does now, so probably something changed on the KoL-side, because I think I also tried a build from last year and this time it no longer worked, either.

Right I see, so it sends that macro then calls your consult script? I wonder if a message will be annoying then. We shall see, and can always remove it.

gausie (approved):
Using the debug log is smart, thanks.

gausie:
Would like a test for this if there's an obvious way to do so. It looks like the closest public caller of this method is `FightRequest.runOnce`

buttercookie42:
Makes sense, though I might need a little bit to figure out your testing framework and how a test could look like.

PR #414, opened 2021-12-24
Primary lab

Veracity0:
Add tiny plastic primary lab. It has no visible Enchantments. If they fix that, we'll have to update.
Add support for Choice Adventure 1461 in the Site Alpha Primary Lab.
Note that option #5 is "Grab the Cheer Core". If I recall, you get that once - and it is your only choice.
Therefore, if you are automating and that choice comes up and that option is present, we select it, regardless of how you have the choice configured.
Note also that this choice adventure does not have a blue "Encounter" when you find it.
I bug reported it to KoL. Surely they'd prefer to have something funny, rather than nothing.

Also added two properties:

primaryLabGooIntensity (default 10)
primaryLabCheerCoreGrabbed (default false)

### DIFF
--- a/src/data/adventures.txt
+++ b/src/data/adventures.txt
@@ -312,6 +312,7 @@ Tammy's Offshore Platform	adventure=538	Env: underwater Stat: 0	The Impenetrable
 Crimbo21	adventure=554	Env: indoor Stat: 0	Site Alpha Dormitory
 Crimbo21	adventure=555	Env: indoor Stat: 0	Site Alpha Greenhouse
 Crimbo21	adventure=556	Env: outdoor Stat: 0	Site Alpha Quarry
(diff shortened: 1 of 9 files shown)


gausie:
Thanks V. We could use deriving the current intensity from the construct names too.

Veracity0:
> Thanks V. We could use deriving the current intensity from the construct names too.

Well, if it says 10-ton, that means intensity is no greater than 10, but it could be less.
Do I need to do this before you'll approve? :)

Veracity0:
Also added the name change: inert grey goo ring -> ert grey goo ring.
Without this, cannot do mall search for it.

gausie:
> > Thanks V. We could use deriving the current intensity from the construct names too.
> 
> Well, if it says 10-ton, that means intensity is no greater than 10, but it could be less. Do I need to do this before you'll approve? :)

Of course not!

gausie (approved):

PR #418, opened 2021-12-25
refactor: remove BooleanArray

midgleyc:
Replace with Map<Integer, Boolean>.

Add tests to cover the affected areas.


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -2072,8 +2071,8 @@ void enumerateEquipment(int equipScope, int maxPrice, int priceLevel)
     StringBuilder outfitSummary = new StringBuilder();
     outfitSummary.append("Outfits [");
     int outfitCount = 0;
-    for (int i = 0; i < usefulOutfits.size(); i++) {
-      if (usefulOutfits.get(i)) {
+    for (Integer i : usefulOutfits.keySet()) {
+      if (usefulOutfits.getOrDefault(i, false)) {
```
gausie, line 2075: Surely this can't ever be non existent

midgleyc: Pretty sure it can't, but this is how it worked before, so I just changed it.

Same comment for the other loop.

I could remove it, I think we'll hear pretty quickly if it breaks something ;)

gausie (approved):

PR #420, opened 2021-12-26
Support more than 2 sessions in parallel

oxc:
The restriction to two sessions in parallel seems a bit arbitrary. I sometimes have 3 characters running in parallel, especially if one of them got stuck unexpectedly :D

This change makes the number of possible parallel sessions a bit more dynamic, although still with a hard upper limit to avoid endless resource allocation.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafia.java
+++ b/src/net/sourceforge/kolmafia/KoLmafia.java
@@ -142,6 +142,7 @@ public abstract class KoLmafia {
 
   public static final int[] initialStats = new int[3];
 
+  private static final int MAX_SESSIONS = 10;
   private static FileLock SESSION_HOLDER = null;
   private static FileChannel SESSION_CHANNEL = null;
   private static File SESSION_FILE = null;
@@ -283,9 +284,14 @@ public abstract class KoLmafia {
     if (SwinglessUIUtils.isSwingAvailable()) {
       KoLmafia.initLookAndFeel();
     }
-    if (!KoLmafia.acquireFileLock("1") && !KoLmafia.acquireFileLock("2")) {
-      System.out.println("Could not acquire file lock");
-      System.exit(-1);
+    for (int i = 1; i <= MAX_SESSIONS; i++) {
+      if (KoLmafia.acquireFileLock(String.valueOf(i))) {
+        break;
+      }
+      if (i == MAX_SESSIONS) {
+        System.out.println("Could not acquire file lock");
+        System.exit(-1);
+      }
     }
 
     FlaggedItems.initializeLists();


oxc:
Looks like a missed a spot :) Will fix right away.

jaadams5:
In Ye Olden Days the decision was not arbitrary.  The primary reason - to prevent abuse of KoL's servers seems to me to remain.  I would prefer that there be some consensus among the devs before this is approved especially @Veracity0 's opinion.

heeheehee-kolmafia:
I was under the impression that the issue is actually that we don't perform locking file I/O, e.g. when updating your settings.txt, and as such file corruption (including data loss) can and will happen, given enough time. Having more concurrent sessions only accelerates that issue.

heeheehee-kolmafia:
See this thread for some more details: https://kolmafia.us/threads/6044/#post-45232

I don't think anything's changed in the intervening decade.

jaadams5:
File I/O noted.  But when I implemented it long ago I was asked not to commit it because of the specter of a KoLmafia driven bot army.  If it was A Good Idea I would have been asked to deal with the file locking first is the way I remember it.  Of course I also remember when Hola would refuse to let anything that could automate Crimbo until after the event ;-)

Veracity0:
Thanks for finding the legacy thread! I agree that nothing has been changed in the last decade.

I looked a bit, yesterday, and noticed we had file locking for the session file, but perhaps that was for multiple threads, not processes.

Hola's comment was illuminating.

I don't want to allow some arbitrary number of parallel characters logging in and simultaneously writing to GLOBAL prefs. Not without substantial stress testing across all supported OSs.

Veracity0:
Regarding automating Crimbo, eventually I reached out to Jick. He thanked me for asking and said he had no problems with it and he was no longer worried about KoLmafia users overwhelming the KoL servers.

I also expect their server infrastructure has modernized over the years - and perhaps the number of active users has flattened out, although I do not have that data.

jaadams5:
Since there are good technical reasons for not commiting this and perhaps some social reasons as well I'm inclined to close it in the next few hours.

oxc:
This sounds reasonable. Thanks for all the interesting information! Merry Crimbo :)

oxc:
Do I see this correctly that as a workaround, I can set `useCWDasROOT` to use different folders for different characters (instead of all of them using `~/.kolmafia`)?

heeheehee-kolmafia:
> Do I see this correctly that as a workaround, I can set `useCWDasROOT` to use different folders for different characters (instead of all of them using `~/.kolmafia`)?

Yes, that's a valid / more robust workaround, and one that doesn't run into the same data race issues.

BadHorseMonkey:
> > Do I see this correctly that as a workaround, I can set `useCWDasROOT` to use different folders for different characters (instead of all of them using `~/.kolmafia`)?
> 
> Yes, that's a valid / more robust workaround, and one that doesn't run into the same data race issues.

Yep.  You won't be able to use global settings and you'll fetch certain data once for each data root, but as long as that's OK with you, that's viable.  You could script up something that set a per-character folder as the CWD (and therefore root).  

I'm not keen on using the application's directory as the writable user directory, but some people have been doing that just fine for years.

jaadams5:
I'm going to close this because we are not going to implement the suggest change and the workarounds and alternatives can be discussed elsewhere.

PR #424, opened 2021-12-28
add available_combat_skill ASH function.

Veracity0:
KoLConstant.availableCombatSkills is maintained by FightRequest; when you are in a fight, it is populated with the skills listed in the skill dropdown. This can change each round.

Added KoLCharacter.availableCombatSkill( skillId ) and KoLCharacter.availableCombatSkill( UseSkillRequest ) to check if the supplied skill is currently available.

Added ASH function: boolean combat_skill_available( skill ) to do the same thing.
Note that this is only accurate while you are in a fight; it is intended for use in consult scripts and combat filters.

Also, KoL has finally given a name to the Choice Adventure in the Site Alpha Primary Lab: Hello Knob My Old Friend
Not that this is useful, but it's nice to have the comments be correct, at least.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -4063,6 +4063,19 @@ public abstract class KoLCharacter {
     return list.contains(skill);
   }
 
+  public static final boolean availableCombatSkill(final int skillId) {
+    UseSkillRequest skill = UseSkillRequest.getUnmodifiedInstance(skillId);
+    return KoLCharacter.availableCombatSkill(skill);
+  }
+
+  public static final boolean availableCombatSkill(final UseSkillRequest skill) {
+    if (skill == null) {
+      return false;
+    }
+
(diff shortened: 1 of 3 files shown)


gausie:
Why not name it `have_combat_skill` or `combat_skill_available`? `available_combat_skill` to me sort of sounds like something that would return an available skill, rather than return a boolean

Veracity0:
OK. have_skill() returns true if the skill is on your charsheet.
Transcendent Olfaction is a combat skill, so I would think have_combat_skill() would return true for that.
But combat_skill_available() recognizes that combat skills can become situationally unavailable.

gausie (approved):
Sounds good to me! I assume it doesn't consider rectifiable unavailability like insufficient MP?

Veracity0:
It looks only what what the combat skill dropdown on fight.php says.
If KoL itself filters out skills without sufficient MP, it will return false.
As I mentioned, it is intended for consult scripts and combat filters - i.e., things where you have programmable control about using MP producing items before casting a skill next round.

PR #427, opened 2021-12-30
Allow -tie to affect hardcoded tiebreakers.

heeheehee-kolmafia:
In particular, this affects item droppers / meat droppers / rollover
effects, but not Bs in Bees Hate You or breakables (although could
control those cases too; I figured those were generally desirable
though).

Context: https://kolmafia.us/threads/the-maximizer-ignores-tie.27073/


gausie (approved):
Makes sense and thanks for adding a test case. This will probably break someone's misuse of -tie in maximizer, but I think that's fine.

PR #429, opened 2021-12-31
Language Server - makeChild

fredg1:
Adds a way for the language server to supply not only the content of the current script, but also all the script that will be imported by it


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -391,6 +393,29 @@ private Scope importFile(final String fileName, final Scope scope, final Locatio
...
+      return childConstructor.newInstance(scriptFile, stream, this.imports);
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      // The caller didn't want us to be able to do this, so don't.
+    } catch (InstantiationException e) {
+      // Shouldn't happen; it can't be an abstract class since we got it from an instance.
+    } catch (InvocationTargetException e) {
+      throw new RuntimeException(e);
+    }
+
+    return new Parser(scriptFile, stream, this.imports);
```
gausie, line 412: Under what circumstances should we get here?

fredg1: this happens if the subclass doesn't have a
`public <class>(File, InputStream, Map) {}`
constructor.

This happens, for example, if we make an anonymous subclass
`new Parser() {};`
whose class becomes `Parser$1` and doesn't have any constructor.
(though I reckon that we might want to recursively retry with the superclass rather than immediately go for Parser...)

fredg1: update: we now recursively retry with the parent class.

gausie: Ok now add a test for that loop please

gausie: Now we're in a situation where if you let the while loop run until it hits the top, line 425 is the same as line 415. You should switch the while to be while the current class has a parent at all, and then `return null` or perhaps better throw some exception at the end of the function since we should never get there

fredg1: > Ok now add a test for that loop please

I... did?

fredg1: > Now we're in a situation where if you let the while loop run until it hits the top, line 425 is the same as line 415.

...no?
The condition is `currentClass != Parser.class`, meaning line 415 will never make a new instance of `Parser`. If the class is `Parser`, we just fall through to line 425

gausie: Yes and now 425 is identical to 415 if the loop were allowed to take one step further up.

fredg1: ...but it's not?

gausie: Yes it is

fredg1: I mean that it's not allowed to take one step further

gausie: Thats why I'm saying you should allow it to go one step further and we can avoid code duplication.

gausie: * Make the loop stop when currentClass doesn't have a parent
* Have some sort of return or throw outside the loop

fredg1: >     * Make the loop stop when currentClass doesn't have a parent
>     * Have some sort of return or throw outside the loop

It's **already** effectively done. There's a ClassCastException from `Class.asSubclass(Class)` if we "somehow" miss Parser's constructor.

fredg1: Besides, you *just* asserted that the now-removed static call, *which cannot fail*, is identical to line 415 if we allow one more iteration
> Yes and now 425 is identical to 415 if the loop were allowed to take one step further up.

So *why*, once I humoured you, are you suddenly saying we need to consider the possibility of it failing??
Making that call completely safe *was the whole point* of the static call.

gausie: Feel free to edit or delete those messages

fredg1: Alright, you're definitely just trolling me at this point.

@lostcalpolydude help please.

jaadams5:
Request for help noted,

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -398,6 +400,34 @@ private Scope importFile(final String fileName, final Scope scope, final Locatio
...
+  /**
+   * Makes a new instance of the first class we can find that has a public constructor expecting
+   * File + InputStream + Map
+   */
+  private Parser makeChild(final File scriptFile) {
+    final InputStream stream = this.getInputStream(scriptFile);
+
+    Class<? extends Parser> currentClass = this.getClass();
+
+    while (true) {
```
jaadams5, line 412: Please rewrite using some other means of control than while (true).

jaadams5 (approved):

PR #435, opened 2022-01-01
Fix codeql warning when casting a float to an int

gausie:


jaadams5 (approved):
Surprised I missed that this had been flagged.

PR #437, opened 2022-01-01
intrinsics in charpane have span directive just like effects

Veracity0:
decorated effects in charpane have a span directive surrounding the (DURATION).
That includes intrinsics, although the DURATION is &infin;
intrinsics, however, short circuited the processing and did not generate a /span to close it.

Note that this does not fix the CharPaneDecoratorTest - which includes four intrinsics; the output files for that did not include the span and /span. Those tests are disabled at the moment. Further work is required.


Veracity0:
That commit does fix the CharPaneDecoratorTest. As it had been, even though it had intrinsics in the CharPane, it didn't actually test decorating intrinsics; just regular effects.

PR #438, opened 2022-01-02
Language Server - interruptions

fredg1:
The language server will run the parsing of the various scripts in separate threads.

This allows us to save resources by allowing those threads to be interrupted when we no longer need them (if, for example, a file is modified during parsing).

Is this a behaviour that needs to be tested? Not sure how I'd even test that.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/AshRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/AshRuntime.java
@@ -133,6 +133,9 @@ public class AshRuntime extends AbstractRuntime {
     try {
       this.parser = new Parser(scriptFile, stream, null);
       this.scope = parser.parse();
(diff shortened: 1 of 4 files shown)


gausie:
Fred the answer is always yes. Definition of insanity and all that...

fredg1:
but *how*

I can't just start the parsing in a new thread and "hope" I interrupt the parsing before it's over.

gausie:
By mocking importFile to throw the exception

fredg1:
> By mocking importFile to throw the exception

which? one is not called during parsing, and the other is private.

fredg1:
(can you mock a method even if it's private? I forgot)

gausie (approved):
Not the test I expected but that'll do I guess

PR #439, opened 2022-01-02
remove the concept of clover protection. Hermit has 11-leaf clovers

Veracity0:
The Hermit has 11-leaf clovers. The "hermit" command and menu item will fetch those, as will breakfast.
Do not search for disappearing clover messages. Do not worry about protecting ten-leaf clovers.

### DIFF
--- a/src/net/sourceforge/kolmafia/RequestEditorKit.java
+++ b/src/net/sourceforge/kolmafia/RequestEditorKit.java
@@ -58,7 +58,6 @@ import net.sourceforge.kolmafia.session.Limitmode;
 import net.sourceforge.kolmafia.session.NemesisManager;
 import net.sourceforge.kolmafia.session.OceanManager;
 import net.sourceforge.kolmafia.session.RabbitHoleManager;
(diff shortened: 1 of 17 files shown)


src/net/sourceforge/kolmafia/RequestEditorKit.java
```diff
@@ -334,6 +323,7 @@ protected static final void applyPageAdjustments(
     } else if (location.startsWith("bigisland.php")) {
       IslandDecorator.decorateBigIsland(location, buffer);
     } else if (location.startsWith("casino.php")) {
+      // Should this be Lucky! instead?
```
Rinn, line 326: Yeah, it's confirmed the big rock is still from the casino and requires Lucky! now.

Rinn (approved):

Veracity0:
OK. I adjusted the message in the Relay Browser - and also got rid of the remnants of the clover warning.
The seimrare warning will be a better model for the eventual Lucky! warning

PR #444, opened 2022-01-05
add data for new ascension equipment, AEDs

midgleyc:
The items tend to grant skills, but the skills aren't public, so they'll have to trickle in one by one.

Added AEDs to spleenhits, discriminated by id. Don't know if that works.


jaadams5:
> Added AEDs to spleenhits, discriminated by id. Don't know if that works.

I'm not sure what I am supposed to do with that comment.  Is it the equivalent of "untested" when making a commit in Ye Olde Days?  Does it mean this is really a Draft and you are still working on something?  Does it mean this is your best guess based upon spading to date?

jaadams5 (approved):

jaadams5:
OK.  I understand the comment and you did what I would have with a similar level of uncertainty.

PR #445, opened 2022-01-05
dead code

fredg1:
removes some dead code.

Does *not* touch dead code caused by hard-coded booleans
(example: "true || condition")

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
@@ -2391,11 +2391,7 @@ public class ItemDatabase {
   }
 
   public static int getNoobSkillId(final int itemId) {
-    Integer skillId = ItemDatabase.noobSkillIdByItemSource.get(itemId);
-    if (skillId == null) {
-      return -1;
-    }
-    return skillId.intValue();
+    return ItemDatabase.noobSkillIdByItemSource.get(itemId);
   }
 
   public static int[] getItemListByNoobSkillId(final int skillId) {
--- a/src/net/sourceforge/kolmafia/request/ShrineRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ShrineRequest.java
@@ -179,7 +179,7 @@ public class ShrineRequest extends GenericRequest {
       return false;
     }
 
-    String message = null;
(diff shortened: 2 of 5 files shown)


src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
```diff
@@ -2391,11 +2391,7 @@ public static boolean isAllowedInStandard(final int itemId) {
   }
 
   public static int getNoobSkillId(final int itemId) {
-    Integer skillId = ItemDatabase.noobSkillIdByItemSource.get(itemId);
-    if (skillId == null) {
-      return -1;
-    }
-    return skillId.intValue();
```
jaadams5, line 2398: Maybe I'm not digging deep enough but I cannot find a guarantee that ItemDatabase.noobSkillIdByItemSource.get(itemId) will never be null.  If it's not there ArrayList will throw an IndevOutOfBoundsException but null is AFAIK a perfectly reasonable element for an ArrayList.

fredg1: the return value of `get(itemId)` is also an `int`, so we just casting it to an Integer, so we were basically doing
```
int -> Integer -> null check
```

fredg1: this is because we use `IntegerArray`
```java
  public int get(final int index) {
    return index < 0 || index >= this.internalList.size() ? 0 : this.internalList.get(index);
  }
```

While it is true that `internalList.get(index)` could return `null`, it will produce a NullPointerException when trying to cast it to an int

jaadams5: While I am not certain I agree with your analysis, I am convinced that noobSkillIdByItemSource will only contain ints and adding/setting a null would cause the add/set to error.  So, for different reasons, we agree that skillId will never be null.

However in the process of tracking this down I note that ItemFinder has code that processes a -1 returned from getNoobSkillId.  I'd love to see that code removed because getNoobSkillId will never return a -1 or have an example where it will return a -1.  But I won't mark this as unresolved.

jaadams5 (approved):

PR #448, opened 2022-01-08
2022 changes to optimal dog, lucky lindy, pillkeeper, Marmot. Improve…

Veracity0:
… poolskill calculation.

The remaining changes from gausie's "lucky" PR that have nothing to do with "encounters".

- optimal dog gives Lucky!
- Lucky Lindy gives Lucky!
- pillkeeper gives Lucky! rather than forcing a semirare
- Marmot no longer gives a clover
- And simplify (rather than duplicate) code for pool skill calculation - in the process, not remerring to the pool shark as a "semirare" monster


gausie (approved):
Thanks for moving these changes over. Sorry I never finished that PR, new years until now has been a huge mess of covid, international travel and jetlag

PR #450, opened 2022-01-08
The Shore can redirect to shop.php

Rinn:


src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -1065,6 +1065,12 @@ public static final void handleServerRedirect(String redirectLocation) {
       return;
     }
 
+    if (redirectLocation.startsWith("shop.php")) {
+      // The Shore Inc. can redirect to the gift shop.
+      AdventureRequest.ZONE_UNLOCK.run();
+      return;
+    }
+
```
BadHorseMonkey, line 1073: I wish we could handle this (and the three identical code blocks above it) with something like a switch.  Are we far enough along on the Java version to get a better switch statement?

BadHorseMonkey (commented):
Is there a unit test that could be written for this function?

src/net/sourceforge/kolmafia/request/AdventureRequest.java
Rinn: > Is there a unit test that could be written for this function?

I'm not writing a unit test.

gausie: I have tried, there's no good test here for now.

gausie (approved):

PR #451, opened 2022-01-09
count fights against void monsters from the cursed magnifying glass with _voidFights

horrible-little-slime:


src/data/defaults.txt
```diff
@@ -2068,6 +2068,7 @@ user	_villainLairWebUsed	false
 user	_vitachocCapsulesUsed	0
 user	_vmaskAdv	0
 user	_vmaskBanisherUsed	false
+user _voidFights 0
```
Rinn, line 2071: need a tab here not a space

Rinn (changes requested):

src/data/defaults.txt
horrible-little-slime: Thanks!

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -2641,6 +2646,10 @@ public static final void updateCombatData(
             }
             break;
 
+          case VOID:
+            Preferences.increment("_voidFights");
```
gausie, line 2650: Can you max this out at 5 (there is an overload for doing so)? And can you add a check for the "time seems to stop" message, the failure of which should set this straight to 5.

horrible-little-slime: aye aye, cap'n

gausie (approved):
Thank you!

gausie:
Now before I merge @horrible-little-slime would you consider writing a test for FightRequest? There are some in the source to base it off

PR #453, opened 2022-01-10
Give full function prototype to all JS function stubs

gausie:
Now you can run `.apply`, `.call` etc on js functions that are generated from ash


phulin:
Only question - did you test in all of: a command-line context, as part of a main script, and as part of a module required by a main script?

gausie:
> Only question - did you test in all of: a command-line context, as part of a main script, and as part of a module required by a main script?

Yes, yes and (no before but now) yes!

PR #454, opened 2022-01-10
Add test for WitchessRequest

gausie:


gausie:
> Not failing locally for me. Messages talk about unnecessary stubbings. Any relationship to #453 ? Also not failure on ubuntu but not mac. (I'm on Windows.)

What message are you seeing that says that? No it can't be related to #453 I think

jaadams5:
In the raw output at https://github.com/kolmafia/kolmafia/runs/4766170533?check_suite_focus=true

Basically I went Checks->clicked on Check in the left pane, selected the failing Test Report (ubuntu) and poked around the failed tests.

That said there could also be some timeouts happening (according to an incomplete understanding of the stack trace)  so I just manually re-ran all checks.  We'll see if it does it again.

jaadams5:
Whoops.  I had not intended to push main, even though it will be relevant later.  Pushed the wrong button.  Revert if needed.

gausie:
Can't be bothered fixing this, it was just meant to be a way to spend 5 mins

PR #455, opened 2022-01-10
track magnifying glass state

horrible-little-slime:


gausie:
For consistency with this sort of pref I might subtract these from 13 and make the charges count up from 0 or -1

gausie:
Throw on a test for the manager and for FightRequest and we're good here!

jaadams5 (approved):

PR #457, opened 2022-01-10
HermitRequest limits clovers to number currently available.

Veracity0:
hermit * clover
=> get all remaining 11-leaf clovers
hermit 1 clover
=> get 1 clover if available
=> HermitRequest gives an error if ask for more than are available.


jaadams5 (approved):
Clearly dug deeper than I was willing to.  Thanks.

PR #462, opened 2022-01-11
Track beard buff

horrible-little-slime:
~~Creating this as a draft; I had a lot of issues setting up the test correctly because the wiki is inaccurately reporting one of the messages--in the message I found, it uses a double space after a period. Waiting to see if additional spading can help.~~

The message I used happened to be the only one that has a period mid-message. Everything is fine. Ignore me.

This tracks the last buff given by the Daylight Shavings Helmet as `lastBeardBuff`, saving the effect ID.


src/net/sourceforge/kolmafia/session/DaylightShavingsHelmetManager.java
```diff
@@ -0,0 +1,30 @@
+package net.sourceforge.kolmafia.session;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import net.sourceforge.kolmafia.preferences.Preferences;
+
+public class DaylightShavingsHelmetManager {
+  public static final Pattern[] DAYLIGHT_SHAVINGS_PATTERNS = {
+    Pattern.compile(
+        "Your helmet shoots some lasers at your face\\.  You smell burning hair\\.<center><table><tr><td><img class=hand src=\"\\/images\\/itemimages\\/dsstache\\d+\\.gif\" onClick='eff\\(\"[^\"]+\"\\);' width=30 height=30 alt=\"([^\"]+)\""),
```
gausie, line 10: With `\\.  ` the common way to handle it in mafia is `\\. +` to cover single and double space weirdness

test/root/request/test_fight_daylight_shavings_buff.html
```diff
@@ -0,0 +1,29 @@
...
...class="item" style="float: none" rel="id=787&s=35&q=0&d=1&g=0&t=1&n=1&m=0&p=0&u=b"><tr><td><img src="/images/itemimages/bottle.gif" alt="bottle of rum" title="bottle of rum" class=hand onClick='descitem(124295227)' ></td><td valign=center class=effect>You acquire an item: <b>bottle of rum</b> <font size=1>[<a href="craft.php?mode=cocktail&a=787&pwd=329b071e34c5164426d877eca2e0bca0">mix</a>]</font></td></tr></table></center><p><table><tr><td width=40><img src='/images/itemimages/songboombox.gif' /></td><td align='center'>You catch a line from the song playing on your SongBoom&trade; BoomBox:<br />&quot;And then I see the meat in your pants&quot;</td></tr></table><p><a name="end"></a><p><a href="adventure.php?snarfblat=451">Adventure Again (The Bubblin' Caldera)</a><p><a href="place.php?whichplace=airport_hot">Go back to That 70s Volcano</a></center><script>parent.charpane.location.href="charpane.php";</script></td></tr></table></center></td></tr><tr><td height=4></td></tr></table><table><tr><td>[<a href="/account_combatmacros.php">edit macros</a>]</td></tr></table></center></div></div></div></body><script src="/onfocus.1.js"></script><script src="/combatfilter.1.js"></script></html>
```
gausie, line 29: Probably remove the pwd from this just for peace of mind and consistency, even though it couldn't possibly do anything bad

gausie (commented):
Do we need to include the HTML at the end of each pattern?

horrible-little-slime:
Right now, the HTML is used to identify the message itself. We don't want
to just check to seenif you got the effect, because someone could have the
crazy horse out.

gausie:
> Right now, the HTML is used to identify the message itself. We don't want to just check to seenif you got the effect, because someone could have the crazy horse out.

Right but why not just have the messages about a hat that cuts your hair for you, without the HTML.

src/net/sourceforge/kolmafia/session/DaylightShavingsHelmetManager.java
```diff
@@ -0,0 +1,32 @@
...
+    "A nozzle emerges from your helmet and sprays a pattern of depilatory foam on your face.",
+    "A couple of straight razors snake out of a panel on the side of your helmet and quickly give you a shave.",
+    "A clippers-tipped robotic arm emerges from your helmet and gives you a quick face trim.",
+  };
+
+  public static void updatePreference(String responseText) {
+    for (String msg : DAYLIGHT_SHAVINGS_MESSAGES) {
+      int position = responseText.indexOf(msg);
+      if (position > -1) {
+        Matcher matcher = Pattern.compile("onClick='eff\\(\"(.*?)\"\\);'").matcher(responseText);
```
gausie, line 21: Compile this pattern as a static on the class.

src/net/sourceforge/kolmafia/session/DaylightShavingsHelmetManager.java
```diff
@@ -0,0 +1,32 @@
...
+        if (matcher.find(position)) {
```
gausie, line 22: I wonder if this is better than substringing the responseText in the previous line.

horrible-little-slime: I also wonder that!

gausie (approved):

PR #463, opened 2022-01-11
Orb changes

gausie:
The orb now

* Has a [ponder] link
* Has its predictions cleared on starting a fight with a previously predicted monster

so we react accordingly.

I think this is it implemented right, would appreciate someone more spadey checking it over.


midgleyc:
Some monsters don't have a/an pronouns (here's blank but I suspect "the" might exist?):
```
<li>  Sloppy Seconds Burger in Sloppy Seconds Diner</li>
```
(Note: space <pronoun would go here but doesn't> space <monster>)

Could you ask CDM if he'd be willing to add some HTML like `<!-- choice: XXXX -->`, like the item / monster ids, for less breakable mafia detection than matching on title? Unlikely to conflict I suspect, but still possible.

gausie:
My plan is to add a few tests to this and then merge

gausie:
> Some monsters don't have a/an pronouns (here's blank but I suspect "the" might exist?):
> 
> ```
> <li>  Sloppy Seconds Burger in Sloppy Seconds Diner</li>
> ```
> 
> (Note: space [ pronoun would go here but doesn't ] space [ monster ])

Sorted

> 
> Could you ask CDM if he'd be willing to add some HTML like `<!-- choice: XXXX -->`, like the item / monster ids, for less breakable mafia detection than matching on title? Unlikely to conflict I suspect, but still possible.

This should be coming whenever he has the time

midgleyc:
> Sorted

Might want to check whether "the" happens as well. My guess would be that the Hole in the Sky monsters have it.

> This should be coming whenever he has the time

Ace! These direct expositions of ids make things a lot more convenient.

PR #259, opened 2021-11-25
Raw arrays - InventoryPanel

fredg1:

PR #256, opened 2021-11-25
Raw arrays - SkillDatabase

fredg1:

PR #465, opened 2022-01-11
mood clear command fixed

jaadams5:
#338  eliminated an overload of removeTriggers.  One version operated on a Collection, the other on an array of objects.  It turns out that code now calls removeTriggers with the collection it is iterating over and deleting from.  That has some unpredictable and undesired side effects.  The easy solution was to have removeTriggers make a copy of the input before it iterated over it.  This PR does that plus adds some tests.  The Draft status will be removed when I stop writing the tests.


jaadams5:
@fredg1 - Let's talk if it is not clear what I think was broken and why.  This is an example of why I am wary of "delinting" for its own sake although I also think the real problem is ancient code that exposes collections that ought to be private.  If you think of any place something similar might have happened call it out and one of us can confirm and fix.  Thanks.

jaadams5:
Got a couple tests in.

fredg1:
Looking for more of these.

Preliminary search:
I looked at every single PR I made, and the only ones that also dealt with an array-list transition not contained within a single method are
https://github.com/kolmafia/kolmafia/pull/255
https://github.com/kolmafia/kolmafia/pull/256
https://github.com/kolmafia/kolmafia/pull/257
https://github.com/kolmafia/kolmafia/pull/258
https://github.com/kolmafia/kolmafia/pull/259
https://github.com/kolmafia/kolmafia/pull/260
https://github.com/kolmafia/kolmafia/pull/261
https://github.com/kolmafia/kolmafia/pull/262
https://github.com/kolmafia/kolmafia/pull/266
https://github.com/kolmafia/kolmafia/pull/267
https://github.com/kolmafia/kolmafia/pull/268

https://github.com/kolmafia/kolmafia/pull/298 also dealt with turning an array into a list, but I clearly remember *really* making sure it would be safe.

src/net/sourceforge/kolmafia/moods/MoodManager.java
```diff
@@ -155,9 +155,10 @@ public static MoodTrigger addTrigger(final String type, final String name, final
     return trigger;
   }
 
-  /** Removes all from the current displayList. */
+  /** Removes all triggers from the current mood and displayList. */
   public static void removeTriggers(final Collection<MoodTrigger> triggers) {
-    for (MoodTrigger trigger : triggers) {
+    MoodTrigger[] local = triggers.toArray(new MoodTrigger[0]);
```
Veracity0, line 160: Surely there is a way to do this that doesn't require you to copy the input list. For example, using an explicit Iterator.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/moods/MoodManager.java
jaadams5: Yes.  Is this conversation about "better" or just "different"?

Previously one of the removeTrigger routines was passed an array of Objects.  It iterated over the array and so there were different structures - one controlling the iteration and one holding the data being removed.  

The delinting that turned some arrays into lists created this bug.  The decision to copy to an array and iterate over that was made because it restored (a close approximation to) the previous implementation, in the hope that restoring the previous implementation was less likely to was less likely to introduce a new bug.  I also figured performance wasn't an issue since 1000 is certainly an upper bound on the size of the trigger list.

So the array copy was a deliberate choice but if I'm the only one who thinks so then an iterator will be used.

midgleyc:
#142 also dealt with a method that used to take an array, now takes a List, and iterates over it removing things. I don't think `  public static final void removeBuffs(final List<Offering> buffs)` ever gets passed the dangerous `BuffBotManager.buffCostTable` though.

src/net/sourceforge/kolmafia/moods/MoodManager.java
jaadams5: After a good nights sleep....

The "contract" for  removeTriggers is that it accepts a collection or triggers and then deletes any trigger that is in the collection from the (should be) internal list of mood triggers.  As such there is no requirement that the collection be a subset of the mood triggers so using the collection iterator to delete from the internal list is not going to work if the collection is just an arbitrary list of triggers.

So if we are going to allow removeTriggers to accept an arbitrary collection then we cannot use the iterator to delete.  So a defensive copy is appropriate.

I can imagine new KoLmafia code that makes a list of triggers that cannot be used and removing all those triggers.  It might allow for some automation when ascending.  But no one has written that yet.  

I think the core problem is that the mood command "knows" that there are functions to get a list of all triggers and to delete all the triggers in a list and composes them tom delete all triggers.  It seems like a "delete all triggers" function would expose not as many details of the MoodManager and that might be the best way to go.

Veracity0: You are right - the method cannot "know" that the collection it is passed is the internal list of triggers. But it can detect when it is - since it is explicitly deleting from that list - and use the iterator to delete, in that case.

`  /** Removes all from the current displayList. */
  public static void removeTriggers(final Collection<MoodTrigger> triggers) {
    Iterator<MoodTrigger> it = triggers.iterator();
    while (it.hasNext()) {
      MoodTrigger trigger = it.next();
      if (MoodManager.currentMood.removeTrigger(trigger)) {
        if (triggers == MoodManager.displayList) {
          it.remove();
        } else {
          MoodManager.displayList.remove(trigger);
        }
      }
    }
  }
`

Veracity0: Regarding "better" vs. "different":

Method A iterates through a list, search/removing elements from two other comparable lists. It copies the input list. So, the algorithm is O(n squared) in computation and O(n) in memory.

Method B does the same with no data copying. It is O(0) in memory.

They are "different". Is one "better"? Seemed obvious to me, which is why I phrased my suggestion in terms of "without allocating memory"

PR #467, opened 2022-01-12
avoidable typecasts

fredg1:


midgleyc:
With the discussion on #465 about efficiency: there are a lot of cases in Mafia where something is taken, converted to an array, then iterated over. Would it be possible to avoid the array conversion in most cases (i.e. when not deleting from the iterated list)? For example, could

```
String[] keys = map.keySet().toArray(new String[0]);
for (String name : keys) {
```

be

```
for (String name : map.keySet()) {
```

I understand you've been trying for minimal changes to increase the chance of these being accepted, but might these perf improvements (if they are indeed improvements) be worth going for additionally?

jaadams5 (approved):

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -908,9 +907,8 @@ private static void checkEquipmentMap(final PrintStream report, ItemMap imap) {
     report.println("# " + tag + " section of equipment.txt");
     report.println();
 
-    Object[] keys = map.keySet().toArray();
-    for (Object key : keys) {
-      String name = (String) key;
+    String[] keys = map.keySet().toArray(new String[0]);
+    for (String name : keys) {
```
gausie, line 911: Can we not iterate the keyset here too?

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -1097,10 +1095,9 @@ private static void checkItemModifierMap(
     report.println("# " + tag + " section of modifiers.txt");
     report.println();
 
-    Object[] keys = map.keySet().toArray();
+    String[] keys = map.keySet().toArray(new String[0]);
     int type = imap.getType();
-    for (Object key : keys) {
-      String name = (String) key;
+    for (String name : keys) {
```
gausie, line 1100: And here

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
fredg1: wtf; did I really forget to continue past the first one??
o///o

gausie (approved):

PR #468, opened 2022-01-12
Add heist CLI command

midgleyc:
For cat burgling convenience.

Displays valid heists or heists given item.


gausie:
Can we restructure these so you can get at them from ash? In particular, listing the possible heist targets.

test/net/sourceforge/kolmafia/textui/command/HeistCommandTest.java
```diff
@@ -0,0 +1,116 @@
...
+import net.sourceforge.kolmafia.objectpool.FamiliarPool;
+import net.sourceforge.kolmafia.request.GenericRequest;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class HeistCommandTest extends AbstractCommandTestBase {
+  @BeforeEach
+  public void initEach() {
+    KoLCharacter.reset("testUser");
+    new HeistCommandFakeRequest().register("heistFake");
```
gausie, line 20: This is one way of doing this for sure! I think there's a way of sneaking in to GenericRequest and telling it what to return, but I think to have easy access to those you need to be extending RequestBase

midgleyc: yeah, I'm already extending `AbstractCommandTestBase` here so I just went for a factory method pattern. Works well. Before DI, this was one popular way of testing classes with dependencies (in languages that let you extend classes and override methods).

src/net/sourceforge/kolmafia/textui/command/HeistCommand.java
```diff
@@ -0,0 +1,82 @@
...
+      return;
+    }
+
+    parameter = parameter.trim();
+
+    if (parameter.equals("")) {
+      // show all items
+      StringBuilder output = new StringBuilder();
+
+      var heistManager = heistManager();
```
gausie, line 31: Are we using `var` in non-test code yet? I've been avoiding it but I'm not sure if being so careful is actually required.

gausie: I like it too, we should allow it.

gausie (approved):

PR #471, opened 2022-01-13
feat: add parameter to not steal familiar equipment

midgleyc:
There are certain cases where Mafia switches to a familiar, not so that it can be used in adventuring, but so you can do something involving the familiar. In those cases, don't steal the familiar equipment (and then steal it back immediately after).

These are:
* using the Space Jellyfish in Breakfast to grab sea jelly
* using the Intergnat to summon a demon
* using the Cat Burglar to heist an item


gausie (approved):
This is great, thank you!

PR #473, opened 2022-01-14
Change rollover from time in seconds to rollover index

s-k-z:
This is a proposed fix to this issue reported on the forum [here](https://kolmafia.us/threads/some-users-reporting-failure-to-reset-daily-flags-on-rollover.27131/).


s-k-z:
Issue was resolved, API now returns a unix epoch timestamp again.

PR #474, opened 2022-01-14
Use API's coolitems key to track iotms instead of having dedicated GUI

gausie:


jaadams5 (approved):

jaadams5:
Do we have any insight into how KoL sets coolitems?  Does it indicate ownership, regardless of location (i.e. Hagnks or the Display Case) or is there an availability aspect to how it is set?  Is it dynamic, changing when something is sold?  Just curious.  Not relevant to this PR.

gausie:
> Do we have any insight into how KoL sets coolitems? Does it indicate ownership, regardless of location (i.e. Hagnks or the Display Case) or is there an availability aspect to how it is set? Is it dynamic, changing when something is sold? Just curious. Not relevant to this PR.

It indicates the fact that a given item has been bound to the account

PR #475, opened 2022-01-14
Make reading effect_desc, outfit_desc, and skill_desc thread safe.

Veracity0:
do not use/reuse static global request objects for fetching skill, outfit, and effect descriptions

We used to do that for desc_item calls, but fixed it when we had multiple threads trying to read item descriptions and they were stomping on each other.

Make the same fix for effect, skill,, and outfit descriptions.


gausie (approved):
This is very clearly good, thanks!

PR #477, opened 2022-01-15
checkrepo command

jaadams5:
Looks at the files in the local repo and flags likely duplicates when they are copied to KoLmafia's directories.

Update svnrepo.json for two scripts that have moved to GitHub.  Might require delete and reinstall for users of the SourceForge version if they want updates.

Figured out how to Mock file system access.  This works but much room for improvement.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
@@ -3793,4 +3793,6 @@ public class DebugDatabase {
       }
     }
   }
+
+  public static void checkLocalSVNRepository() {}
 }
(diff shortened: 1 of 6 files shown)


src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -92,7 +84,7 @@ private static String readWikiData(String url) {
           byte[] bytes = ByteBufferUtilities.read(istream);
           return StringUtilities.getEncodedString(bytes, "UTF-8");
         }
-
+        // JAA - Should this be && and not || ?
```
jaadams5, line 87: I'd like someone else's opinion.  Should this be (responseCode >= 301 && responseCode < 308) ?

gausie: I mean I suppose you could write `floor(responseCode/100) == 3`

gausie (approved):

PR #480, opened 2022-01-15
Wrong check for redirect?

jaadams5:
Old code actually triggers for non-redirect status codes, I believe.  Am I right?


src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -84,8 +84,7 @@ private static String readWikiData(String url) {
           byte[] bytes = ByteBufferUtilities.read(istream);
           return StringUtilities.getEncodedString(bytes, "UTF-8");
         }
-        // JAA - Should this be && and not || ?
-        if (responseCode >= 301 || responseCode < 308) {
+        if (301 <= responseCode && responseCode <= 308) {
```
jaadams5, line 87: Old code triggered the conditional if responseCode <= 301.  So the comparison to 308 was never made.

Veracity0 (approved):

PR #481, opened 2022-01-15
track combat usage of cosmic bowling ball

Veracity0:
- When you throw your cosmic bowling ball in combat, increment hiddenBowlingAlleyProgress
- When you use a bowling skill, remove it from inventory, increment _cosmicBowlingSkillsUsed, and set cosmicBowlingBallActive to true.
- When you receive a cosmic bowling ball in combat, set cosmicBowlingBallActive to false
- When you ascend, set cosmicBowlingBallActive to false

And when I inserted the case block to support the bowling skills in FightRequest.payActionCost, I noticed that the code following the case label for SkillPool.FIRE_EXTINGUISHER__ZONE_SPECIFIC ("Fire Extinguisher: Zone Specific") was a complete mess. Whoever wrote it apparently assumed that the code following the case label was an implicit block - defining local variables and omitting the break statement at the end. The result being, it then fell through the case label for  SkillPool.BE_GREGARIOUS.

I stuck {} around the "implicit block" and added a break;

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -440,6 +440,7 @@ user	controlPanelOmega	0
 user	copperheadClubHazard	none
 user	cornucopiasOpened	0
 user	corralUnlocked	false
+user	cosmicBowlingBallActive	false
 user	cozyCounter6332	0
 user	cozyCounter6333	0
 user	cozyCounter6334	0
@@ -1560,6 +1561,7 @@ user	_companionshipCasts	0
 user	_confusingLEDClockUsed	false
 user	_controlPanelUsed	false
 user	_corruptedStardustUsed	false
+user	_cosmicBowlingSkillsUsed	0
 user	_cosmicSixPackConjured	false
 user	_crappyCameraUsed	false
 user	_creepyVoodooDollUsed	false
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -440,7 +440,7 @@ user	controlPanelOmega	0
 user	copperheadClubHazard	none
 user	cornucopiasOpened	0
 user	corralUnlocked	false
-user	cosmicBowlingBallActive	false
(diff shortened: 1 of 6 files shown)


thedictatorkol:
Good to see bowlingball tracking and support, but I was hoping/expecting a countdown timer for how much longer your current bowling activity will be active. Something like bowlingBallReturnCombats instead of cosmicBowlingBallActive.

Veracity0:
I thought about that a bit.

If the goal is to indicate how many combats remain before your missing bowling  ball returns, then when you throw the ball as a combat item, that counter should be 0, since you will get it back at the beginning of the next turn.

That implies that the default value needs to be -1, and we count it down at the beginning of every combat, if it is non-negative.

I'll experiment tomorrow.

Veracity0:
OK, I implemented it before "tomorrow".

cosmicBowlingBallReturnCombats defaults to -1
is set to 0 when used as a combat item
is set to (calculated) number, based on skill usages so far today, when use a combat skill.
decrements by one (to a minimum of -1) at the start of each combat.

So, if that property is >= 0, you own a bowling ball, but it is not in inventory
If the property is == 0, you get it back at the beginning of next combat. Time to switch zone, perhaps!
If it is >0, cool your heels. Or your jets. Patience!

I'll test this after rollover.

Veracity0:
I tested it and discovered that setting the cool-down period to ( 2 * casts ) +3 is one too high; it counted down combats just fine, but the bowling ball returned when the cool-down count was 1, not 0. Perhaps the combat you are in when you cast the bowling skill counts as the first combat with the bowling ball out?

In any case, when you cast the skill, if I set the cooldown period to (calculation) - 1, my next test had the bowling ball return exactly when I expected - when the return count was 0, not 1.

jaadams5 (approved):

PR #482, opened 2022-01-16
Language Server - Outer Shell (StateCheckWrappers)

fredg1:
The skeleton of the Language Server, as well as StateCheckWrappers, which, like the name suggests, is a wrapper for the server, which takes care of accepting the messages in function of the server's state.

I thought making the tests for it would be simple. HAH. HAHAH. HAHAHAH. I laugh.

HAAAHAHAHA. I laugh some more.

HAAAAAAAAAAAAHHHAHAHAHA.
Kill me.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/langserver/AshLanguageServer.java
+++ b/src/net/sourceforge/kolmafia/textui/langserver/AshLanguageServer.java
@@ -1,0 +1,134 @@
+package net.sourceforge.kolmafia.textui.langserver;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import net.sourceforge.kolmafia.StaticEntity;
+import net.sourceforge.kolmafia.textui.langserver.textdocumentservice.AshTextDocumentService;
+import net.sourceforge.kolmafia.textui.langserver.workspaceservice.AshWorkspaceService;
+import org.eclipse.lsp4j.ClientCapabilities;
+import org.eclipse.lsp4j.InitializeParams;
+import org.eclipse.lsp4j.InitializeResult;
+import org.eclipse.lsp4j.ServerCapabilities;
+import org.eclipse.lsp4j.ServerInfo;
+import org.eclipse.lsp4j.jsonrpc.Launcher;
+import org.eclipse.lsp4j.launch.LSPLauncher;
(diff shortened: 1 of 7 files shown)


jaadams5:
Tests dump a stack trace to the console when running.  Since the exceptions are expected, is it possible to swallow the stack trace instead of dumping it?

`    private enum ServerState {
      STARTED,
      INITIALIZED,
      SHUTDOWN
    }`

occurs in at least two places.  Any reason not to share one instance?

fredg1:
> Tests dump a stack trace to the console when running. Since the exceptions are expected, is it possible to swallow the stack trace instead of dumping it?

Sadly, not to my knowledge. This is something that I tried to fix, and I just couldn't find a way to even access those errors.

The logging happens at `org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor`, line 115

```java
import java.util.logging.Logger;

[...]

	private static final Logger LOG = Logger.getLogger(ConcurrentMessageProcessor.class.getName());

        [...]

	public void run() {
		processingStarted();
		try {
			messageProducer.listen(messageConsumer);
		} catch (Exception e) {
			LOG.log(Level.SEVERE, e.getMessage(), e);
		} finally {
			processingEnded();
		}
	}
```

The logger the stack is pushed to is just not something I seem to have access to.
I'm new to the `java.util.logging.Logger` interface, so I may be wrong, but if so, I still just lack the knowledge.


> ` private enum ServerState { STARTED, INITIALIZED, SHUTDOWN }`
> 
> occurs in at least two places. Any reason not to share one instance?

I... don't exactly get what you mean?
Are you saying that I' declaring the enum at two different locations?

jaadams5:
OK.  I have not had to manage log4j for a very long time.  I won't distract you by asking you to dig deeper.  In the past I have done two things.  One was to set the logging level so that whatever was cluttering things would not be displayed.  The other was to tell log4j that /dev/null (or equivalent) was the log file.  We can defer investigating either until we learn that other people get concerned when there are stack traces in test output.

jaadams5:
Nevermind on ServerState.  I was looking at the same file in two different places and got confused.  :-(

jaadams5 (approved):

BadHorseMonkey:
> OK. I have not had to manage log4j for a very long time. I won't distract you by asking you to dig deeper. In the past I have done two things. One was to set the logging level so that whatever was cluttering things would not be displayed. The other was to tell log4j that /dev/null (or equivalent) was the log file. We can defer investigating either until we learn that other people get concerned when there are stack traces in test output.

More exciting than Log4J, it's java.utils.logging.Logger!
You need to switch from ConsoleHandler to StreamHandler than throw away the stream (or do whatever you want to do with it).

fredg1:
Is this something that requires access to the Logger that made the LogRecords? If so, it doesn't seem to be accessible...

PR #483, opened 2022-01-16
Add Sauce Spell Damage modifier

Rinn:
Make a new modifier for Sauce Spell Damage, this currently only applies to Intrinsic Spiciness.

image


gausie:
Looks good! I'd like to see a test on the modifier patterns. What do you see this as being for? Just to stop it showing up as a Spell Damage modifier?

jaadams5 (approved):

Rinn:
Main thing right now is this mod doesn't count towards the CS spell damage task so anyone writing a script has to compensate for it, but I do want to get skill damage into an internal mafia data file eventually and this is info that will be needed.

jaadams5 (approved):

PR #487, opened 2022-01-17
Update campaway ownership prop on visit attempt

soolar:
Currently campaway ownership only updates on visit to the forgotten forest, which took me a while to realize, after a few times manually setting the ownership prop after standard runs when going to visit it via a shortcut in my top bar didn't update it. Saw someone else having the same issue on the ASS discord so I figure it's not an entirely uncommon thing to be confused by, so why not change that?

Open to suggestions on better ways to check than lack of Uh Oh!, and I haven't checked if this works properly in run yet (I assume proper behavior would be the setting being false, so I'll have to check that next time I ascend before I mark this as a non-draft PR).


gausie:
I might prefer a positive check for the campfire image rather than a negative check for Uh-oh - especially if there's some path that disallows it with a different message or something.

PR #488, opened 2022-01-17
Scrapbook text

thedictatorkol:
Add missing messages for scrapbook, both collecting and tracking of the banish.


gausie (approved):
Thanks dict!

PR #489, opened 2022-01-17
RollingLinkList Test

jaadams5:
As implemented RollingLinkedList is potentially buggy because the implementation of addAll is provided by LinkedList.  That implementation can change the size of the list beyond the RollingLinkedList size limit.

This PR is part one of two.  The previous version of RollingLinkListTest did cover the code but was clearly driven by relearning how to write tests and by coverage reports.  It was less comprehensible in terms of RollingLinkedList functionality.

This rewrites the test in hopes of improving comprehensibility.  It includes a disabled test that fails because of addAll.

Part two will implement addAll in RollingLinkedList and enable the test.

The goal of part one is to get feedback on the test.

Part two will be along shortly.

### DIFF
--- a/test/net/sourceforge/kolmafia/utilities/RollingLinkedListTest.java
+++ b/test/net/sourceforge/kolmafia/utilities/RollingLinkedListTest.java
@@ -2,90 +2,91 @@ package net.sourceforge.kolmafia.utilities;
 
 import static org.junit.jupiter.api.Assertions.*;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 /**
- * As implemented RollingLinkedList will fail if passed an initial length that is negative. addAll
- * is not implementd and so the list can be longer than the declared length.
+ * A RollingLinkedList is bounded above in size. Items are added to the end of the list until the
+ * size is reached, at which point the first element is deleted. The update method is used as an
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/utilities/RollingLinkedListTest.java
```diff
@@ -2,90 +2,91 @@
...
+      removedValue = rll.update(testValues[i]);
+      assertNull(removedValue, "Undeleted element returned");
+    }
+    removedValue = rll.update(testValues[3]);
+    assertEquals(testValues[0], removedValue, "Unexpected value for delete element.");
+  }
+
+  @Test
+  @Disabled("RLL needs to implement addAll, not LinkedList")
+  public void addAllAndIteratedAddsShouldDoTheSameThing() {
```
jaadams5, line 85: This is wrong but will be fixed in part 2.

jaadams5:
The error in the disabled test was caused by running lint on the code and not paying attention.

The only place in KoLmafia that uses update never uses add or addAll so there should be no changes in behavior.

addAll is used in numerous places.  This fix will mean that RollingLinkedList actually enforces the limit.  That may change behavior but since it is incorrect behavior this might as well get fixed and then see what breaks.

gausie (approved):
Looks good. Hopefully we don't have anything behaving differently after this but, as you say, if we do it was broken anyway.

PR #494, opened 2022-01-19
ShopCommandTest

jaadams5:
Test for ShopCommand

Delete a file that leaks from a currently disabled test.

I think there is an error with "reprice d20 @ 1,333 limit 1" that results in a repriced item with no limit.  I will look at that later.

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/command/ShopCommandTest.java
+++ b/test/net/sourceforge/kolmafia/textui/command/ShopCommandTest.java
@@ -4,8 +4,10 @@ import static org.junit.jupiter.api.Assertions.*;
 
 import internal.helpers.Player;
 import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.KoLConstants;
 import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
+import net.sourceforge.kolmafia.persistence.ItemFinder;
 import net.sourceforge.kolmafia.preferences.Preferences;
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/textui/command/ShopCommandTest.java
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  @Test
+  public void itShouldPutAnItemFromStorage() {
+    int itemID = ItemPool.GUITAR_4D;
+    String itemName = ItemDatabase.getItemDataName(itemID);
+    String output = execute("put using storage " + itemName);
+    String expected = "Skipping '4-dimensional guitar', none found in storage." + LS;
+    assertEquals(expected, output, "Item not in storage.");
+    // Better way to convert item id or name to AdventureResult?
```
gausie, line 95: There are a bunch of ways: `ItemPool.get(int itemId)`, `AdventureResult.tallyItem(String itemName)` are the two that come to mind

jaadams5:
I fixed the parsing error although it is hard to prove without actually running the request.

I feel like ShopCommand needs to be refactored.  I'd have some kind of input validation and abort if there were problems.  I'm not sure whether that is Better or just Different, but for the moment I'm not tackling it.

jaadams5:
> I would rather we tested based on what the state of our shop was after running certain commands, rather than the logging output which could easily change. But mocking the HTTP traffic would be very complex and this is a good bit of testing to have. Thank you!

Agreed.  But at under 20% coverage every little bit helps.  And I did discover and fix a bug, although in the absence of mocking the request the debugger was required to understand that.  But this does make me want to look at ShopManager for testing :-)

PR #495, opened 2022-01-20
tests: add tests for plumber keywords, add help

midgleyc:
Further maximizer tests, add some help about previously undocumented keywords.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -67,7 +67,7 @@ public class Evaluator {
   private boolean retroCapeNeeded = false;
   private boolean backupCameraNeeded = false;
 
-  /** Whether or not we should consider or not consider certain slots when maximizing */
+  /** if slots[i] >= 0 then equipment of type i can be considered for maximization */
   private final int[] slots = new int[EquipmentManager.ALL_SLOTS];
 
   private String weaponType = null;
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -69,6 +69,7 @@ public class Evaluator {
 
   /** Whether or not we should consider or not consider certain slots when maximizing */
   private final int[] slots = new int[EquipmentManager.ALL_SLOTS];
+
   private String weaponType = null;
   private int hands = 0;
   int melee = 0; // +/-2 or higher: require, +/-1: disallow other type
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -67,6 +67,7 @@ public class Evaluator {
   private boolean retroCapeNeeded = false;
   private boolean backupCameraNeeded = false;
 
+  /** Whether or not we should consider or not consider certain slots when maximizing */
   private final int[] slots = new int[EquipmentManager.ALL_SLOTS];
   private String weaponType = null;
   private int hands = 0;
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
@@ -110,18 +110,20 @@ public class MaximizerTest {
   @Test
   public void clownosityTriesClownEquipment() {
     canUse("clown wig");
-    assertFalse(maximize("clownosity"));
+    assertFalse(maximize("clownosity -tie"));
     // still provides equipment
     recommendedSlotIs(EquipmentManager.HAT, "clown wig");
+    assertEquals(50, modFor("Clowniness"), 0.01);
   }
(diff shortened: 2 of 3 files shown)


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -67,7 +67,9 @@
   private boolean retroCapeNeeded = false;
   private boolean backupCameraNeeded = false;
 
+  /** Whether or not we should consider or not consider certain slots when maximizing */
```
jaadams5, line 70: I don't understand the comment,

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -110,18 +110,20 @@ public void startingMinKeywordPassesMaximizationIfHit() {
   @Test
   public void clownosityTriesClownEquipment() {
     canUse("clown wig");
-    assertFalse(maximize("clownosity"));
+    assertFalse(maximize("clownosity -tie"));
```
jaadams5, line 113: What is different because of -tie?  Since tests are "no cost" why not add an additional test if clownosity and clownosity -t do different things?

jaadams5 (changes requested):

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
midgleyc: I realised that the test wasn't actually testing anything. The only items in inventory were clown items, and the maximizer tries to equip any items it finds according to the tiebreaker, so the test passes if you maximize "-club" or "current" or "empty" or any other valid string that doesn't actually do anything.

To test without the tiebreaker, I'd have to add items the tiebreaker considered "better" than the clown wig, then maximize clownosity and confirm the wig was preferred. I thought it was easier to just disable it.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
midgleyc: It's a note for how `slots` works, because I found it confusing when reading through the file. It's a little bit of documentation.

To give an example, if `this.slots[EquipmentManager.HAT]` is non-negative, hats are considered, and if negative, hats aren't considered.

jaadams5: How about 

if slots[i] > 0 then equipment of type i can be considered for maximization

That is much clearer to me and since you were confused enough to comment it seems reasonable to have a comment that informs both of us ;-)

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
jaadams5: The fact that you are changing an existing test causes concern.  What this says to me is that your changes broke an existing test so you change the test.   There are cases where the test was flawed and the test needs to be fixed.  But I wonder if this is just leakage and would not be necessary if your new tests just cleaned up after themselves.

midgleyc: It's not leakage, the test was flawed. It still passes without the `-tie` change. But this pass doesn't guarantee that the clownosity keyword actually works. The test existed, but failed to test what it was supposed to test.

For example, here is one of the tests as it originally was:
```
  @Test
  public void clownositySucceedsWithEnoughEquipment() {
    canUse("clown wig");
    canUse("polka-dot bow tie");
    assertTrue(maximize("clownosity"));
    recommendedSlotIs(EquipmentManager.HAT, "clown wig");
    recommendedSlotIs(EquipmentManager.ACCESSORY1, "polka-dot bow tie");
  }
```

however this "version" of the test also passes:
```
  @Test
  public void clownositySucceedsWithEnoughEquipment() {
    canUse("clown wig");
    canUse("polka-dot bow tie");
    assertTrue(maximize("current"));
    recommendedSlotIs(EquipmentManager.HAT, "clown wig");
    recommendedSlotIs(EquipmentManager.ACCESSORY1, "polka-dot bow tie");
  }
```
(I have swapped out `maximize("clownosity")` for `maximize("current")`). I intended to test that the clownosity keyword favoured clowny equipment. However, what I actually tested was that maximizing favours some equipment over no equipment.

Happy to revert the change if it would make you happier :)

gausie: Honestly this explanation makes sense to me, @jaadams5. I vote to keep the `-tie`.

jaadams5: Nope (to reverting).  If the test under question was flawed then the change isn't hurting.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
midgleyc: Clarified to:

> /** if slots[i] >= 0 then equipment of type i can be considered for maximization */

jaadams5 (approved):

PR #496, opened 2022-01-20
"lint" removal from Parser.java

jaadams5:
Ran the IntelliJ static analyzer in hopes that reviewing Parser in the future would be easier.  There are several things that can be removed - notably empty conditionals - but I tried to err on the side of being a work in progress.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/Parser.java
+++ b/src/net/sourceforge/kolmafia/textui/Parser.java
@@ -379,9 +379,7 @@ public class Parser {
     Parser parser = this.makeChild(scriptFile);
     Scope result = parser.parseFile(scope);
 
-    for (AshDiagnostic diagnostic : parser.diagnostics) {
-      this.diagnostics.add(diagnostic);
-    }
+    this.diagnostics.addAll(parser.diagnostics);
 
     if (parser.mainMethod
         != null) { // Make imported script's main() available under a different name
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -4934,7 +4929,7 @@ private void readToken() throws InterruptedException {
 
   /**
    * If we have an unread token saved in {@link #currentToken}, null the field, and delete it from
-   * its {@link Line#tokens}, effectively forgetting that we saw it.
+   * its {@link Line}, effectively forgetting that we saw it.
```
gausie, line 4932: This should probably stay. If we can't get it to ignore this type of error, I might consider removing the `@link` instead

jaadams5: It generates a JavaDoc ERROR (!!!!!) that says something about protection and accessibility.  I couldn't figure it out in 5 minutes since JavaDoc is not a major part of my skill set.  I'd rather fix it than hide it.  Maybe @fredg1 (as author) has a suggestion?

gausie: Semantically it doesn't make sense just saying "delete it from its Line" - its specifically deleting it from the line's list of tokens. I suppose we could say

```suggestion
its {@link Line#getTokensIterator()}, effectively forgetting that we saw it.
```

or

```suggestion
the tokens list on its {@link Line}, effectively forgetting that we saw it.
```

jaadams5:
Javadoc fixed but using GitHub to make the change bypassed spotless.

PR #501, opened 2022-01-21
Add Preference to Suppress the Dialog for items with malignant status effects

BadHorseMonkey:
Creates a new boolean preference "suppressNegativeStatusPopup".  If true, the Description frame that shows the status results in the SwingX GUI is not displayed.  If not present or false, the dialog is displayed as it is now.


src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -278,6 +278,10 @@ public GeneralOptionsPanel() {
         {"saveSettingsOnSet", "Save options to disk whenever they change"},
         {},
         {"removeMalignantEffects", "Auto-remove malignant status effects"},
+        {
+          "suppressNegativeStatusPopup",
+          "Suppress popup dialog for items which cause harmful effects"
+        },
```
BadHorseMonkey, line 284: I swear this was normal before I spotless-ified it...

gausie (approved):

PR #503, opened 2022-01-21
Handle wiki links with question marks

gausie:
Fixes https://kolmafia.us/threads/drippy-plum-doesnt-correctly-link-to-the-wiki.27146/


midgleyc (approved):
Initially I thought of recommending an HTML escape library instead, but most of these are specific to what the wiki does (e.g. dropping HTML tags and # signs).

Looks good to me. Could do with a comment, or maybe even a test for drippy plum(?).

jaadams5 (approved):

jaadams5:
WikiUtilities has no test coverage at the moment so I'm ok with adding tests later.

PR #419, opened 2021-12-26
Language Server - prototype

fredg1:
Not a pull request. Just a proof of concept/prototype display.

PR #513, opened 2022-01-21
Language Server - FilesMonitor

fredg1:
Early implementation of FilesMonitor, where only the saved version of the scripts are looked at
(next version will allow the client to send unsaved versions and to ask the server to refresh)


src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
```diff
@@ -0,0 +1,103 @@
...
+        script = new Script(this.parent, file);
+        this.parent.scripts.put(file, script);
+      }
+
+      return script;
+    }
+  }
+
+  /** Returns the currently existing {@link Script.Handler handler(s)} for the given file. */
+  List<Script.Handler> findHandlers(final File file) {
```
midgleyc, line 35: `public`, given `findOrMakeHandlers` is public?

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    return handlers;
+  }
+
+  /**
+   * Checks every sub-directory under our authority as a KoLmafia root folder, and ensures that
+   * every {@code .ash} file in them has a {@link Script.Handler handler}.
+   */
+  void scan() {
```
midgleyc, line 76: `public`, please.

src/net/sourceforge/kolmafia/textui/langserver/Script.java
```diff
@@ -0,0 +1,193 @@
...
+  Handler makeHandler() {
+    this.handler = new Handler();
+
+    this.parent.executor.execute(
+        () -> {
+          this.handler.parseFile();
+        });
+
+    return this.handler;
+  }
```
midgleyc, line 46: Unless you have a good reason to, I'd prefer to avoid the default visibility modifier in favour of one of public, protected and private.

Good reasons I'm aware of include:
* wanting a field / method to be accessible in a test (accompanied with a comment saying it should be private)
* splitting a series of former nested classes into their own namespace, and not wanting them to be generally public

src/net/sourceforge/kolmafia/textui/langserver/Script.java
```diff
@@ -0,0 +1,193 @@
...
+
+      this.waitForParsing();
+
+      if (Script.this.handler != this || this.parser == null || Thread.interrupted()) {
+        // We've been kicked out
+        return;
+      }
+
+      final String previousThreadName = Thread.currentThread().getName();
+      // Will technically send the diagnostics of every file it imports, but
```
midgleyc, line 115: What is "it"? Does this thread send the diagnostics of the assigned file (`Script.this.file`), plus the diagnostics of every file that file imports?

src/net/sourceforge/kolmafia/textui/langserver/AshLanguageServer.java
```diff
@@ -61,6 +66,9 @@ public static AshLanguageServer launch(final InputStream in, final OutputStream
       new StateCheckWrappers.AshWorkspaceService(this);
 
   public final ExecutorService executor = Executors.newCachedThreadPool();
+  public final FilesMonitor monitor = new FilesMonitor(this);
+
+  public final Map<File, Script> scripts = Collections.synchronizedMap(new Hashtable<>(20));
```
midgleyc, line 71: Why Hashtable? Might be the right move, but when HashMap is the standard anything else will look odd.

src/net/sourceforge/kolmafia/textui/langserver/Script.java
```diff
@@ -0,0 +1,193 @@
...
+      // don't change the name accordingly; it would change too fast.
+      Thread.currentThread().setName(Script.this.file.getName() + " - Diagnostics");
+
+      synchronized (this.parser.getImports()) {
```
midgleyc, line 119: synchronizing on a mutable map? Could something go wrong here?

src/net/sourceforge/kolmafia/textui/langserver/Script.java
midgleyc: This also applies to everywhere else in this commit you've used default visibility modifiers. For example, in `AshLanguageServerTest` I'd expect most things to go to `protected`, because the class is designed to be inherited.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
```diff
@@ -0,0 +1,66 @@
...
+        "This test is currently based on the existence of test_function_coercion_override_builtin.ash and test_function_coercion.ash. If removing them, please update this test.");
+
+    Assertions.assertDoesNotThrow(
+        (ThrowingSupplier<InitializeResult>) proxyServer.initialize(new InitializeParams())::get,
+        "Initialization failed");
+
+    // Once initialized, the server will automatically scan the root for .ash scripts. Just give it
+    // some time.
+
+    // Note: time may need to be increased if big scripts are added to test/root
```
midgleyc, line 35: Is there a plan for, and is it possible, to have something like a `waitForScanComplete` method? Having a test break because more scripts were tested is going to be painful. Similarly, waiting longer than necessary is also painful. Is it possible to only scan the folder containing the two scripts that are desired to be tested?

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
```diff
@@ -0,0 +1,103 @@
...
+  /** Returns the currently existing {@link Script.Handler handler(s)} for the given file. */
+  List<Script.Handler> findHandlers(final File file) {
+    final List<Script.Handler> handlers = new LinkedList<>();
+
+    synchronized (this.parent.scripts) {
+      for (final Script script : this.parent.scripts.values()) {
+        if (script.handler != null
+            && script.handler.parser != null
+            && script.handler.parser.getImports().containsKey(file)) {
+          handlers.add(script.handler);
```
midgleyc, line 43: IntelliJ tells me this line is uncovered. If true, could you add a test for successfully finding a handler?

midgleyc (changes requested):
First, thanks for the well-documented code and tests! Excellent coverage.

Second, could you rebase on `main`? [A commit](https://github.com/kolmafia/kolmafia/commit/63ddea279c58173f2047590496e0434489463460) has gone in to remove some lint. GitHub says it's okay to merge, but worth doing anyway I think.

Third, IntelliJ prompts me with some nits on your code (e.g. using method references instead of lambdas). I think these would improve things slightly, but I've not called them out: if you're not using IntelliJ yourself, these might be annoying to find.

Fourth, many many comments on various lines. Some apply to multiple parts of this PR, mostly the default visibility comments.

Fifth, are you also testing this manually (i.e. the language server overall)? Does it work?

Thanks Fred!

src/net/sourceforge/kolmafia/textui/langserver/Script.java
Veracity0: I completely agree with wanting explicit private, protected, or public for methods and fields.

fredg1: these are all on purpose, though.

These are purposefully package-only visibility.
There's no point in them being protected since this is not a class meant to be extended.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: Script.Handler has a reason to look for the existence of a Handler for a given File (to see if they have been replaced), but not the rest. They shouldn't have access to this method.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: Only AshLanguageServer and Script are meant to have the authority to trigger a scan.

src/net/sourceforge/kolmafia/textui/langserver/Script.java
Veracity0: You're not listening. Use explicit keywords, rather than omitting them.

fredg1:
> Fifth, are you also testing this manually (i.e. the language server overall)? Does it work?

I do/did! And it does! [See the prototype](https://github.com/kolmafia/kolmafia/pull/419)

src/net/sourceforge/kolmafia/textui/langserver/Script.java
fredg1: It does. That's the point of Handler. (i.e. to be in charge of a File, plus every script it imports)

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: (FilesMonitor was originally in the same file as Script, but was separated due to it getting somewhat bulky. Otherwise the method would have been private)

src/net/sourceforge/kolmafia/textui/langserver/Script.java
fredg1: Given that we're iterating on that map's `entrySet` on the next line, yes, we want to prevent someone from accessing and modifying the map while we're doing this (Not sure "why" they would alter `parser.imports`, but it's just good practice)

src/net/sourceforge/kolmafia/textui/langserver/Script.java
fredg1: :-(

src/net/sourceforge/kolmafia/textui/langserver/AshLanguageServer.java
fredg1: HashMap permits null values, which we don't want here. Otherwise it could have been a HashMap (given that we're doing the synchronization ourselves)

src/net/sourceforge/kolmafia/textui/langserver/Script.java
midgleyc: I appreciate that was probably more of a response to me, giving your reasons, but I think Veracity would prefer we forbid the default modifiers entirely in this case. I think, stylistically, I agree: `protected` should serve the same purpose (of forbidding use outside the package) if the class is never extended, and reads more easily to me (I, being somewhat old-school, don't see many things use the package local modifier). It also means that if someone wanted to extend the class for some reason, everything would work as hoped for.

src/net/sourceforge/kolmafia/textui/langserver/AshLanguageServer.java
midgleyc: Could you please add a comment to the code with that reasoning?

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: Mhh... this was mostly included "just in case". It will be formally used in the next step, where the server will start being able to update scripts' content.

midgleyc: I am happy for this method to be a stub for now with something like `return new LinkedList<>()`, and a comment saying the real implementation is coming in the next step, if it will be easier to write a test in the next step.

But if it is coming in this step, I would rather a test be written here, so we don't forget about it.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
fredg1: > Is it possible to only scan the folder containing the two scripts that are desired to be tested?

No.

> Is there a plan for, and is it possible, to have something like a `waitForScanComplete` method?

No... and no.
While there is a method that I could prematurely expose that would effectively allow us to wait for the end of the parsing of a Handler, it, well, only works for a single Handler at a time.
In order to check every script, we would need to iterate on `AshLanguageServer.scripts`, which might get modified during the iteration.
If we synchronize on it to prevent the concurrent modification, we effectively freeze the server (and effectively deadlock).

src/net/sourceforge/kolmafia/textui/langserver/Script.java
midgleyc: Sorry for the lack of precision here: I meant "is it possible that `this.parser.getImports()` could return two different things for two different threads?".

I've checked, and it seems that `parser.imports` is only assigned in the constructor. So I think this is fine.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
midgleyc: I'm less concerned with a fully general solution than I am with one that will work for `FilesMonitorTest`. To reframe: I don't want this test to break in the future for reasons other than the LSP breaking, and I don't want the fix to be to repeatedly increase the time it waits for the scanner to finish (which will, in effect, slow down the build for everything else).

Where does the "root folder" get set? Can it be changed for just this test?

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: The thing is: it *is* currently possible to reach this line, but it is only possible as some sort of race condition.

Given a very big folder containing a *lot* of scripts.
`FilesMonitor.scan()` will iterate over its files and make sure there's a Handler for each of them.

If one of the very *first* script was to import one of the very *last* script, it would be possible for a Handler to exist for that script before `FilesMonitor.scan()` gets to it.

The thing is, due to the speed at which `FilesMonitor.scan()` operates, there would need to be a ***LOT*** of scripts for this to happen, so testing it would require cluttering `test/root`, a lot. Probably hundreds of files.

src/net/sourceforge/kolmafia/textui/langserver/Script.java
fredg1: Just to clarify: I thought what you were originally saying was "I don't get why we need to synchronize here, what's the use?"

If I understand correctly, what you meant was: "I see we synchronize on *this*. Are we certain that *this* will always be what we think?"
Am I correct?

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
fredg1: > Where does the "root folder" get set? Can it be changed for just this test?

See "path": https://kolmafia.us/threads/ash-language-server-features.26098/post-166081

The root is effectively set in `build.gradle`, line 158
```gradle
test {
	useJUnitPlatform()
	systemProperty 'line.separator', '\n'
	systemProperty 'junit.jupiter.extensions.autodetection.enabled', true
	systemProperty 'useCWDasROOT', true
	workingDir 'test/root' // <<<<<<<<<<<<

	testLogging.showStandardStreams = true

	reports {
		html.required = true
		junitXml.required = true
	}
}
```

And the fact that it can't be modified has been brought up as an issue, is not currently changing.

src/net/sourceforge/kolmafia/textui/langserver/Script.java
fredg1: > I agree: `protected` should serve the same purpose (of forbidding use outside the package) if the class is never extended, and reads more easily to me. It also means that if someone wanted to extend the class for some reason, everything would work as hoped for.

The only issue I have with this is that seeing "protected" *implies* that the class *can* be extended. So if the class can't or is not meant to, it creates some confusion IMO.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
midgleyc: Here's a POC that changes the scanned folders for that test:

https://github.com/midgleyc/kolmafia/commit/7a6999fcf0b9a031e43b8595253e533e9a7aba17

Feel free to pull that in as-is, make changes if you like, or propose another way of doing it.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: Will you be able to add a test for this in one of the upcoming steps, or is it untestable in totality? If it will be testable later, add a `// TODO: test (step X)` and that'll be fine by me.

Is it testable as a non-integration test, if you mock out enough things?

src/net/sourceforge/kolmafia/textui/langserver/Script.java
midgleyc: Pretty much. More precisely, it's 

"I see we're synchronising on the output of a function. Are we sure the output of the function won't be different objects in different threads?"

src/net/sourceforge/kolmafia/textui/langserver/Script.java
midgleyc: I acknowledge that that could be an issue, but I'd still like you to avoid using the default modifiers, as Veracity requested.

Veracity0: If the intent is that the class cannot be extended, why is it not final?

fredg1: The point was that even if you were to try, you wouldn't be able to replace the one the server is using, making it effectively pointless to extend, but I guess also marking it as final can't hurt.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
fredg1: It would probably be better to implement a "filter" of some sort, rather than this (I can take care of it)

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: This line will end up getting tested in *the next* step. Right after this PR is accepted.

test/net/sourceforge/kolmafia/textui/langserver/FilesMonitorTest.java
fredg1: (done)

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: Okay, great. Please add a comment along the lines of:
```
// TODO: currently untested, will be tested after server can update scripts' content
```
and then remove the comment in the upcoming PR, after the line is tested.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: re: the below discussion, can you make this non-package local?

fredg1: but doesn't specifically fall in the "good reason to have something be package local" that you listed? (namely, the second one)

> Good reasons I'm aware of include:
> 
>     wanting a field / method to be accessible in a test (accompanied with a comment saying it should be private)
>     splitting a series of former nested classes into their own namespace, and not wanting them to be generally public

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: ...could I please not?
As you can see below, there's currently some jankiness going on with tests (unrelated to this PR) making them fail inconsistently, and now that I finally got them to work, I'd prefer to not jeopardize them again just for a comment :-(

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: It does, but later on Veracity posted:
> I completely agree with wanting explicit private, protected, or public for methods and fields.

We're both maintainers. If I say "Can you do X, unless Y?" and Veracity says "Please do X", the result is that you should do X, even if Y.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: If a comment breaks the tests, they are broken, and should be fixed.

If the tests don't consistently pass, they are broken, and should be fixed.

If this PR gets merged with tests that fail randomly, they will break the build, and we will be unable to release new versions or see whether anything has broken. Please fix the tests.

fredg1: The failures are unrelated, but I can't tell what's wrong on my end since the failure only seems to happen on macOS, and test results aren't available on github

jaadams5: TypedValueTest > testScriptValidity(ScriptData) > net.sourceforge.kolmafia.textui.parsetree.TypedValueTest.testScriptValidity(ScriptData)[12] FAILED
    org.opentest4j.AssertionFailedError at TypedValueTest.java:121

CustomScriptTest > testScript(String) > net.sourceforge.kolmafia.CustomScriptTest.testScript(String)[3] FAILED
    org.opentest4j.AssertionFailedError at CustomScriptTest.java:67

Above from one of the recent failures.

If you want your commits to continue to be reviewed and submitted a better attitude towards tests might help.

fredg1: at the part exposing how this is actually related to this PR is..?

Both these tests are related to $class[] and $monster[] datatypes, which were updated in r26160 and r26161.

fredg1: I agree that it was presumptuous to phrase it as if it was "definitely" unrelated, but surely you can agree that these errors don't "seem" to come from this PR, right?

jaadams5: Your assertion that the failures are unrelated to your changes may be correct but it may reflect a lack of imagination on your part.  Your unwillingness to try and resolve the problem whether or not your code was the direct cause is concerning.

fredg1: No, it was... it was due to the fact that it only seemed to happen on macOS, so I couldn't dig in any further by myself...

BadHorseMonkey helped me with this, so I was able to dig a little further in, and I even think I found a fix, so calling this "unwillingness to try and resolve the problem" kinda hurts... :<

midgleyc (approved):

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
midgleyc: Thanks for finding and submitting a fix, fred.

To dig into your earlier statements, while GitHub only allows us to require that the latest attempt at running tests works, in fact we want the tests to pass consistently: so if you ran them three times, say, they would all pass every time. You statement "I finally got them to work" above was not true under this paradigm: you got them to pass on the server once, but not by fixing the problem, instead just running them over and over until they passed once. This isn't great.

I wasn't convinced the failures were unrelated because:
* the failures were all in the parser
* this test modifies something related to the parser
* the tests had passed consistently before this PR

I felt it was more likely this PR had broken something accidentally than a sporadic failure, targeting only parser tests, came out of nowhere during this PR that affects the parser.

I am going to run the tests three times, and if they pass every time, I will merge.

PR #515, opened 2022-01-22
Contract Checker

jaadams5:
I wanted to validate another compareTo contract.  I copied existing code and tried to parameterize it.  There are a couple of "how to code" questions in the code but there are some general questions we can converse about.

This is more of a data validation exercise than a test.  As implemented if there are contract violations something needs to traverse a list and tell a user.  We also know that 10,000+ plus concoctions is too much for tests that get run frequently.  But I like the idea of having this capability available.

There is a checkconcoctions command that could be modified to preserve the parameterized version.  That will force me to make some decisions about/provisions for reporting.  If I do that, do I make a checkadv command?  I can remove checkconcoctions from the codebase and restore it to tests and then disable it.  Other ideas?

I almost certainly will not submit this PR as is.


test/internal/helpers/CompareContractValidator.java
```diff
@@ -0,0 +1,110 @@
...
+    FUNCTION_EQUALITY,
+    OPERATOR_EQUALITY,
+    TRANSITIVITY
+  }
+
+  // Code intended to verify that compareTo() meets its contract.
+  // public static <T extends Comparable<T>> List<T> checkForContractViolations(List<T> inputList) {
+  public static <T> List<Violator> checkForContractViolations(List<T> inputList) {
+
+    List<Violator> violators = new ArrayList<>();
```
jaadams5, line 35: This list and the associated structures were the first pass at capturing violations in the context of a unit test.  Not sure it should persist in this form or not.

test/internal/helpers/CompareContractValidator.java
```diff
@@ -0,0 +1,110 @@
...
+    SYMMETRY,
+    FUNCTION_EQUALITY,
+    OPERATOR_EQUALITY,
+    TRANSITIVITY
+  }
+
+  // Code intended to verify that compareTo() meets its contract.
+  // public static <T extends Comparable<T>> List<T> checkForContractViolations(List<T> inputList) {
+  public static <T> List<Violator> checkForContractViolations(List<T> inputList) {
+
```
jaadams5, line 34: Types are underutilized by me - one of the reasons for trying this.  But I could not figure out how to declare things so that the only allowed types implemented Comparable.

test/internal/helpers/CompareContractValidator.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+    if (!canCompare) {
+      Violator faux = new Violator("bad", "input", "fail");
+      violators.add(faux);
+      return violators;
+    }
+
+    for (i = 0; i < maxIndex; ++i) {
+      @SuppressWarnings("unchecked")
+      Comparable<T> objectThatCanCompare = (Comparable<T>) ids.get(i);
```
jaadams5, line 64: The above check makes this a safe cast but...

PR #520, opened 2022-01-24
If the relay server is opened on start up then wait until SVN updates have finished.

jaadams5:
https://kolmafia.us/threads/delay-automatic-opening-of-relay-browser-until-after-svn-update-on-login-completes.27172/

Put a flag around SVNUpdate running at login and then wait before launching the relay browser.

If this "kinda works" the wait times could be a preference but let's confirm that this works elsewhere and is actually useful first.


gausie:
Something is wrong with the testing here - its failing catastrophically but we're not getting a red ❎  for some reason. It looks like something to do with the language server is crapping out.

midgleyc:
That's not a problem: the running of `StateCheckWrappersTest` winds up printing a bunch of errors to the console, but that's expected.

No tests are failing.

jaadams5:
> Something is wrong with the testing here - its failing catastrophically but we're not getting a red ❎ for some reason. It looks like something to do with the language server is crapping out.

986 tests run, 6 skipped, 0 failed.

Are you perhaps getting mislead by a whole bunch of stack traces that are generated by some language server testing but are actually expected?  The output looks scary but the tests expect the exception.  The problem is that some of the third party code is dumping the trace anyway.

https://kolmafia.us/threads/r26142-language-server-outer-shell-statecheckwrappers-by-fredg1.27160/#post-166546

There was some talk of getting the 3rd party stuff to log to /dev/null but nothing happened yet and I couldn't find it quickly.

jaadams5:
Maybe https://github.com/kolmafia/kolmafia/pull/482#issuecomment-1019403014 is the comment I was thinking of?

gausie:
Oh. I'm confused why the coverage didn't work then

jaadams5:
> Oh. I'm confused why the coverage didn't work then

[2022-01-24T20:52:54.971Z] ['error'] There was an error running the uploader: Error while cleaning paths. No paths matched existing files!

Does that help?  I found it in some recent checks for other PRs as well.

src/net/sourceforge/kolmafia/webui/RelayLoader.java
```diff
@@ -36,11 +37,27 @@ protected RelayLoader(final String location, final boolean isRelayLocation) {
     }
   }
 
+  private void waitForSVNUpdateToFinish() {
+    // This first loop is to wait for LoginManager to start processing updates, if it is going to
+    PauseObject pauser = new PauseObject();
+    for (int i = 0; i < 50 && !RelayServer.isRunning(); ++i) {
+      pauser.pause(200);
```
gausie, line 44: There are a bunch of (repeated) magic numbers here, it might be good to pull them out to named static props on the class or something

jaadams5: Moved them to named constants.  Do I want to make numTries a preference or is this too niche to bother?

gausie:
I think something about this implementation just seems overcomplicated to me - I'd expect this to be possible without multiple wait loops. But if it works, I'm happy to merge it.

jaadams5:
> I think something about this implementation just seems overcomplicated to me - I'd expect this to be possible without multiple wait loops. But if it works, I'm happy to merge it.

Definitely a judgement call since it is kludgy and in response to perhaps one user.  I tried moving various calls earlier or later and doing so broke something or had a side effect that I didn't like.  The precedent for the unconditional loop came from elsewhere in the code but the waiting is basically because I didn't really figure out the threading well enough to use thread communication tools.  Without the initial wait the launch code checked the flag before it had been set.  

I am also uncomfortable just because the timing works for me on a mid-range Windows laptop but will that timing work elsewhere?

I'm fine with your call - approve it or close it.  If closed then maybe we look at the core symptom - something in relay/ gets updated but a running relay browser doesn't respond to the change.

Thanks.

gausie:
Maybe we could do something like -> if Relay browser wants to open but SVN
update is running, you can register a callback on an "on completion"
listener that is poked when SVN is over. Or something.

jaadams5:
OK.  I think the thing to do is to close this because no one is really happy with the implementation.   I'll look at listeners since that should eliminate all of the timing constants.

PR #521, opened 2022-01-24
Language Server - FilesMonitor - script content updating

fredg1:


fredg1:
The tests... didn't run?

gausie:
It's because this is a PR from a remote branch. Need to make a change to the workflow to go on pull_request too, and limit push to just the main branch (probably)

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
```diff
@@ -18,6 +20,44 @@ protected FilesMonitor(final AshLanguageServer parent) {
...
+
+      if (handlers.size() > 0
+          && ((text != null && script.version >= version)
+              || (script.text == null && text == null))) {
+        // We already have a working handler using an
+        // up-to-date version of that file.
+        return;
+      }
+
+      if (script.text == null || script.version < version || script.text != null && text == null) {
```
midgleyc, line 42: `script.text != null` is unnecessary because it would be caught by the first check `script.text == null`.

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
```diff
@@ -18,6 +20,44 @@ protected FilesMonitor(final AshLanguageServer parent) {
...
+        script.text = text;
+        script.version = version;
+      }
+
+      if (handlers.size() == 0) {
+        // make a new handler
+        handlers.add(script.makeHandler());
```
midgleyc, line 49: If this line is straightforward to write a test for, would you mind?

Additionally, does creating a new handler do the equivalent of "refreshParsing"?

src/net/sourceforge/kolmafia/textui/langserver/textdocumentservice/AshTextDocumentService.java
```diff
@@ -30,14 +37,55 @@ public final void setCapabilities(final ServerCapabilities capabilities) {
...
+          if (changes.size() == 0) {
+            // Nothing to see here
+            return;
+          }
+
+          File file = FilesMonitor.UriToFile(document.getUri());
+
+          // We don't support incremental changes, so we expect the client
+          // to put the whole file's content in a single TextDocumentContentChangeEvent
+          this.parent.monitor.updateFile(file, changes.get(0).getText(), document.getVersion());
```
midgleyc, line 67: Could you raise an error if multiple changes are raised, as it's not supported? As you only handle the first, and assume it's the whole text.

src/net/sourceforge/kolmafia/textui/langserver/textdocumentservice/AshTextDocumentService.java
```diff
@@ -30,14 +37,55 @@ public final void setCapabilities(final ServerCapabilities capabilities) {
...
+  }
 
   @Override
-  public void didSave(DidSaveTextDocumentParams params) {}
+  public void didSave(DidSaveTextDocumentParams params) {
+    // What we had in memory should now match the content of the file.
+    // Should we just ignore these notifications?
+    // We could set the Script's text and version back to null and -1, respectively. If so, either
+    // A- the two are indeed identical, and this is just a waste of processing time, or
+    // B- the two are different, which should not happen; the issue to solve would be elsewhere...
```
midgleyc, line 89: This sounds like a design question which should be sorted as part of this PR.

I think ignoring the notifications is fine.

midgleyc (changes requested):
Thanks fred!

gausie:
The `Files Changed` tab now shows annotations for lines missing coverage, so you can check against that for getting 100%

src/net/sourceforge/kolmafia/textui/langserver/textdocumentservice/AshTextDocumentService.java
fredg1: Since this is a notification, the client won't get a response. As such, raising an error is pointless.

Besides, more than a single change simply means "undefined behaviour". It's not stated that it needs to result in an error.
We said that we supported `TextDocumentSyncKind.Full`. This means we just look at the first element.
They didn't put any? IndexOutOfBoundsException and we silently move on.
They put more? We don't care.

fredg1:
> The `Files Changed` tab now shows annotations for lines missing coverage, so you can check against that for getting 100%

fancy!

src/net/sourceforge/kolmafia/textui/langserver/FilesMonitor.java
fredg1: Not quite.
First, both submit a different boolean to `Handler.parseFile(initialParsing)`.
Second, `makeHandler` also additionally causes the Handler to get registered for future use.

midgleyc (approved):

PR #526, opened 2022-01-25
Disable logging during testing StateCheckWrappers

gausie:
This should resolve the issue noted in #482 whereby the test log gets flooded with exception logs that do not indicate test failure


jaadams5 (approved):
Nice.  I assume it works :-)  It never occurred to me to go down that path.  Thanks.

gausie:
@fredg1 this has uncovered an actual, albeit minor, synchronisation issue:

https://github.com/kolmafia/kolmafia/runs/4940055759?check_suite_focus=true#step:6:74

If a handler has been `close()`ed before it can be executed, `parseFile()` is called on `null`. Can you fix that please?

PR #527, opened 2022-01-25
fix: rare NPE in getCount

midgleyc:
Don't know what causes this, probably a threading issue where `KoLConstants.inventory` is changing out from under `getCount`.


gausie (dismissed):

gausie:
@midgleyc is there a way to add a regression test here?

midgleyc:
I don't think so. It's in threading hell.

Here's what I think is happening.

First
```
int index = list.indexOf(this);
```
runs. `this` cannot be null. In the error case, `this` exists in the list, and we get an index.

Then another thread modifies the inventory list. The list is a `LockableListModel`, so if the index is greater than the new size of the list, we get null. This is what I think happens. Then it NPEs.

A regression test would have to do this. I could fake it by passing a mocked list that returns something for `indexOf` and always `null` for `get`, but that's testing implementation details, and I don't think that's very useful.

PR #533, opened 2022-01-25
Exclude lib from coverage

gausie:


heeheehee-kolmafia (dismissed):
missing newline? otherwise looks reasonable

PR #535, opened 2022-01-26
More language level changes

gausie:
This clears up all the language warnings in IntelliJ <= Java 7. I had some fun composing some streams too.

### DIFF
--- a/src/net/sourceforge/kolmafia/AreaCombatData.java
+++ b/src/net/sourceforge/kolmafia/AreaCombatData.java
@@ -1305,9 +1305,7 @@ public class AreaCombatData {
         break;
       case "An Overgrown Shrine (Northwest)":
         // Assume lianas are dealt with once Apartment opened. Player may leave without doing so,
-        // but
-        // that's
-        // abit niche for me to care!
+        // but that's a bit niche for me to care!
         if (monster.equals("dense liana")
             && Preferences.getInteger("hiddenApartmentProgress") > 0) {
           return 0;
@@ -1315,25 +1313,21 @@ public class AreaCombatData {
         break;
       case "An Overgrown Shrine (Northeast)":
         // Assume lianas are dealt with once Office opened. Player may leave without doing so, but
-        // that's
-        // abit niche for me to care!
+        // that's a bit niche for me to care!
         if (monster.equals("dense liana") && Preferences.getInteger("hiddenOfficeProgress") > 0) {
           return 0;
         }
         break;
       case "An Overgrown Shrine (Southwest)":
         // Assume lianas are dealt with once Hospital opened. Player may leave without doing so, but
-        // that's
-        // abit niche for me to care!
+        // that's a bit niche for me to care!
         if (monster.equals("dense liana") && Preferences.getInteger("hiddenHospitalProgress") > 0) {
           return 0;
         }
         break;
       case "An Overgrown Shrine (Southeast)":
         // Assume lianas are dealt with once Bowling Alley opened. Player may leave without doing
-        // so,
-        // but that's
-        // abit niche for me to care!
+        // so, but that's a bit niche for me to care!
         if (monster.equals("dense liana")
             && Preferences.getInteger("hiddenBowlingAlleyProgress") > 0) {
           return 0;
@@ -1341,8 +1335,7 @@ public class AreaCombatData {
         break;
       case "A Massive Ziggurat":
         // Assume lianas dealt with after 3 turns, won't always be right, but this is a bit niche
-        // for
-        // special tracking
+        // for special tracking
         int zoneTurns = AdventureSpentDatabase.getTurns(zone);
         if (monster.equals("dense liana")
             && (zoneTurns >= 3 || QuestDatabase.isQuestFinished(Quest.WORSHIP))) {
--- a/src/net/sourceforge/kolmafia/AreaCombatData.java
+++ b/src/net/sourceforge/kolmafia/AreaCombatData.java
@@ -1178,454 +1178,521 @@ public class AreaCombatData {
 
   private static int adjustConditionalWeighting(String zone, String monster, int weighting) {
     // Bossbat can appear on 4th fight, and will always appear on the 8th fight
-    if (zone.equals("The Boss Bat's Lair")) {
-      int bossTurns = AdventureSpentDatabase.getTurns(zone);
-      if (monster.equals("Boss Bat")) {
-        return bossTurns > 3 && !QuestDatabase.isQuestLaterThan(Quest.BAT, "step3") ? 1 : 0;
-      } else {
-        return bossTurns > 7 || QuestDatabase.isQuestLaterThan(Quest.BAT, "step3") ? -4 : 1;
-      }
-    } else if (zone.equals("The Hidden Park")) {
-      if (monster.equals("pygmy janitor")
-          && Preferences.getInteger("relocatePygmyJanitor") != KoLCharacter.getAscensions()) {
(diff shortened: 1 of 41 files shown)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -4326,62 +4326,39 @@ public static final void setArenaWins(final int wins) {
...
+
+  private static final FamiliarData findFamiliar(final Predicate<FamiliarData> familiarFilter) {
+    // Quick check against NO_FAMILIAR
+    if (familiarFilter.test(FamiliarData.NO_FAMILIAR)) return FamiliarData.NO_FAMILIAR;
 
     // Don't even look if you are an Avatar
-    if (KoLCharacter.inAxecore() || KoLCharacter.isJarlsberg() || KoLCharacter.isSneakyPete()) {
-      return null;
-    }
+    if (KoLCharacter.getPath().isAvatar) return null;
```
midgleyc, line 4347: While a good idea, this doesn't mean what you think it means ;)

To be more precise, some avatar paths, like Plumber, have familiars. I'd recommend splitting out another boolean in the `Path` constructor.

src/net/sourceforge/kolmafia/KoLDesktop.java
```diff
@@ -250,14 +251,9 @@ static final void updateTitle() {
...
-    for (int i = 0; i < frames.length; ++i) {
-      if (frames[i] instanceof GenericFrame) {
-        GenericFrame frame = (GenericFrame) frames[i];
-
-        frame.setTitle(frame.getLastTitle());
-      }
-    }
+    Arrays.stream(Frame.getFrames())
+        .filter(f -> f instanceof GenericFrame)
+        .forEach(f -> f.setTitle(((GenericFrame) f).getLastTitle()));
```
midgleyc, line 256: In `Kolmafia.java`, you use `GenericFrame.class::isInstance`, etc. Better to be consistent I think.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1618,37 +1611,30 @@ public static final boolean checkRequirements(final List<AdventureResult> requir
...
-
     long actualCount = 0;
 
     // Check the items required for this quest,
     // retrieving any items which might be inside
     // of a closet somewhere.
 
-    for (int i = 0; i < requirementsArray.length; ++i) {
-      if (requirementsArray[i] == null) {
+    for (AdventureResult adventureResult : requirements) {
```
midgleyc: I think this loop removes things from the list it's iterating over. That's probably why it was converted to an array first.

Please use iterators instead.

src/net/sourceforge/kolmafia/StaticEntity.java
```diff
@@ -643,26 +639,12 @@ public static final void generateHeapDump() {
...
-      }
-    }
-
-    String[] pastUsers = new String[pastUserList.size()];
-    pastUserList.toArray(pastUsers);
-    return pastUsers;
+    return Arrays.stream(DataUtilities.listFiles(KoLConstants.SETTINGS_LOCATION))
+        .map(f -> f.getName())
+        .filter(u -> !u.startsWith("GLOBAL") && u.endsWith("_prefs.txt"))
+        .map(u -> u.substring(0, u.length() - 10))
```
midgleyc, line 645: You seem to have lost the "user != 'GLOBAL'" check.

Might be intended because surely that's covered by "starts with GLOBAL", but worth checking.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1236,32 +1238,49 @@ public static final String getDerivedModifierName(final int index) {
...
+  private static String modifierName(final Object[][] table, final int index) {
+    return modifierRow(table, index, Modifiers::modifierName);
+  }
+
+  private static String modifierName(final Object[] tableRow) {
+    return (String) tableRow[0];
+  }
+
+  private static Pattern[] modifierDescPattern(final Object[][] table, final int index) {
+    return modifierRow(table, index, Modifiers::modifierDescPattern);
```
midgleyc, line 1259: Please ensure this file is tested more thoroughly.

I know it wasn't tested before, but this is some extensive refactoring.

src/net/sourceforge/kolmafia/AscensionClass.java
```diff
@@ -47,26 +47,25 @@
...
-      if (ascensionClass.getName().toLowerCase().contains(name.toLowerCase())) {
-        return ascensionClass;
-      }
-    }
-    return null;
+  public static AscensionClass find(String name) {
+    if (name == null || name.equals("")) return null;
+
+    return Arrays.stream(values())
+        .filter(a -> a.getName().toLowerCase().contains(name.toLowerCase()))
```
midgleyc, line 62: If you bring `name.toLowerCase` out of the stream, you can avoid calculating it each time.

This might be what happens anyway, but I don't think it is.

Same for the other instances of this.

src/net/sourceforge/kolmafia/AreaCombatData.java
```diff
@@ -1178,454 +1178,521 @@ public String getZone() {
...
+        // abit niche for me to care!
+        if (monster.equals("dense liana") && Preferences.getInteger("hiddenHospitalProgress") > 0) {
+          return 0;
+        }
+        break;
+      case "An Overgrown Shrine (Southeast)":
+        // Assume lianas are dealt with once Bowling Alley opened. Player may leave without doing
+        // so,
+        // but that's
+        // abit niche for me to care!
```
midgleyc, line 1336: spotless has really knackered this file's comments.

Can you go through and combine them then reformat? Or make spotless not do this?

midgleyc (changes requested):
You're still making changes while I'm doing the review, but some comments.

src/net/sourceforge/kolmafia/CreateFrameRunnable.java
```diff
@@ -36,35 +38,22 @@ public CreateFrameRunnable(final Class<?> creationType) {
...
+            .filter(
+                c -> {
+                  var constructorParamTypes = c.getParameterTypes();
+                  if (constructorParamTypes.length != parameters.length) return false;
+                  return IntStream.range(0, parameterTypes.length)
+                      .filter(j -> parameterTypes[j] != null)
+                      .allMatch(j -> constructorParamTypes[j].isAssignableFrom(parameterTypes[j]));
+                })
+            .reduce((a, b) -> b)
+            .orElse(null);
```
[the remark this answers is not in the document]
gausie: Ah see to me this is much simpler easier to understand.

src/net/sourceforge/kolmafia/KoLCharacter.java
gausie: Ah yeah good point

src/net/sourceforge/kolmafia/KoLmafia.java
gausie: Already did this before the review!

src/net/sourceforge/kolmafia/StaticEntity.java
gausie: Yeah I did it on purpose, it was redundant

src/net/sourceforge/kolmafia/Modifiers.java
gausie: I've added loads of tests to it

midgleyc (approved):

PR #536, opened 2022-01-26
In Ronin, you can only pull one of an item

Veracity0:
When creating the Storage panel for the Item Manager, if you are in Ronin, set the "movers" radio button to "exactly one" rather than the default "multiple".


gausie (approved):
Good start!

PR #538, opened 2022-01-26
tests: speed up maximizer tests

midgleyc:
`KoLCharacter.reset(true)` is pretty slow (in part due to `KoLCharacter.currentModifiers.resetModifiers()`, which parses the entire `modifiers.txt` file again).

For maximizer tests (the ones I work on the most), do a more targeted teardown, focusing on restoring exactly what was set up to the baseline.

Also split out the regression tests to a different file.

### DIFF
--- a/test/internal/helpers/Cleanups.java
+++ b/test/internal/helpers/Cleanups.java
@@ -1,0 +1,35 @@
+package internal.helpers;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class Cleanups {
+  private final List<Runnable> cleanups = new ArrayList<>();
+
+  public Cleanups() {}
+
+  public Cleanups(Runnable r) {
+    cleanups.add(r);
+  }
+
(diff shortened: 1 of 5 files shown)


gausie:
Commenting here with my suggestion I made via DM to make the cleanup concept (which is good) instead be something around `try with resources`

jaadams5 (approved):
I'm OK with this.  I'm not sure about try with resources but I have no problem with doing so.

gausie:
Sorry to add a little bit of conflict res to Player.java, it should hardly be anything

midgleyc:
Rebased.

try-with-resources isn't as good as hoped, because I have to create a variable to store the cleanup in -- I can't just leave it anonymous.

Also `spotless` insists on putting each variable in the block on its own line, which I'm really not keen on.

I could move the variable declarations inside the block. I initially thought it was cleaner outside but I'm not sure.

jaadams5 (approved):

PR #539, opened 2022-01-26
In Ronin, you can only pull one of an item

Veracity0:
This adds a new property - _roninStoragePulls - which contains the item number of each item you have pulled today while in Ronin.

Will grey that item out in the Storage panel of Inventory manager if you've already pulled one.
Huh. I should make it not visible so you can't even select it. Stay tuned.

StorageRequest will prune out all items you've already pulled today and will not attempt to pull such. Here's what that looks like:

```
> pull 1 jar of squeeze

You've already pulled one 'jar of squeeze' today. Skipping...
No items could be removed from storage.
```

When you do succeed in pulling items, the property is updated.


jaadams5:
My experience is that it was one pull of each type per run and not per day.  Can't check at the moment but maybe someone can chime in.

Veracity0:
I guarantee that is not true.

January 24

pull: 1 jar of squeeze
[190] The Defiled Nook
drink 1 jar of squeeze

January 25

pull: 1 jar of squeeze
[417] The Hidden Temple
drink 1 jar of squeeze

midgleyc (approved):

PR #541, opened 2022-01-26
Remove improper handling for the refresh command

s-k-z:
Currently, typing `help refresh` does not work as expected. These lines of code improperly intercept the keyword refresh and alter the inputted command in undesirable ways for all other commands as well such that any command of the form `command refresh` will attempt to run the refresh command instead.

Currently:
```
> help refresh

help cannot be refreshed.

> genie refresh

genie cannot be refreshed.
```

With this change:
```
> help refresh

refresh all | [status | effects] | [gear | equip | outfit] | inv | camp | storage | [familiar | terarrium] | stickers | quests | shop - resynchronize with KoL.

> refresh genie

genie cannot be refreshed.

> genie refresh

(no output)
```

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/command/HelpCommandTest.java
+++ b/test/net/sourceforge/kolmafia/textui/command/HelpCommandTest.java
@@ -23,9 +23,6 @@ public class HelpCommandTest extends AbstractCommandTestBase {
     String output = execute("refresh");
 
     assertContinueState();
-    assertThat(
-        output,
-        containsString(
-            "refresh all | [status | effects] | [gear | equip | outfit] | inv | camp | storage | [familiar | terarrium] | stickers | quests | shop - resynchronize with KoL."));
+    assertThat(output, containsString("refresh all |"));
   }
 }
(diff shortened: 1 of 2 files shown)


gausie:
Presumably this meant you could do `status refresh` instead of `refresh status`.

I agree that is silly and should go away. Let's merge this. Can you add a regression test please?

test/net/sourceforge/kolmafia/textui/command/HelpCommandTest.java
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  @Test
+  void Refresh() {
+    String output = execute("refresh");
+
+    assertContinueState();
+    assertThat(
+        output,
+        containsString(
```
gausie, line 28: Tiny thing but could you change this to just be startsWith `refresh all |`. That would reasonably cover the regression while also not make us have to change this test every time we add a new option to refresh

gausie: Can you make this `startsWith` though, I'm so sorry!

s-k-z: Sure, I overlooked that at first, haha.

gausie (approved):

PR #543, opened 2022-01-26
Actually use java version in matrix.

heeheehee-kolmafia:
Previously we were inadvertently hard-coding 17 as the Java
version. This was not great for maintaining Java 11 compatibility
as we purported to do.


heeheehee-kolmafia:
Confirmed that logs for Run Checks (macos-latest, 11) has:

```
Java configuration:
  Distribution: temurin
  Version: 11.0.13+8
```

gausie (approved):

PR #547, opened 2022-01-27
Ronin storage

Veracity0:
Free pulls in Ronin have to go into their own requests, 1 at a time, not shared with other items.


gausie (approved):
Looks great. Well done on these (and the previous set of) tests too!

PR #549, opened 2022-01-27
Expose JS enumerated types on runtime library.

phulin:
This is in addition to global scope; will allow usage in environments where Element and Location cause conflicts with globally-defined DOM constructors.


gausie (approved):
Looks good, thanks!

PR #550, opened 2022-01-28
Delay relay browser startup until SVN Update finished

jaadams5:
I thought about it.  Changed the logic so it didn't need two wait loops and lowered the wait to 30 seconds.


src/net/sourceforge/kolmafia/webui/RelayLoader.java
```diff
@@ -36,20 +39,32 @@ protected RelayLoader(final String location, final boolean isRelayLocation) {
     }
   }
 
+  private void waitForSVNUpdateToFinish() {
+    PauseObject pauser = new PauseObject();
+    int triesLeft = 6;
+    while ((triesLeft > 0) && LoginManager.isSvnLoginUpdateUnfinished()) {
```
gausie, line 45: Why not make this a for loop like the loop below?

jaadams5: The "loop below" was existing code that I didn't think about changing.  I tend to use a for when the number of times I do something is important and a while when there is some condition that I expect to change.  I can certainly make a decision and be consistent.  Stand by.

src/data/defaults.txt
```diff
@@ -120,6 +120,7 @@ global	relayAddsUseLinks	true
 global	relayAddsWikiLinks	false
 global	relayAllowRemoteAccess	false
 global	relayBrowserOnly	false
+global  relayDelayForSVN    5
```
midgleyc, line 123: you've used spaces instead of tabs here.

jaadams5:
I reworked the pause loops and added a preference.  5 seconds is a little too short on my system so it might as well be a preference and not hardwired.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/webui/RelayLoader.java
```diff
@@ -36,22 +37,40 @@ protected RelayLoader(final String location, final boolean isRelayLocation) {
     }
   }
 
+  private void pauseAndWaitForRelayAction(int timeInMilliseconds, boolean waitForStart) {
+    // Pause for input value of seconds, waking up to check relay server
```
jaadams5, line 41: waitForStart is always true in this code but I have an uncommitted test that, if I keep, used the same routine to  wait for shutdown.  Not sure it's a good test but the parameter here defers the discussion.

src/data/defaults.txt
jaadams5: will fix.  how did you find that so fast

midgleyc: if you view it on GitHub it looks obviously misaligned, and then I checked by highlighting the characters.

midgleyc (approved):

PR #552, opened 2022-01-28
Replace mod expression with new `Multiplies` and `Multiplies By` modifier

gausie:
We were pretending to do this two ways before, which confused me. 

Still need to add tests to make sure this all correctly applies when speculating.


gausie:
Yeah this indeed needs changes to work with the maximizer 😢

gausie:
This is too much to do for now

PR #554, opened 2022-01-28
No ListCellRenderer for non-lists

Veracity0:
We use ListCellRendererFactory for creating ListCellRenderer objects for JLists. It is full of specialized renderers for AdventureResults, Concoctions, PurchaseRequests, and so on.

It also has specialized renderers for Storage and FreePulls.

Back in 2012, we changed most of the panels on the Item Manager Frame to no longer use JLists. Instead, they use JXTables. Therefore, many of the renderers in ListCellRendererFactory are never used anymore.

The ItemManagerFrame uses (mostly) InventoryPanels - which extend ItemTableManagePanel - which wraps the elementModel in a JXTable. Therefore, I made InventoryPanel - and HagnkStoragePanel, which extends it - NOT set a ListCellRenderer.

I also pulled the PulverizationRenderer out of ListCellRendererFactory and put it into PulverizationPanel, since it uses none of the framework of the Factory. (We have precedent for this; the CoinMasterFrame uses its own ListCellRenderer.)

I am considering removing StorageRenderer and FreePullsRenderer from ListCellRendererFactory, since we never create one any more.


jaadams5 (approved):
Of course my opinion of "Swing code" is suspect.

PR #555, opened 2022-01-29
Prevent buying two oversized sparklers in the Maximizer

midgleyc:
Tests are in their own file because the setup is completely different for testing creatables.

Includes one disabled test to do with meat pasting. I don't see why it doesn't work, so comments appreciated. I had added it as a "basic" test for creatables, but couldn't get it to work.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/CheckedItem.java
+++ b/src/net/sourceforge/kolmafia/maximizer/CheckedItem.java
@@ -10,6 +10,7 @@ import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase.FoldGroup;
 import net.sourceforge.kolmafia.persistence.MallPriceDatabase;
+import net.sourceforge.kolmafia.persistence.NPCStoreDatabase;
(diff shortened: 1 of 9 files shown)


src/net/sourceforge/kolmafia/maximizer/CheckedItem.java
```diff
@@ -229,6 +229,14 @@ private static int limitBuyable(final int itemId) {
       case ItemPool.REFLEX_HAMMER:
       case ItemPool.MAYO_LANCE:
         return Preferences.getBoolean("_mayoDeviceRented") ? 0 : 1;
+      case ItemPool.FEDORA_MOUNTED_FOUNTAIN:
```
gausie, line 232: Perhaps we could use `NPCPurchaseRequest.getQuantity()` here (and move over the mayo stuff?) to reduce this sort of duplication

midgleyc: Done, but I put it in `NPCStoreDatabase`, which has yet another, presumably slightly different, method.

The change in `RequestProcessor` is to fix what I think has happened in some of my runs: if you buy the hat using e.g. `/buy fedora-mounted fountain` Mafia doesn't load fwshop and so doesn't record the purchase.

gausie (approved):

PR #556, opened 2022-01-29
Finish support for 2022 changes to storage in Ronin.

Veracity0:
The "empty storage" button on the Storage panel of the Item Manager is now disabled unless:

- You are not in Hardcore
- You are not in Ronin
- You actually have items in Storage

Leaving Hardcore or Ronin (via freeing the king, for example) will fire a state change which will make the button enable itself.

New CLI command - test toggle {ronin|hardcore} - let me test that functionality.

Additionally, the StorageRequest now forbids the EMPTY_STORAGE action if you are in Ronin.

I rationalized the run() method of StorageRequest. I also wrote tests for all seven cases that will cause that method to abort without submitting a request to KoL.


Veracity0:
And now I have souped the response text parsing to notice "You already pulled one of those today" and add the corresponding item to the roninStoragePulls set.

midgleyc (approved):

PR #557, opened 2022-01-29
fix: use try-with-resources for closing non-finally-closed BufferedReaders

midgleyc:
Where a `BufferedReader` was closed, but a `finally` block was not used, there is the possibility of the reader not being closed if an exception occurs in the middle of execution.

Change almost everywhere that closes a reader to use a try-with-resources block. Leave places that return a `BufferedReader`, places that do nothing but close the reader, and `RuntimeLibrary` (which both does something more complicated and correctly uses the finally block).

Some places don't close readers at all (e.g. `TCRSDatabase.load`). Those were untouched, but should probably change in a future PR.

This changeset should be easier to read if you download the branch and use `git show --ignore-all-space` (which ignores the indentation changes). There might be a way of doing this in GitHub but I don't know it.

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -2,6 +2,7 @@ package net.sourceforge.kolmafia;
 
 import java.io.BufferedReader;
 import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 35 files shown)


gausie (approved):
Looks great, thank you!

PR #559, opened 2022-01-30
track number of oyster eggs found

horrible-little-slime:
Per the wiki, the probability of finding an oyster egg exponentially decreases by the number of eggs you've found; this adds tracking for the number of oyster eggs encountered, (although not their location)


horrible-little-slime:
Drafting this because I've discovered you can encounter eggs outside of combat, and I'm not entirely sure where to move the code if that's the case.

gausie:
Check for `adventureResults` boolean and add some tests please!

horrible-little-slime:
> Check for `adventureResults` boolean and add some tests please!

Added the boolean check; begging for HTML now so I can write a test

gausie (changes requested):
Awaiting that test

test/net/sourceforge/kolmafia/session/OysterEggTest.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.GregorianCalendar;
+import net.sourceforge.kolmafia.AdventureResult;
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.preferences.Preferences;
+
+public class OysterEggTest {
+  public void OysterEggObtainingTest() {
```
gausie, line 13: These should go in `ResultProcessorTest`

test/net/sourceforge/kolmafia/session/OysterEggTest.java
```diff
@@ -0,0 +1,29 @@
...
+    KoLCharacter.reset("the Tristero");
```
gausie, line 14: Instead of this, add a `@BeforeEach` to the class that runs `KoLCharacter.reset(<className>)` and Preferences.reset(<className>)`. I started using the class name for a username because it limits any unexpected statefullness to that single class.

test/net/sourceforge/kolmafia/session/OysterEggTest.java
```diff
@@ -0,0 +1,29 @@
...
+    ResultProcessor.processResult(new AdventureResult("magnificent oyster egg", 1));
+    assertEquals(Preferences.getInteger("_oysterEggsFound"), 0);
```
gausie, line 16: These should be three separate tests imo

horrible-little-slime:
Wrote the test, ran into some issues. Cannot seem to get it to think that it's Oyster Egg Day. Pushing it with the failing test so as to milk every last drop of wisdom from each of you.

gausie (approved):

PR #560, opened 2022-01-30
InventoryManager.priceToAcquire as retrieve_price

Rinn:
This will return the price that is used when calling `acquire` or `retrieve_item` for the "cheapest" way to acquire said item.

Some things to be aware of how acquire works:
* `valueOfInventory` is taken into account for items in your inventory that determines how an item in your inventory affects the retrieve price. This is a float value that defaults to 1.8 so you may want to set this to 1.0 so it matches mall price.
* historical price is usually used (7 day max) but it will mall search on occasion if the prices are within a 2x threshold of each other to get a more exact difference.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -4628,7 +4628,7 @@ public abstract class RuntimeLibrary {
     int item = countThenItem ? arg2Value : arg1Value;
 
     if (count <= 0) {
-      return RuntimeLibrary.continueValue();
+      return new Value(0);
     }
 
     return new Value(
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -4628,7 +4628,7 @@ public abstract class RuntimeLibrary {
     int item = countThenItem ? arg2Value : arg1Value;
 
     if (count <= 0) {
-      return new Value(0);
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -4608,6 +4617,28 @@ public static Value retrieve_item(ScriptRuntime controller, final Value item) {
...
+    int arg2Value = (int) arg2.intValue();
+
+    boolean countThenItem =
+        arg1.getType().equals(DataTypes.INT_TYPE) || arg1.getType().equals(DataTypes.FLOAT_TYPE);
+
+    int count = countThenItem ? arg1Value : arg2Value;
+    int item = countThenItem ? arg2Value : arg1Value;
+
+    if (count <= 0) {
+      return RuntimeLibrary.continueValue();
```
Veracity0, line 4631: This can't be right. This method returns an int, but the continueValue is a boolean.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
Rinn: oh my bad I copied that from retrieve_item, it should return 0 or -1 I'll fix it.

Veracity0 (changes requested):
Please use DataTypes.ZERO_VALUE

Rinn:
> Please use DataTypes.ZERO_VALUE

Ah right, sorry. Done.

Veracity0 (approved):

PR #562, opened 2022-01-30
Implement wrapped tabs

libraryaddict:
This applies to any areas of the application which implements tabs. Not just chat.

image


gausie:
Please don't make any changes outside of `src`. In any case, the plan is to remove the dependency you've edited very soon!

jaadams5:
> Please don't make any changes outside of `src`. In any case, the plan is to remove the dependency you've edited very soon!

So we should close this?

PR #564, opened 2022-01-31
Allow testing of HolidayDatabase-related functions (and add relevant tests to Modifiers)

gausie:


src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
```diff
@@ -912,12 +913,17 @@ public String toString() {
     }
   }
 
+  public static Date getDate() {
+    return new Date();
+  }
+
   public static final String getHoliday() {
-    return HolidayDatabase.getHoliday(new Date(), false);
+    Date d = getDate();
```
midgleyc, line 921: feel like this function may as well be `getHoliday(false)`.

midgleyc (dismissed):

PR #566, opened 2022-02-01
Move ROLLOVER prop to before static initialization block

gausie:
Is this a thing? Does the order here matter for some operating systems?


src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
```diff
@@ -22,6 +22,7 @@
   private static int RONALD_PHASE = -1;
   private static int GRIMACE_PHASE = -1;
   private static int HAMBURGLAR_POSITION = -1;
+  private static final TimeZone ROLLOVER = TimeZone.getTimeZone("GMT-0330");
```
Veracity0, line 25: Yes. Static initialization is done in order. Therefore, ROLLOVER has to be defined before getPhaseStep (which uses it) is called in a static block.

Veracity0 (approved):

PR #575, opened 2022-02-04
Reminisce command

gausie:


src/net/sourceforge/kolmafia/utilities/StringUtilities.java
```diff
@@ -889,6 +889,19 @@ public static final String leetify(final String text) {
...
   }
 
+  public static final boolean isVowel(char letter) {
+    switch (Character.toLowerCase(letter)) {
+      case 'a':
+      case 'e':
+      case 'i':
+      case 'o':
+      case 'u':
+        return true;
```
midgleyc, line 899: may as well cover this I think, but up to you

midgleyc (approved):

src/net/sourceforge/kolmafia/utilities/StringUtilities.java
gausie: Sure, will do

midgleyc (approved):

PR #576, opened 2022-02-04
update combat text for scrapbook and magnifying glass

thedictatorkol:
Some more text for the scrapbook banish. And some fixed text for the cursed magnifying glass.


gausie (approved):
Thank you!

PR #577, opened 2022-02-04
Add full testing suite for BanishManager

gausie:
I want to refactor BanishManager, so here are some tests for it to make sure nothing breaks.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/BanishManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/BanishManagerTest.java
@@ -12,6 +12,7 @@ import net.sourceforge.kolmafia.combat.MonsterStatusTracker;
 import net.sourceforge.kolmafia.persistence.MonsterDatabase;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.StandardRequest;
+import org.junit.jupiter.api.AfterAll;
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/session/BanishManagerTest.java
```diff
@@ -0,0 +1,365 @@
...
+    assertTrue(BanishManager.isBanished("paper towelgeist"));
+
+    assertFalse(BanishManager.isBanished("zmobie"));
+  }
+
+  @Test
+  void saveBanishedMonsters() {
+    assertThat("banishedMonsters", isSetTo(""));
+
+    // This will be removed because its run out.
```
midgleyc, line 71: `it's`, as in `it has`. Or just in long.

midgleyc (approved):
wow, good job with these

midgleyc:
You might want to add a test to check the Banish Manager doesn't error horribly in case of an unparseable preference, but I don't think it's necessary to block refactoring work (as I don't think it needs to work the same as it did, necessarily -- just "don't crash").

test/net/sourceforge/kolmafia/session/BanishManagerTest.java
gausie: Oh yeah, I'll fix this next round

PR #579, opened 2022-02-05
More Maximizer tests: synergies & gelatinous noob

midgleyc:
Move the tests into nested classes for better ordering. Add tests for synergies and gelatinous noob.

Some tests are disabled. 
* At least one is definitely a Mafia bug (outfits should work in gel noob, but don't).
* one is very weird (gel noob sometimes doesn't recommend absorbing items, but I can't see why). To be more specific, it recommends absorbing a disco mask on "moxie", but not a helmet turtle on "muscle".
* the remaining one looks to also be a Mafia bug: Mafia no longer recommends Flaskfull of Hollow, even on a simple "muscle" maximization. It definitely used to, though.


gausie (approved):
This looks good. I agree that I used to see Flaskful of Hollow recommended and no longer do. Writing these tests is the first step to fixing it - I wish there was a good way of marking a test as an undesirable but known failure.

PR #582, opened 2022-02-05
Item finder fixes

Veracity0:
I added an extensive test suite for ItemFinder
I may add to it, once coverage runs, and I can see where it falls short.

Fixed ItemFinder issues:

- N meat -- "meat" is no longer case sensitive
- Fix NOOB skill filter
- Remove unreachable code (low level "parse one item name" code no longer worries about receiving a comma separated list of names)
- When temporarily changing a preference, put inside try/finally
- When looking for items in storage, they might be in freepulls

Fixed StorageCommand issue:

- In Hardcore, only search freepulls list
- Otherwise, search storage list - and assume ItemFinder will find it on freepulls, if necessary

### DIFF
--- a/test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
+++ b/test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
@@ -11,19 +11,27 @@ import net.sourceforge.kolmafia.KoLmafia;
 import net.sourceforge.kolmafia.StaticEntity;
 import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.persistence.ItemFinder.Match;
+import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.CharPaneRequest;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class ItemFinderTest {
 
-  @BeforeEach
-  private void itemFinderSetup() {
+  @BeforeAll
+  private static void itemFinderClassSetup() {
     // Simulate logging out and back in again.
     KoLCharacter.reset("");
     KoLCharacter.reset("item finder user");
-    // Reset preferences to defaults.
-    KoLCharacter.reset(true);
+  }
+
+  @BeforeEach
+  private void itemFinderSetup() {
+    // Reset exactly the preferences used by this package to defaults
+    Preferences.setString("_roboDrinks", "");
+    Preferences.setBoolean("autoSatisfyWithNPCs", false);
 
     // Not in error state
     StaticEntity.setContinuationState(MafiaState.CONTINUE);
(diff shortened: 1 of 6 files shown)


gausie:
Not sure why no actions are running right now!

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -0,0 +1,833 @@
...
+
+    // Not in error state
+    StaticEntity.setContinuationState(MafiaState.CONTINUE);
+
+    // Default is not in Ronin or Hardcore
+    KoLCharacter.setRonin(false);
+    KoLCharacter.setHardcore(false);
+    CharPaneRequest.setCanInteract(true);
+
+    KoLConstants.inventory.clear();
```
midgleyc, line 36: The log out / log in above should have already cleared the inventory (and everything else). There shouldn't be any need to do it twice.

Some work was done (mostly in MaximizerTests) in an attempt to avoid using `reset` between tests, and do the smallest amount of setup / teardown possible. This might be worth looking into, but I don't think it's required at the moment.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -0,0 +1,833 @@
...
+    // that it will not restrict its matches to what is available in a
+    // specific list. The only exception is if you use wildcard counts -
+    // * or -N - in which case, it assumes inventory. We'll test wildcard
+    // counts separately.
+
+    AdventureResult item;
+    String message;
+    String parameter;
+
+    // Test successful exact match for entire string.
```
midgleyc, line 137: If these were their own tests, you could have a name instead of a comment. You could use `@Nested class SingleItemParsing` to group all of them.

Not particularly fussed about this, I've never really agreed with the "each test should only test one thing" idea.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -0,0 +1,833 @@
...
+    item = ItemFinder.getFirstMatchingItem("toilet paper", false, null, Match.ANY);
+    assertEquals(StaticEntity.getContinuationState(), MafiaState.CONTINUE);
+    assertTrue(item != null);
+    assertTrue(item.getItemId() == ItemPool.TOILET_PAPER);
```
midgleyc, line 141: `assertEquals(ItemPool.TOILET_PAPER, item.getItemId())` should give a slightly better message in the case of failure. I think it's better to default to using `assertEquals(x, y)` over `assertTrue(x == y)`.

midgleyc (approved):
Various comments, all optional. The tests pass (and are very well-commented), so I don't think it's necessary to change them.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
gausie: > Not particularly fussed about this, I've never really agreed with the "each test should only test one thing" idea.

I've definitely found it help later down the line when you make a change and suddenly see a random test break - you know exactly what behaviour has changed.

midgleyc: yeah, but if you look at the stack trace (or look at the assertion error message, if you provided one), you also know exactly which behaviour changed ;). You don't get to see multiple failures up-front, but given that you have to fix all the failures always anyway, it all comes out in the wash for me.

jaadams5 (approved):

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
Veracity0: I want to clear everything before the whole suite runs, so the login/etc. could go in a @BeforeAll.
I want to be free to set up various item lists in each (group of) tests, so reseting items lists (and hardcore/ronin/interaction) could go in @BeforeEach or @AfterEach.
Ditto for continuation state; don't want to stay in ERROR, for example.

I'll try splitting into @BeforeAll and @BeforeEach, at least, to minimize the mega-reset

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
Veracity0: Seeing multiple failures would have been helpful, rather than see one, fix it, run again, see next, etc.
Not familiar with @Nested.

As I mentioned in a previous comment, I want to rested the items lists before each group of tests (each test, as currently coded). And perhaps add items to the lists, custom per group of tests (or at the top of the function, as currently coded).

I want to get this in, so will study @Nested later - and try it out for my NEXT major test suite.
Or maybe I'll follow up in a different PR by refactoring this one, as a learning experience. :)

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
Veracity0: Yes, this makes sense. Does assertEquals actually use the equals() method? (One would hope)
I'll experiment and convert, as appropriate.

Veracity0: All done.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
Veracity0: Not changing structure at this time, as mentioned.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
Veracity0: Made @BeforeAll and @BeforeEach, as discussed.
Reset exactly the two Preferences that ItemFinder uses in the @BeforeEach.

Veracity0:
I'm going to force merge this. It was previously approved, the test suggestions were optional - although I did adopt some of them :) - so, time to get it in to the wild.

PR #584, opened 2022-02-07
Fixes to InventoryManager.priceToAcquire

Veracity0:
I noticed these bugs:

- When calculating cost to make an item, there is at least one concoction with an ingredient which is Meat.
   In particular, rare Meat isotope can be purchased from Cosmic Ray's Bazaar - a CoinMaster.
   Don't call InventoryManager.costToAcquire for Meat. We know the cost.
- Cosmic Ray's Bazaar is only available in Kingdom of Exploathing. the "Coinmaster" crafting type is
   generally available, but individual Coinmasters may or may not be available.
   Do not consider a concoction available to make if the Coinmaster is not accessible.
- priceToAcquire is willing to use items in inventory and to buy or make more to fill the desired quantity.
   If the cost to buy is too close to the cost to make, it tries again using exact pricing - i.e., fetching current
   mall prices - but only checked the prices for the not-in-inventory quantity needed, as if that was enough
   for the whole quantity needed. Check instead for the entire quantity needed.

I'm not sure how to write tests for this; the methods are public, but depend on the Mall Price Database and doing MallSearchRequests. Seems like a lot of mocking would be needed.

Suggestions welcome.


jaadams5 (dismissed):

gausie:
> I'm not sure how to write tests for this; the methods are public, but depend on the Mall Price Database and doing MallSearchRequests. Seems like a lot of mocking would be needed.

 I think it would indeed just have to be a fair bit of mocking - though it might not be loads!

jaadams5 (approved):

PR #585, opened 2022-02-07
getAdjustedHitStat return max for all stats if ATTACKS_CANT_MISS

ckb11:
Move ATTACKS_CANT_MISS check before getHitStatType switch so that it applies for any hit stat (muscle, mysticality, or moxie)


jaadams5:
I'm approving because it does what it says it will do.  But I am not merging since I just don't know that it is the right thing.  So someone who is more certain than I am that the modifier applies to non-Muscle attacks can merge.

jaadams5 (approved):

gausie:
Let's merge it and see if anyone complains :)

PR #586, opened 2022-02-08
Use image caching in GCLI

gausie:
I had a chat with @cdmoyer who noticed that for some reason the top 3 resource requests (by a huge margin) were to the images for hp, new age healing crystals and gold nuggets. I realised that said images show up in the GCLI when you are volcano farming with preference logging activated.

GCLI was not using the cached image renderer and I think every time it was redrawn it was pulling images from the KoL servers for all the images it was displaying. This PR should fix that.


jaadams5 (approved):

cdmoyer:
This is exciting!  Although, I'm worried that ImageCachingViewFactory will be mad at me for forcing its privacy to be violated.

PR #588, opened 2022-02-08
feat: extend absorb to allow 'absorb X item'

midgleyc:
When [farming in Gel Noob](http://forums.kingdomofloathing.com/vb/showthread.php?t=248197) you want to kick off the day by absorbing 15 of a single item. Extend the "absorb" command to allow this (most of the logic was there already).

Use the same item parsing function in `heist`, now that I know what it is.

Tested manually.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
@@ -38,6 +38,8 @@ public class AbsorbCommand extends AbstractCommand {
 
     AdventureResult match = ItemFinder.getFirstMatchingItem(parameters, Match.ABSORB);
     if (match == null) {
+      // for backwards compatibility: previously no message was displayed, so do not error here
+      KoLmafia.updateDisplay("What item is " + parameters + "?");
       return;
     }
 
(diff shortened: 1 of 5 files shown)


src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
```diff
@@ -52,15 +55,19 @@ public void run(final String cmd, String parameters) {
...
     }
 
-    // Absorb the item
-    RequestThread.postRequest(
-        new GenericRequest("inventory.php?absorb=" + itemId + "&ajax=1", false));
+    // Absorb the item(s)
+    for (int i = 0; i < count; i++) {
+      RequestThread.postRequest(
+          new GenericRequest("inventory.php?absorb=" + itemId + "&ajax=1", false));
+    }
```
Veracity0, line 62: Is it necessary to allocate a new request each time through the loop, or could you make one and reuse it?
I don't feel strongly about it, but I'd like to hear your rationale if you do feel strongly.

midgleyc: sure, I can re-use it. I didn't spot that it was creating a new object.

Veracity0 (approved):

PR #589, opened 2022-02-08
Fix a few bugs in Player

Veracity0:
Having just started investigating the unit testing classes we provide, I noticed a couple bugs in Player.java.

AdventureResult.addResultToList will take an AdventureResult, complete with count, and will add (or remove) that object to the desired list, adding the count to an existing object on the list (adding it if it was not present). If the count is negative, adding the count will decrease the existing count (obviously), and if the decremented count becomes 0, the object is removed from the list.

Player.addItem could have used that. Now it does.

### DIFF
--- a/test/internal/helpers/Player.java
+++ b/test/internal/helpers/Player.java
@@ -162,4 +162,9 @@ public class Player {
     mocked.when(HolidayDatabase::getDate).thenReturn(cal.getTime());
     return new Cleanups(mocked::close);
   }
+
+  public static Cleanups usedAbsorbs(int absorbs) {
+    KoLCharacter.setAbsorbs(absorbs);
+    return new Cleanups(() -> usedAbsorbs(0));
+  }
 }
--- a/test/internal/helpers/Player.java
+++ b/test/internal/helpers/Player.java
@@ -32,25 +32,30 @@ public class Player {
     return cleanups;
   }
 
-  public static Cleanups addItem(String item) {
-    return addItem(item, 1);
+  public static Cleanups addItem(String name) {
+    return addItem(name, 1);
   }
 
-  public static Cleanups addItem(String item, int count) {
+  public static Cleanups addItem(String name, int count) {
+    int itemId = ItemDatabase.getItemId(name, count, false);
+    return addItem(ItemPool.get(itemId, count));
+  }
+
+  public static Cleanups addItem(int itemId) {
+    return addItem(itemId, 1);
+  }
+
+  public static Cleanups addItem(int itemId, int count) {
+    return addItem(ItemPool.get(itemId, count));
+  }
+
+  public static Cleanups addItem(AdventureResult item) {
     var cleanups = new Cleanups();
-    AdventureResult parsed = AdventureResult.tallyItem(item, count, true);
-    AdventureResult negated = parsed.getNegation();
-    AdventureResult.addResultToList(KoLConstants.inventory, parsed);
-    cleanups.add(() -> AdventureResult.addResultToList(KoLConstants.inventory, negated));
+    AdventureResult.addResultToList(KoLConstants.inventory, item);
+    // Per midglec: "All the cleanups I wrote assume you're reverting back to a reset character"
+    // Therefore, simply remove this item from inventory.
+    cleanups.add(() -> AdventureResult.removeResultFromList(KoLConstants.inventory, item));
     return cleanups;
-  }
-
-  public static void addItem(int itemId) {
-    addItem(itemId, 1);
-  }
-
-  public static void addItem(int itemId, int count) {
-    AdventureResult.addResultToList(KoLConstants.inventory, ItemPool.get(itemId, count));
   }
 
   public static int countItem(int itemId) {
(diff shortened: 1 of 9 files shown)


test/internal/helpers/Player.java
```diff
@@ -51,7 +50,7 @@ public static void addItem(int itemId) {
   }
 
   public static void addItem(int itemId, int count) {
-    AdventureResult.addResultToList(KoLConstants.inventory, ItemPool.get(itemId));
+    AdventureResult.addResultToList(KoLConstants.inventory, ItemPool.get(itemId, count));
```
gausie, line 53: Not introduced in this PR but I think this function and the other should probably both point to a third function that takes an `AdventureResult`

test/internal/helpers/Player.java
```diff
@@ -38,11 +38,10 @@ public static Cleanups addItem(String item) {
 
   public static Cleanups addItem(String item, int count) {
     var cleanups = new Cleanups();
-    AdventureResult parsed = AdventureResult.tallyItem(item);
-    for (int i = 0; i < count; i++) {
-      AdventureResult.addResultToList(KoLConstants.inventory, parsed);
-      cleanups.add(() -> AdventureResult.removeResultFromList(KoLConstants.inventory, parsed));
-    }
+    AdventureResult parsed = AdventureResult.tallyItem(item, count, true);
```
gausie, line 41: Not introduced in this PR but one issue I noticed with this cleanup is that it is misleading for tests that remove the item added during the course of execution. A better cleanup would be to count the items before and restore to that count in the cleanup

gausie (approved):

test/internal/helpers/Player.java
Veracity0: Well, the other one takes a name and makes a tallyItem. That can be used for things that are not actually items. For example, "pirate insult" or "Arena flyer ML". Those (intentionally) end up with itemId = -1, but have ITEM_PRIORITY, so work well on lists of AdventureResults that hold items. Like KoLConstants.tally - the "results" list that is shown in the GUI. Or even KoLConstants.inventory, although the non-items I mentioned are not put there by the actual KoLmafia code that creates them.

Perhaps a better API would have addItem use ItemPool.get and add to inventory and addTallyItem create a tallyitem and add to tally.

The other one also generates Cleanups and this one does not. I suppose the Cleanups version could call the other one to add to the list.

gausie: I think we want this one to also make cleanups. It was made before Cleanups existed. As far as I know, all usages use it for items - we don't have any testing that uses tallyitems

test/internal/helpers/Player.java
Veracity0: Are there such tests? If so, they can't use this. If not, as soon as the first such test is written, it will be time to fix this. :)

test/internal/helpers/Player.java
Veracity0: OK. I'll refactor.

test/internal/helpers/Player.java
midgleyc: All the cleanups I wrote assume you're reverting back to a reset character -- for example, `canEquip` and `setStats` just set the stats back to zero, instead of whatever they were before.

test/internal/helpers/Player.java
midgleyc: When I created these I didn't (and still don't) really understand the difference between each of the different methods -- I was just trying to find one that worked.

`tallyItem` was used because I was previously using `AdventureResult.parseResult("baconstone")` but that gave the wrong result while `tallyItem` worked. If there's a better "this is an item, parse this exactly" method that should probably be used instead.

test/internal/helpers/Player.java
Veracity0: If that is the assumption we want, then we actually do want AdventureResult.removeFromList().

test/internal/helpers/Player.java
Veracity0: new AdventureResult("baconstone") should work. But if yu want to guarantee you get an item, ItemDatabase.getItemId("baconstone", 1, false) will look up the name, assuming it is not a plural (count===1) without fuzzy matching (false).

OK. I am submitting an improved refactoring. The tests still pass, so I haven't broken anything, it seems.

Note that all variants addItem always returns Cleanups, so you can use it in try-with-resources - or do a bit of unnecessary allocation if you don't do that.

Veracity0: By the way - perhaps we want ItemPool.get(String name, int count) which calls ItemDatabase, in order to make it very clear which function you should call to get an AdventureResult.

test/internal/helpers/Player.java
gausie: I've written plenty - testing quests that remove a quest item or whatever. It hasn't had an issue because as ryo says, we just reset the character afterwards. I'm just imagining more elaborate tests in the future - but resetting to zero is probably better. Remove from list sounds good. But we can do this later

gausie (approved):

PR #580, opened 2022-02-05
Refactor BanishManager + add Cosmic Bowling Ball support

gausie:

PR #590, opened 2022-02-09
Fix recording banished monster from Ice House in museum

s-k-z:
Broke in #580 with case-sensitive monster name searching.
Added a test to simulate visiting the museum ice house.


s-k-z:
Not sure why the test doesn't work consistently.

gausie (dismissed):
Thank you!

gausie:
> Not sure why the test doesn't work consistently.

Yes I see this now, will look into it.

PR #594, opened 2022-02-10
Security Scan flagged casts

jaadams5:
The security scan flagged three casts as being problematic.  This fixes them and cleans up some lint.

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -1479,9 +1479,8 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
   private int safeDowncastLongToInt(long value) {
     int retValue;
     try {
-    retValue = Math.toIntExact(value);
-    }
-    catch (ArithmeticException e) {
+      retValue = Math.toIntExact(value);
+    } catch (ArithmeticException e) {
       if (value < 0) {
         retValue = -1 * Integer.MAX_VALUE;
       } else {
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -1049,7 +1049,7 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
         item.id = this.id;
       }
 
-      item.count = (int) quantity;
+      item.count = safeDowncastLongToInt(quantity);
       return item;
     }
 
@@ -1064,11 +1064,11 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
         effect.id = this.id;
       }
 
-      effect.count = (int) quantity;
+      effect.count = safeDowncastLongToInt(quantity);
       return effect;
     }
 
-    return new AdventureResult(this.name, (int) quantity);
+    return new AdventureResult(this.name, safeDowncastLongToInt(quantity));
   }
 
   public AdventureResult getInstance(final int[] quantity) {
@@ -1474,5 +1474,20 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
       }
       return null;
     }
+  }
+
+  private int safeDowncastLongToInt(long value) {
+    int retValue;
+    try {
+    retValue = Math.toIntExact(value);
+    }
+    catch (ArithmeticException e) {
+      if (value < 0) {
+        retValue = -1 * Integer.MAX_VALUE;
+      } else {
+        retValue = Integer.MAX_VALUE;
+      }
+    }
+    return retValue;
   }
 }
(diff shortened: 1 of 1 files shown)


jaadams5:
At some point we added a scan by CodeQL for vulnerabilities and possible coding errors.  The code had about 70 when CodeQL was first run (on GitHub) we drove that down to zero (actually one but it dealt with an outdated encryption that we have to use because KoL requires it so we ignore it) and a recent scan found some more.

Almost at the top of this page is a tab(?) called Security,  It currently has a 4 in a circle meaning 4 unresolved items.

If you click on it you go to https://github.com/kolmafia/kolmafia/security which has an overview.

Click on Code Scanning Alerts there to get to https://github.com/kolmafia/kolmafia/security/code-scanning which lists the open alerts and has options for viewing history.  You can click on an alert to get more information about the alert, root cause and possible fixes.

If you go to Actions (in the same top level "tab" list) you can go to https://github.com/kolmafia/kolmafia/actions

If you select the bottom CodeQl workflow you can get an option to run the workflow on main or a branch of your choosing.  I did that so I could keep this as a draft until I know it addressed the problem. 

OK?

jaadams5:
"Casting a user-controlled numeric value to a narrower type can result in truncated values unless the input is validated."

There were about 15 previous instances that were corrected so there is precedent for just fixing newly discovered instances.

I won't self approve this as is.

Veracity0:
The key is "user controlled int field".

Is there anywhere in the code where user input will result in a long value which is out of range for an int.

I claim "no". I think it's up to you to fond a case - and write a test for it - that contradicts that.

Veracity0:
To make it more clear:

If we parse Meat, HP, or MP, we need a long, and we pass in a real long,

If we parse literally any other quantity, ant int is fine, and we pass in an int to s method which takes a long.

Focus on the "user input" part pf the "issue".

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1475,4 +1472,18 @@ public static WildcardResult getInstance(String text) {
       return null;
     }
   }
+
+  private int safeDowncastLongToInt(long value) {
+    int retValue;
+    try {
+      retValue = Math.toIntExact(value);
```
gausie, line 1479: Just return here

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1475,4 +1472,18 @@ public static WildcardResult getInstance(String text) {
...
+    } catch (ArithmeticException e) {
+      if (value < 0) {
+        retValue = -1 * Integer.MAX_VALUE;
```
gausie, line 1482: Just return here

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1475,4 +1472,18 @@ public static WildcardResult getInstance(String text) {
...
+      } else {
+        retValue = Integer.MAX_VALUE;
```
gausie, line 1484: Just return here!

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1475,4 +1472,18 @@ public static WildcardResult getInstance(String text) {
       return null;
     }
   }
+
+  private int safeDowncastLongToInt(long value) {
```
gausie, line 1476: Should this be in a utility class elsewhere? Can you add tests for it?

gausie: But for some reason I didn't see that V has already questioned the necessity of the change in the first place.

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1,9 +1,6 @@
 package net.sourceforge.kolmafia;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.StringTokenizer;
+import java.util.*;
```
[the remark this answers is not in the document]
jaadams5: Either Spotless or my IDE did that.  I will check the latter but we may want to check our Spotless configuration.

src/net/sourceforge/kolmafia/AdventureResult.java
jaadams5: Spaghetti code.  I have a distinct preference for code segments that have only one exit path.  Do I need to abandon that preference?

src/net/sourceforge/kolmafia/AdventureResult.java
jaadams5: > Should this be in a utility class elsewhere? Can you add tests for it?

Yes.  But the equivalent is Ints.saturatedCast in Google's Guava and I wasn't sure I wanted to pull in the entire library for one routine.  But at least the routine would have been tested :-)

jaadams5:
The sole motivation for doing this was to address an alert flagged as "Critical" by CodeQL.  Since both the motivation and the implementation are under discussion, I am going to close this.

PR #595, opened 2022-02-10
Address security scanner issue with certificate trust

jaadams5:


src/net/sourceforge/kolmafia/utilities/NaiveSecureSocketLayer.java
```diff
@@ -79,11 +77,15 @@ public static void uninstall() {
   private static class NaiveTrustManager implements X509TrustManager {
     @Override
     public void checkClientTrusted(X509Certificate[] chain, String authType)
-        throws CertificateException {}
+        throws CertificateException {
+      throw new CertificateException("This is client unreal");
```
gausie, line 81: What does this mean? I don't think I understand the error message. Same with the other

jaadams5: I don't remember either.  I was thinking it didn't matter since they are never called in our code and exist only to pass static analysis.  But the message has been changed to something that should make sense if it were ever displayed.

midgleyc:
Checking the code, the `NaiveSecureSocketLayer` is installed only if the preferences `useNaiveSecureLogin` or `connectViaAddress` are true, and they both default to `false`.

If this leads to a security issue, I would consider just removing this feature.

gausie:
> Checking the code, the `NaiveSecureSocketLayer` is installed only if the preferences `useNaiveSecureLogin` or `connectViaAddress` are true, and they both default to `false`. `connectViaAddress` uses the IP address instead of the hostname (so the cert won't match) and `useNaiveSecureLogin` doesn't validate certificates (which is faster but less secure).
> 
> If this leads to a security issue, I would consider just removing this feature. I assume nowadays Java can validate certificates pretty quickly, and the DNS is stable.

Support for deleting code :)

jaadams5:
> > Checking the code, the `NaiveSecureSocketLayer` is installed only if the preferences `useNaiveSecureLogin` or `connectViaAddress` are true, and they both default to `false`. `connectViaAddress` uses the IP address instead of the hostname (so the cert won't match) and `useNaiveSecureLogin` doesn't validate certificates (which is faster but less secure).
> > If this leads to a security issue, I would consider just removing this feature. I assume nowadays Java can validate certificates pretty quickly, and the DNS is stable.
> 
> Support for deleting code :)

That addresses all of my preferences and concerns as well.  I will do so and push again.  I want to do some research to see if I can understand what problem this was supposed to solve when it was added.

PR #598, opened 2022-02-13
Refactor: move everything to do with mall prices out of StoreManager into new MallPriceManager

Veracity0:
I want to work with InventoryManager.priceToAcquire and related methods.
I'm also looking at the mall price cache and such and am itching to clean it up.

But before I can do that, I want a robust set of tests.

To facilitate that, I have moved all the Mall Price cache stuff out of StoreManager (which deals with YOUR store in the mall - inventory, store logs, repricing, etc.) into a new package: MallPriceManager.

Note that I observed a bug in Inventory Manager; priceToAcquire should assume all "accessible" items are "onhand" - not just those which are in inventory - but I have that as a *** comment, rather than actually making the change, yet. See above comment about need for tests...

No new code was added nor existing code changed. Just moved. One would hope for zero change to coverage.


src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -0,0 +1,497 @@
...
+import net.sourceforge.kolmafia.request.CoinMasterPurchaseRequest;
+import net.sourceforge.kolmafia.request.GenericRequest;
+import net.sourceforge.kolmafia.request.MallPurchaseRequest;
+import net.sourceforge.kolmafia.request.MallSearchRequest;
+import net.sourceforge.kolmafia.request.PurchaseRequest;
+import net.sourceforge.kolmafia.utilities.IntegerArray;
+
+public abstract class MallPriceManager {
+
+  private static final IntegerArray mallPrices = new IntegerArray();
```
midgleyc, line 31: the corresponding fields should be removed from StoreManager, but don't seem to have been

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -117,8 +117,33 @@
 import net.sourceforge.kolmafia.request.CampgroundRequest.CropType;
 import net.sourceforge.kolmafia.request.DeckOfEveryCardRequest.EveryCard;
 import net.sourceforge.kolmafia.request.FloristRequest.Florist;
-import net.sourceforge.kolmafia.session.*;
+import net.sourceforge.kolmafia.session.BanishManager;
```
midgleyc, line 120: not a concern for this PR, but some people's editors appear to be collapsing many imports into *, while this here is expanding * into many. Can we agree on a standard setting (e.g. if greater than X, star, otherwise not, tests can always use * if that would be more convenient?)

midgleyc (dismissed):

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: Oops. Could have sworn I did that. Fixed.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
Veracity0: My impression is that we don't want stars - although, why not for tests?
For actual code, make sure that you know you are adding a new dependency.
But let's discuss this elsewhere.

Veracity0: My argument: if you see a method call for another package, you can see exactly where to find it in the imports - unless it is hiding behind a star.

I use Emacs, not an IDE, so this is rather important for me.

midgleyc (approved):

PR #600, opened 2022-02-15
Basic support for Grey You

gausie:
- Teach enums about the path and class
- Add all currently known skills, effects, modifiers
- Fix the CharSheetRequest status parser - we have to move away from StringTokenizer because it doesn't give us the ability to peek a token without consuming it, so our stat parsing was broken with this path not having substats. The class is "not advised" by Java anyway (for this exact reason!), though not deprecated. Added some tests because this change is non-trivial
- Also updated some skill images that mafia flagged as incorrect


midgleyc (approved):
I think tests for Vampyre + Zombie would be nice, but not worth blocking on, and will be identified pretty quickly in the wild if relevant.

gausie:
> I think tests for Vampyre + Zombie would be nice, but not worth blocking on, and will be identified pretty quickly in the wild if relevant.

Yeah agreed. If I had a spare account in one of the paths I'd have done this.

PR #601, opened 2022-02-15
Add testing to Tuesday's Ruby + refactor some Calendar things

gausie:
This is mostly to stop coverage changing every day as the Tuesday's Ruby script executes a different line. However it also:

- Removes some items that didn't need to be in the overrideItem function
- Centralises date/calendar code in HolidayDatabase in case someone ever feels like moving us over to the new time API.


src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
```diff
@@ -1120,14 +1130,26 @@ public static final String getRealLifeOnlyHoliday(final String stringDate) {
...
   }
 
+  private static final boolean withCalendar(final Date date, Predicate<Calendar> predicate) {
+    var cal = getKoLCalendar();
+    cal.setTime(date);
+    return predicate.test(cal);
+  }
+
+  public static final boolean isMonday() {
+    return isDecember(getDate());
```
midgleyc, line 1140: you don't want `isDecember` here

midgleyc (changes requested):

src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
gausie: Yeah I spottted this

midgleyc (approved):

PR #605, opened 2022-02-16
Add spells to dailylimits db and use that resource where appropriate

gausie:
We already have maximum usage for items tracked in a more repeatable way - this does the same for skill use. I dropped it for a few months (I got made redundant and it threw me a bit!) but I've rebased and improved it here. Luckily I wrote enough tests in the past that it was easy to jump back into.


src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
```diff
@@ -128,44 +138,55 @@ public int getId() {
...
-          String tag = data[0];
-
-          DailyLimitType type = DailyLimitDatabase.tagToDailyLimitType.get(tag.toLowerCase());
-          DailyLimit dailyLimit = parseDailyLimit(type, data);
-          if (dailyLimit == null) {
-            RequestLogger.printLine("Daily Limit: " + data[0] + " " + data[1] + " is bogus");
-            error = true;
-          }
+
+    BufferedReader reader =
```
midgleyc, line 171: You've dropped a `try (...)` here.

test/net/sourceforge/kolmafia/persistence/DailyLimitDatabaseTest.java
```diff
@@ -0,0 +1,211 @@
...
+    @Test
+    void canGetUsesRemaining() {
+      Preferences.setBoolean("_floundryItemUsed", true);
+      var limit = DailyLimitType.USE.getDailyLimit(ItemPool.FISH_HATCHET);
+
+      assertThat(limit.getUsesRemaining(), equalTo(0));
+    }
+
+    @Test
+    void canGetUsesRemainingForUsedBoolean() {
```
midgleyc, line 77: This isn't a boolean, but an integer.

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
```diff
@@ -112,7 +122,7 @@ public int getMax() {
         }
       }
 
-      return 1;
+      return -1;
```
jaadams5, line 125: This looks like a magic number change to me, having not looked at every file yet.  Given my expectations for something named getMax I would expect it always returned a non-negative integer.  If it is used in the context of "can I do this" then I would expect zero to mean No.  So I'd like to see this return a zero (if things fall through to here) or make a named constant so it is clear that -1 has a special meaning.

test/net/sourceforge/kolmafia/request/BasementRequestTest.java
```diff
@@ -19,19 +17,16 @@ protected static void injectPreferences() {
...
-        Arguments.of("Stone Golem'", "X Stone Golem"),
-        Arguments.of("Hydra'", "X-headed Hydra"),
-        Arguments.of("Toast that Ghost'", "Ghost of Fernswarthy's Grandfather"),
-        Arguments.of("Bottles of Beer on a Golem", "X Bottles of Beer on a Golem"),
-        Arguments.of("Collapse That Waveform!", "X-dimensional horror"));
-  }
-
   @ParameterizedTest
-  @MethodSource("monsterFights")
+  @CsvSource({
```
midgleyc, line 21: while personal preference, I prefer the external strongly-typed argument to the stringly-typed csv source ;)

midgleyc (changes requested):

jaadams5 (dismissed):

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
gausie: > Given my expectations for something named getMax I would expect it always returned a non-negative integer.

Your expectations are correct - this is what is returned if we have invalid underlying data. I think it's fine to keep it as is.

test/net/sourceforge/kolmafia/request/BasementRequestTest.java
gausie: I think this just lends itself to really easy test writing, and it's surprisingly hard to make mistakes with it

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
gausie: On purpose! Unless I'm mistaken this exception is not thrown, its caught in FileUtilities.readLine

midgleyc: it's not about the exception, it's about closing the reader.

gausie: Ah so we need the try but not the catch. Will fix this later, I'm off to a concert!

jaadams5: Performing or listening?

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
jaadams5: I would prefer, then, a named constant or a comment that said something like "-1 is understood elsewhere to mean that there was invalid underlying data".

I also wonder, because I have not spent a lot of time on context for this, if "invalid underlying data" should be logged as an error so the underlying data could be corrected?

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
midgleyc: you need the catch because the `close` method, by Java's exceptions, can throw. Realistically I don't think it's going to but that's Java for you.

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
gausie: It is logged, it's just done in the calling function which is where the data is actually being loaded. I could always make it throw or something but this seems fine

gausie: I'll add a comment

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
midgleyc, line 171: still no `try` here

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
midgleyc, line 125: still no explanation of magic number here

midgleyc (changes requested):

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
gausie: Wait what

src/net/sourceforge/kolmafia/persistence/DailyLimitDatabase.java
gausie: Wait what

gausie: Forgot to push

gausie: Please make note of the commit time for that commit! Haha

midgleyc (approved):

PR #607, opened 2022-02-16
Preliminary unit tests for MallPriceManager

Veracity0:
Work in progress: write tests for MallPriceManager in preparation for (eventual) refactoring.
But, this is as much as I have patience/time for, at the moment, so time to submt it.

This package "timestamps" MallPurchaseRequests. Rather than using System.currentTimeMillis(), it now uses a (Java 8) Clock object, which it can mock during tests, to allow testing saved PurchasesRequests to age out, as required.

I fixed a bug in MallSearchRequest when getting the results for a whole "category" items; since KoL returns three times as many items as we thought per page, we ended up making three times too many requests.

Minor style cleanups: we don't need to require ArrayList arguments or return explicit ArrayList results; a simple List will do, since we are not using any features specific to ArrayList.

And I split some methods up so that I could test the guts of them without actually making - or mocking - MallSearchRequests.

And while testing getting the mall prices for "allitems", I notice that the lucky pill is now the lucky-ish pill


test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
```diff
@@ -0,0 +1,287 @@
...
+    KoLCharacter.reset("");
+    KoLCharacter.reset("mall price manager user");
+    CharPaneRequest.setCanInteract(true);
+  }
+
+  @BeforeEach
+  private void beforeEach() {
+    // Stop requests from actually running
+    GenericRequest.sessionId = null;
+    KoLCharacter.setAvailableMeat(1_000_000);
```
jaadams5, line 63: Never ran across that "format" for a literal before :-)

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -113,12 +142,12 @@ public static final void flushCache(final int itemId) {
 
   public static final void flushCache() {
     long t0, t1;
-    t1 = System.currentTimeMillis();
+    t1 = MallPriceManager.currentTimeMillis();
     t0 = t1 - 15 * 1000;
```
jaadams5, line 146: Perhaps make 15 seconds a named constant?

jaadams5 (approved):

Veracity0:
I figured out a peachy way to mock/stub a MallSearchRequest so that MallPriceManager can actually call (what it thinks is) such a request and get back the results loaded from an actual responseTest.

This opens up the opportunity for testing all the rest of the methods that actually do a (single) mall search.

Later, perhaps.

jaadams5 (approved):

PR #608, opened 2022-02-17
Fix ninja snowman assassin in expected_damage

Alium58:
Updated expected_damage to be more correct for ninja snowman assassins. Used formula from https://kol.coldfront.net/thekolwiki/index.php/Ninja_snowman_assassin

Assumed high end of damage (120 in the 100-120 range). Can see in tests below that expected and actual damage match - average actual is around10 less than expected. Only exception is at low damage, which is not a huge concern.

Partially resolves https://kolmafia.us/threads/expected_damage-can-be-very-wrong-in-some-cases.25221/

Tests:
Expected damage: 9
Actual damage: 11, 13, 11, 12, 11, 13

Expected damage: 73
Actual damage: 65, 62, 62, 55, 60, 52, 58, 55, 51

Expected damage: 84
Actual damage: 74, 76, 73, 72, 73, 69, 68

Expected damage: 127
Actual damage: 122, 120, 115, 116, 125


Alium58:
I'm going to write a unit test for this before merging. Will try to find time in the next couple days to start working on that. Going to be my first unit test in Java so may have to upgrade from VSCode haha. Anyways goal is to get general feedback on change itself while I work on testing.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -8205,6 +8205,14 @@ private static Value expected_damage(
         1.0 - (Math.sqrt(Math.min(1000, KoLCharacter.getDamageAbsorption()) / 10.0) - 1.0) / 10.0;
     double elementAbsorb =
         1.0 - KoLCharacter.getElementalResistance(monster.getAttackElement()) / 100.0;
+
+    // https://kol.coldfront.net/thekolwiki/index.php/Ninja_snowman_assassin
+    if (monster.getName().equals("ninja snowman assassin")) {
+      baseValue = Math.max(0, attack - defenseStat) + 120;
```
midgleyc, line 8211: There is no point doing the initial calculation in 8201 if you are going to override it here. Declare the variables ahead of time (`int baseValue; double elementAbsorb`) then do `if...else`

midgleyc, line 8211: Going by the wiki formula, Hero of the Half-Shell doesn't apply, so don't run that in the assassin case.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
Alium58: good catch. I'm going to test to confirm after RO since wiki isn't explicit

Alium58: Hero of the Half-Shell applies to NSA's. Only change in tests below is having a shield equipped or not. Will update wiki

Mus 350
Mys 90
Mox 50

Expected damage WITH wicker shield: 103
Actual: 85, 77, 83, 86, 83, 71, 71, 80, 77

Expected damage with OUT shield: 189 (code currently respects hero shell for NSA)
Actual: 171, 166, 179, 166

midgleyc: awesome, thanks for checking!

midgleyc (approved):

PR #616, opened 2022-02-19
StringUtilsTest Part 1

jaadams5:
Following up on a suggestion elsewhere to  write tests for StringUtilities.  I should be able to get coverage to 100% but just felt like taking a break and getting some questions answered about the intent of basicTextWrap.

Draft for now.  Will see how this goes before deciding to do the rest as a new PR or just add to this.


src/net/sourceforge/kolmafia/utilities/StringUtilities.java
```diff
@@ -781,20 +781,31 @@ public static final double parseDouble(String string) {
   }
 
   public static final String basicTextWrap(String text) {
+    return basicTextWrap(text, 80);
+  }
+  /*
+  Overloaded to facilitate hand crafted test strings where it is easier to follow what is happening with
```
gausie, line 787: If adding a comment to the top of a function, better to just go the whole way and make it a jsdoc (with `/**` and the like)

midgleyc:
I think we may as well get these tests in incrementally, instead of waiting for the end.

I've added a couple of tests to StringUtilities as part of another bit of work, and you might have conflicts. No harm in getting it in earlier, I think?

jaadams5:
OK.  Real Life has been getting in the way.  I have pushed what I have.  We can see whether you merge into this, we merge this and then you open a PR or what.  I can be flexible :-)

A lot of my previous comments have been OBE.  I will review.

src/net/sourceforge/kolmafia/utilities/StringUtilities.java
```diff
@@ -832,6 +840,14 @@ public static final String lookupPrepositions(final String text) {
     return rv == null ? text : rv;
   }
 
+  public static HashMap<String, String> getCopyOfPrepositionsMap() {
```
midgleyc, line 843: I'm not sure what the point of taking a copy of the map is? Does it get changed in the test somehow?

`new HashMap<>(prepositionsMap)` should avoid the cast.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
```diff
@@ -13,8 +15,292 @@ void isVowel() {
...
+  This inserts a break into a string
+   */
+  private String manuallyInsertBreak(String input, int breakPos) {
+    if (breakPos >= input.length()) {
+      return input;
+    } else {
+      return input.subSequence(0, breakPos) + BREAK + input.subSequence(breakPos, input.length());
+    }
+  }
+  // End tests for basicTestWrap
```
midgleyc, line 102: you can use `@Nested` to separate tests without having to put comments to break. See e.g. `MaximizerTest.java`

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
```diff
@@ -13,8 +15,292 @@ void isVowel() {
...
+    "'not an id', '-1'",
+    "'[not an id]', '-1'",
+    "'123','-1'",
+    "'[123', '-1'",
+    "'123[', '-1'",
+    "'123', '-1'",
+    "'[1337]2468 googles', '1337'",
+    "'[123]unreal item', '123'"
+  })
+  public void itShouldExerciseGetBracketedID(String name, String id) {
```
midgleyc, line 115: "parse" rather than "exercise" is more descriptive I think

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
```diff
@@ -13,8 +15,292 @@ void isVowel() {
...
+    String testString = "folder (cyan), folder (magenta), folder (yellow)";
+    List<String> expected = new ArrayList<>();
+    expected.add("folder (cyan)");
+    expected.add("folder (magenta)");
+    expected.add("folder (yellow)");
+    try {
+      List<String> result = StringUtilities.tokenizeString(testString);
+      assertEquals(expected, result, "Lists are not the same.");
+    } catch (Exception e) {
+      fail(e.getMessage());
```
midgleyc, line 145: I think if a method throws the test will fail by default. Not sure, but I think that would simplify things?

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
```diff
@@ -13,8 +15,292 @@ void isVowel() {
...
+    }
+  }
+
+  @Test
+  public void itShouldHandleGoodAndBadEscape() {
+    String testString = "a,\tb";
+    List<String> expected = new ArrayList<>();
```
midgleyc, line 152: I think `List.of()` would be clearer than making a new list and adding to it.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
```diff
@@ -13,8 +15,292 @@ void isVowel() {
...
+    expected.add("a");
+    expected.add("\tb");
+    try {
+      List<String> result = StringUtilities.tokenizeString(testString, ',', '\\', false);
+      assertEquals(expected, result, "Lists are not the same.");
+    } catch (Exception e) {
+      fail(e.getMessage());
+    }
+    char[] test = new char[4];
```
midgleyc, line 161: Does `"a,\\g"` not work?

Also maybe split this into three tests? They don't seem to rely on each other.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
jaadams5: One test case was designed to trigger the exception and a thrown exception means the test passed.  An exception in the other cases was a test failure.  I was trying to make that distinction clear in the code.  Since I had packed distinct cases in one test I also did not want an unexpected exception to be a failure.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
jaadams5: The code under test is parsing a delimited string into a list.  I chose to use a single string with the delimiter exactly were I expected it as the input rather than build the string from tokens.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
jaadams5: "a.\\g" did not work because I am explicitly trying to make a string that has invalid escape sequences.  The complier wouldn't let me.  I will look at breaking things up.

src/net/sourceforge/kolmafia/utilities/StringUtilities.java
midgleyc: can you avoid the cast by using `new HashMap`?

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
midgleyc: Talking about `expected` here: so `List.of("a", "\tb")`.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
midgleyc: Needs to be `a,\\g`, because `\g` is an invalid escape. To make a literal backslash in a string you have to double the character.

As far as `tokenizeString` is concerned, escaping is only relevant for commas: it's used for item names with commas in them (e.g. "tiny plastic Hank North, Photojournalist")

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
midgleyc: Also, if this is supposed to be a string with five characters: `a <comma> \ t b` it is not: it is a string with four characters: `a <comma> <tab> b`.

jaadams5: Changed to List.of().  String length is not relevant here.

test/net/sourceforge/kolmafia/utilities/StringUtilitiesTest.java
jaadams5: testString = "a,\\\\g"; is what my IDE substitutes when I try and make a string a comma slash slash g.  The test then fails whereas building the character array, passes.  This test was driven by a desire to get 100% coverage of tokenizeString.  This is the first attempt that succeeded in covering all the branches.  tokenizeString accepts both a separator and an escape character.  The focus on command and slash is just me focusing on coverage.  Perhaps different separators and escapes need to be tried to make things clearer?

PR #618, opened 2022-02-21
Check if we've used a day pass, before declaring we own the iotm

libraryaddict:
## Issue

If the user uses a day pass, then logs out. On login, mafia will declare that they now own the iotm itself.

## Fix

Check if the "today" preference was set, indicating they used a day pass to get access to the iotm.


jaadams5:
I'm confused about why this is needed.  @gausie changed a number of preferences so that they were set by whether KoL thought the account owned the cool item, or not.  That is clearly more reliable than trying to guess or just believing what the user claimed.

If a user doesn't have a cool item, uses a day pass, logs out and then KoLmafia claims the user has the cool item it seem there might be a better fix.  My first question is just to ask how you know that "mafia will declare that they now own the iotm itself"?  Mafia should only make such a declaration if KoL tells KoLmafia via an API call that the item is owned.  So the mafia code or script that makes the declaration could be incorrect and a better (IMO) fix would be at that point.  If the code making the declaration is correct then perhaps when the API is called and the cool items updated needs to be moved?  Logically, it seems to me that if the item is owned then the day preference is not relevant.  

I note there is a reduction in test coverage.  A failing test that shows what doesn't work without this fix would certainly be helpful.

midgleyc:
The problem is that `http://127.0.0.1:60080/api.php?what=status&for=me` returns a `coolitems` array that contains whether or not you have access to the item at the moment, but the Mafia implementation assumes that `coolitems` showed whether you have the item unlocked permanently.

jaadams5:
> The problem is that `http://127.0.0.1:60080/api.php?what=status&for=me` returns a `coolitems` array that contains whether or not you have access to the item at the moment, but the Mafia implementation assumes that `coolitems` showed whether you have the item unlocked permanently.

But #474 says about cool items that 

> It indicates the fact that a given item has been bound to the account

Before we change the code perhaps we should come to an agreement on what the preferences are supposed to mean?

To me, "bound to the account" is not the same as "accessible".  

The behavior as described - unbound, use day pass, logout, bound, does seem buggy but that is because the only things that logically should change the "bound status" are a call to API or doing something that makes the IOTM show up in inventory (and thus KoL will eventually bind it to the account).

midgleyc:
Yeah, I think `coolitems` was supposed to be "bound to the account" but actually is "can access at the moment". So the API is less useful than thought.

jaadams5:
> Yeah, I think `coolitems` was supposed to be "bound to the account" but actually is "can access at the moment". So the API is less useful than thought.

So what we have is a fix to KoLmafia to deal with a KoL bug or a lack of communication on KoL's part?

So if cool items means accessible NOW...

I use a day pass.  If that is successful it means the area is available.  Presumably if path or other restrictions applied mafia would not let you use the pass.  I log out.  I log back in.  Something says the area is available.  That is correct since the pass was used.  There is no error mafia to be fixed unless KoLmafia overrides what it read from the API.  If for some reason KoL returned different cool item values then maybe KoL should be petitioned to fix their bug?

So is the proposed fix for something that is not a bug?

midgleyc (commented):
Initially I thought we should bring back the preferences panel, but I see this one sets both prefs to false if the item isn't returned from the API, which should correct mistakenly applied IoTMs.

I think we still need to check what gets returned from cool items while in a standard run that doesn't have access to e.g. the airports to confirm behaviour: if it removes areas you can't access because they're in standard, this would introduce a bug where "Always" is removed even though it shouldn't be.

After that I think this is a good solution for the current API. Broken prefs will be fixed by checking the following day.

Veracity0:
I'm in a run which is under Standard restrictions, although it is Grey Goo, not Standard.
Here is an extract from my api.php status:

"pathname":"Grey You",
"coolitems":"airport1,airport2,airport3,airport4,airport5,gingerbreadcity,spacegate,fantasyrealm,piraterealm,neverendingparty,voterregistered,boxingdaycare,hascosmicball"

On the Wrong Side of the Tracks, the Boxing Daycare is not present - but the Neverending Party is.
If I try to enter, "That party is so last year. You don't want to visit on this path."
(The other "old" locations - Detective Precinct, L.O.V., etc. - are not present. Nor are the Voting Booth and LT&T and Horsery on the Right Side of the Track.)

Spacegate and Chateau Montegna are visible in the Mountains. Gingerbread City is not.

Lyle's Monorail can go to FantasyReal and PirateRealm.
I can go to the Welcome Center and get an item - which is too old to use.
I can step into Port Beginning and get the eyepatch - which is too old.

I can go to the airport.
I can go to Barf Mountain
I can enter the Maintenance Tunnels and the Kiosk and the Store
If I try to adventure, "This zone is too old to visit on this path."

midgleyc (approved):
this approach looks good to me, then.

PR #620, opened 2022-02-21
MallPriceManager and MallSearchRequest API improvements

Veracity0:
When I looked at MallPriceManager recently, I noticed a lot of room for improvement - including bugs.
I also noticed that MallSearchRequest's API was hard to understand - partly because it had changed but the documentation had not.

--------------------
Rationalizing MallSearchRequest's API:

Old:

MallSearchRequest(int storeid) - get a list of PurchaseRequests for a single store
MallSearchRequest(String category, String tiers) - get a list of PurchaseRequests for all items in category/tier
MallSearchRequest(String search, int cheapestCount, List<PurchaseRequest> results, boolean retainAll)

Search for an exact item (if search is in "") or all matches (otherwise).

Ask for cheapestCount matches - or as many as KoL wants to give, if 0. (Seems to be 60)

You must supply the list to hold the results - although request.getResults() will return it

And retainAll is documented to mean "whether the list should be cleared before searching", but it means nothing of the sort. It is used only when searching a store. true means search just that store, otherwise, search all your "favorite" stores. The kicker is that the code which handles retainAll as false is unreachable unless you specify a storeId - and you can't do that and also set retainAll. That is obviously a bug.

Observations:

1) You must supply a "results" list when passing in a search string, even though you can get it from the request via a method call. I suppose it can be convenient , on occasion, but it should be optional; a list will be created for you as needed - just as in the category and store search calls.

2) The code to get the results for all your "favorite" stores - by concatenating the results of a mall search of each of them individually - is unreachable.

3) There are exactly 5 places in KoLmafia that create a MallSearchRequest:
```
RuntimeLibrary.java:     MallSearchRequest msr = new MallSearchRequest(item, 20, results, false);
ChatFrame.java:          MallSearchFrame.searchMall(new MallSearchRequest(StringUtilities.parseInt(playerId)));
MallSearchFrame.java:    new MallSearchRequest(searchText, searchCount, MallSearchFrame.results, false));
MallPriceManager.java:   return new MallSearchRequest(searchString, maximumResults, results, true);
MallPriceManager.java:   return new MallSearchRequest(category, tiers);
```

MallSearchFrame wants to pass in its own results.
Everywhere else could easily fetch them via request.getResults().
"retainAll" is only used via its obsolete documented usage

I'm thinking this:

MallSearchRequest(int storeId)
MallSearchRequest(String searchString, int maximumResults)
MallSearchRequest(String searchString, int maximumResults, List<PurchaseRequest> results)
MallSearchRequest(String category, String tiers)

And maybe ... something ... for the "all the results from your favorite stores" feature - although that is never used. As coded, aside from being unreachable, it concatenates the store inventories. One might think it would group items together - so item X from all your favorite stores, then item Y, and so on.

I moved the unreachable code into an unused function until/unless we want to work on that feature. WON'T
I rationalized the constructors as proposed. DONE.

--------------------
Rationalizing MallPriceManager API and internals:

1) IntegerArray mallPrices

Why an IntegerArray? Why not a simple Map<Integer> accessed via mallPrices.getOrDefault(itemid, 0)? DONE.

2) LinkedHashMap<Integer, List<PurchaseRequest>> mallSearches

Why a LinkedHashMap? That preserves the iteration order, but I see no code anywhere in this package that depends on that. Changed to a HashMap. DONE

3) flushCache() iterates over ALL the saved searches and removes those which are more than 15 seconds old.

Why is this ever necessary? Seems like when we look up a mall price - for a single item - we can flush exactly that one, if it is too old. Why the overhead of looking at all of them?

This method is removed. Now, when we retrieve a saved search for a single item, we see if it is "too old". DONE.

4) flushCache(item) and flushCache(item, shopid) iterate over all saved searches.

Why not simply look up the search for the specified item? Well, flushCache(-1, shopid) wants to remove that shop from all items, so it is necessary there, but otherwise, lots of unnecessary overhead.

Both methods now do a single lookup, when appropriate. DONE

5) - getMallPrice(item) - returns the "current" mall price, flushing searches that are "too old" and re-searching, if necessary
   - getMallPrice(item, maxAge) - returns the "currently cached" mall price - unless it is "too old", re-searching if necessary

The definition of "too old" differs for those: in one, it uses the timestamp of the most recent actual mall search, in the other, the timestamp of the price from the MallPriceDatabase.

Can we rationalize this, somehow?

Answer: I think not. The first looks up mall searches, the second is an interface to MallPriceDatabase, falling back on the first if "too old". WON'T

Edit: This is good to go. The version of getMallPrice with a maxAge is not tested - but all I did was to rename some variables to agree with the other getMallPrice. I'll consider testing that method later, if/when I move to MallPriceDatabase.


src/net/sourceforge/kolmafia/request/MallSearchRequest.java
```diff
@@ -254,7 +233,10 @@ public void run() {
     // If an exact match, we can think about updating mall_price().
     if (this.searchString.startsWith("\"") && this.results.size() > 0) {
       AdventureResult item = this.results.get(0).getItem();
-      MallPriceManager.maybeUpdateMallPrice(item, new ArrayList<PurchaseRequest>(this.results));
+      // This was "maybe" updateMallPrice - i.e., only do so if there is not
```
midgleyc, line 236: I feel like this comment is only valuable in the context of this PR -- seeing what the code was, and what it changed into.

I don't feel like it adds any value to the codebase after this PR goes in. Can it be removed?

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -154,92 +154,81 @@ public static MallSearchRequest newMallSearchRequest(String category, String tie
...
-        i.remove();
-        MallPriceManager.updateMallPrice(ItemPool.get(itemId), search);
-        return;
-      }
-      break;
+    List<PurchaseRequest> search = MallPriceManager.mallSearches.get(itemId);
+    if (search != null) {
+      MallPriceManager.mallSearches.remove(itemId);
+      // *** Previous code had this. Really?
+      // MallPriceManager.updateMallPrice(ItemPool.get(itemId), search);
```
midgleyc, line 210: I think removing the code entirely is better than commenting it out with a questioning comment.

If a bug occurs after this PR, someone can look at the code changes in the git history to potentially identify any issues.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -154,92 +154,81 @@ public static MallSearchRequest newMallSearchRequest(String category, String tie
...
-      if (t < t0 || t > t1) {
-        i.remove();
-        continue;
-      }
-      break;
+  public static boolean searchIsTooOld(List<PurchaseRequest> search) {
+    if (search == null || search.size() == 0) {
+      return true;
     }
+    long t1 = MallPriceManager.currentTimeMillis();
```
midgleyc, line 219: While I realise this was a pre-existing problem, can we have better names than `t0` and `t1`? `freshnessLimit` and `now`?

Also the search is too old if the timestamp is after the current time? Why is that? (Again a pre-existing issue though).

midgleyc (changes requested):
Some comments, mostly about comments.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: Done.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: Done.

src/net/sourceforge/kolmafia/request/MallSearchRequest.java
Veracity0: Done.

midgleyc (approved):

PR #621, opened 2022-02-21
Create HTTP connections only through HttpUtilities

midgleyc:
Just the one place missing.

Going to start looking at https://kolmafia.us/threads/migration-from-http-1-1-to-http-2-0.27303/, this was necessary cleanup.


midgleyc:
oh, lib, I see...?

PR #622, opened 2022-02-21
tests: only extend requesttestbase where used

midgleyc:
https://kolmafia.us/threads/migration-from-http-1-1-to-http-2-0.27303/ will involve redoing some of the base class, better to have it used in fewer places.


Veracity0:
I see that I am responsible for three of those. 😆

Veracity0 (approved):

PR #624, opened 2022-02-22
Revert TTT for HC Maximizer

BadHorseMonkey:
Since Maximizer (or something it calls) doesn't properly parse the formula in modifiers.txt and detects the Time-Twitching Toolbelt as a Free Pull, it causes unexpected aborts in HC runs. 

Revert modifiers.txt until it's fixed.


midgleyc (changes requested):
You need to also add back the code that was removed from Modifiers, or this won't be seen as a free pull ever (not that the TTT has been around lately)

BadHorseMonkey:
> You need to also add back the code that was removed from Modifiers, or this won't be seen as a free pull ever (not that the TTT has been around lately)

Doesn't the quest manager resolve this?  Looking at this, I'm not sure why we need anything else.

The same code that sets the timeTowerAvailable sets the free pull setting.  I'm thinking all the code that got yanked out was unnecessary.

Of course, I can't really test this live.
 ```
 public static void handleTimeTower(final boolean available) {
    if (Preferences.getBoolean("timeTowerAvailable") == available) {
      return;
    }

    Preferences.setBoolean("timeTowerAvailable", available);

    // time-twitching toolbelt is a free pull if the time tower is
    // available. Place it in correct storage list.

    Modifiers.getModifiers("Item", "time-twitching toolbelt");
    AdventureResult toolbelt = ItemPool.get(ItemPool.TIME_TWITCHING_TOOLBELT, 1);
    List<AdventureResult> source = available ? KoLConstants.storage : KoLConstants.freepulls;
    List<AdventureResult> dest = available ? KoLConstants.freepulls : KoLConstants.storage;
    int index = source.indexOf(toolbelt);
    if (index > -1) {
      AdventureResult item = source.get(index);
      source.remove(item);
      dest.add(item);
    }

    ConcoctionDatabase.setRefreshNeeded(false);
  }
```

**Comment by midgleyc** [comment]
There's also `StorageRequest.isFreePull`. I don't know which the maximizer uses: can you confirm that calling `isFreePull` on the belt returns true if `handleTimeTower(true)` is called?

**Comment by BadHorseMonkey** [comment]
Sounds like an opportunity for me to make a test case. :)

**Review by midgleyc — APPROVED** [review]
I checked the maximizer code, and I think the relevant code is:

CheckedItem.java
```java
this.initial = InventoryManager.getAccessibleCount(itemId);
```

InventoryManager.getAccessibleCount
```java
count += item.getCount(KoLConstants.freepulls);
```

so I agree that it should use the free pulls section.

PR #629, opened 2022-02-25
Raise command history limit, don't store empty or duplicate commands

libraryaddict:
20 is probably too high granted, 10 may be better given we're no longer storing duplicates or empty lines.

That said, I don't think there's a real reason to keep the limit at 10.


midgleyc (approved):
Looks fine to me. As a stretch, you might want to make the amount a preference, so that folk can choose to store an awful lot of history if they want.

jaadams5 (approved):

libraryaddict:
> Looks fine to me. As a stretch, you might want to make the amount a preference, so that folk can choose to store an awful lot of history if they want.

I do like the idea of making it a preference, seeing as it hasn't been merged in yet.

But the issue is that I think unless we changed a bunch of code, this would only be effective on a restart of the command input screen.

PR #635, opened 2022-02-27
Wiki check improvements

midgleyc:
Simplify the wiki name lookup by using `modifiers.txt`

Share Http Clients between calls where reasonable.

Delete commented out code.

Also mark "bad rum and good cola" as giving the effect it gives, instead of the effect the description says it gives. There's a note in `items.txt` about this.


Veracity0 (approved):
The comment about bad rum and good cola was in modifiers.txt, not items.txt.
That was my bug report - and followup comment - by the way.

PR #639, opened 2022-02-28
Change mall price updates to use Java 9 HttpClient

midgleyc:
More HttpClient to prep for https://kolmafia.us/threads/migration-from-http-1-1-to-http-2-0.27303/, again unnecessary as https://kolmafia.us/scripts/updateprices.php doesn't use HTTP/2.

Could do with somebody with access to https://kolmafia.us/scripts/updateprices.php confirming that the calls work completely correctly. I've tested it on a fake server, and it seems to send and receive the right data, but without access to the real server I can't be sure. I have sent some data to the real thing, and got successful responses back.

One more difference found: HttpClient doesn't read `http.agent` for the user agent; it has to be set manually.

On the plus side, the code is slightly neater, and we can choose to read the whole body in case of error, instead of just the first line, more easily.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
@@ -80,7 +80,8 @@ public class DebugDatabase {
       return "";
     }
 
-    var request = HttpRequest.newBuilder(uri).build();
+    var request =
+        HttpRequest.newBuilder(uri).header("User-Agent", GenericRequest.getUserAgent()).build();
     HttpResponse<String> response;
     try {
       response = client.send(request, BodyHandlers.ofString(StandardCharsets.UTF_8));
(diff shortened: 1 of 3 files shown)


Veracity0 (approved):
I don't have access to the .php file, but I did this:

In the original version of MallPriceDatabase, I changed submitPrices to use this:
```
try (OutputStream o = DataUtilities.getOutputStream("mallprices.old.txt");
     BufferedWriter w = new BufferedWriter(new OutputStreamWriter(o))) {
```

and in the new one, writing to mallprices.new.txt

I ran vcon and used "spade prices https://kolmafia.us/scripts/updateprices.php" with both versions of the module.

No network traffic ensued; we didn't really submit data to the server.
However, I had two data files.

Modulo differences in timestamps, the generated data was exactly in the same format.

I'm happy.

midgleyc:
Thanks! I will get this in, then look at creating a test for `getPostData` as a follow-up.

PR #640, opened 2022-03-03
Parse stats from api.php in Grey You

lostcalpolydude:


Veracity0 (approved):
I'm approving this - although I'd like a test - because you're fixing a major issue. And the new code is clearly not going to change anything out of Grey You.

In your ample free time, consider writing a test for CharSheetRequest.parseStatus(JSON).

PR #641, opened 2022-03-04
Add 'gooskills' command

Veracity0:
gooskills

Will display an HTML table with all Grey You skills - name, type, which monster and zone grants it, the effect.

gooskills needed

The same as above but filters out all skills you already have.

gooskills id (default)
gooskills name
gooskills monster
gooskills zone

(You can also include 'needed' if you want).
How to sort the skills before generating the table.

I haven't actually ascended into a Grey You run yet, so haven't tested "needed".
I expect to ascend today - which is why I made this little helper.
Sort of like we have "accordions" and "latte" commands.


src/net/sourceforge/kolmafia/objectpool/SkillPool.java
```diff
@@ -454,6 +454,69 @@
...
+  public static final int NANTLERS = 27039;
+  public static final int NANOSHOCK = 27040;
+  public static final int AUDIOCLASM = 27041;
+  public static final int SYSTEM_SWEEP = 27042;
+  public static final int DOUBLE_NANOVISION = 27043;
+  public static final int INFINITE_LOOP = 27044;
+  public static final int PHOTONIC_SHROUD = 27045;
+  public static final int UNUSED_COMBAT_SELFBUFF = 27046;
+  public static final int STEAM_MYCELIA = 27047;
+  public static final int SNOW_COOLING_System = 27048;
```
midgleyc, line 505:
```suggestion
  public static final int SNOW_COOLING_SYSTEM = 27048;
```

midgleyc (changes requested):

src/net/sourceforge/kolmafia/objectpool/SkillPool.java
Veracity0: Right. Done.

midgleyc (dismissed):

PR #644, opened 2022-03-05
Refactor into GreyYouManager and add absorptions command

Veracity0:
I created GreyYouManager and moved the GooSkill data structure and data into it from GooSkillsCommand.
Added GooAbsorption data structure and added the various non-skill absorptions.
Created "absorptions" command:

absorptions

lists everything, including skills, sorted by zone in which the monster appears.

absorptions needed

lists only things you have not absorbed

absorptions skills
absorptions adventures (or advs)
absorptions muscle (or mus)
absorptions mysticality (or myst or mys)
absorptions moxie (or mox)
absorptions maxmp
absorptions maxhp

(optionally with needed) list only the absorptions of the specified type

gooskills is still superior for looking at skills, since it gives the enchantments or combat effects of the skill.
This command simply gives the skill name.


src/net/sourceforge/kolmafia/session/GreyYouManager.java
```diff
@@ -0,0 +1,736 @@
...
+
+  public static void refreshAbsorptions() {
+    CharSheetRequest request = new CharSheetRequest();
+    request.run();
+  }
+
+  // Absorbed 5 adventures from a warwelf.<!-- 199 -->
+  private static final Pattern ABSORPTION_PATTERN =
+      Pattern.compile(
+          "Absorbed (.*?) from (?: an?| the |)(.*?)\\.<!-- ([\\d]+) -->", Pattern.DOTALL);
```
midgleyc, line 39: seeing as you don't use the displayed name at all, I feel like you could merge the pronoun with the name (as in, just match it all under the same `(.*?)`).

This would avoid a potential issue where a monster has a strange pronoun. Also I think you have a potential double space issue with the pronouns as-is (it would only match "from \<space> \<space> the", for example).

midgleyc (dismissed):
One comment, non-blocking.

Also I think a test to confirm character sheet parsing works would be nice (again non-blocking).

src/net/sourceforge/kolmafia/session/GreyYouManager.java
Veracity0: I fixed that. And I added tests to make sure I found it and added other error checks which revealed that I had a fair number of erroneous monster names. Tests help. ;)

Veracity0:
I now parse absorptions and goo skills right from the CharSheet.
If you learn a new goo skill after battle, we update the set of learned skills.
There are tests for parsing the charsheet!

For now, we don't recognize significant absorptions after battle, but the absorptions command updates from the charsheet, so it will still be up to date.

midgleyc (approved):

PR #646, opened 2022-03-07
register absorptions after a won fight

Veracity0:
Minor cleanups.

1) When in Grey You, after a fight which you win, register the monster absorption, as needed.
2) In the following cases, log a message both to gCLI and session log:
- Learned an unknown Grey You skill
- A Grey Goo "passive" skill does not have an entry in modifiers.txt 
- Found an unknown "significant" Grey Goo absorption on the charsheet

These all require a KoLmafia fix.

3) Don't need to refresh absorptions in "absorptions" command, since we read them from char sheet and update them appropriately when we absorb a monster.
4) When you try to visit a zone which is too old (per Standard restrictions), recognize that and don't count it as a visited location.
5) When free king after Grey You, you go through a choice adventure to choose a new class. Handle it in the usual way for Avatar paths.
6) Grey You ascensions tracked in greyYouPoints

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -669,6 +669,7 @@ user	goreCollected	0
 user	gourdItemCount	5
 user	grabCloversHardcore	false
 user	grabCloversSoftcore	true
+user	greyYouPoints	0
 user	grimoire1Summons	0
 user	grimoire2Summons	0
 user	grimoire3Summons	0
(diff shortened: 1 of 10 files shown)


midgleyc (dismissed):
Looks good to me.

A note for future reviewers: if you click on the file and press "a", or click the "..." next to "Viewed" and uncheck "Show annotations", the notes about coverage all vanish and the code is easier to read.

PR #648, opened 2022-03-08
Things I noticed while looking at potions in You, Robot

Veracity0:
I did a single Hardcore run of You, Robot a year ago,  back when it was new. Before the potted power plant. It took me a week and I didn't like it much. But, now it is time to finish off the factoids, so here I am - and it's going a lot smoother now. I just defeated the Sorceress on day 3 of my Hardcore run. I haven't freed the King yet, but will do so after I sleep, assuming I don't come up with something to try out which has to happen before I ascend again.

Here are fixes for some things I noticed which confused me and/or slowed me down.

I can see potions. And I can try using them in the Item Manager and nothing happens. If I try it in the gCLI, it says it failed because of a "daily limit". If I use one in the Relay Browser, KoL tells me I don't know where to put it and does not consume it - but KoLmafia decrements the inventory count, until I refresh inventory.

I cannot eat, drink, or spleen, but there's no indication of this in the Item Manager.

1) UseItemRequest correctly sets a max use count of 0, without the CPU upgrade, but doesn't give a reason. Fix the reason to be "lack of necessary CPU upgrade".
2) When parsing the response to trying to use a potion - including from the Relay Browser or a hand-crafted URL - notice the failure message and actually report "You need the Biomass Processing Function CPU upgrade to use potions."
3) I souped up the Item Manager so that the food/booze/spleen/potion panels disable the "consume" button if you can't eat/drink/spleen/use potions.
4) I did NOT grey out the "enqueue" button, thinking you could use the "create" button up top, if you really wanted to, but you actually can't enqueue, since ItemManagePanel.getDesiredItemAmount will set not let you choose items you can't actually consume - even though there is a "create" button. That's an existing bug. I did not change it. It needs thought.
5) I added a lot of comments to ScrapHeapRequest to document what the body part options and cpu upgrades are. We currently use them strictly via KoL request parameters - "4" for the fourth Left Arm option (Vice Grips), "robot-shirt" for the Topology Grid option, and so on. Unfortunately, we expose those integers and keywords right in the preferences. That seems non-user-friendly. Not to mention coder-friendly; we have Magic Numbers all over the place referring to body parts. I did not change anything about that - yet - but now it's documented to make it easier.
6) When I was looking at potions, the code reminded me that the SythensizePanel and the CandyDatabase want to do mass searches on candies (many of which are potions). There was a big comment in the SynthesizePanel descriping this - and then calling the CandyDatabase to do the work - in a different way. I moved the comment from one file to the other and corrected it.
7) While investigating that, I notice that ASH has a proxy field saying if an item is candy, but not if it is a potion. It now has boolean item.potion
8) When parsing the CharSheet in You, Robot, we now fetch all five of the images for the avatar. The DailyDeeds frame (which displays your avatar) now knows how to overlay multiple images to make a composite. I currently have:
otherimages/robot/left4.png
otherimages/robot/right4.png
otherimages/robot/top4.png
otherimages/robot/bottom4.png
otherimages/robot/body6.png
9) Bug fix: property "adding" scrolls take an integer from 0 to 3. The default is -1. This is not valid for a property which is configurable via the GUI; when creating the dropdown, the current value is used to choose the index into the dropdown - which is 0-indexed. The default is now 1, which is probably what people want: if you set a condition to make a particular scroll, automate it, otherwise punt to your CCS.
10) When you use your spinal-fluid-covered emotion chip, you learn the Emotionally Chipped skill.
11) I started refactoring by moving a bunch of existing code into YouRobotManager
12) And I added 12 HTML responseText files to enable testing.

I noticed that ScrapheapRequest will parse changes to Robot parts - 12 of which grant a combat skill - and updates various preferences, but it does not update available combat skills. It should. Next PR.


Veracity0:
I uploaded the following HTML responses to various actions in You, Robot:

test/root/request/test_scrapheap_charpane.html
--- Refreshing the charpane. We parse the Avatar, Energy, and Scrap

test/root/request/test_scrapheap_charsheet.html
--- Visiting the charsheet. We parse the Avatar.

test/root/request/test_scrapheap_reassembly_station.html
--- Visiting the Reassembly Station at the scrapheap

test/root/request/test_scrapheap_show_cpus.html
--- Asking to see the CPU upgrade options

test/root/request/test_scrapheap_cpu_upgrade.html
--- Buying a CPU upgrade for 40 energy (Weather Control Algorithms -> Resist All +2)

test/root/request/test_scrapheap_show_top.html
--- Asking to see the Top equipment options

test/root/request/test_scrapheap_add_skill.html
-- Buying a combat-skill granting item for 5 scrap (Pea Shooter -> +Shoot Pea)

test/root/request/test_scrapheap_add_skill_lose_skill.html
-- Buying a combat-skill granting item for 30 scrap (Tesla Blaster -> -Shoot Pea, +Tesla Blast)

test/root/request/test_scrapheap_lose_skill.html
-- Buying a non-combat-skill granting item for 5 scrap (Solar Panel -> -Tesla Blast, Energy +1)

I haven't started writing tests yet - and may not, in this PR - but this should cover the tests I am envisioning.
Now I can ascend and not worry about having to go make some more DEBUG logs before I start working on tests. :)

Veracity0:
I need to move on. This was all a result of my first (of five) You, Robot runs.

I saw a bunch of stuff which wasn't quite right. I fixed it.
I studied the existing code in detail and am itching to refactor and fix it.
I did some of that - I moved all the internals into a new YouRobotManager and made the rest of KoLmafia use methods from there, rather than knowing about those internals.

I want tests for YouRobotManager.
I want to refactor it to not depend on user-visible preferences.
I want to make the internals not depend so much on magic numbers right from KoL.
I'd like to clean up the user-visible preferences, but existing scripts use the magic numbers and such. It's not a user-friendly API, but changing it now is not backwards compatible. One the internals are cleaned up, we can consider it, after public discussion.

I just ascended into another You, Robot run and am ready to go with all of that.

I want that to be in a new PR. This one has been waiting for review for two days. No apparent interest, so far. I'm going to skip approvals and merge it. Further development will be in a new (draft) PR which I will open almost immediately.

midgleyc:
I was waiting for changes to stop being regularly made to this PR before I started the review. I can review it as you go along next time, if you'd prefer.

About my only thought was on
```
private static String[] avatar = new String[0];
```

I thought this could be an empty `List` instead of an array, because it's not of fixed length (as in - sometimes it is one, sometimes more) and is always overridden in entirety (replaced), so I didn't think there was any benefit to it being an array.

jaadams5:
I too was watching and started a review, didn't finish it and found more changes when I returned.  I elected to wait.

Veracity0:
Well, that's fair.

I made my new PR a draft - so you can expect it to change a lot.
I seem do be doing 3-day Hardcore runs, so I will continue to improve while in-run, and then when I free the king (and can no longer-test in-path), I will mark it ready for review.

(Although, if your review comments expect me to improve - and test, in-path - oops.)

I like the idea of the avatar being a List, not an Array.
My rationale for my initial implementation is that avatar images correspond to "layers " - each of which is a JLabel - and I am uneasy about adding and deleteing layers on the fly. It may work!

There are other uses for this technology.
- Melodramadery uses multiple images - but they are concatenated horizontally.
- Left-Hand Man uses multiple images (overlays item image on top of headless familiar) - concatenated vertically.
I am expecting, eventually, to use this technology elsewhere.

But, yes - for familar images, we currently have either 0 or 5 images, but there is really no reason to assume that.

Thank! I'll add this to my to-do list.

PR #652, opened 2022-03-12
refactor: remove always false logic in GenericRequest

midgleyc:
`processOnFailure()` always returns false, so don't check it.


midgleyc:
okay, used in subclass.

PR #653, opened 2022-03-12
Use HttpClient in GenericRequest

midgleyc:
Finish off https://kolmafia.us/threads/migration-from-http-1-1-to-http-2-0.27303/.

Can log in and navigate around in the relay browser, but could do with more testing. Some errors had to be moved because e.g. IO exception on request / response is now one call.

Really not sure whether proxy will continue to work (using the system properties). Looking at the code, not sure it ever worked, though: it sets "http.proxyUser" and "http.proxyPassword" which don't seem to be used.

Tests not moved over yet but the way they're written they don't actually call the connection functions, which is not great for porting. Plan to migrate if reasonable.


midgleyc:
Also not sure exactly where the 504 logic should be. I know I've hit it before. In HttpClient, an IOException means you don't have a response object which means you don't have a response code you can read. This means either:
* the failure was in converting the gzip istream, and the new code doesn't handle that
* the new code doesn't throw an exception at all, and the logic is in the right place

midgleyc:
Migrated tests, ran a full session with this and everything went fine.

libraryaddict:
This is amazing work!

jaadams5 (approved):
Placing a lot of faith in

> Migrated tests, ran a full session with this and everything went fine.

:-)

midgleyc:
The test migration gives me practically no confidence (not running in real conditions), but running this for a session and having it work was great!

I wasn't able to set up a proxy, but I can confirm it crashes in the same way. If it breaks I expect users will report, and then I'll have somebody who can walk me through how to set up a proxy.

Once this gets merged, it'll see a lot more usage, and I'll be on the lookout for any bugs arising out of this.

PR #655, opened 2022-03-14
When calculating HP, Muscle has no effect for Robots

Veracity0:
1) When calculating HP, you have a (supposedly) class specific base HP value and only items and effects that provide Maximum HP boosts affect it. The maximize didn't understand that, so wanted to recommend things that provide a Muscle boost.
For my AT Robot, my base HP is 30; "modtrace Maximum HP" says I currently have +295 worth of Maximum HP boosts, and +66% Maximum HP Percent, and my current Maximum HP is 345. 
The BUFFED_HP calculation in Modifiers already had special handling for Vampyres. This PR adds special handling for Robots.
2) If you cannot equip hats, no Mohawk wig warning in Relay Browser
3) When available equipment slots change, refresh Equipment lists so Gear Changer updates.
4) When available equipment slots change, refresh standard outfits so Gear Changer updates.
5) Moved ConcoctionType from Concoction to ConcoctionDatabase, since it pertains to how GUI elements interact with the latter package, as opposed to the former. (I did that refactoring in another PR first, since I want to make UseItemEnqueuePanel and UseItemDequeuePanel listen for (potions) - If it is a ConcoctionType.POTION panel - and I got annoyed with looking at the cruddy code which I was going to modify.)
6) When your robot acquires the CPU upgrade to use potions, fire (potions).
UseItemEnqueuePanel and UseItemDequeuePanel listen for (potions) and enable the "consume" button as appropriate.
I wrote a test which confirms that we fire that listener and the consume buttons get enabled.
7) UseItem{Enqueue,Dequeue}Panel now registers a listener for all types: (food), (booze), (spleen), (potions), as appropriate. These could all be used when you liberate the king.
8) I wrote a test that confirms that (avatar) listeners only fire if the avatar actually changes.
9) When you ascend and continue in the same session, the info from get_campground() is now reset.
10) Robots CAN eat - size-0 magical sausages - but have no fullness. This change allows magical sausages to be created and eaten from the Foods panel pf the Item Manager.
11) Per midgleyc's earlier code review, the avatar is no longer a String[]. It is a List<String>.
12) Now that all the GUI consumption panels are listening for signals when consumption status changes, when you free the king and you no longer have any path restrictions, poke all those listeners.
13) Since api.php doesn't have energy and scrap, we used charpane in its place. Unfortunately, charpane doesn't have MP. This meant that we completely failed to track MP changes - changing equipment, casting any of the (few) spells that are still available in this path, and so forth. I did this:
- We use api.php.
- At login, we look at charpane.php to get current scrap & energy
- SCRAP and ENERGY are now AdventureResults and we parse gaining them after battles.
- YouRobotManager and ScrapheapRequest now account for energy or scrap costs for all the things you can do in the Scrapheap.
14) 9 of the attachments grant combat skills that cost 1 scrap or 1 energy to use. Account for that when processing a fight.
15) When you use Statbot, update equipment lists so that newly equipable items are shown.
-------------

Here are additional things I'd like to know, but do not intend to spade myself.

- What is base HP for Robot Seal Clubbers, Turtle Tamers, Pastamancers, Saucerors, and Disco Bandits?
- What is the failure message if you are out of scrap or energy and try to use a combat skill that requires it?


Veracity0:
Having completed four Hardcore You, Robot ascensions, this fixes every issue, annoyance, and bug I encountered along the way. I have one more run to go to get my last factoid, which I will start tomorrow.

I do confess to surprise that my "Free King" warning did not fire. But, there is literally nothing I can do to debug that until I am about to free the king again - next run. So, not QUITE right.

But, let's get this in now, please. Thanks.

Code review issues and suggestions welcome, as always. :)

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -1504,6 +1504,11 @@ private boolean sendMohawkWigWarning() {
       return false;
     }
 
+    // If they can't equip Wig, no problem
```
midgleyc, line 1507: As a minor note, I would prefer to see a warning that an item I temporarily can't equip could speed up a zone. I could then respond to that by deciding to be able to equip hats, for example.

midgleyc (approved):
One comment, not blocking.

Nicely commented code :)

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -578,6 +592,14 @@ public static final AdventureResult parseResult(final String s) {
         statname = parsedGain.nextToken();
       }
 
+      if (statname.startsWith("energy") || statname.startsWith("Energy")) {
+        return new AdventureResult(AdventureResult.ENERGY, (int) modifier);
```
jaadams5, line 596: Non-blocking comment.  This could be one clause - statname.toLowerCase().startsWith("energy")

Similar comment for scrap.

jaadams5 (approved):

src/net/sourceforge/kolmafia/request/RelayRequest.java
Veracity0: That's an interesting idea. Note that there are several reasons one might not be able to equip the item:

- Moxie less than 55 (Not a problem for my AT at level 10, but in You, Robot, one's off-stats will heavily lag one's main stat unless you intentionally spend energy to improve them.)
- You don't have the Mannequin Head equipped - something you can fix with 15 scrap.

Along the same line, one might think that the warning for equipping an antique machete for fighting lianas would have the same issue. Except, I notice that sendMacheteWarning doesn't even check whether you can equip the machete - which is why I learned to make a point up upgrading muscle to at least 62 before starting to open the zones in the Hidden City.

I'm now thinking, why not write a unit test package for RelayRequest "warnings"? That will allow me to work on your suggestion without using, say, debug prints and such. Also my "free king warning" issue.

OK. I'll work on that. Next PR. :)

src/net/sourceforge/kolmafia/AdventureResult.java
Veracity0: OK.

PR #656, opened 2022-03-15
Simplify Panels that handle ConcoctionDatabase.usables

Veracity0:
Refactor GUI panels to have a ConcoctionType enum rather than three mutually exclusive booleans

UseItemEnqueuePanel and UseItemDequeuePanel handle the list of "usables" from ConcoctionDatabase.
Each such panel can (mutually exclusively) be {food, booze, spleen, potions}.

They - and ConctionDatabase - are full of methods like this:

method(boolean food, boolean booze, boolean spleen)

FOOD = true, false, false
BOOZE = false, true, false
SPLEEN = false, false, true
POTION = false, false, false

I made an enum:
```
  public static enum ConcoctionType {
    FOOD,
    BOOZE,
    SPLEEN,
    POTION;
  }
```
and did a mass refactoring for all those methods to simply take one of these as an argument.

I have no idea how to write unit tests for GUI classes.

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/Concoction.java
+++ b/src/net/sourceforge/kolmafia/objectpool/Concoction.java
@@ -29,6 +29,14 @@ import net.sourceforge.kolmafia.utilities.StringUtilities;
  * actually make the item.
  */
 public class Concoction implements Comparable<Concoction> {
+  // For queueing
+  public static enum ConcoctionType {
+    FOOD,
+    BOOZE,
+    SPLEEN,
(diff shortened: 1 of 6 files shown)


jaadams5 (approved):
I don't really know how to test Swing components either.

midgleyc:
I put in a test for `DailyDeedsPanel` in #610: find a function you want to test and use or expose a function that lets you test it.

So ignore it's a GUI component and test it as a class, assuming it has meaningful algorithms.

PR #658, opened 2022-03-15
tests: simplify some string utilities tests

midgleyc:
Some comments I had on #616, implemented.

* replace char[] with String
* remove the explicit catch-failure on exceptions, and just let them be thrown to fail the test
* use `List.of` for expected


jaadams5:
> Some comments I had on #616, implemented.
> 
> * replace char[] with String

I deliberately did not do that because "your way" on my system created a failing test and reduced coverage.  Since this doesn't do either of those, I must have misunderstood "your way".

> * remove the explicit catch-failure on exceptions, and just let them be thrown to fail the test

I chose not to do this because I think the other implementation was easier to quickly read and understand what the test was, and was not doing, but not worth arguing over a preference.

> * use `List.of` for expected

Whoops.  I either missed some or something got lost in conflict resolution.  Thanks.

jaadams5 (approved):

PR #662, opened 2022-03-18
Fix maximizer behavior with force-equipping empty bjorns (and crowns)

horrible-little-slime:
For a long time, I've encountered a weird issue involving the buddy bjorn and the maximizer that I've had a hard time describing and duplicating. No longer!

I've also had issues when assigning a bonus value to the buddy bjorn rather than using +equip, although I had a much harder time writing a failing test for that.

### DIFF
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
@@ -8,9 +8,12 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.beans.Transient;
+
 import net.sourceforge.kolmafia.FamiliarData;
 import net.sourceforge.kolmafia.KoLCharacter;
 import net.sourceforge.kolmafia.objectpool.FamiliarPool;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.session.EquipmentManager;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -107,6 +110,15 @@ public class MaximizerRegressionTest {
     assertEquals(25, modFor("Buffed Muscle"), 0.01);
     // Actually equipped the buddy bjorn
     assertEquals(25, modFor("Meat Drop"), 0.01);
+  }
+
+  @Test
+  public void equipBjornWhenEmpty() {
+    equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
+    KoLCharacter.addFamiliar(new FamiliarData(FamiliarPool.MOSQUITO));
+    KoLCharacter.setBjorned(FamiliarData.NO_FAMILIAR);
+    
+    assertTrue(maximize("item, -buddy-bjorn, +equip buddy bjorn"));
   }
 
   // https://kolmafia.us/threads/27073/
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
@@ -113,12 +113,15 @@ public class MaximizerRegressionTest {
   }
 
   @Test
+  @Disabled("Refuses to equip a bjorn without a familiar in it, even when the equip keyword is used.")
   public void equipEmptyBjornNoSlot() {
     equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
     KoLCharacter.addFamiliar(new FamiliarData(FamiliarPool.MOSQUITO));
     KoLCharacter.setBjorned(FamiliarData.NO_FAMILIAR);
     
-    assertTrue(maximize("item, -buddy-bjorn, +equip buddy bjorn"));
+    // Here, buddy-bjorn refers to the slot, while Buddy Bjorn refers to the item associated with that slot.
+    // We've earlier forced that slot to be empty, and here we're asking mafia not to fill it when maximizing, but to still equip the bjorn.
+    assertTrue(maximize("item, -buddy-bjorn, +equip Buddy Bjorn"));
   }
 
   // https://kolmafia.us/threads/27073/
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
```diff
@@ -109,6 +112,15 @@ public void keepsBonusBjornUnchanged() {
     assertEquals(25, modFor("Meat Drop"), 0.01);
   }
 
+  @Test
+  public void equipEmptyBjornNoSlot() {
```
jaadams5, line 116: If we agree that this is a good test and the fact that is fails indicates an error to be fixed, then we should probably add the Disabled annotation here with a comment to that effect.

test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
```diff
@@ -109,6 +112,15 @@ public void keepsBonusBjornUnchanged() {
...
+    equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
+    KoLCharacter.addFamiliar(new FamiliarData(FamiliarPool.MOSQUITO));
+    KoLCharacter.setBjorned(FamiliarData.NO_FAMILIAR);
+    
+    assertTrue(maximize("item, -buddy-bjorn, +equip buddy bjorn"));
```
jaadams5, line 121: I am lazy and could look things up myself but are "Buddy Bjorn", "buddy-bjorn" and "buddy bjprn" all supposed to be the same item?  Perhaps using the canonical name in all three places would be appropriate?

jaadams5 (changes requested):
I'll add the annotation and approve once I understand the name issue.  Or you can add the annotation and I'll approve.  Thank you.

test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
horrible-little-slime: `buddy-bjorn` refers to a slot, rather than an item, which is a good shout; I'll be consistent about the capitalization across the other two, and add a comment to that effect. Here, what's happening is that the buddy-bjorn slot is empty, and the maximizer is told not to fill that slot, so it refuses to equip the item entirely. Maybe this has to do with $familiar[none] not being the same as null, which I assume is what it defaults to. I can investigate it further, but the maximizer code is pretty labyrinthine to me.

horrible-little-slime: I think "buddy bjprn" is a typo on your end because I do not see it in my code, but I am not 1000% confident of that.

horrible-little-slime:
I will fix whatever is failing about this in the morning.

test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
jaadams5: Yes, typo.  The last parameter in "assertTrue(maximize("item, -buddy-bjorn, +equip buddy bjorn"));" since the name of the item was capitalized elsewhere.

"Here, what's happening is that the buddy-bjorn slot is empty, and the maximizer is told not to fill that slot, so it refuses to equip the item entirely" that sounds to me like correct behaviour.  I wonder if there is some "confusion" between using the back slot for the bjorn and changing the familiar in a bjorn that is already equipped?  Regardless I look forward to your next attempt.  Thank you.

horrible-little-slime:
I didn't mean to close this, and must have hit the wrong button when going to bed. Whoops!

jaadams5:
Well it's back opened again so no harm, no foul.

test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
horrible-little-slime: I'd like mafia to be willing to equip an empty bjorn because I plan on manually filling it myself. Because there's not a way to assign bonus values (or force-equips) to particular bjornified familiars, there aren't any other great wolutions. In particular, I don't want to constantly seesaw between mafia bjornifying a familiar and me replacing it basically once a turn, because that's two unnecessary server hits per adventure.

It feels like kind of a frivolous thing, but that's enough to make me feel like this is probably bad behavior--just possibly low-priority bad behavior. I've adjusted the capitalization, added comments, and disabled the test, so I'll undraft this now.

midgleyc:
This only happens if no familiar is assigned to the bjorn when you do `-buddy-bjorn`, right? Due to the code starting line 387 in `MaximizerSpeculation.java`.

The easiest fix for exactly this case is to change that line to `if (useBjornFamiliar != FamiliarData.NO_FAMILIAR || item.requiredFlag) {`.

horrible-little-slime:
> This only happens if no familiar is assigned to the bjorn when you do `-buddy-bjorn`, right? Due to the code starting line 387 in `MaximizerSpeculation.java`.
> 
> The easiest fix for exactly this case is to change that line to `if (useBjornFamiliar != FamiliarData.NO_FAMILIAR || item.requiredFlag) {`.

I've also encountered this in situations where I assign a large bonus to the bjorn, rather than making it a requirement. That being said, this does seem to stop this test from failing without making any others fail as well. Thanks!

midgleyc (changes requested):
To give more detail here:
* `-buddy-bjorn` says "don't change the Bjorn's familiar". There is already some code specifically dealing with this case
* if you can equip a familiar that contributes to the maximization string, or that contributes to the tiebreaker, `bonus` or `equip` should be sufficient to favour the bjorn
* if you have no familiar, and no bjornable familiars you hit the error case we see here
* if you have a familiar which does not contribute to the maximization string or the tiebreaker, then perhaps the initial check for "is this item relevant" says it isn't, and then "bonus" fails to be applied at all?

Even though the PR title says "and crowns", you've not fixed the crown here. Could you make an equivalent test for the crown, and fix it in presumably the same way?

jaadams5:
I feel like this is a feature request and not a bug.  As I understand it, the desire is to force an empty bjorn to be equipped.  This is not easily done because an empty bjorn is almost never going to be chosen in comparison to a full one.  Similarly an empty bjorn is unlikely to be a preferred back item.  My very quick and superficial look at the code suggests that the distinction between forcing or equipping a bjorn in the back slot and equipping a familiar in the bjorn slot, if a bjorn is equipped is not made.  Perhaps it should be.

midgleyc (approved):

midgleyc:
Fixed this myself. I see that there was a commit fixing this for the crown, but not adding the test, which I just did.

jaadams5 (approved):

horrible-little-slime:
Hey y'all, thanks for taking over on this; I became very busy at work this week and had to step away. Great work!

PR #665, opened 2022-03-20
Do not finish Q5 with Goblin Water unless in Heavy Rains

BadHorseMonkey:
Mr. Cheeng's spectacles can cause Goblin Water to drop, causing the ResultProcessor to finish the Goblin King quest.  Without fixing it comepletely, I closed the hole for all other paths.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ResultProcessor.java
+++ b/src/net/sourceforge/kolmafia/session/ResultProcessor.java
@@ -2066,7 +2066,8 @@ public class ResultProcessor {
         break;
 
       case ItemPool.GOBLIN_WATER:
-        if (adventureResults) {
+        if (adventureResults
+            && KoLCharacter.inRaincore()) { // because you can now get the Goblin Water otherwise...
           QuestDatabase.setQuestProgress(Quest.GOBLIN, QuestDatabase.FINISHED);
         }
         break;


src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -2066,7 +2066,8 @@ private static void gainItem(boolean adventureResults, AdventureResult result) {
         break;
 
       case ItemPool.GOBLIN_WATER:
-        if (adventureResults) {
+        if (adventureResults
+            && KoLCharacter.inRaincore()) { // because you can now get the Goblin Water otherwise...
```
Veracity0, line 2070: What if you are wearing Mr. Cheeng's spectacles in Heavy Rains? :)

BadHorseMonkey: Then it would incorrectly mark the quest completed.

If I were to solve for that case, I would have to think about a better solution for longer than I thought about this one. 

The reports of the problem were coming from Casual, and I thought the hole would be at least smaller.  I would happily implement a better solution.

midgleyc (approved):
I think this is a good incremental improvement to a bug seen in the wild.

PR #666, opened 2022-03-20
chore: lint generic request

midgleyc:
Clean up the easy warnings IntelliJ warns about.


jaadams5 (approved):
I have never felt comfortable enough to accept suggested changes to a regex :-)

midgleyc:
In this particular case, `/` never needs to be escaped because it's not a special character, so this is an easy one ;) (and `\/` isn't anything special).

PR #671, opened 2022-03-21
Initial support Grey Goose and combat skills

Veracity0:
If your Grey Goose weighs at least (base) 6 pounds, combat skills are available.

- Re-Process Matter is available only in Grey You. Using it resets the Grey Goose to 1 pounds.
- Meatify Matter is available once per day. Using it resets the Grey Goose to 5 pounds.
- Emit Matter Duplicating Drones, Convert Matter to Protein, Convert Matter to Energy, and Convert Matter to Pomade have no daily limit, although they reset the Grey Goose to 5 pounds.

1) When you take a Grey Goose out of the Terrarium, if it is at least 6 pounds, add combat skills, as appropriate
2) When you put Grey Goose back in the Terrarium, if it is at least 6 pounds, remove combat skills
3) When your Grey Goose gains weight from 5 to 6 or more, add combat skills.
4) When your Grey Goose loses weight to less than 6 pounds, lose combat skills
5) Added tests for FamiliarData

### DIFF
--- a/src/net/sourceforge/kolmafia/FamiliarData.java
+++ b/src/net/sourceforge/kolmafia/FamiliarData.java
@@ -15,6 +15,7 @@ import javax.swing.SwingConstants;
 import net.java.dev.spellcast.utilities.JComponentUtilities;
 import net.sourceforge.kolmafia.objectpool.FamiliarPool;
 import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.objectpool.SkillPool;
 import net.sourceforge.kolmafia.persistence.EquipmentDatabase;
 import net.sourceforge.kolmafia.persistence.FamiliarDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
@@ -188,6 +189,7 @@ public class FamiliarData implements Comparable<FamiliarData> {
   private boolean favorite;
   private int charges;
   private int pokeLevel;
+  private boolean active = false;
 
   public FamiliarData(final int id) {
     this(id, "", 1, EquipmentRequest.UNEQUIP);
@@ -344,12 +346,11 @@ public class FamiliarData implements Comparable<FamiliarData> {
   }
 
   public final void loseExperience(int exp) {
-    setExperience(this.experience - exp);
-    setWeight();
+    this.setExperience(this.experience - exp);
   }
 
   public final void loseExperience() {
-    loseExperience(this.experience);
+    this.loseExperience(this.experience);
   }
 
   public final int determineTestTeachExperience() {
@@ -441,7 +442,8 @@ public class FamiliarData implements Comparable<FamiliarData> {
   }
 
   private void setWeight() {
-    this.weight = Math.max(Math.min(this.getMaxBaseWeight(), (int) Math.sqrt(this.experience)), 1);
+    this.setWeight(
+        Math.max(Math.min(this.getMaxBaseWeight(), (int) Math.sqrt(this.experience)), 1));
   }
 
   public final void checkWeight(final int weight, final boolean feasted) {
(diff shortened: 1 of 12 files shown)


src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -783,6 +847,10 @@ public String getRace() {
     return this.race;
   }
 
+  public boolean getActive() {
```
midgleyc, line 850: Naming conventions elsewhere would have this as `isActive` as it's a boolean, e.g. https://stackoverflow.com/questions/5322648/for-a-boolean-field-what-is-the-naming-convention-for-its-getter-setter#5322666

I don't know what we do in the rest of the codebase, so this is just a thought.

src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -617,6 +619,67 @@ public boolean getFeasted() {
...
+  }
+
+  private void changeWeight(int oldWeight, int newWeight) {
+    if (oldWeight == newWeight) {
+      return;
+    }
+    switch (this.id) {
+      case FamiliarPool.GREY_GOOSE:
+        if (this.active) {
+          if (oldWeight < 6 && newWeight >= 6) {
```
midgleyc, line 651: `addAvailableCombatSkill` and the `remove` version seem idempotent to me, so you could do without the `oldWeight` checks. I'm not sure whether, overall, that's better or worse.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9554,6 +9560,48 @@ private static void payActionCost(final String responseText) {
...
+        }
+        break;
+
+      case SkillPool.MEATIFY_MATTER:
+        // NAME detaches a big chunk of itself, slaps it onto your opponent's
+        // torso, and atomically reconfigures the whole mess into Meat.
+        if (responseText.contains("reconfigures the whole mess into Meat")) {
+          // This skill can only be used once per day.
+          Preferences.setBoolean("_meatifyMatterUsed", true);
+          KoLCharacter.addAvailableCombatSkill(SkillPool.MEATIFY_MATTER);
```
midgleyc, line 9580: perhaps you meant remove? Despite that, I don't think this is necessary: when the familiar weight is reset to 1, the skill will disappear anyway, right?

Veracity0: Right on both accounts.

src/net/sourceforge/kolmafia/FamiliarData.java
Veracity0: I agree.

src/net/sourceforge/kolmafia/request/FightRequest.java
Veracity0: (The thinking was that since it's a once-per day skill, explicitly make it unavailable, rather than having it become unavailable as a side effect of the weight change. But the comment will suffice. :)

src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -641,16 +641,14 @@ public void activate() {
...
+  public void setWeight(final int weight) {
+    this.weight = weight;
     switch (this.id) {
       case FamiliarPool.GREY_GOOSE:
         if (this.active) {
-          if (oldWeight < 6 && newWeight >= 6) {
+          if (weight >= 6) {
             addGreyGooseSkills();
-          } else if (oldWeight >= 6 && newWeight < 6) {
+          } else if (weight < 6) {
```
midgleyc, line 651: tiny comment: if weight is not >= 6, it is definitely less than 6 (right? there's no int NaN?), so you can do without the second "if".

midgleyc (dismissed):

src/net/sourceforge/kolmafia/FamiliarData.java
Veracity0: Fixed

midgleyc (approved):

PR #672, opened 2022-03-22
rationalize availableCombatSkills

Veracity0:
KoLConstants.availableCombatSkills was a LockableListModel of UseSkillRequests.
It was never used in a JList, so need not be a ListModel.
Combat skills cannot be cast using a UseSkillRequest.
Its ONLY use is to pass the list of skillIds from the fight.php skill dropdown to StationaryButtonDecorator.
KoLConstants.availableCombatSkillsMap was an IdentityMap of skillIds.
KoLConstants.availableSkillsMap was an IdentityMap of skillIds.
They could simply be sets.
KoLCharacter.hasSkill is ONLY used to look up skills in availableSkills.
KoLCharacter.availableCombatSkills is ONLY used to look up combat skills.
In both cases, UseSkillRequests were generated and passed to a method which looked up skillIds in the skill Maps.

1) KoLConstants.availableSkillsMap is now KoLConstants.availableSkillsSet
KoLConstants.availableCombatSkillsMap is now KoLConstants.availableCombatSkillsSet
2) KoLCharacter.hasSkill and KoLCharacter.availableCombatSkill now only have variants that take an int skillId or a String skillName. They go straight to the appropriate set of skillIds
3) KoLCharacter.addAvailableSkill and KoLCharacter.removeAvailableSkill still have variants which takes a UseSkillRequest, since KoLConstants.availableSkills is still a list of such.
4) KoLCharacter.addAvailableCombatSkill and KoLCharacter.removeAvailableCombatSkill now only add or remove a value to KoLConstants.avaliableCombatSkillSet. No need to manipulate KoLConstants.availableCombatSkillList.
5) FightRequest and StationaryButtonDecorator manipulate KoLConstants.availableCombatSkillList, whose only purpose is to communicate between those two modules.
6) The handful of places in KoLmafia which looked directly at KoLConstants.availableSkills now call KoLCharacter.hasSkill.
7) The handful of places in KoLmafia which create a UseSkillRequest in order to call KoLCharacter.hasSkill now call the variant with a skillId. Some of them no longer actually need a UseSkillRequest. Some - like the Maximizer - want to maintain a list of UseSkillRequests as a way to cache a skillName/skillId pair. I did not refactor that code, although I find the Maximizer's use of UseSkillRequests ONLY to cache Passive skills not exactly clean.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -3905,9 +3905,7 @@ public abstract class KoLCharacter {
 
   /** Adds a single skill to the list of skills temporarily possessed by this character. */
   public static final void addAvailableCombatSkill(final int skillId) {
-    if (!KoLConstants.availableCombatSkillsSet.contains(skillId)) {
-      KoLConstants.availableCombatSkillsSet.add(skillId);
-    }
+    KoLConstants.availableCombatSkillsSet.add(skillId);
   }
 
   public static final void addAvailableCombatSkill(final String skillName) {
@@ -3916,9 +3914,7 @@ public abstract class KoLCharacter {
   }
 
   public static final void removeAvailableCombatSkill(final int skillId) {
-    if (KoLConstants.availableCombatSkillsSet.contains(skillId)) {
-      KoLConstants.availableCombatSkillsSet.remove(skillId);
-    }
+    KoLConstants.availableCombatSkillsSet.remove(skillId);
   }
 
   public static final void removeAvailableCombatSkill(final String skillName) {
@@ -4068,16 +4064,8 @@ public abstract class KoLCharacter {
   }
 
   /** Utility methods which looks up whether or not the character has a particular skill. */
-  public static final boolean hasAvailableSkill(final int skillId) {
-    return KoLConstants.availableSkillsSet.contains(skillId);
-  }
-
-  public static final boolean hasCombatSkill(final int skillId) {
-    return KoLConstants.availableCombatSkillsSet.contains(skillId);
-  }
-
   public static final boolean hasSkill(final int skillId) {
-    return hasAvailableSkill(skillId);
+    return KoLConstants.availableSkillsSet.contains(skillId);
   }
 
   public static final boolean hasSkill(final String skillName) {
@@ -4086,7 +4074,7 @@ public abstract class KoLCharacter {
     return KoLCharacter.hasSkill(skillId);
   }
 
-  public static final boolean availableCombatSkill(final int skillId) {
+  public static final boolean hasCombatSkill(final int skillId) {
     return KoLConstants.availableCombatSkillsSet.contains(skillId);
   }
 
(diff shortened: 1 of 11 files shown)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -4090,54 +4067,27 @@ public static final boolean hasAmphibianSympathy() {
...
-    return KoLCharacter.availableCombatSkill(skill);
-  }
-
-  public static final boolean availableCombatSkill(final UseSkillRequest skill) {
-    if (skill == null) {
-      return false;
-    }
-
-    return KoLConstants.availableCombatSkillsMap.containsKey(skill);
+    return KoLConstants.availableCombatSkillsSet.contains(skillId);
```
midgleyc, line 4090: `availableCombatSkill` is identical to `hasCombatSkill`. It would be nice to only have one.

Similarly, `hasSkill` just calls `hasAvailableSkill`.

I like the symmetry in the names `hasSkill` and `hasCombatSkill`.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -3905,47 +3905,25 @@ private static void addAvailableSkill(final UseSkillRequest skill, final boolean
...
-    }
-
-    KoLConstants.availableCombatSkills.add(skill);
-    KoLConstants.availableCombatSkillsMap.put(skill, null);
+    KoLCharacter.addAvailableCombatSkill(SkillDatabase.getSkillId(skillName));
   }
 
   public static final void removeAvailableCombatSkill(final int skillId) {
-    KoLCharacter.removeAvailableCombatSkill(UseSkillRequest.getUnmodifiedInstance(skillId));
+    if (KoLConstants.availableCombatSkillsSet.contains(skillId)) {
```
midgleyc, line 3919: I think you can skip the contains: if you call `remove`, and the item isn't present, the function returns false (which you can just ignore).

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -3905,47 +3905,25 @@ private static void addAvailableSkill(final UseSkillRequest skill, final boolean
 
   /** Adds a single skill to the list of skills temporarily possessed by this character. */
   public static final void addAvailableCombatSkill(final int skillId) {
-    KoLCharacter.addAvailableCombatSkill(UseSkillRequest.getUnmodifiedInstance(skillId));
+    if (!KoLConstants.availableCombatSkillsSet.contains(skillId)) {
```
midgleyc, line 3908: similarly to the remove comment, all should work if you just call `add` here without contains: if it's already present, it won't be added again (as the container is a set), and if it isn't it will be.

src/net/sourceforge/kolmafia/KoLCharacter.java
Veracity0: Done.

src/net/sourceforge/kolmafia/KoLCharacter.java
Veracity0: Done.

src/net/sourceforge/kolmafia/KoLCharacter.java
Veracity0: Done.

midgleyc (approved):

PR #679, opened 2022-03-28
treat unassigned bjorn/crown slots differently from those assigned to $familiar[none]

horrible-little-slime:
This fixes the issue described here:

https://kolmafia.us/threads/assigning-the-buddy-bjorn-a-bonus-while-using-buddy-bjorn-results-in-failure.27419/

and obviates the earlier fix done here:
https://github.com/kolmafia/kolmafia/pull/662

Essentially, the root issue is that the value FamiliarData.NO_FAMILIAR was being used to communicate two distinct ideas: bjorn-slots that were unassigned (which is now assigned a value of null), and bjorn-slots that were actively set to $familiar[none] (which are now continued to be called FamiliarData.NO_FAMILIAR). This allows the maximizer to better handle $familiar[none] as though it was a regular familiar.

We've added two tests that deal with this.

I swear to god this is the last PR I write about this.


heeheehee-kolmafia:
I have a reproducing example for "unequip bjorn":

```java
  @Test
  public void keepFilledBjornEquippedWithBonus() {
    equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
    KoLCharacter.setBjorned(new FamiliarData(FamiliarPool.DICE));

    assertTrue(maximize("adventures, -buddy-bjorn, +25 bonus Buddy Bjorn"));
    recommendedSlotIs(EquipmentManager.CONTAINER, "Buddy Bjorn");
  }
```
(Namely, bjorn needs an inhabitant.)

Looking for a reproducing test case for the CoT case.

horrible-little-slime:
> I have a reproducing example for "unequip bjorn":
> 
> ```java
>   @Test
>   public void keepFilledBjornEquippedWithBonus() {
>     equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
>     KoLCharacter.setBjorned(new FamiliarData(FamiliarPool.DICE));
> 
>     assertTrue(maximize("adventures, -buddy-bjorn, +25 bonus Buddy Bjorn"));
>     recommendedSlotIs(EquipmentManager.CONTAINER, "Buddy Bjorn");
>   }
> ```
> 
> (Namely, bjorn needs an inhabitant.)
> 
> Looking for a reproducing test case for the CoT case.

Interestingly, this build seems to be failing that test. I'll investigate why. Under what circumstances does `maximize` typically return false?

Testing it in my built copy appears to be succeeding fine with or without a familiar currently in the bjorn.

heeheehee-kolmafia:
And, reproducing example for "illegal bjornification":

```java
  @Test
  public void keepFilledBjornEquippedAndUnchangedWithCrownAndBonus() {
    canUse("Crown of Thrones");
    canUse("time helmet");

    KoLCharacter.addFamiliar(new FamiliarData(FamiliarPool.DICE));
    equip(EquipmentManager.CONTAINER, "Buddy Bjorn");

    assertTrue(maximize("adventures, -buddy-bjorn, +25 bonus Buddy Bjorn"));

    recommendedSlotIs(EquipmentManager.HAT, "time helmet");
    recommendedSlotIs(EquipmentManager.CONTAINER, "Buddy Bjorn");
    // Fuzzy dice adds +10 to all stats. Make sure we didn't just equip it.
    assertEquals(0, modFor("Muscle"), 0.01);
  }
```

Would be good to make sure that your fix addresses both of these cases.

heeheehee-kolmafia:
> Under what circumstances does `maximize` typically return false?

Per the code, two cases: Mafia encountered an error (!KoLmafia.permitsContinue()), or the recommended maximization failed.

horrible-little-slime:
Fiddling around with this, it seems like `recommendedSlotIs` just doesn't work for things that are already equipped?

I fiddled around with `keepEmptyBjornEquippedAndUnchangedWithCrownAndBonus`, and tested a few cases:
1. using `canUse` instead of `equip` for the bjorn was a success
2. adding a `recommendedSlotIs` for the time helmet did not add a failure (that is, it failed later, when it got to the bjorn's `recommendedSlotIs`
3. adding a `recommendedSlotIs` for the time helmet _and_ equipping it instead of using `canUse` got us a failure

Prior to adding these tests, the only place `recommendedSlotIs` gets used on this file is for the basic meat fez in `noTiePrefersCurrentGear`, which does _not_ start off equipped.

That being said, I'm not sure how actually to test it if we can't use `recommendeSlotIs`. `KoLCharacter.hasEquipped` seems not particularly useful because my understanding is that the maximize function we're calling doesn't equip anything.

midgleyc:
If you debug the test and freeze it at the end, you can inspect `Maximizer.boosts` to see what Maximizer is recommending. That's how I wrote the helper functions in the first place.

test/net/sourceforge/kolmafia/maximizer/MaximizerRegressionTest.java
```diff
@@ -132,30 +132,27 @@ public void equipEmptyCrownNoSlot() {
...
+  public void keepBjornEquippedWithBonus() {
+    // Provide a helpful alternative to the bjorn.
+    canUse("makeshift cape");
     equip(EquipmentManager.CONTAINER, "Buddy Bjorn");
-    KoLCharacter.setBjorned(new FamiliarData(FamiliarPool.DICE));
 
-    assertTrue(maximize("adventures, -buddy-bjorn, +25 bonus Buddy Bjorn"));
-    // Fuzzy Dice adds +10 to all stats. Make sure it's still on.
-    assertEquals(10, modFor("Muscle"), 0.01);
+    assertTrue(maximize("-item, -buddy-bjorn, +25 bonus Buddy Bjorn"));
```
midgleyc, line 140: you're maximizing `-item`, you probably want `item` for a meaningful test?

horrible-little-slime: Yep! I'm switching this back to adventures to be consistent with the other test. I dont remember when I changed that. Good catch!

horrible-little-slime:
> If you debug the test and freeze it at the end, you can inspect `Maximizer.boosts` to see what Maximizer is recommending. That's how I wrote the helper functions in the first place.

I ended up hashing this out in the forums with heeheehee. (This may be an incorrect number of hees). The new setup _should_ be appropriate.

midgleyc (approved):
one more test for a fringe case, but I think all else is covered?

PR #680, opened 2022-03-29
a finally clause is allowed to do a function return

Veracity0:
```
int test( int input )
{
    try {
        print("You gave me " + input);
    } finally {
        if (input == 100) {
            int output = input + 1;
            print("I prefer " + output);
            return output;
        }
    }
    print("Good enough!");
    return input;
}

print("I received " + test(50));
print("I received " + test(100));
```
yields
```
> finally.ash

You gave me 50
Good enough!
I received 50
You gave me 100
I prefer 101
Good enough!
I received 100
```

I don't like that. Now it yields
```
> finally.ash

You gave me 50
Good enough!
I received 50
You gave me 100
I prefer 101
I received 101
```
which is what I expected, when I wrote that code.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/parsetree/Try.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/Try.java
@@ -54,14 +54,27 @@ public class Try extends Command {
           interpreter.trace("Entering finally, saved state: " + oldState);
         }
 
-        this.finalClause.execute(interpreter);
+        Value fresult = this.finalClause.execute(interpreter);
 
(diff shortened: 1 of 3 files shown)


midgleyc:
Could you add that example as a test? I'm pretty sure we already have (ash, output) tests somewhere.

midgleyc (approved):

Veracity0:
I slightly augmented that test to include a test of the "abort" code path in a "finally" clause.
That gives me 100% code coverage of the code I changed in this PR. A first for me!

PR #682, opened 2022-03-30
Fix retrieve_price() when asking for more than 5 of an item.

Veracity0:
(I created this branch to work on a particular issue, described below. In order to create a draft PR, I started with a commit which is not related to this issue, but is simple code cleanup that doesn't actual change any executable code.)

### The Issue:

KoLmafia only exposes the "5th-lowest" mall price to user scripts. The goal of that is to prevent the creation of mall-bots using unmodified KoLmafia source/jars. Yes, it is easy to modify the code to bypass that restriction. But, cheaters gonna cheat. We can't stop that.

For honest folk, the contract is "if you want to buy such-and-such an item in the mall, it will cost no more than this (the 5th lowest) price". A wrinkle appears, however, if you want to buy more than 5 of the item. You can use retrieve_price(int count, item it) to get what is, essentially, "count" times the 5th-lowest price. (That function also takes into account the "cost to make" and will return the lowest of the "cost to buy" and the "cost to make".)

Consider what happens if the actual mall price increases after the 5th available item.
```
count  mall price    actual   current   better
-----  ----------  ---------   ------   ------
  1        100         100       500      500
  2        200         300      1000     1000
  3        300         600      1500     1500
  4        400        1000      2000     2000
  5        500        1500      2500     2500
  6       1000        2500      3000     3500
  7       1000        3500      3500     4500
  8       1000        4500      4000     5500
  9       1000        5500      4500     6500
 10       1000        6500      5000     7500
 11       5000       11500      5500    12500
 12       5000       16500      6000    17500
 13       5000       21500      6500    22500
 14       5000       26500      7000    27500
 15       5000       31500      7500    32500
```
That script that expects to be able to buy 15 items for 7,500 will be disappointed when it actually spends 24,000 more than that.

The "better" approach I propose is to use the "5th-lowest" mall price when acquiring from 0 to 5 of the item, and the actual mall price for the 6th on up item.

Note that "accessible" items - in inventory, in storage or closet (if you have enabled those as item sources) - are accounted for first, using your "valueOfInventory" setting to price them vis a vis actual (5th lowest) mall prices.

Note also that calculating "cost to make" is recursively called by this cost function, so the same algorithm will apply to "cost to make".

An interesting refinement might be handling a case where buying 5 is cheaper than making 5, but making 6-15 is cheaper than buying 6-15. But, that would require a change to retrieve_item to actually use that algorithm. Likely beyond the scope of this proposed fix.


Veracity0:
There are two kinds of mall prices available in MallPriceManager.

MallPriceManager.getMallPrice(item) - the "exact" price found via a mall search
MallPriceManager.getMallPrice(item, maxAge) - the "historical" price from the MallPriceDatabase.

Even though "item" is an AdventureResult, both use only the "itemId" and ignore the "count".

I have changed the semantics.

- I added versions of those that take int itemId. Those behave just like the original versions.
- I changed the version which take an AdventureResult to use the "count" field to mean the cost to buy that many of the item. For now, they simply multiply the amount returned by the "itemId" versions by the "count".

I looked at all the places that call MallPriceDatabase.getCount. Almost all of them pass in an AdventureResult with a count of 1. That is exactly equivalent to calling the itemId version, but to make it crystal clear, I changed all those calls to pass in an itemId.

The few places that actually have a count really do want to calculate the price to buy that many of the item. They are all in InventoryManager and are associated with priceToAcquire - the exact thing I want to fix in this PR. Therefore, I simplified all the internal methods in InventoryManager that pass around an AdventureResult AND a "quantity" to simply pass around an AdventureResult where the "count" field is the desired quantity - and to expect that the mall price they get back will, in fact, be the price to buy that many.

This is most of what I'm trying to achieve here. Now I think it's down to making MallPriceManager look at actual saved searches when the number to buy is greater than 5; otherwise simply multiplying by the "5th cheapest" price is still correct.

I am very happy there was an extensive test suite for priceToAcquire. Since it was mocking getMallPrice, I did have to change the mocked method to have the new semantics.

Veracity0:
The MallPriceManager.getMallPrice() methods that take an itemId return an int; the maximum item price in the mall is one shy of a billion. However, the ones that take an AdventureResult - itemId + count - now return a long. Try asking for the mall price of 100 Mr. Accessories. :)

The only places that use those methods is the priceToAcquire methods in InventoryManager.
I also had to tweak the tests that passed in an AdventureResult (but not the ones that passed in an itemId).

A couple more tweaks:

ASH retrieve_price() now asks for "exact" prices, rather than historical prices.
(Perhaps I should make a 3-argument version which lets you specify if you want that, defaulting to historical? Probably.)

And the time before a saved mall search is "too old" is now 60 seconds, rather than 15. I noticed that vcon takes longer than 15 seconds to run and therefore might force extra mall searches. :(

Veracity0:
Turns out that the ResultProcess test for verifying that finding Oyster Days when it is not Easter is broken on Easter. :)

I also added the three-argument form of int retrieve_price(int count, item it, boolean exact).
It defaults to "false", which uses historical prices. With "true", it uses cached search results.

I also added a unit test to MallPriceManager.getMallPrice(AdventureResult item) which tests retrieving from 1 to 15 of an item with varying mall prices beyond the 5th-cheapest.

src/net/sourceforge/kolmafia/session/InventoryManager.java
```diff
@@ -1155,33 +1156,34 @@ public static int itemValue(final AdventureResult item, final boolean exact) {
...
+    int itemId = item.getItemId();
+    int needed = item.getCount();
+
     // Not just inventory; include anything our setting allow to be retrieved
     int onhand = Math.min(needed, InventoryManager.getAccessibleCount(item));
+    long price = 0;
 
     if (onhand > 0) {
-      if (item.getItemId() != ItemPool.PLASTIC_SWORD) {
+      if (itemId != ItemPool.PLASTIC_SWORD) {
```
midgleyc, line 1182: The change here is fine, but why is the TPS special-cased here anyway?

Might be worth adding the comment from the commit history:

    Added a special case to the create vs. buy decision-making code: an on-hand
    tiny plastic sword is valued at 0 meat, since you get it back from the drink.
    This should avoid undesirable behavior if the historical price of the TPS is
    ever greater than one of the items containing a TPS.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -146,13 +146,16 @@ public static MallSearchRequest newMallSearchRequest(String category, String tie
   // sort you could buy. In particular, the 5th cheapest.
   public static int NTH_CHEAPEST_PRICE = 5;
 
-  // How many stores to request results for in a mall search.  It should be at
-  // least NTH_CHEAPEST_PRICE, but stores can have limits and can ignore you
-  // (or vice versa), it could be higher.
-  public static int MALL_SEARCH_RESULTS = 5;
+  // How many stores to request results for in a mall search. We'd like enough to find at least
+  // NTH_CHEAPEST results. Stores frequently have more than one of an item for sael, but stores also
```
midgleyc, line 150:
```suggestion
  // NTH_CHEAPEST results. Stores frequently have more than one of an item for sale, but stores also
```

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -415,64 +418,195 @@ public static final void searchMall(
...
+    int price = -1;
+
+    long total = 0;
+
+    for (PurchaseRequest req : results) {
+      price = req.getPrice();
+      int available = req.getLimit();
+
+      // If we are still looking for the "nth-cheapest price" ...
+      if (found < NTH_CHEAPEST_PRICE) {
```
midgleyc, line 538: I think inside an algorithm like this, `NTH_CHEAPEST_PRICE` isn't as clear as it could be. My brain wants it to be a price (e.g. 300 meat) but it's actually a number (e.g. 5). `NTH_CHEAPEST_COUNT`?

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -415,64 +418,195 @@ public static final void searchMall(
...
+        // If this store does not have it, simply account for available inventory
+        if ((found + available) < NTH_CHEAPEST_PRICE) {
+          // Don't add price until we have at least 5
+          found += available;
+          continue;
+        }
+
+        // Otherwise, we have found the 5th-cheapest price!
```
midgleyc, line 546:
```suggestion
        // Otherwise, we have found the nth-cheapest price!
```
nth cheapest vs 5th cheapest. Probably nth cheapest is more consistent.

src/net/sourceforge/kolmafia/session/InventoryManager.java
Veracity0: Yeah. Fortunately, I have the verbose "svn log" output saved, so it was easy to discover this was r8873 by jasonharper from 2011-01-04. I have another comment from him in InventoryManager.itemValue - which I added when I was finding it difficult to understand what that method was calculating. He was very good at writing detailed commit messages. Not so much with in-line comments.

Comment added.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: Done.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: This makes a lot of sense. It used to be all magic numbers - i.e., "5" - but "count" is better than "price".
Changed to NTH_CHEAPEST_COUNT.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
Veracity0: Yes. I clarified all comments. The magic number "5" is now only mentioned in the declaration of the constant.

PR #687, opened 2022-04-01
Remove mallprices.txt from jar & override delete list

Rinn:
- Remove mallprices.txt from jar, it is terribly out of date
- Remove mallprices.txt from override list so it stops being deleted on app update
- Stop updating prices from non-override mallprices.txt as it no longer exists


src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
```diff
@@ -40,7 +40,6 @@
   private static final int CONNECT_TIMEOUT = 15 * 1000;
 
   static {
-    updatePrices("mallprices.txt", false);
     updatePrices("mallprices.txt", true);
```
Veracity0, line 43: This looks right. I looked at FileUtilities/DataFileUtilities, and this is the argument that allows it to look at the file system, rather than for a built-in file.

It seems like MallPriceDatabase.updatePrices could lose its boolean parameter and always call the lower level method with "true".

Veracity0: Or maybe not; there is another updatePrices(filename) which does extra logging and writes data to disk (assuming it had come from the server).

The parameter irritates me, and you get rid of it by renaming a method, but it's not critical.

Rinn: Yeah that's why I didn't remove the parameter I didn't want to merge the functions, I can rename it.

jaadams5 (dismissed):

jaadams5 (approved):

PR #690, opened 2022-04-05
Drones n banish

Veracity0:
- Melodramadery's image is composed of three images. As a result, we did not recognize its combat actions by image - as opposed to by a familiaraction HTML comment. Fix this.
- System Sweep is a banishing skill.

And apparently if I start a new branch when I am in another branch, rather than main, it is now based on the other branch?

Sheesh. Well, at least the other branch had nothing that shouldn't be committed yet.


midgleyc:
> And apparently if I start a new branch when I am in another branch, rather than main, it is now based on the other branch?

Yes. git on your machine by default creates branches off whichever commit you're currently on, even if you're not on a branch. If you want to create a new branch from main, run `git checkout -b new-branch main` (or `git switch -c new-branch main`).

midgleyc (approved):

PR #693, opened 2022-04-06
restore (accidentally deleted) factor in getAdverageML

Veracity0:
A for loop was replaced by a stream in PR 326.
It's pretty, but it dropped a factor in the calculation.
This restores it.


gausie (approved):
My bad! Thanks

PR #694, opened 2022-04-06
fix: avoid unexpected error in error case with no stream

midgleyc:
Relates to https://kolmafia.us/threads/starting-mafia-with-a-script-on-the-command-line.27442/#post-167643

This should allow the error message to be printed to the screen.


heeheehee-kolmafia (approved):
ha, this is basically the change I was going to propose in https://github.com/kolmafia/kolmafia/commit/f9bd02e53d8d48703c030a2de9893a6ea47a8286.

PR #695, opened 2022-04-07
Additional Umbrella Implementation

horrible-little-slime:
This more or less apes the Saber code beat-for-beat. I hope that, at the very least, this is of use to someone.

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -436,7 +436,7 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
       case ItemPool.PUNCHCARD_WALL:
       case ItemPool.PUNCHCARD_SPHERE:
         return AdventureResult.punchCardName(this.id);
-      
+
       case ItemPool.UNBREAKABLE_UMBRELLA:
         return this.name + "(" + Preferences.getString("umbrellaState") + ")";
 
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -438,7 +438,7 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
         return AdventureResult.punchCardName(this.id);
 
       case ItemPool.UNBREAKABLE_UMBRELLA:
-        return this.name + "(" + Preferences.getString("umbrellaState") + ")";
+        return this.name + " (" + Preferences.getString("umbrellaState") + ")";
 
       default:
         return this.name;
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -436,6 +436,9 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
       case ItemPool.PUNCHCARD_WALL:
       case ItemPool.PUNCHCARD_SPHERE:
         return AdventureResult.punchCardName(this.id);
+      
+      case ItemPool.UNBREAKABLE_UMBRELLA:
+        return this.name + "(" + Preferences.getString("umbrellaState") + ")";
 
       default:
         return this.name;
--- a/src/net/sourceforge/kolmafia/textui/command/UmbrellaCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/UmbrellaCommand.java
@@ -28,22 +28,22 @@ public class UmbrellaCommand extends AbstractCommand {
       return;
     }
 
-    Integer umbrellaForm = null;
+    Integer umbrellaState = null;
     if (parameter.equals("ml")) {
-      umbrellaForm = 1;
+      umbrellaState = 1;
     } else if (parameter.equals("dr")) {
-      umbrellaForm = 2;
+      umbrellaState = 2;
     } else if (parameter.equals("item")) {
-      umbrellaForm = 3;
+      umbrellaState = 3;
     } else if (parameter.equals("weapon")) {
-      umbrellaForm = 4;
+      umbrellaState = 4;
     } else if (parameter.equals("spell")) {
-      umbrellaForm = 5;
+      umbrellaState = 5;
     } else if (parameter.equals("nc")) {
-      umbrellaForm = 6;
+      umbrellaState = 6;
     }
 
-    if (umbrellaForm == null) {
+    if (umbrellaState == null) {
       KoLmafia.updateDisplay(
           MafiaState.ERROR, "I don't understand what Umbrella form " + parameter + " is.");
       return;
@@ -55,7 +55,7 @@ public class UmbrellaCommand extends AbstractCommand {
 
     request = new GenericRequest("choice.php");
     request.addFormField("whichchoice", "1466");
-    request.addFormField("option", Integer.toString(umbrellaForm));
+    request.addFormField("option", Integer.toString(umbrellaState));
     request.addFormField("pwd", GenericRequest.passwordHash);
     RequestThread.postRequest(request);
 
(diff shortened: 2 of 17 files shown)


horrible-little-slime:
Very rudimentary testing says that this is working. A few things could be better, though:

- Is `umbrellaState` the best name for the property?
- maximizer is currently not willing to fold this, presumably because I copied all the saber code and saber can't be reconfigured during the day--how do I resolve that?
- Mafia always calls it `unbreakable umbrella (broken)` despite the item's name changing periodically--I want to declare this out of scope of this PR
- the CLI command's inputs are case sensitive and possibly counterintuitive. Does that matter if the `help` feature exists?

jaadams5 (dismissed):
Approved but see non blocking comments at the top level

jaadams5:
umbrellaState seems fine to me.  I couldn't think of anything I liked more but maybe someone else can.

Have not looked at the maximizer issue but if it can be addressed that would be great.  I don't think that is blocking as long as it ends up on someone's TODO list.

I'm fine with name changes being out of scope.  Indeed I think I'd prefer a constant name for the item and to have to use unbrellaState to determine the state.

I feel like mafia cli commands are inconsistent in terms of case sensitivity and exact vs. fuzzy matching for parameters.  The only reason not to defer action on the command is the highly likely situation where someone writes a script using V1.0 of the command parameters and the V2.0 version deliberately or otherwise breaks the script.

Might you consider using an ENUM for the umbrella states?  It can make code easier to read.  You could go one step further and make unbrellaState externally a string.  That would make scripts potentially easier to read.

gausie:
This is great! Don't use an integer for the states, use the word that's used in kol. That's easier for everyone. Enum still a good idea. Write some tests please also

horrible-little-slime:
I'll switch over to an enum later today, and use words like "broken" and
"cocoon". May take me longer on the tests

Veracity0:
> * Is `umbrellaState` the best name for the property?
This seems fine.

> * maximizer is currently not willing to fold this, presumably because I copied all the saber code and saber can't be reconfigured during the day--how do I resolve that?

There are other items which change names and/or effects via "using" - (Sneaky Pete's leather jacket, Lucky Tam O'Shanter) -> (Sneaky Pete's leather jacket (collar popped), Lucky Tam O'Shatner) - but those are actually different objects. Ditto for the items in foldgroups.txt, which fold (in sequence) into distinct items.

On the other hand, you CAN configure the backup camera multiple times per day - and maximizer supports that.

> * Mafia always calls it `unbreakable umbrella (broken)` despite the item's name changing periodically--I want to declare this out of scope of this PR

It changes name, but does not actually change itemID or itemDescid?
When I looked at mine - in the (broken) state - I noticed that its image was umbrella7.gif. Considering that umbrella1.gif is the un-bound IOTM, that implies the existence of umbrella2.gif - umbrella6.gif
Huh. I just configured it into "unbreakable umbrella (bucket style)" and it has the same descid and the image is umbrella5.gif

I've seen this for the latte lovers member's mug: mine currently has the name "Latte: Guarna and oil paint Latte with a shake of cinnamon".

Well, combining thoughts on those last two:

What are the six possible names for this item?
What if the "base name" in items.txt and modifiers is simply "unbreakable umbrella".
modifiers.txt would have modifiers for unbreakable umbrella as you have it - based on the umbrellaState property
AdventureResult .getName() could choose the appropriate name for the item based on the property.
- that just to make it pretty for the user - you could see in your inventory how it was configured.

modifiers.txt has a BackupCamera section with the three configuration options.
Presumably, that is how the Maximizer knows it can switch between them.

Add an UnbreakableUmbrella section with the six options.
Teach Maximizer to use that, regarding the "unbreakable umbrella" equipment, the same way it deals th the backup camera.

> * the CLI command's inputs are case sensitive and possibly counterintuitive. Does that matter if the `help` feature exists?
Case sensitive is fine. We make item and effect names and such insensitive - we do have the infrastructure to do that - but we never bother with command parameters.

Strive to make the options intuitive. You can provide multiple variants of the same parameter name.

Veracity0:
> It does not change itemID, although its description changes. Weird little item! I'll look at the latte code to figure out exactly how you do that. Now that the umbrellaState property just steals the name directly, it'll be easier.

We don't actually do anything with the latte mug - that is how KoL reports it to my main, who owns one and last configured it that way.

I just looked at the unbreakable umbrella with a character who doesn't own one. It shows me a description of an "unbreakable umbrella (broken)". With my main, it shows me "umbreakable umbrella (bucket style)". I expect because there is no such thing as an "unconfigured" version of the item. Presumably, "(broken)" is the "default". You should use that in defaults.txt, where you declare the property.

I like the idea of changing the name in items.txt, modifiers.txt, and equipment.txt to simply "unbreakable umbrella". Look at this:

```
> /closet 2 unbreakable umbrella

You don't have that many unbreakable umbrellas, you've only got 1.
```

That is KoL's response to a chat command. I think that tells us that KoL's internal name really is "unbreakable umbrella".

If we wanted a purely cosmetic change, we could add a case clause to AdventureResult.getName() which would return one of six names for display purposes - in the GUI. We do that for El Vibrato punchcards, for example.

If we wanted the user to actually be able to type in and use such names, we can declare "aliases" in ItemDatabase. That is necessary for Dungeon of Doom bang potions, for example: if the user wants to put "use 1 potion of detection" into a mood, say, that has to resolve to the correct (identified) potion.

But for this item, if we declared "unbreakable umbrella" and also 6 aliases, would that cause ambiguity in the "use" command in the CLI, say? So we probably don't want aliases.

As I said, I like changing the base name in the three data files. I think that is the correct thing, based on my chat command example.

And maybe the display name via AdventureResult. I don't think that would do any harm. I'll try it myself, if you decide not to, once you release whatever you end up doing. :)

horrible-little-slime:
Alright I _think_ I'm done, assuming me kicking the maximizer can down the road is chill.

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -437,6 +437,9 @@ public String getName() {
       case ItemPool.PUNCHCARD_SPHERE:
         return AdventureResult.punchCardName(this.id);
 
+      case ItemPool.UNBREAKABLE_UMBRELLA:
+        return this.name + "(" + Preferences.getString("umbrellaState") + ")";
```
Veracity0, line 441: Presumably you want " (" rather than "("

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/AdventureResult.java
horrible-little-slime: Incredible catch.

src/net/sourceforge/kolmafia/textui/command/UmbrellaCommand.java
```diff
@@ -0,0 +1,66 @@
...
+      umbrellaState = UmbrellaRequest.Form.TWIRL.id;
+    } else if (parameter.equals("nc") || "cocoon".startsWith(parameter)) {
+      umbrellaState = UmbrellaRequest.Form.COCOON.id;
+    }
+
+    if (umbrellaState == null) {
+      KoLmafia.updateDisplay(
+          MafiaState.ERROR, "I don't understand what Umbrella form " + parameter + " is.");
+      return;
+    }
```
Veracity0, line 52: This is OK, but it could just be the "else" clause of the preceding if/elseif chain.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/command/UmbrellaCommand.java
horrible-little-slime: I was copying the saber code, but it is cleaner looking to tag it onto the end there.

Veracity0 (approved):

jaadams5 (approved):

PR #696, opened 2022-04-08
feat: stop zero meaning all for meat

midgleyc:
Relates to https://kolmafia.us/threads/csend-with-commas-acts-unexpectedly-when-sending-meat.27447/

"0 meat" now parses as no meat, instead of all your meat. * continues to be all. -X continues to be all but X.


Veracity0 (approved):

jaadams5 (approved):
Would a test showing 1,000 meat is an error or that showed 1,000 meat and 1000 meat were the same thing be appropriate?

midgleyc:
I think any test would go against one of the commands.

1,000 meat is an error. Parses as "1" (error) and "000 meat" (0 meat).

PR #697, opened 2022-04-08
Two tests and lint for SendCommand

jaadams5:
I started this to explore csend 1,000,000 meat sending all available meat.  While it was a good way to get into the debugger it ultimately has little to no bearing on the csend situation.  I'm proposing committing it anyway:

1. Sometimes it is easier to add to an existing test class than start from scratch
2. The scaffolding for command tests is new to me so there might be a learning opportunity


test/net/sourceforge/kolmafia/textui/command/SendMessageCommandTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+class SendMessageCommandTest extends AbstractCommandTestBase {
+
+  public SendMessageCommandTest() {
+    this.command = "csend";
+  }
+
+  @Test
+  public void itShouldParseMeatWithCommas() {
+    String output = execute(" 1,000,000 meat to buffy");
```
midgleyc, line 16: this should send no meat and just error, I think. we don't support numbers with comments

jaadams5: The problem with these tests is what I have written only captures that a message was sent and not the contents of the message.  If you can point me at how to get the message I will gladly expand.  Alternatively I can shift the focus to increasing coverage according to jacoco in which case deleting this test makes sense.

jaadams5:
I'm going to close or revert this to draft.  The user level problem has been fixed and tested elsewhere and #699 looks to be a good example of getting to the request sent by a command.  I will follow that example and redirect this towards generating test coverage and not a specific issue.

jaadams5:
I'm going to undraft this momentarily in search of advice or the other pair of eyeballs.  I see no point in pushing until I give up or have 100% coverage.

I will also annotate code but...

I mocked MoodManager as an alternative to changing access or adding a function for testing only.

The test works but when it was in the same test class as others they failed.  The MoodManager mock was leaking over to other tests.  I'm not sure how to "unmock".  MoodManager being static may be an issue.

I had a conditional (a && (b || c)).  Coverage said a branch was being missed and I could not figure it out.  I rewrote it into what I think was a functionally equivalent set of nested ifs and the same tests generate 100% coverage.  I am either wrong in thinking my rewrite was functionally equivalent or Jacoco and the associate byte code are not doing what I expect.

src/net/sourceforge/kolmafia/textui/command/SendMessageCommand.java
```diff
@@ -65,17 +65,29 @@ public static void send(final String parameters, final boolean isConvertible) {
         ItemFinder.getMatchingItemList(itemList, KoLConstants.inventory);
 
     if (attachments.length == 0
-        && (itemList.length() > 1 || message == KoLConstants.DEFAULT_KMAIL)) {
+        && (itemList.length() > 1 || message.equals(KoLConstants.DEFAULT_KMAIL))) {
       return;
     }
+    /*
```
jaadams5, line 71: This is the code where the rewrite has different coverage than the original and I can't figure out why.

test/net/sourceforge/kolmafia/textui/command/SendMessageCommandTestAlso.java
```diff
@@ -0,0 +1,60 @@
...
+  private MockedStatic<MoodManager> mockMoodManager() {
+    var mocked = mockStatic(MoodManager.class, Mockito.CALLS_REAL_METHODS);
+    mocked.when(MoodManager::isExecuting).thenReturn(true);
+    return mocked;
+  }
+
+  @Test
+  public void itShouldNotSendDuringMoodSwings() {
+    String output;
+    MockedStatic<MoodManager> mockery = mockMoodManager();
```
jaadams5, line 51: The mock leaks into other tests in the same file and I couldn't figure out how to "unmock" after one test.

midgleyc:
Added straight to the branch, but:
* I think you were missing the case where the user tries to send no items with the default message
* static mocks are unset by calling `.close()` on the object returned from the setup

jaadams5:
> Added straight to the branch, but:
> 
> * I think you were missing the case where the user tries to send no items with the default message
> * static mocks are unset by calling `.close()` on the object returned from the setup

I'll check the coverage again but the case was covered in the rewrite.  Thanks.

the annotation at nested is new to me.  next time I have a static mock and use Cleanups it will be better.  thank you.

jaadams5:
> Added straight to the branch, but:
> 
> * I think you were missing the case where the user tries to send no items with the default message

Coverage is there so my claim of equivalence must have been incorrect,  Thanks.  Will revert to draft until I have 100% or give up.

jaadams5:
I'm giving up.  100% coverage is going to require mocking two request types and most of the tests I can think of are meaningless in terms of testing functionality.  I think there are better ways to spend my time and better places to increase overall coverage.  Indeed I started this because of a parsing issue that might have been related to this class but isn't.  Un-drafting and will push or close in response to feedback.

midgleyc (approved):
more tests = more good :)

PR #701, opened 2022-04-10
tests: cover closet command, refactor

midgleyc:
There was a small problem with #699: I didn't reset the client between calls, so somehow it was winding up with different clients. Don't understand that because the builder was instantiated once in the test class (and the client against never reinstantiated), so it should have been the same client, but it wasn't.

That's fixed, enough tests have been added to cover the class (though this doesn't cover all the functionality -- e.g. passing multiple items, meat counts is not done because I haven't yet managed to get the request text, and I need that to write a meaningful test).

Small refactoring to use streams.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
@@ -64,7 +64,8 @@ public class ClosetCommand extends AbstractCommand {
     var items = split.get(false);
 
     if (meat.size() > 0) {
-      int meatCount = meat.stream().map(AdventureResult::getCount).mapToInt(Integer::intValue).sum();
+      int meatCount =
+          meat.stream().map(AdventureResult::getCount).mapToInt(Integer::intValue).sum();
       if (meatCount > 0) {
         int moveType = isTake ? ClosetRequest.MEAT_TO_INVENTORY : ClosetRequest.MEAT_TO_CLOSET;
         RequestThread.postRequest(new ClosetRequest(moveType, meatCount));
--- a/src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia.textui.command;
 
+import java.util.Arrays;
 import java.util.List;
 import net.sourceforge.kolmafia.AdventureResult;
 import net.sourceforge.kolmafia.KoLConstants;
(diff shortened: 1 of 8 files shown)


jaadams5:
Thank you.  I was working last night on tests for csend and discovered two things.  The first is that my tests would work when I ran a single test in my IDE, work when I ran the complete test class in the IDE but fail when they were run as part of the entire test suite.  I think we now know why.  I also realized my tests were somewhat superficial because I could not actually see the request test.  I will wait for you to sort that out.  I think the work is with POST requests but some of my superficial research noted that the uri is available and request parameters can be parsed from that.  But thank you.

src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
```diff
@@ -57,35 +59,28 @@ public void run(final String cmd, String parameters) {
...
-      if (item.getName().equals(AdventureResult.MEAT)) {
-        meatCount += item.getLongCount();
-        meatAttachmentCount += 1;
-        itemList[i] = null;
-      } else if (EquipmentDatabase.isHat(item)) {
-        hatCount += 1;
+    if (meat.size() > 0) {
+      int meatCount =
+          meat.stream().map(AdventureResult::getCount).mapToInt(Integer::intValue).sum();
+      if (meatCount > 0) {
```
jaadams5, line 69: Will this still work if the object is an AdventureLongCountResult?  Line 67 in the deleted code handled that but I not comfortable enough with streams to be certain the new line 68 will.  I will consider this blocking until I am told I am wrong :-)

jaadams5 (changes requested):

src/net/sourceforge/kolmafia/textui/command/ClosetCommand.java
midgleyc: good spot, I think this one should also be `getLongCount`. I will add a test.

midgleyc: Fixed, and also had to adjust `ItemFinder`, which was parsing the number as an int. I guess we didn't spot this before because it's extremely rare, and the previous behaviour went 3 billion -> 0 -> all meat, which was probably close to what people wanted out of this.

jaadams5: IIRC we found characters with meat quantities that overflowed an int.  The fix was AdventureLongCountResult.  There was a deliberate decision not to just make the count long in AdventureResult and a deliberate decision not to change every reference to a long, only the ones we expected might someday overflow.  Since ItemFinder can parse meat that was probably missed.

jaadams5 (approved):

PR #702, opened 2022-04-10
optionally save battle details in Bastille.battles.txt

Veracity0:
I've been playing Bastille Battalion with two characters and have been manually recording details of battles: which round, which stance I choose, whther I ended up being offensive or defensive, castle type, and so on.

It would be nice for KoLmafia to save that data for me in a file which I can load into an ASH script to analyze.

Assuming you have set "logBastilleBattalionBattles" to true, "data/Bastille.battles.txt" will be loadable via file_to_map.

```
record battle {
    int number;         // Affects strength of enemy
    int [6] stats;      // MA/MD/CA/CD/PA/PD
    string boosts;      // MCP
    string enemy;       // {frenchcastle,masterofnone,bigcastle,berserker,shieldmaster,barracks}
    string stance;      // {offensive,waiting,defensive}
    boolean aggressor;  // as opposed to defender
    boolean military;   // true if won
    boolean castle;     // true if won
    boolean psych;      // true if won
    int cheese;         // if won
};

battle [string] battles;
file_to_map( "Bastille.battles.txt", battles );
```

The (string) key is: DATE.PLAYERID.GAME.BATTLE

Since a player can have five games with up to 5 battles each, up to 25 entries per day per player.
Yes, all your multis will write to the same file; you don't need to analyze the data player by player.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/BastilleBattalionManager.java
+++ b/src/net/sourceforge/kolmafia/session/BastilleBattalionManager.java
@@ -1,6 +1,8 @@
 package net.sourceforge.kolmafia.session;
 
+import java.io.PrintStream;
 import java.util.Arrays;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.TreeMap;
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/session/BastilleBattalionManager.java
```diff
@@ -928,4 +1147,111 @@ public static final boolean registerRequest(final String urlString) {
...
+    buf.append(".");
+    buf.append(game);
+    buf.append(".");
+    buf.append(number);
+    return buf.toString();
+  }
+
+  private static String generateFields(Battle battle) {
+    StringBuilder buf = new StringBuilder();
+    buf.append(battle.number);
```
midgleyc, line 1204: I would probably make a `String makeTabSeparated(String... fields)` method instead of manually putting a TAB between the fields.

Veracity0: That is a fine idea. I extended it and it's a lot easier to read now.

jaadams5 (approved):

PR #704, opened 2022-04-11
Stabilize appName for packaged builds.

heeheehee-kolmafia:
Users of the .exe should be able to keep their settings across version
upgrades without having to manually copy them over from previous
installs.

https://kolmafia.us/threads/exe-and-dmg-builds-dont-work-anymore.27408/#post-167737
describes my experience with the latest commit. This PR should
address one of my concerns (namely, around the install path / setting
persistence).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -280,6 +280,7 @@ tasks.jpackage {
 	input = 'dist'
 	destination = 'build/releases'
 	mainClass = mainClassName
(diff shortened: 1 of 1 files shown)


Veracity0:
Not that I ever will use Windows, I can't really assess this, but Windows users will like it, so, why not?

PR #707, opened 2022-04-13
create get_permed_skills ash function

horrible-little-slime:
https://kolmafia.us/threads/get_permed_skills-ash-function.27465/

This adds tracking for hardcore permed skills atop our current tracking for permed skills, and exposes the information in a RuntimeLibrary function.


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -8121,6 +8128,19 @@ public static Value get_locket_monsters(ScriptRuntime controller) {
...
   }
 
+  public static Value get_permed_skills(ScriptRuntime controller) {
+    AggregateType type = new AggregateType(DataTypes.BOOLEAN_TYPE, DataTypes.SKILL_TYPE);
+    MapValue value = new MapValue(type);
+
+    for (var permedSkill : KoLConstants.permedSkills) {
+      var skill = DataTypes.makeSkillValue(permedSkill.getSkillId(), true);
+      var hardcore =
+          DataTypes.makeBooleanValue(KoLConstants.hardcorePermedSkills.contains(permedSkill));
```
midgleyc, line 8138: As the only thing you're doing with this list is checking whether an element is within it, it may as well be a `Set<int>`.

midgleyc:
One comment, plus a test might be nice. You can look at `CustomScriptTest` (to call a script) or a variant of `XCommandTest` that uses the `ash` (or `ashq`) command to find a base that should work.

The general existence of this function and the map<skill, boolean> seems reasonable to me.

horrible-little-slime:
Will work on test in the morning; I think I'll probably want to use CommandTest so that I can have an easier time giving mafia a charsheet to read.

midgleyc:
tbh I prefer the proxy records as dumb bags of constant data, independent of whatever character is currently logged in. I know this is already broken with `skill`'s `timescast', but eh.

horrible-little-slime:
> It might make more sense to make this a field on the skill proxy record

This isn't anywhere near my bailiwick (or even my wheelhouse), but this doesn't feel right just based on the way the information is harvested. As Midgley points out, most proxy records are unchanging, and the information comes from the bowels of mafia itself. Even `timescast` is, as far as I'm aware, tracked purely internally, rather than using information from KoL. Perms, on the other hand, are scraped from `charsheet.php`. I don't know if any of this actually matters, that's just what makes it feel inconsistent to me with the other proxy records. Also, on a personal note, I'd like to be able to iterate directly over perms.

I absolutely can switch it over to a proxy record if that's the verdict, though. I'll look into how that might work after I finish figuring out this test.

midgleyc (approved):

PR #710, opened 2022-04-13
create '_robortDrops' pref to track number of robortender booze/mixer drops in a day

lewismd13:
By popular demand in the ASScord, this pref tracks the number of times your robortender drops a special mixer, base booze, or basic mixer. Spading suggests that all three affect the drop rate, so this pref will allow scripters to go for more valuable drops when the probability is higher. 

It's in FightRequest.java as opposed to ResultProcessor.java, which would have been my preference, because of the need to listen for the drop messages for the base booze and mixers.

This is my first kolmafia PR, so I am more than open to suggestions, critique, etc. I will shortly be making a feature thread in the forums for the same.


midgleyc (changes requested):
My only comment is on the preference name: can it be either `_roboDrops` (similar to the already existing `_roboDrinks`) or `_robortenderDrops` (so as to not cut the familiar name off in the middle)? Probably the first is better for consistency.

lewismd13:
> My only comment is on the preference name: can it be either `_roboDrops` (similar to the already existing `_roboDrinks`) or `_robortenderDrops` (so as to not cut the familiar name off in the middle)? Probably the first is better for consistency.

Fair point! I have now made that change.

midgleyc (approved):

PR #712, opened 2022-04-16
Don't send a request when autoselling zero items

midgleyc:
Many CLI commands parse an item list and need to handle the zero case specially. Here's autosell.

Follow-up to #699.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/AutoSellRequest.java
+++ b/src/net/sourceforge/kolmafia/request/AutoSellRequest.java
@@ -117,6 +117,10 @@ public class AutoSellRequest extends TransferItemRequest {
         continue;
       }
 
+      if (item.getCount() == 0L) {
+        continue;
+      }
+
       if (ItemDatabase.getPriceById(item.getItemId()) <= 0) {
         continue;
       }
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/request/AutoSellRequest.java
```diff
@@ -108,15 +108,17 @@ public ArrayList<TransferItemRequest> generateSubInstances() {
...
-    for (int index = 0; index < this.attachments.length; ++index) {
-      AdventureResult item = this.attachments[index];
+    Set<AdventureResult> others = new HashSet<>();
 
+    for (AdventureResult item : this.attachments) {
       if (item == null) {
         continue;
       }
 
+      if (item.getCount() == 0L) {
```
jaadams5, line 118: Given AdventureResult and the long version, this doesn't seem aesthetically correct.  I think using getCount and not getLongCount is correct but I would test against 0 and not 0L

jaadams5 (approved):

src/net/sourceforge/kolmafia/request/AutoSellRequest.java
midgleyc: good spot! yes, had this as `getLongCount` to start with, adjusted to long, then didn't re-adjust.

PR #715, opened 2022-04-18
allow gregarious monsters to enter the queue

horrible-little-slime:
Gregarious monsters enter the queue. It's a pretty big deal!

I've hardcoded this basically as a special case, because I feel like touching `ignoreSpecialMonsters()` is a recipe for disaster, and I feel like this will probably stay as basically one special case for an extremely long time.


horrible-little-slime:
This mostly matters for the Time-Spinner,but it also matters for queue math.

midgleyc:
A test would be appreciated, but let me know here if it's too hard / you can't figure out where to start.

horrible-little-slime:
I gave it the ol' college try but came up emptyhanded; this code is in AdventureRequest.registerEncounter, and I can't quite figure out how to fake an AdventureRequest properly. I spent a while doing it with EncounterManager.registerEncounter  before realizing that that'll never actually update the queue. Which is a shame, because EncounterManager is relatively easy to write tests for.

midgleyc:
okay, so I've put together the following test:
```java
  @Test
  public void gregariousMonstersAreQueued() throws IOException {
    KoLAdventure.setLastAdventure(AdventureDatabase.getAdventure("Barf Mountain"));
    MonsterStatusTracker.setNextMonster(MonsterDatabase.findMonster("Knob Goblin Embezzler"));

    var req = new GenericRequest("fight.php");
    req.setHasResult(true);
    req.responseText = Files.readString(Path.of("request/test_fight_gregarious_monster.html"));
    req.processResponse();

    assertThat(AdventureQueueDatabase.getZoneQueue("Barf Mountain"), contains("Knob Goblin Embezzler"));
  }
```

which fails because a Knob Goblin Embezzler is a lucky monster, so `EncounterManager.isLuckyMonster` is hit and the encounter isn't queued. So I think the `if` chain needs a bit more work.

Veracity0:
You don't need to "fake" an AdventureRequest. You can allocate a request, give it a urlString and a responseText, and call methods in AdventureRequest.java with that request. Even static ones, like registerEncounter.

For example, in my BastilleBattalionManagerTest, I do this:

```
    String urlString = "choice.php?whichchoice=1314&option=1";
    String responseText = loadHTMLResponse("request/test_bastille_game1_1.html");
    GenericRequest request = new GenericRequest(urlString);
    // If you want to reuse the request with different URL/responseText pairs:
    // request.constructURLString(urlString);
    request.responseText = responseText;
    ...
    assertNull(AdventureRequest.parseChoiceEncounter(urlString, 1315, responseText));
    assertEquals(38, Preferences.getInteger("_bastilleLastCheese"));
    assertEquals(38, Preferences.getInteger("_bastilleCheese"));
```
Notice that I called a method that AdventureRequest.registerEncounter calls and verified that state -- properties, in this case - had been modified as expected.

midgleyc (approved):

PR #717, opened 2022-04-20
make Combat Rate case insensitive

horrible-little-slime:
Right now,` numericModifier("Combat Rate")` and `numericModifier("combat rate")` return different values when you're past the cap of 25, because we only are special casing the input "Combat Rate".

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -1498,10 +1498,10 @@ public class Modifiers {
   }
 
   public double get(final String name) {
-    if (name.equals("Prismatic Damage")) {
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1498,10 +1498,10 @@ public double get(final int index) {
   }
 
   public double get(final String name) {
-    if (name.equals("Prismatic Damage")) {
+    if (name.toLowerCase().equals("prismatic damage")) {
```
midgleyc, line 1501: `equalsIgnoreCase` is probably clearer in intent

jaadams5 (approved):

PR #718, opened 2022-04-20
refactor: skeleton command

midgleyc:
Similar to #714, this is a proof of concept for what I'd eventually like `Modifiers` to look like: only allowing enum values instead of ints or strings.

Avoid casts by moving away from `Object[][]`.


jaadams5 (approved):

Veracity0:
If your enum is inside the class, you can declare a map from String to Skeleton - and populate it from the enum constructor - and obviate iterating through Skeleton.values() in order to lookup by name.

You do need to invoke Skeleton.values() in a static block to force all those constructors, if you want to reference that map from tests, which apparently do not necessarily build the while enum.

midgleyc:
I think this one is better as a map with no enum involved. The values aren't exposed outside the class, and the only use of the enum is
* iterate over all values
* if the name is identical (`.equals`)
* then return the number

This is a perfect use-case for a map!

Veracity0 (approved):

Veracity0 (approved):

PR #720, opened 2022-04-21
add pref hasCosmicBowlingBall parsed from coolitems section of api.php

Malibu-Stacey:
The bowling ball was added to coolitems as when on cooldown it is not in the players inventory so there's no easy way to always know if the player owns one. This exposes the entry in coolitems as the preference `hasCosmicBowlingBall`.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ApiRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ApiRequest.java
@@ -396,7 +396,8 @@ public class ApiRequest extends GenericRequest {
           Map.entry(
               "neverendingparty", Map.entry("neverendingPartyAlways", "_neverendingPartyToday")),
           Map.entry("voterregistered", Map.entry("voteAlways", "_voteToday")),
-          Map.entry("boxingdaycare", Map.entry("daycareOpen", "_daycareToday")));
+          Map.entry("boxingdaycare", Map.entry("daycareOpen", "_daycareToday")),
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/request/ApiRequest.java
```diff
@@ -396,7 +396,8 @@ public static final void parseStatus(final JSONObject JSON) {
           Map.entry(
               "neverendingparty", Map.entry("neverendingPartyAlways", "_neverendingPartyToday")),
           Map.entry("voterregistered", Map.entry("voteAlways", "_voteToday")),
-          Map.entry("boxingdaycare", Map.entry("daycareOpen", "_daycareToday")));
+          Map.entry("boxingdaycare", Map.entry("daycareOpen", "_daycareToday")),
+          Map.entry("hascosmicball", Map.entry("hasCosmicBowlingBall", "_null")));
```
midgleyc, line 400: `Map.entry("hasCosmicBowlingBall", "_null")` indicates that `PREF_TO_COOL_ITEM` probably has the wrong type. I would use a class with properties `coolItem`, `alwaysPref` and `todayPref`, all Strings. It only is called using `forEach`, so a list of those objects works.

That should avoid the special casing `if (coolItem != "hascosmicball")`: you can check whether `todayPref` is null instead.

midgleyc (changes requested):
A test would also be good. A good entrypoint is probably `parseStatus`.

midgleyc:
Obsoleted by #767.

PR #722, opened 2022-04-21
refactor: remove integer pool

midgleyc:
IntegerPool was an eagerly computed cache of values between -2 and 13000. Where used, it prevented creating more than one of the same integer, and would allow them to be compared using `==` (although I couldn't find any uses of this). However, some of the memory saving was lost by precomputing all 13000 integers, and there was a slight perf hit every use (by checking whether the int was too big or small and computing where it was in the cache).

In newer Java, I don't think this is necessary: there is lots of memory available for modern computers, and most integers are short-lived (and integers between -127 and 128 are cached anyway).

Plus, I think the code looks cleaner with the integers alone.

I removed it using IntelliJ's refactoring: inline function.


src/net/sourceforge/kolmafia/session/HaciendaManager.java
```diff
@@ -379,25 +378,25 @@ private static int countString(String inString, String lookFor) {
 
   private static final Object[][] RECORDINGS = {
     {
-      IntegerPool.get(EffectPool.THE_BALLAD_OF_RICHIE_THINGFINDER), "_thingfinderCasts",
+      (Integer) EffectPool.THE_BALLAD_OF_RICHIE_THINGFINDER, "_thingfinderCasts",
```
Veracity0, line 381: Why do these need casts?

src/net/sourceforge/kolmafia/swingui/GearChangeFrame.java
```diff
@@ -1494,15 +1493,15 @@ public FakeHandsSpinner() {
     public void stateChanged(final ChangeEvent e) {
       int maximum = this.availableFakeHands;
       if (maximum == 0) {
-        this.setValue(IntegerPool.get(0));
+        this.setValue((Integer) 0);
```
Veracity0, line 1496: Casts?

src/net/sourceforge/kolmafia/swingui/panel/AdventureSelectPanel.java
```diff
@@ -637,15 +636,15 @@ public AdventureCountSpinner() {
     public void stateChanged(final ChangeEvent e) {
       int maximum = KoLCharacter.getAdventuresLeft();
       if (maximum == 0) {
-        this.setValue(IntegerPool.get(0));
+        this.setValue((Integer) 0);
```
Veracity0, line 639: Casts?

src/net/sourceforge/kolmafia/swingui/widget/AutoHighlightSpinner.java
```diff
@@ -18,7 +17,7 @@ public AutoHighlightSpinner() {
   }
 
   public void setValue(int value) {
-    this.setValue(IntegerPool.get(value));
+    this.setValue((Integer) value);
```
Veracity0, line 20: Casts? Swing widget thing?

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5002,22 +5001,27 @@ public static Value get_related(ScriptRuntime controller, Value item, Value type
       ArrayList<Integer> elems = new ArrayList<Integer>();
       boolean clusters = (pulver & EquipmentDatabase.YIELD_1C) != 0;
       if ((pulver & EquipmentDatabase.ELEM_HOT) != 0) {
-        elems.add(IntegerPool.get(clusters ? ItemPool.HOT_CLUSTER : ItemPool.HOT_WAD));
+        int i = clusters ? ItemPool.HOT_CLUSTER : ItemPool.HOT_WAD;
```
Veracity0, line 5004: I suppose. But "i" is an "itemId", right?

src/net/sourceforge/kolmafia/textui/command/SkeletonCommand.java
```diff
@@ -22,11 +21,11 @@ public SkeletonCommand() {
 
   public static final Object[][] SKELETONS =
       new Object[][] {
-        {"warrior", IntegerPool.get(WARRIOR)},
-        {"cleric", IntegerPool.get(CLERIC)},
-        {"wizard", IntegerPool.get(WIZARD)},
-        {"rogue", IntegerPool.get(ROGUE)},
-        {"buddy", IntegerPool.get(BUDDY)},
+        {"warrior", WARRIOR},
```
Veracity0, line 24: Ya know, these scores of Object[][] are really calling to be Enums or something. Some other PR. :)

src/net/sourceforge/kolmafia/utilities/HashMultimap.java
```diff
@@ -11,14 +10,14 @@
  */
 public class HashMultimap<V> extends HashMap<Integer, ArrayList<V>> {
   public final void put(int key, V value) {
-    Integer okey = IntegerPool.get(key);
+    Integer okey = key;
```
Veracity0, line 13: I initially thought "okay", but it is "o-key", isn't it.

src/net/sourceforge/kolmafia/swingui/widget/AutoHighlightSpinner.java
midgleyc: My guess is that this one predates autoboxing. If you remove the case straight it just calls itself. I think removing the function entirely should work, though.

src/net/sourceforge/kolmafia/utilities/HashMultimap.java
midgleyc: A lot of these can be removed because they'll be inferred using autoboxing.

The trick is knowing which ones are safe to remove, because there is Mafia code that treats ints and Integers differently.

src/net/sourceforge/kolmafia/request/CrimboCafeRequest.java
```diff
@@ -13,22 +12,22 @@ public class CrimboCafeRequest extends CafeRequest {
   public static final Object[][] MENU_DATA = {
     // Item, itemID, price
     {
-      "Peppermint Nutrition Block", IntegerPool.get(-104), IntegerPool.get(50),
```
Veracity0, line 16: A topic for another PR, maybe, but do we want to put pseudo-itemids into ItemPool?

src/net/sourceforge/kolmafia/request/DwarfFactoryRequest.java
```diff
@@ -1093,7 +1092,7 @@ public DwarfNumberTranslator(final String digits) {
 
     private void mapCharacter(final char c, final int i) {
       Character code = Character.valueOf(Character.toUpperCase(c));
-      Integer val = IntegerPool.get(i);
+      Integer val = i;
```
Veracity0, line 1095: Do we need a temporary variable?

src/net/sourceforge/kolmafia/request/UneffectRequest.java
```diff
@@ -308,170 +307,170 @@ public static final void reset() {
...
-    removableEffects.add(IntegerPool.get(EffectPool.SOMEWHAT_POISONED));
-    removableEffects.add(IntegerPool.get(EffectPool.REALLY_QUITE_POISONED));
+    removeWithItemMap.put(ItemPool.ANTIDOTE, removableEffects);
+    removableEffects.add(EffectPool.HARDLY_POISONED);
+    removableEffects.add(EffectPool.MAJORLY_POISONED);
+    removableEffects.add(EffectPool.A_LITTLE_BIT_POISONED);
+    removableEffects.add(EffectPool.SOMEWHAT_POISONED);
+    removableEffects.add(EffectPool.REALLY_QUITE_POISONED);
 
     removableEffects = new HashSet<Integer>();
```
Veracity0, line 318: Outside the scope of this PR, but HashSet<>(), everywhere.

Veracity0 (approved):

Veracity0:
I left a bunch more comments. None are critical, so you can shrug and ignore, if you wish. If you agree and make some minor changes, assume that my approval still stands. I'm not sure I want to scroll thriugh this all yet another time. ;)

src/net/sourceforge/kolmafia/request/UneffectRequest.java
midgleyc: I will file a follow-up that just makes my IDE do it everywhere possible :P

src/net/sourceforge/kolmafia/request/CrimboCafeRequest.java
midgleyc: I think it is a good idea to avoid magic numbers.

src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
```diff
@@ -88,136 +87,131 @@ public class ClanLoungeRequest extends GenericRequest {
...
+          -103,
+          3,
           ItemPool.get(ItemPool.GAMEPRO_MAGAZINE, 3),
           ItemPool.get(ItemPool.DEFECTIVE_TOKEN, 1),
         },
       };
 
   public static final Object[][] SPEAKEASY_DATA =
       new Object[][] {
-        {"glass of &quot;milk&quot;", IntegerPool.get(1), IntegerPool.get(1), IntegerPool.get(250)},
```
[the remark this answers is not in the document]
midgleyc: Some lines are probably very slightly too long.

I assume spotless' default line length is 80, and there is perhaps an argument for making it slightly longer.

src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
```diff
@@ -81,54 +80,54 @@ public class ItemDatabase {
 
   public static Object[][] PUNCHCARDS = {
     // Verbs
-    {IntegerPool.get(3146), "El Vibrato punchcard (115 holes)", "El Vibrato punchcard (ATTACK)"},
-    {IntegerPool.get(3147), "El Vibrato punchcard (97 holes)", "El Vibrato punchcard (REPAIR)"},
-    {IntegerPool.get(3148), "El Vibrato punchcard (129 holes)", "El Vibrato punchcard (BUFF)"},
-    {IntegerPool.get(3149), "El Vibrato punchcard (213 holes)", "El Vibrato punchcard (MODIFY)"},
-    {IntegerPool.get(3150), "El Vibrato punchcard (165 holes)", "El Vibrato punchcard (BUILD)"},
+    {3146, "El Vibrato punchcard (115 holes)", "El Vibrato punchcard (ATTACK)"},
```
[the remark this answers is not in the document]
midgleyc: I'll file a follow-up tonight.

midgleyc:
I will file some follow-ups. I think the ideas are good, but I don't want to scope creep this PR too far.

src/net/sourceforge/kolmafia/request/CrimboCafeRequest.java
midgleyc: Having actually started it... they're used in exactly one place, it's reasonably clear what they are (and should be more so after we stop using `Object[][]`) so I'm less keen.

src/net/sourceforge/kolmafia/request/UneffectRequest.java
midgleyc: IDE has only an option for "in file", so let's keep going with the original method of fixing lints as we edit files containing them.

PR #723, opened 2022-04-22
Update jenkinsfile to build against Java 17

fewyn:
This will make Jenkins start building against Java 17 instead of 16.


jaadams5 (approved):
I answered my own question.  Adoptium's last release was 16.  They have rebranded and rehomed and temurin is the successor ;-)

PR #731, opened 2022-04-26
chore: ignore blame rev for spotless

midgleyc:
GitHub offers [a standardized file to ignore certain revisions when running blame](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view).

Ignore the mass application of spotless.


Veracity0 (approved):
Thank you thank you thank you!

PR #732, opened 2022-04-26
feat: make zap command aware of counts

midgleyc:
Make zap command accept counts of items, and not make requests for 0 items.

While doing this, I noticed that the zap command silently fails if the user doesn't have a wand. There's a check in `ZapRequest` that would cover this, but it isn't hit due to an earlier error case being hit. I left it as it was.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ZapRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ZapRequest.java
@@ -29,7 +29,7 @@ public class ZapRequest extends GenericRequest {
   private static final Pattern OPTION_PATTERN =
       Pattern.compile("<option value=(\\d+) descid='.*?'>.*?</option>");
 
-  private static final Map<Integer, List<String>> zapGroups = new HashMap<Integer, List<String>>();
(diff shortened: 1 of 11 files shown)


Veracity0 (approved):
I confess that I do not understand all the changes to the testing infrastructure, but it looks sound. I will try it if/when I want to test a command without mockin requests. Lets give it a go.

Veracity0 (approved):
Did my previous comment get submitted?

I don't understand the testing framework changes, but they look sound, and I will eventually use them.

Edit. Yes, my comment was submitted, but the github app would not show it to me until I exited and restarted.

midgleyc:
To summarize the testing framework changes:

* First, we did `GenericRequest.sessionId = null` to prevent sending any requests, so the code never entered that branch.
* Now, we set up a fake HttpClient to send the requests, and we can check on the requests that were sent. The response is always null, though. A fallback response for all requests can be set up, but most often I don't.
* The next step would be to set up responses for each call, and also test the parsing of responses.

PR #735, opened 2022-04-29
tests: add test for item trace

midgleyc:
I changed it so you can trace any item, not just those currently in your inventory.

Item Trace had some interesting behaviour: cancelling a previous trace didn't take effect until garbage collection ran. I fixed this too.

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
+++ b/test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
@@ -1,6 +1,7 @@
 package net.sourceforge.kolmafia.textui.command;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.not;
 import static org.hamcrest.core.StringContains.containsString;
 
 import internal.helpers.Cleanups;
@@ -32,5 +33,12 @@ public class ItemTraceCommandTest extends AbstractCommandTestBase {
     String output = execute("");
 
     assertThat(output, containsString("Previously watched items have been cleared"));
+
+    RequestLoggerOutput.startStream();
+    Cleanups cleanups = Player.addItem("hair spray");
+    try (cleanups) {
+      var text = RequestLoggerOutput.stopStream();
+      assertThat(text, not(containsString("itrace")));
+    }
   }
 }
--- a/test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
+++ b/test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
@@ -37,5 +37,4 @@ public class ItemTraceCommandTest extends AbstractCommandTestBase {
 
     assertThat(output, containsString("Previously watched items have been cleared"));
   }
-
 }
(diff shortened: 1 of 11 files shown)


Veracity0 (approved):
I like this.

RequestLoggerOutput looks like a good simplification of the previous "custom stream" feature of RequestLogger, which I just discovered a few days ago.

And allowing Listeners to be unregistered is good new functionality. I looked at that when I wanted to created NamedListeners in my YouRobotManager test suite - for "(potions)" and "(avatar)" - and concluded I didn't need to clean them up since there was no harm if a freshly allocated Listener went out of scope and was cleaned up later - but I can easily envision a Cleanup that registers a NamedListener and unregisters it on close.

test/net/sourceforge/kolmafia/textui/command/PrefTraceCommandTest.java
```diff
@@ -0,0 +1,53 @@
...
+  public void tracesPreferencesChanged() {
+    execute("_VYKEACompanionType");
+
+    RequestLoggerOutput.startStream();
+    Cleanups cleanups = Player.setProperty("_VYKEACompanionType", "couch");
+    try (cleanups) {
+      var text = RequestLoggerOutput.stopStream();
+      assertThat(text, containsString("ptrace: _VYKEACompanionType = couch"));
+    }
+  }
```
Veracity0, line 37: I notice that you are leaving the PreferenceListener active, rather than clearing it via execute("").
Intended?

test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
```diff
@@ -0,0 +1,59 @@
...
+    RequestLoggerOutput.startStream();
+    Cleanups cleanups = Player.addItem("beefy nigiri");
+    try (cleanups) {
+      var text = RequestLoggerOutput.stopStream();
+      assertThat(text, not(containsString("itrace")));
+    }
+
+    String output = execute("");
+    assertThat(output, containsString("Previously watched items have been cleared"));
+  }
```
Veracity0, line 31: This cleans up item listener at end of test.

test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
```diff
@@ -0,0 +1,59 @@
...
+  public void tracesItemsAddedToInventory() {
+    execute("hair spray");
+
+    RequestLoggerOutput.startStream();
+    Cleanups cleanups = Player.addItem("hair spray");
+    try (cleanups) {
+      var text = RequestLoggerOutput.stopStream();
+      assertThat(text, containsString("itrace: hair spray = 1"));
+    }
+  }
```
Veracity0, line 43: This does not clean up item listener at end.

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/textui/command/ItemTraceCommandTest.java
midgleyc: This is to cover the case where an item listener is `unregister`ed without having been registered in the first place -- which happens if the item added has a negative id (i.e. is not a real item).

test/net/sourceforge/kolmafia/textui/command/PrefTraceCommandTest.java
midgleyc: It was intended when I wrote it. The tests that clear it do so to test particular code paths.

But I think you're right, I should be clearing up the listeners. As it stands, whichever test runs last will leak its listeners into the remaining tests.

PR #736, opened 2022-04-29
Add tests for recipe command, improve consistency between `recipe` and `ingredients`

midgleyc:
Consistently don't bold the colon, and add a number to the start of the list even in the "uncreatable" condition.


Veracity0 (approved):
This looks fine.

PR #739, opened 2022-05-02
ape backup code for maximizing umbrella

horrible-little-slime:


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -1414,6 +1416,12 @@ void enumerateEquipment(int equipScope, int maxPrice, int priceLevel)
           this.backupCameraNeeded = true;
         }
 
+        if (id == ItemPool.UNBREAKABLE_UMBRELLA
+            && (this.slots[EquipmentManager.OFFHAND] + this.slots[EquipmentManager.FAMILIAR]
```
midgleyc, line 1420: Probably not familiar given the implementation? Does this work, if the Left-Hand Man is equipped?

Not sure though because Maximizer does shuffle the backup camera into slot 1 despite the whole "accessory3" thing. Familiars are a bit different though.

midgleyc (changes requested):
Can you make a test for the maximizer bit, please?

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
horrible-little-slime: I'm not sure if this is necessary, but I can say that as-is, if I do `maximize monster level percent, equip big stick` with my LHM out, it does not have a fun time. I mean, it doesn't _crash_, it just doesn't work.

midgleyc (approved):

gausie:
We need to make some easier way to add configurable items to the maximiser

PR #742, opened 2022-05-04
Add gooseReprocessed to remember what monsters has been reprocessed by grey goose

libraryaddict:


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9705,6 +9705,20 @@ private static void payActionCost(final String responseText) {
         if (responseText.contains("launches almost all of its body mass at your foe")) {
           // It resets the weight of the Grey Goose to 1 lb.
           KoLCharacter.getFamiliar().setExperience(0);
+
+          // Load the monsters that we've already reprocessed
+          String reprocessed = Preferences.getString("gooseReprocessed");
```
jaadams5, line 9710: I don't understand the mechanics of the goose, but two things concern me.  Is the fact that a monster was reprocessed something that is of interest forever, i.e. across Ascensions?  If not then there needs to be a way to clear this.  Can a goose ever reprocess the same monster, possibly across Ascensions?  If so then adding a monster needs to be conditional.  I consider getting the answers blocking but it is certainly possible that no changes are needed, just education on my part.

jaadams5 (changes requested):

src/net/sourceforge/kolmafia/request/FightRequest.java
libraryaddict: This is cleared in preferences.java on ascension

jaadams5: OK.  The diff context mislead me but you are correct.  I will assume that in context your new code will never be given a duplicate.

jaadams5 (approved):

PR #746, opened 2022-05-07
AutoFilterTextField case sensitivity

holatuwol:
I was trying out something with InputFieldUtilities and a bunch of strings and noticed weird behavior in AutoFilterTextField, where it was skipping over any upper case letters in the match. I'm pretty sure this was an accident back when it was written, but I admit to not knowing whether anything relies on this odd case-sensitive behavior now.


midgleyc (approved):
Seems reasonable given the others all call `.toLowerCase()`.

Also welcome back, hola!

PR #747, opened 2022-05-07
fix: goose reprocess pref does not contain spaces

midgleyc:
As reported in https://kolmafia.us/threads/grey-you-challenge-path.27308/page-2

I don't have a debug log with a third monster being reprocessed, so the test isn't as good as it could be, but still shows the error.

As we never add spaces to the pref, it could split on just `,` instead of `\s*,\s*`, but I left it similar to how it was.


jaadams5 (approved):

Veracity0:
> As we never add spaces to the pref, it could split on just `,` instead of `\s*,\s*`, but I left it similar to how it was.

True. But since properties are user editable, we can't guarantee that the user did not put in some spaces, unlikely as that may be, so it is defensive.

Thanks for fixing this.

PR #750, opened 2022-05-08
Before 10k server requests are sent, reset our connection.

libraryaddict:
https://kolmafia.us/threads/ioexception-goaway-received.27397/#post-167957

### DIFF
--- a/src/net/sourceforge/kolmafia/request/GenericRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenericRequest.java
@@ -194,7 +194,9 @@ public class GenericRequest implements Runnable {
   }
 
   public static void resetClient() {
-    GenericRequest.client.resetClient();
+    if (GenericRequest.client != null) {
+      GenericRequest.client.resetClient();
+    }
   }
 
   private Builder getRequestBuilder(URI uri) {
--- a/src/net/sourceforge/kolmafia/request/GenericRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenericRequest.java
@@ -148,7 +148,7 @@ public class GenericRequest implements Runnable {
   public String redirectMethod;
 
   private static HttpClient client;
-  private final static AtomicInteger clientRequestsSent = new AtomicInteger();
+  private static final AtomicInteger clientRequestsSent = new AtomicInteger();
   private HttpRequest request;
   protected HttpResponse<InputStream> response;
 
--- a/src/net/sourceforge/kolmafia/request/GenericRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenericRequest.java
@@ -31,6 +31,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
@@ -147,6 +148,7 @@ public class GenericRequest implements Runnable {
   public String redirectMethod;
 
   private static HttpClient client;
+  private final static AtomicInteger clientRequestsSent = new AtomicInteger();
   private HttpRequest request;
   protected HttpResponse<InputStream> response;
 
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -1599,6 +1605,12 @@ private boolean sendRequest() {
       RequestLogger.printLine(MafiaState.ERROR, message);
       this.timeoutCount = TIMEOUT_LIMIT;
       return true;
+    } finally {
+      // At 10k client requests, the server will send a GOAWAY exception.
+      // We reset our connection before that to avoid this.
+      if (clientRequestsSent.incrementAndGet() >= 9900) {
```
Veracity0, line 1611: No magic number, please. Make a static constant - and put the comment there.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/GenericRequest.java
libraryaddict: Fixed in https://github.com/kolmafia/kolmafia/pull/750/commits/c7fd8de2136549feb5271298c3970b452d93445d

src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -177,15 +184,19 @@ public static URL getSecureRoot() {
...
 
   private static HttpClient getClient() {
-    if (GenericRequest.client != null) {
-      return client;
-    }
+    // Wrap this in a synchronized block to avoid multiple clients being built in a rare condition
+    synchronized (clientRequestsSent) {
+      if (GenericRequest.client != null) {
+        return client;
+      }
```
midgleyc, line 191: You can shift the "!= null" part outside the synchronization.

midgleyc: On a second note, does it really matter if multiple clients get built? We always return the one attached to the class for new requests, and creating a client is not expensive. With the reset, it would reset X requests earlier than it has to, but over 10k requests I don't see much impact.

But equally with "not null" moved synchronizing shouldn't slow down too much.

midgleyc (changes requested):
I would consider putting this in its own class, something like `ResettingHttpClient`, and having `GenericRequest` store one of those. It should implement `send`, pass on the request to a stored HttpClient, and store a count of requests, creating a new HttpClient as it needs. Not sure if that would work as implemented, but it would divorce the case where `GenericRequest` wants a new client because a relevant preference changed from the case where `GenericRequest` wants a new client because it's hitting the request limit. 

Does this also happen with Cloudflare accessing images? If so it may want to be implemented in `FileUtilities.java`, but I imagine that's safer as all those requests can be retried.

src/net/sourceforge/kolmafia/request/GenericRequest.java
libraryaddict: Moving it outside the block should be harmless in almost all cases, though given it wouldn't be thread safe as client is set to null when reset() is called; It'd best to have the field declared volatile.

After all, http requests are not single thread.

libraryaddict:
Regarding FileUtilities, I would be surprised if it was a non-issue there. But at the same time, as you said it's also pretty lenient there.

It should be only new installations of mafia that could hit that limit.

In regards to GenericRequest. I'm not sure moving it to its own class is needed. Although it could be prettier if this code is used elsewhere; Eg FileUtilities.

If you'd like to make the changes for moving it to its own class, you're more than welcome.

src/net/sourceforge/kolmafia/request/GenericRequest.java
midgleyc: oh yeah, forgot it got reset -- mixed it up with singleton pattern.

This could still lead to a rare NPE, though? Hits the sync block here, confirms non-null, reset runs in another thread and sets the field to null, `getClient` returns null.

jaadams5:
> It should be only new installations of mafia that could hit that limit.

Maybe I'm missing something but since there is a user initiated cache clearing command it seems like the limit would also be approached after the cache was cleared in an old installation.

midgleyc (approved):
shifted resetting out to a separate class, applied to FileUtilities

not tested (to 10k), but the images are served using Cloudflare, and our initial discussion of the issue mentioned that Cloudflare and nginx had the GOAWAY behaviour.

slight change in behaviour: now a reset eagerly creates a new client, instead of lazily waiting for the next call. Requires a slight test change, but avoids the client being null.

PR #751, opened 2022-05-09
Add quick access dialog to avoid having to search through menus to open windows

holatuwol:
https://kolmafia.us/threads/ctrl-k-search-pop-up.27530/


midgleyc (changes requested):
Checking it on my instance, some items are duplicated:

image

The duplicated items are those I've chosen to start in tabs, plus "KoLMafia Forums" (which is present in both "General" and "Help").

As clicking either of the options seems to do the same thing, I think these could be filtered out?

Code + idea seems nice overall.

midgleyc (approved):

PR #754, opened 2022-05-10
If the client isn't static, neither should the AtomicInteger be

libraryaddict:
If the client isn't static, neither should the AtomicInteger be.

Received a GOAWAY just earlier.

It could be the case that this entire fix doesn't actually work. But it can easily be that the wrong client was reset.

And yes, it was on an up to date build.

I think actually testing this would mean spamming KOL ten thousand times, which for obvious reasons isn't really something I'd want to do.


midgleyc (approved):

midgleyc:
Yes, oops, thanks for testing.

PR #759, opened 2022-05-14
Merge bookmarks menu with other menu-related functionality

holatuwol:
### gCLI Toolbar

This feature existed to give you the option to run com gCLI commands from a quick access menu in the relay browser (and then had a sidebar equivalent in the main interface, if you turned it on), but it doesn't display correctly at all anymore with awesomemenu taking over the top menu.

I could be incorrect in my assumption that people are not using it due to the display issues, so if you know people who use it (or if you yourself use it), I can just drop this commit.

### Script Menu

This update trims down the script menu to only show a script manager and the ability to load a script from the local file system, without showing the full folder hierarchy of scripts. Additionally, the MRU list is updated to only include scripts that you accessed from the "Load Scripts..." menu option to address the issue of delegated script invocations appearing in that menu.

### Bookmarks Menu

I thought it would make sense to allow people to bookmark gCLI commands (replacing the gCLI toolbar with being able to select a command from Ctrl+K), and add scripts to quick access separately from the MRU list, which would cover for any functionality that was lost by the previous two changes.


jaadams5:
>gCLI Toolbar

Is this invoked by pressing the link above "Type KoLmafia scripting commands in your browser!" in the RHS relay browser pane?  If so that is used by people who run KoLmafia on one machine but actually play using the relay browser from a different machine.  I seem to recall some people also use it when running headless.  If not, then never mind.

jaadams5:
>Script Menu

Could you elaborate?  The ability to run a script in the gCLI and have it appear in the Script MRU is a Feature.  The ability to select a script in /planting or in a subdirectory of /scripts is also a feature, driven, in part, by the decisions script authors made to avoid name collisions and menu bloat.  The presence of a script that is invoked via "call" from another script is sometimes useful when figuring out whether something had run, or not.  That, however is a pleasant side effect and not a desired Feature :-)  At one point the thinking was that if a script had a main function and was invoked in a way that executed main then it had a place on the Script MRU menu.

jaadams5 (commented):
Code looks OK.  Need some time to decide whether the change in functionality is something I "approve" of ;-)

Veracity0:
gCLI Toolbar.

I never use this in the Relay Browser, but I do have 11 scripts configured for use in the GUI sidebar. I use them All The Time; I have four multis that do not ascend - and almost never use the Relay Browser - simply run VeracityMeatFarm, collecting Meat which they don't use - except when I need to spend it to test a KoLmafia feature. :)

So, I log in, press sidebar button #3, wait for the script to finish, and log out.

For my character who still adventure in the Relay Browser - my main, who still ascends when a new challenge path comes out or, rarely, when the PVP seasonal reward is interesting - I don't use the "awesome menu". I use the old "links" style menu. I never automate adventuring when I ascend - it's all Relay Browser adventuring - but I use the GUI and/or gCLI to handle everything else: item management, equipment, familiars, skill casting, coinmasters, and so on.

I use the "run script" dropdown - which holds relay scripts - for PVP Bookkeeper, and formerly for Psychose-a-matic. I expect that is how people start Guide, although I have never used that script.

For the dropdown menu of scripts with an "exec" button next to it - I never use it, because I use the GUI sidebar. As long as you don't get rid of the latter, I don't much care about it in the relay browser. The default "restore HP" and "restore MP" scripts can be invoked from the charpane - which is how I've always done it.

src/net/sourceforge/kolmafia/swingui/GenericFrame.java
```diff
@@ -475,81 +467,6 @@ public void addCompactPane() {
     this.framePanel.add(this.sidepane, BorderLayout.WEST);
   }
 
-  public void addScriptPane() {
```
Veracity0, line 478: I use this all the time. I don't care if scriptList is used in the Relay Browser, but I depend on it being on the right side of the GUI.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -119,7 +119,6 @@ public OptionsFrame() {
     customDeedPanel.add(new CustomizeDailyDeedsPanel());
     selectorPanel.addPanel(" - Daily Deeds", customDeedPanel, true);
 
-    selectorPanel.addPanel("Script Buttons", new ScriptButtonPanel(), true);
```
Veracity0, line 122: I use this.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -430,94 +428,6 @@ public void contentsChanged(final ListDataEvent e) {
     public abstract void saveSettings();
   }
 
-  private class ScriptButtonPanel extends ShiftableOrderPanel {
```
Veracity0, line 433: I use this.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1671,16 +1609,11 @@ public UserInterfacePanel() {
       // this.toolIcons.addItem( "Use modern toolbar icons" );
       // this.toolIcons.addItem( "Use light toolbar icons" );
 
-      this.scripts = new JComboBox<>();
```
Veracity0, line 1674: I use this.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1634,7 +1572,7 @@ protected class UserInterfacePanel extends OptionsPanel implements Listener {
                   {"allowCloseableDesktopTabs", "Allow tabs on main window to be closed"},
                 };
 
-    private final JComboBox<String> looks, toolbars, toolIcons, scripts;
```
Veracity0, line 1637: I use this.

src/data/defaults.txt
```diff
@@ -139,8 +138,6 @@ global	removeMalignantEffects	false
 global	saveSettingsOnSet	true
 global	saveState
 global	saveStateActive
-global	scriptButtonPosition	0
```
Veracity0, line 142: I use these.

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1711,7 +1644,6 @@ public void actionConfirmed() {
       Preferences.setString("swingLookAndFeel", lookAndFeel);
 
       Preferences.setBoolean("useToolbars", this.toolbars.getSelectedIndex() != 0);
-      Preferences.setInteger("scriptButtonPosition", this.scripts.getSelectedIndex());
```
Veracity0, line 1714: Keep

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1726,7 +1658,6 @@ public void actionCancelled() {
       }
 
       this.toolbars.setSelectedIndex(Preferences.getInteger("toolbarPosition"));
-      this.scripts.setSelectedIndex(Preferences.getInteger("scriptButtonPosition") == 0 ? 0 : 1);
```
Veracity0, line 1729: Keep

Veracity0 (changes requested):

Veracity0:
Regarding the Scripts menu, I'm going to have to think more about this.

Personally, I almost never invoke scripts from the menu; the ones I care about (VMF, missingManuel, smashParty, mostly) are in the script buttons. If I write a script to test something, I invoke it from the gCLI. I have a handful of "utility" scripts - tattoos, trophies, etc. - which I almost never invoke, but find in the Utilities submenu of scripts. Other scripts (supplied by me or other script authors) are either invoked by VMF (beach combing, detective solver, Far Future solver, Bastille, Garden Harvester) or have (rarely used) buttons.

I do not use the scriptMRUList.

Veracity0:
Bookmarks. Huh. I see that you have two kinds:

1) Text - URL or CLI command (which could be an "ashq" command.)
2) File - search the Scripts directory to find/create it.

For me - and for the frono's purported people who want scripts in the relay browser and a headless KoLmafia - having a single button (or relay browser link) to press for my "common" scripts is convenient. Fortunately (for me), most of my scripts use "vprops" for configuration rather than command line arguments. Scripts that require input - like my SpaceGate or BeachComber script, or Ezandora's Bastille or Far Future or KGB Briefcase scripts, or the "combo" beach comb script - will prompt you.

Or, if you have a set of parameters you almost always want, you can make a "text" bookmark.

For me, script buttons (single click invocation) and bookmarks (key press and filtering text field) would be a full replacement for the Script menu - which is unwieldy. Popping up a file selector would be fine for the rare times I want to run an aritrary script.

src/net/sourceforge/kolmafia/swingui/menu/LoadScriptMenuItem.java
```diff
@@ -13,25 +12,70 @@
...
   }
 
-  @Override
-  public String toString() {
-    return this.scriptPath == null ? this.getText() : this.scriptPath;
+  public String getScriptPath() {
+    return this.scriptPath;
+  }
+
+  public static final String getRelativePath(File file) {
```
midgleyc, line 42: For me, on Windows, the relative path of a relay script comes back as `/C:/KOLMAFIA_LOCATION/relay/SCRIPT_NAME.ash` instead of the `SCRIPT_NAME.ash` I think you intended.

Relay scripts don't get added any more by default (it seems) which I like, but I still see them for old scripts (and can make them appear if I call directly by absolute path).

holatuwol: Hm, the logic was originally there to try to preserve as much of the old MRU menu items as possible without modifying their paths, but it doesn't make that much sense in hindsight.

I've simplified the relative path checks, so that ideally, your script would be remembered as `relay/SCRIPT_NAME.ash`. I don't know what relay scripts actually do, so if it's actually ideal that they are not included in MRU, it's a relatively simple fix to make them not be included there, and I can add that before I request another review.

Veracity0: Two kinds of relay scripts:

- If you adventure in the relay browser, want to invoke xyx.php, and xyz.ash exists in the "relay" directory, you script is invoked. It has the opportunity to make the request and then mark up the response any way it wants. Additionally, for choice.php (and place.php and campground workshed items), you can have, for example, choice.1267.php to handle choice #1267 (genie wishing).
- Any script named relay_xxx.ash in the "relay" directory is added to the relay script dropdown on the right side of the topmenu. If you select such a script, it's a standalone app which can make any requests it wants and give you a fancy HTML front end. For example, relay_PVP_Bookkeeper.ash looks up your PVP fite history for the current season, looks at all the minis available, and gives you a table with your win/loss record for each of them, either offensively or defensively.

So, yeah - we don't want any of those appearing on any script history, since it makes no sense to invoke them by themselves.

That said, many such scripts import scripts that are in the scripts menu - and those  generally provide command line support to do this or that.

For example, Ezandora has Bastille Battalion relay support:

choice.1313.ash - choice.1319.ash are the various choices that support that IOTM.
Those import BastilleRelay.ash, which marks up the HTML (importing BastilleRelay.js to detect JS events).
It also calls upon scripts/bastille.ash, to do fancy calculations - which has a main function which parses a command line.

If you invoke bastille.ash directly, I expect it would go on a history list.
But when BastilleRelay imports bastille.ash, the latter's functions are built into the former's namespace, and calling those functions is not invoking the latter as a script.

jaadams5: There was a lot of effort to make sure that everything listed in the MRU list was a script that could be launched from the Scripts menu.  If you could not launch it from there, then it would not be included and that includes relay scripts.  That said there are some bugs.  If you visit a page in the relay browser that has an associated relay override, the override script sometimes appears.  I have turned that into a feature since I find the MRU list serves as a "to do list" or a reminder that I have, or have not, done something.  But it does break the design contract.  Occasionally I find zlib on the list.  But many times I have typed something incorrectly in the gCLI and it has resolved as a request to invoke the script zlib.ash so the gCLI invoked it and hence it ended up on the list.  If one script invokes another using call, and perhaps something else I can't think of then it was included by design.  But if it was just included then it should not appear on the MRU list and if it does that is an unobserved or unreported bug :-)

Veracity0:
I checked out this PR and have been experimenting.
I use the side buttons and do not use the MRU menu.
I have a Mac.

1) The side buttons are present.
- They work work as expected. Good.

2) The Script menu is all I need - Script Manager and the ability to look up and execute a file.
- I looked up and executed bbstyles2.ash, tattoos.ash, and trophies.ash. They worked. Good.

3) I created "file" bookmarks for those three files.
- They appear in the Command-K dialog. Good.
- When I try to execute them from there, they throw an exception. For example:

```
Exception in thread "RelayLoader@/Users/pld/Library/Application Support/KoLmafia/scripts/bbstyles2.ash" java.lang.IllegalArgumentException: Illegal character in path at index 52: http://127.0.0.1:60080/Users/pld/Library/Application Support/KoLmafia/scripts/bbstyles2.ash
	at java.base/java.net.URI.create(URI.java:906)
	at net.sourceforge.kolmafia.webui.RelayLoader.run(RelayLoader.java:87)
Caused by: java.net.URISyntaxException: Illegal character in path at index 52: http://127.0.0.1:60080/Users/pld/Library/Application Support/KoLmafia/scripts/bbstyles2.ash
	at java.base/java.net.URI$Parser.fail(URI.java:2974)
	at java.base/java.net.URI$Parser.checkChars(URI.java:3145)
	at java.base/java.net.URI$Parser.parseHierarchical(URI.java:3227)
	at java.base/java.net.URI$Parser.parse(URI.java:3175)
	at java.base/java.net.URI.<init>(URI.java:623)
	at java.base/java.net.URI.create(URI.java:904)
	... 1 more
```
"Index 52" in that path is the " " character in "Application Support",

4) I deleted those "file" bookmarks and created a "text" bookmarks - "vsg status"
- Invoking it ran my "vsg.ash" script, which does "cli_execute( "VeracitySpacegate " + command );". Good.

5) I made another "text" bookmark - "VeracitySpacegate status" - which sent that (invocation of an ASH script) right to the gCLI, successfully. Good.

One observation about the Command-K menu:

It is sorted case sensitively. Therefore:
- "VeracitySpacegate status" appears between "Untinker Item" and "Violet Fog Mapper".
- "bbstyles2.ash", "tattos.ash", "trophies.ash", and "vsg status" (being lowercase) appear at the end in that order.

That probably doesn't matter; the AutoFIlterTextBox serches case insensitively.

I will resolve the requested change about retaining the buttons.
But the IllegalArgumentException thrown because "Application Support" has a space in it should be fixed.

Everything else is to my taste.

I'll leave it to Frono to comment on whether the MRU list - and no script/folder hierarchy - is to his taste.

jaadams5:
>I'll leave it to Frono to comment on whether the MRU list - and no script/folder hierarchy - is to his taste.

Not working.  It did what I expected with all of the scripts that get run from Login but when I ran a script from the menu list, the list completely disappeared and I had to use a file chooser to pick and run a script.

Veracity0:
I changed the "add (file)" bookmark to take the File (as returned by InputFieldUtilities.chooseInputFile) to use LoadScriptMenuItem.getRelativePath to get rid of all the path stuff, assuming the script is in one of the "executable" directories from which you can "call" a script.

I also made it not try to make a RelayBrowserMenuItem if you tried executing an absolute path starting with a "/". I need to rethink that a bit; I think it was intended to allow you use "/main.php", say, but "/user/..../script.ash" was being caught by that. But trying to turn that into a URI when invoking from a menu caused the GUI to freeze hard when it threw an exception. I could certainly URL encode it, but even if I made it "file:://PATH", is that something we want to pass to the browser?

I think I'll work on the MRU list next. I've never used it. How hard can it be? :)

Veracity0:
ScriptMRUList seems to work for me now.
- I synchronized on it.
- Only add to it in the "call script" command, as before; do not need to update it after invoking that command. Which might have been why it needed to be synchronized before. Hmm.
- Renamed KoLConstants.scriptMList to scriptMRULIst. Because, really.

jaadams5 (approved):

PR #761, opened 2022-05-15
chore: runtime library lints

midgleyc:
In standard fashion, found while I was doing #760.

Worth doing while in the file, but there's a lot so I split them out.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -4789,8 +4789,7 @@ public abstract class RuntimeLibrary {
     KoLConstants.inventory.toArray(items);
 
     for (AdventureResult item : items) {
-      value.aset(
-          DataTypes.makeItemValue(item.getItemId(), true), new Value(item.getCount()));
+      value.aset(DataTypes.makeItemValue(item.getItemId(), true), new Value(item.getCount()));
     }
 
     return value;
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5161,7 +5152,7 @@ public static Value well_stocked(
     if (checkQuant < 6) return DataTypes.FALSE_VALUE;
     if (checkPrice < (2 * ItemDatabase.getPriceById(itemID))) return DataTypes.FALSE_VALUE;
     // get some data
-    SortedListModel<PurchaseRequest> results = new SortedListModel<PurchaseRequest>();
+    SortedListModel<PurchaseRequest> results = new SortedListModel<>();
```
Veracity0, line 5155: Why is this a SortedListModel?
MallSearchRequest takes ```List<PurchaseRequest> results;```
It makes a request of KoL with ```this.addFormField("sortresultsby", "price");```

This method "Assumes sorted by price". Yes. that is the contract of MallSearchRequest.
The "results" list is never shown in a GUI element. It need not be a ListModel.

Why not a simple:
```
List<PurchaseRequest> results = new ArrayList<>()
```

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
jaadams5: I'm going to guess that when I wrote this in 2021 MallPurchaseRequest took something else and when the change happened in 2022 no one revisited this code until now.  Similarly, if the sorting contract was called out in purchase request code, I missed it and so called it out since well_stocked's implementation depended upon that.  I can't see any reason not to do the cleanup.

Veracity0 (approved):

PR #762, opened 2022-05-16
Chat highlighting improvements

libraryaddict:
Highlighting could be loaded multiple times, yet would not be unloaded.
This is generally a non-issue until you remove a chat highlight after reopening mafia's chat relay.

It was also possible to highlight the same string multiple times, the old string is now removed.

Previously, it was possible that highlights that are similar could be removed.

The code was confusing to read, and has been cleaned up a little.

### DIFF
--- a/src/net/sourceforge/kolmafia/chat/ChatFormatter.java
+++ b/src/net/sourceforge/kolmafia/chat/ChatFormatter.java
@@ -1,7 +1,10 @@
 package net.sourceforge.kolmafia.chat;
 
 import java.awt.Color;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.regex.Pattern;
 import net.sourceforge.kolmafia.KoLCharacter;
 import net.sourceforge.kolmafia.KoLConstants;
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/chat/ChatFormatter.java
```diff
@@ -325,11 +328,19 @@ public static final void addHighlighting() {
...
+    addHighlighting(highlight, ChatFormatter.getRandomColor());
+  }
+
+  public static final void addHighlighting(String highlight, Color color) {
     if (highlight == null || highlight.length() == 0) {
       return;
     }
 
-    Color color = ChatFormatter.getRandomColor();
+    removeHightlighting(highlight);
```
midgleyc, line 343: `removeHighlighting` (spelling)

libraryaddict: Well dang, I knew something looked funny.

src/net/sourceforge/kolmafia/chat/ChatFormatter.java
```diff
@@ -359,30 +370,40 @@ public static final void removeHighlighting() {
...
+  public static void removeHightlighting(String selectedValue) {
+    selectedValue = selectedValue.toLowerCase();
 
-        String cleanString = newSetting.toString();
-        cleanString = ChatFormatter.MULTILINE_PATTERN.matcher(cleanString).replaceAll("\n");
-        cleanString = cleanString.trim();
+    String[] patterns = StyledChatBuffer.searchStrings.toArray(new String[0]);
 
-        Preferences.setString("highlightList", cleanString);
+    for (int i = 0; i < patterns.length; ++i) {
```
midgleyc, line 392: I think `for (String pattern : StyledChatBuffer.searchStrings) {` would save having to make a new array.

I'm not certain, but the reason for the original "convert to array, loop over array" is, I think, that the `StyledChatBuffer.searchStrings` can change while being looped over and you could wind up past the end if you loop by indexes. But if you loop by iterators, I don't think that can happen. Not sure though.

libraryaddict: It would definitely throw an error.

midgleyc (approved):

PR #763, opened 2022-05-16
Tests for AdventureSpentDatabase. And fixes.

Veracity0:
The tests revealed four bugs in choice adventures and turns_spent:

- autocrafting kicking in when the choice is finished
- multi-turn choice chains - like the beehive
- fight -> choice (as in Haunted Bedroom) that takes 1 turn, not 2.
- tiles.php (in Hidden Temple) was not counting as a turn when it finished.


midgleyc (approved):

midgleyc:
When I saw 47 files I was expecting something much scarier to review :P

Veracity0:
I just added 28 more files - responseTexts for going through the Hidden Temple to open the Hidden City. :)
I also fixed the turn counting issue from the tile jumping puzzle.

PR #765, opened 2022-05-18
refactor: asdonmartin command

midgleyc:
Similar to #714, add tests and remove an `Object[][]`.


src/net/sourceforge/kolmafia/textui/command/AsdonMartinCommand.java
```diff
@@ -48,23 +60,23 @@ private static String driveStyleName(final int index) {
...
       }
     }
     return -1;
   }
 
   @Override
   public void run(final String cmd, final String parameters) {
-    if (CampgroundRequest.getCurrentWorkshedItem().getItemId() != ItemPool.ASDON_MARTIN) {
+    var workshedItem = CampgroundRequest.getCurrentWorkshedItem();
+    if (workshedItem == null || workshedItem.getItemId() != ItemPool.ASDON_MARTIN) {
```
jaadams5, line 79: Just a personal style preference but I would love using parentheses so I don't have to rely on memory or a web search to confirm that workshedItem of null won't trigger a NPE.  

(workshedItem == null || (workshedItem.getItemId() != ItemPool.ASDON_MARTIN))

That makes it clearer to me :-)  Not a blocking comment, however.

jaadams5 (dismissed):

src/net/sourceforge/kolmafia/textui/command/AsdonMartinCommand.java
midgleyc: happy to, but I think there are other ways to be satisfied in this case:
* there is a test that checks null and passes
* the type of `workshedItem == null || workshedItem.getItemId()` would be `boolean || int` which isn't valid

I was expecting spotless to remove the brackets, but it didn't so I think it's fine to have them, and I've added some

PR #767, opened 2022-05-21
Add preferences maydayContractOwned & cosmicBowlingBallOwned

libraryaddict:
As Map.entry cannot have null as a parameter, an empty string is provided.


midgleyc (approved):
Thanks!

Closes #720.

PR #768, opened 2022-05-21
Better mayday & cosmic ball pref names

libraryaddict:
@midgleyc I realized after seeing Malibu's PR that the name wasn't as good as it should've been.


midgleyc (approved):
sure, but it's not like all the names are consistent everywhere else :P

PR #771, opened 2022-05-22
ChoiceManager refactoring

Veracity0:
ChoiceManager is unwieldy. It is gigantic. I split it into three separate files:

ChoiceManager.java

This contains the various control flow variables - ChoiceManager.handlingChoice, ChoiceManager.lastChoice, ChoiceManager.lastResponseText, and so on - we are familiar with.
It contains the various methods called by GenericRequest - preChoice, postChoice0, postChoice1, postChoice2, visitChoice, registerRequest - to update those variables appropriately.

It also contains the code for automating choices via CHOICE_HANDLER.
All of the choice-specific code has been moved to ChoiceControl, with the exception of the automation code that knows how to choose "appropriate" decisions for specific choices based on current play state.

ChoiceControl.java

This contains the massive switch statements which used to be in postChoice1, visitChoice,  and the other control flow methods called by GenericRequest.

ChoiceAdventures.java

This contains the data for all the choices we support in the GUI (with "whichchoiceXXXX" properties - used for automation) or in the Relay Browser (choice spoilers).

In addition to refactoring the massive file into three coherent modules, I made the following improvements:

- No more storing & passing data around in Object[] and Object[][] structures. We now have actual classes.
- No more array searches of large arrays to locate the choice data for a particular choice number. HashMaps work well.

My testing looks good, so far. The GUI for choices is unchanged. I see choice spoilers in the Relay Browser. I'm going to test more tomorrow (well, today. After I sleep. :)

I'm marking this ready to review, in case anybody wants to get a head start with it and give me comments.


src/net/sourceforge/kolmafia/RequestEditorKit.java
```diff
@@ -1898,31 +1902,29 @@ private static void addChoiceSpoilers(final String location, final StringBuffer
...
-                  "<img src=\"/images/itemimages/magnify.gif\" valign=middle onclick=\"descitem('");
-              spoilerBuffer.append(ItemDatabase.getDescriptionId(item.getItemId()));
-              spoilerBuffer.append("');\">");
-
-              int available = KoLCharacter.hasEquipped(item) ? 1 : 0;
-              available += item.getCount(KoLConstants.inventory);
-
-              spoilerBuffer.append(available);
-              spoilerBuffer.append(" in inventory");
+        ChoiceAdventures.Option option = ((ChoiceAdventures.Option) spoiler);
```
midgleyc, line 1905: probably don't need the cast here? spoiler is already the right type?

src/net/sourceforge/kolmafia/session/WumpusManager.java
```diff
@@ -726,7 +727,7 @@ public static final void decorate(final StringBuffer buffer) {
   public static final String getWumpinatorURL() {
     String code = WumpusManager.getWumpinatorCode();
     String current = WumpusManager.getCurrentField();
-    return "http://www.feesher.com/wumpus/wump_map.php?mapstring=" + code + current;
+    return "https://www.feesher.com/wumpus/wump_map.php?mapstring=" + code + current;
```
midgleyc, line 730: HTTPS here is a different website (which 404s). HTTP is right.

midgleyc: actually, scratch that, I think the website is down in its entirety.

midgleyc (changes requested):
nice to see moving away from `Object[][]` :)

src/net/sourceforge/kolmafia/RequestEditorKit.java
Veracity0: I actually don't need the temporary variable at all; it is only used in the next line to call getItems() on.

Veracity0:
I think this is good to go.

midgleyc (approved):

PR #772, opened 2022-05-23
Allow orb to properly parse pondertext

horrible-little-slime:
resolves the issue brought up here:
https://kolmafia.us/threads/orb-pondering-prefix-trimming.27561/

by accounting for the "some" and "the" articles, as well as providing for support for monsters like "the gunk" whose names unfortunately begin with an article.


midgleyc (approved):
Thanks, nice tests!

gausie:
Worth noting that the monster manual now includes the article for every monster in HTML comments - might be something we could build into a tool for this sort of thing

horrible-little-slime:
I was about to post a thread asking for a monster.article proxy record;
manuel will make spading this easier

PR #773, opened 2022-05-23
Audit fancy ingredients vs concoctions.

heeheehee-kolmafia:
In particular, fancy ingredients indicate that the concoction should
correspond to a fancy CraftingTYpe (MIX_FANCY or COOK_FANCY), and
non-fancy ingredients should correspond to a plain crafting type (MIX
or COOK).

Reported at https://kolmafia.us/threads/27566/ -- thanks TQuilla.

### DIFF
--- a/src/data/concoctions.txt
+++ b/src/data/concoctions.txt
@@ -670,7 +670,7 @@ slap and slap again	MIX	slap and tickle	pretty flower
 
 bungle in the jungle	ACOCK	banana daiquiri	little paper umbrella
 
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Is the difference between "MIX, AC" and "ACOCK" that the latter indicates a recipe takes a turn and the former doesn't?

heeheehee-kolmafia:
Yeah. ACOCK is shorthand for MIX_FANCY, AC.

PR #775, opened 2022-05-25
Enable icon images for windows

libraryaddict:


jaadams5:
Could you comment on KolMafia,java, line 401 which seems to me to already be trying to enable a taskbar icon if supported?  I'd prefer to understand whether this was fixing a bug or adding new functionality before approving.

jaadams5:
Also, would this (and the code in KolMafia.java) be better using the value of the global preference loginWindowLogo and not just hardwiring the Lime Glass?

jaadams5:
Built and ran on my system.  The icon on the top left of the login window and in my system tray did change from the Java cup icon to "something" but that something was mostly gray with a green blob or two.  I don't consider that an upgrade so once we resolve my question about the code in Kolmafia.java there might be some discussion about what makes a good icon for this purpose.

libraryaddict:
> Could you comment on KolMafia,java, line 401 which seems to me to already be trying to enable a taskbar icon if supported? I'd prefer to understand whether this was fixing a bug or adding new functionality before approving.

https://stackoverflow.com/a/51204455

Basically not supported on windows.



> Also, would this (and the code in KolMafia.java) be better using the value of the global preference loginWindowLogo and not just hardwiring the Lime Glass?

That sets the splash image, not the app icon.



> Built and ran on my system. The icon on the top left of the login window and in my system tray did change from the Java cup icon to "something" but that something was mostly gray with a green blob or two. I don't consider that an upgrade so once we resolve my question about the code in Kolmafia.java there might be some discussion about what makes a good icon for this purpose.

Yeah, the scaling is terrible. Poking at other options atm

libraryaddict:
I'm not sure what the dark theme is for honestly, added the images to there to keep inline with wine glass.

jaadams5 (approved):

jaadams5:
My questions have been answered and this is better and seems to work as expected.  I'll push later but just want to see if anyone else cares.  maybe @BadHorseMonkey has some comments with regard to the various L&Fs?

libraryaddict:
The taskbar icon might render differently, but given I'm a windows user I can't exactly test that out.
It might benefit from a multi resolution image? Not sure if that's supported there; But again, can't test it.

PR #784, opened 2022-05-30
feat: blank content types to avoid a Rhino problem

midgleyc:
Relates to https://kolmafia.us/threads/java-bug-on-some-versions-of-java-weird-things-happen-to-strings-with-special-characters.27531

Requires testing on non-Windows platforms.

Uses a blank content-types property file to avoid licensing problems: the [version of the file prior to the introduction of text/javascript](https://github.com/adoptium/jdk/blob/3789983e89c9de252ef546a1b98a732a7d066650/src/java.base/windows/classes/sun/net/www/content-types.properties) is distributed under GPLv2 without the classpath exception: as such, bringing it into KoLMafia would make this a derived work, and the combination would also have to be distributed under GPLv2. Using our own file (with no content) avoids this problem.

After https://github.com/mozilla/rhino/issues/1232 is fixed in some way and the fix released, this code can be removed.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafia.java
+++ b/src/net/sourceforge/kolmafia/KoLmafia.java
@@ -13,6 +13,7 @@ import java.io.RandomAccessFile;
 import java.math.BigInteger;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
+import java.nio.file.Files;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
@@ -111,6 +112,7 @@ import net.sourceforge.kolmafia.swingui.SystemTrayFrame;
 import net.sourceforge.kolmafia.swingui.listener.LicenseDisplayListener;
 import net.sourceforge.kolmafia.swingui.panel.GenericPanel;
 import net.sourceforge.kolmafia.textui.AshRuntime;
+import net.sourceforge.kolmafia.utilities.FileUtilities;
 import net.sourceforge.kolmafia.utilities.LockableListFactory;
 import net.sourceforge.kolmafia.utilities.LogStream;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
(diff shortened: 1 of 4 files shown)


BadHorseMonkey:
This seems like a good fix.  Can we put together a test for this?  I want us to be sure that this is good on our six tested platform/Java version combos and also to catch it when Rhino or Java changes again.

midgleyc:
good idea, that will hopefully obviate the need for manual testing

BadHorseMonkey:
I manually tested it anyway, since I was setting that up when I thought of reasons to make a manual test.  Worked fine on MacOS with JDK 17.0.2 and 17.0.3

I know some JS based scripts use JSON (or so a quick grep tells me), but I think that will be fine.  Do we know if any scripts use binary data files or images/sounds?  I'm not sure it's a real issue, but I'm looking for the edge cases...

I couldn't make it change any behavior on MacOS, but MacOS wasn't exhibiting the symptoms that Win was.  I will approve it, but you may want to find additional reviewers, including the ones who reported the problem.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -130,13 +132,25 @@ public abstract class KoLmafia {
...
 
+  private static void ensureContentTypes() {
+    var contentTypesFile = KoLConstants.DATA_LOCATION.toPath().resolve("content-types.properties");
+    if (!Files.exists(contentTypesFile)) {
+      FileUtilities.loadLibrary(
+          KoLConstants.DATA_LOCATION, KoLConstants.DATA_DIRECTORY, "content-types.properties");
+    }
+    System.setProperty("content.types.user.table", contentTypesFile.toString());
+  }
+
```
BadHorseMonkey, line 153: So, if the file is missing/removed, we just ignore it.

That seems reasonable.

BadHorseMonkey (commented):
Not detecting any harm, either by reading It or by running it.  My platform wasn't affected, so I recommend another reviewer.

Veracity0:
> I know some JS based scripts use JSON (or so a quick grep tells me), but I think that will be fine. Do we know if any scripts use binary data files or images/sounds? I'm not sure it's a real issue, but I'm looking for the edge cases...

Spacegate Sounds:

 https://kolmafia.us/threads/spacegate-sounds.24197/

uses /images/spacegate_activate.mp3

midgleyc:
I tested sounds with the barrel smashing, and it worked fine.

PR #786, opened 2022-05-31
Add coat of paint to mafia defaults

libraryaddict:
Not sure if this is "normal" or a one off tbh. 

To properly check this, would probably need to find every string matching some regex to find every preference set in the code. Then check if it exists in mafia's defaults.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1571,6 +1571,7 @@ user	_clipartSummons	0
 user	_cloudTalkMessage
 user	_cloudTalkSmoker
 user	_coalPaperweightUsed	false
+user	_coatOfPaintModifier	
 user	_cocoaDispenserUsed	false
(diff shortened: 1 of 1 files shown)


Veracity0:
Ha ha. I'm currently working on a PR which will verify that every ChoiceAdventure which is configurable in the GUI via a property has that property listed in defaults.txt. This check will be done at class initialization time. The properties are programatically constructed using the choice adventure number, so searching the entire codebase for property names wouldn't really be the way to go.

I agree that if a property is referenced by name in KoLmafia code, it should be in defaults.txt, if only to indicate this is an "official" property used by KoLmafia itself, not available to be arbitrarily used by user code for some different purpose.

Veracity0 (approved):
This fine. I searched for where it was used:
- set in consequences.txt
- read in InventoryManager.java

PR #787, opened 2022-05-31
More error checking during class initialization of ChoiceAdventures

Veracity0:
ChoiceAdventures has a lot of static initialization: every ChoiceAdventure (configurable via GUI), ChoiceSpoiler (not configurable via GUI, but displayable in Relay Browser), and ChoiceCost (for either) has an object created and added to a map.

There are a number of things that are checked in the constructors. Programmer errors. Some of them, I had as assertions - but upon trying them out, I discovered the following:

- You have to enable assertions by starting your jar with -ea
- If you want to see it, you have to catch AssertionError (which is not an Exception) somewhere and print something
- You get to see one at a time.

I have converted all those assertions to prints to stdout - and saving the choice # in a debugging set, which, if not empty, is printed to the gCLI via RequestLogger. The programmer who inserted the error SHOULD catch it, one way or the other, before releasing the code to the wild. :)

One error not previously checked was to verify that every ChoiceAdventure - which is configured via a property in the GUI - has that property listed in defaults.txt. Even if said default is "0", which means "show in browser".

This would have caught the bug I had which generated incorrect property names for ChoiceAdventures.

As is, it caught 11 ChoiceAdventures with missing defaults, which I fixed here.


src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
```diff
@@ -6052,6 +6110,52 @@ public static final Option findOption(final Option[] options, final int decision
...
+              + ")");
+    }
+    if (duplicateChoiceCosts.size() > 0) {
+      RequestLogger.printLine(
+          "Duplicate ChoiceCosts: ("
+              + duplicateChoiceCosts.stream().map(String::valueOf).collect(Collectors.joining(","))
+              + ")");
+    }
+
+    // System.out.println("ChoiceSpoilers with defaults = " + choiceSpoilersWithDefaults.size());
```
midgleyc, line 6156: rather than having some commented out code, could we have e.g. a `test choices` command that prints them out when run in the CLI?

Veracity0: I will do that. And, per my new top-level comment, I will probably not write a test for it - although I could, if you insist. ;)

Veracity0:
```
> test choicedefaults

There are 54 ChoiceSpoilers with defaults.
{2,3,6,8,9,10,11,12,26,27,28,29,45,86,87,88,89,123,125,127,129,152,162,276,277,278,279,280,281,282,283,284,285,286,287,288,289,290,503,504,505,506,507,549,550,551,552,553,554,583,873,888,889,1392}
There are 12 ChoiceSpoilers without defaults.
{235,240,250,251,252,255,259,264,267,268,914,1063}
```

midgleyc (approved):
no need for a test I think -- this is debugging code.

PR #790, opened 2022-06-01
You can have nulls in an array of Options

Veracity0:
For example, in ChoiceAdventures.dynamicChoiceOptions:

```
      case 781:
        // Earthbound and Down
        result = new Option[6];
        result[0] = new Option("Unlock Hidden Apartment Building");
        result[1] = new Option("Get stone triangle");
        result[2] = new Option("Get Blessing of Bulbazinalli");
        result[5] = SKIP_ADVENTURE;
        return result;
```


midgleyc (approved):
might be nicer this being a map given the array can have gaps but this is a good fix.

Veracity0:
Agreed, but that's another stage of refactoring. This will fix stage open Bug Report.

PR #794, opened 2022-06-02
track Spring Loaded Front Bumper via preference

Tokoeka:
Attempting to fix the bug with Asdon Martin Spring Loaded Front Bumper as listed here:

https://kolmafia.us/threads/banished-monster-list-spring-loaded-front-bumper-tracking-issues.27457/#post-168011


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9312,6 +9312,7 @@ private static void payActionCost(final String responseText) {
 
       case SkillPool.AM_FRONT_BUMPER:
         if (responseText.contains("before flying out of sight") || skillRunawaySuccess) {
+          Preferences.setInteger("_lastSpringLoadedBumperUse", KoLCharacter.getTurnsPlayed());
```
midgleyc, line 9315: This looks like a new preference. Can it be added to `src/data/defaults.txt`?

midgleyc (changes requested):
Could you please update `BanishManagerTest` to cover the new code you've added?

Thanks!

gausie:
I have now understood this and realised that this is not the right fix. If banishes are disappearing incorrectly we need to fix that in the code, not patch around it

PR #795, opened 2022-06-02
Expose get_zap_wand() from KoLCharacter.findWand()

libraryaddict:
Exposes KoLCharacter.findWand() as an ash function.


midgleyc (approved):
Thanks!

PR #796, opened 2022-06-05
Add fix-monster-articles command which fetches articles from Monster Manuel

Veracity0:
When you examine a page in Monster Manuel, the page is parsed and MonsterManuelManager.registerMonster is called with every monster found on it.

I changed that method to extract the "article" and compare it to what is already known for the monster. If it differs, we log a message (as we do for anything discrepancy), and also add an "update" to a map.

I added the "test fix-monster-articles" command which examines every page in Manuel, and writes monsters.txt in your "data" directory, with the Article added to every monster that has one.

I ran this command on my complete Manuel and moved that file to data/monsters.txt

I have a small test suite for MonsterManuelManager.registerMonster.


gausie:
Do we want to consider ommitting a/an where they can be assumed as a default and descriminated by the first letter of the name?

Veracity0:
Unlike plurals, which KoLmafia doesn't "publish" for you, but which you can figure out in various ways, articles are published by KoL in Monster Manuel. If KoL tells us, I see no reason not to use it.

Or are you referring to when we see a new monster for the first time and register a temporary monster? Even there, if we encounter "a new monster", is that the name - with no article - or is "new monster" the name with "a" as the article?

We need Manuel to tell us.

src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
```diff
@@ -801,10 +812,11 @@ public static int fixMonsterAttributes(Map<Integer, Map<Attribute, Object>> upda
     int count = 0;
 
     String filename = "monsters.txt";
+    String outputFilename = "monsters.new.txt";
```
midgleyc, line 815: should this method not be overwriting the original file?

src/data/monsters.txt
```diff
@@ -1107,57 +1107,57 @@ Count Drunkula	1390	drunkula.gif	NOCOPY Atk: 1000 Def: 1000 HP: 8000 Init: 20 P:
...
 Falls-From-Sky	1386	fallsfromsky.gif	NOCOPY Atk: 700 Def: 700 HP: 1500 Init: 50 P: beast	Covers-Your-Head (c0)	Drapes-You-Regally (c0)	Warms-Your-Tush (c0)	Gets-You-Drunk (c0)
 Falls-From-Sky (Hard Mode)	-88	fallsfromsky_hm.gif	NOCOPY NOMANUEL P: beast	Covers-Your-Head (c0)	Drapes-You-Regally (c0)	Warms-Your-Tush (c0)	Helps-You-Sleep (c0)	Quiets-Your-Steps (c0)	Protects-Your-Junk (c0)	Gets-You-Drunk (c0)
-Great Wolf of the Air	1387	wolfoftheair.gif	NOCOPY Atk: 1200 Def: 1200 HP: 5000 Init: 10000 P: beast	Great Wolf's headband (c0)	Great Wolf's left paw (c0)	Great Wolf's right paw (c0)	Hunger&trade; Sauce (c0)
+Great Wolf of the Air	1387	wolfoftheair.gif	NOCOPY Atk: 1200 Def: 1200 HP: 5000 Init: 10000 P: beast Article: The	Great Wolf's headband (c0)	Great Wolf's left paw (c0)	Great Wolf's right paw (c0)	Hunger&trade; Sauce (c0)
 Great Wolf of the Air (Hard Mode)	-87	wolfoftheair_hm.gif	NOCOPY NOMANUEL P: beast	Great Wolf's headband (c0)	Great Wolf's left paw (c0)	Great Wolf's right paw (c0)	Great Wolf's rocket launcher (c0)	Great Wolf's beastly trousers (c0)	Great Wolf's lice (c0)	Hunger&trade; Sauce (c0)
```
midgleyc, line 1111: no "article" entry can mean either:
* the monster has no article
* there was no manuel entry for the monster

the articles for the hard mode "monsters" are the same as those for the normal level.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
Veracity0: KoLmafia is willing to "override" built-in files with versions in the "data" directory.

If we use the same file name, the first time you run it, it reads from the built-in file from the .jar file and writes to "data". You can repeat the process as often as you want in the same session. Subesequent times it will find 0 monsters to change.

If you start a new session without rebuilding your jar, KoLmafia will find the newer file in "data" and use it as an override. If I run the script again, it reads the file from "data" and simultaneously writes the same file. I am unclear exactly why it works this way, but the end result is that rather than stopping at the natural EOF, it continues reading and writing an infinite number of blank lines until I abort the gCLI command.

I considered a couple of solutions:

- Use a "backup" system: if monsters.txt is in "data", rename it to monster.txt.bak (or something) before opening monsters.txt for output.
- Write to a different file name and copy that file to monsters.txt in src/data when it is time to commit.

Considering that even if I did the first approach, KoLmafia would not see it until you start a new session: it decides whether or not to use an "override" when you restart. And if you've rebuild the .jar, it will not use a file in "data".

The second approach seemed simpler. If you want to try out the new file, but it in src/data and rebuild.

If you have a different approach and feel strongly it is better, I'm happy to consider it.
Otherwise, I am content with how it is now.

src/data/monsters.txt
Veracity0: I will be happy to manually add the hard mode Dreadsylvania bosses.

src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
Veracity0: I can force it to always read from the built-in file and write to monsters.txt.
That means that if you restart KoLmafia without rebuilding, it will use the file in data as an override,  but it will still munge the built-in version - until you copy it into src/data.

Let me try that out.

Veracity0: I am happy with that. After running the script, you can quit and restart and KoLmafia will use the file in data as an override and you can see how things worked out.

midgleyc (approved):

PR #798, opened 2022-06-06
Add Journeyman items #10917, #10918

ckb11:


midgleyc (approved):
The change is fine because we're going to squash merge it, but do you think you could reset your `main` branch to the `kolmafia/main` branch? That is, `git reset --hard kolmafia/main`? Each PR you do has every other commit you've made and a bunch of merge commits.

If you set `git config --global pull.rebase true` this should configure `git` to rebase branches instead of merging them when you `pull`, which should avoid this issue in the future.

PR #804, opened 2022-06-08
Track skipped June cleaver choice adventures; switch to tracking fights remaining rather than current charge

spaghetti-squash:
Skipping a June cleaver choice adventure seems to not advance the counter for encounters in terms of cooldown, but does put it on a reduced cooldown for the next one. These numbers have been confirmed by TheDictator, a /dev.

~~One other area to explore is it seems like there is potentially a queue-like mechanic to prevent repeated hits on the same choice adventure for the june cleaver, but also we can burn that bridge when we come to it; it'll take a lot of time and math to determine what information that uses, or if it's just people looking for patterns in a sea of static.~~

There is a queue with a 100% rejection rate and a length of 6. I am very confident that it persists across rollover, but i'm also a very confident person.


midgleyc (changes requested):
please add tests for all the prefs updates, including the queue

midgleyc:
Also, dic's discord comment:

> You get a NC at 60 progress, and gain 60, 10, 6, 5, 4, 3, 2, 2, 2, ... per fight
> Skipping sets you to 45 progress.

midgleyc (dismissed):
you could use something like `.collect(Collectors.joining(",")` to have the queue preference not end in a comma, but I won't block on that because it seems to work.

Thanks!

horrible-little-slime:
> you could use something like `.collect(Collectors.joining(",")` to have the queue preference not end in a comma, but I won't block on that because it seems to work.
> 
> Thanks!

I have nothing else to do today, so I might as well do this.

There are so many types of thing in Java! Wow!

midgleyc (approved):

PR #805, opened 2022-06-08
feat: elemental attacks from wiki element pages

midgleyc:
Add a number of monster elemental attacks from wiki element pages.

Some monsters were using "E:" for the main element without a corresponding attack (e.g. Beaver). Corrected to "ED".


Veracity0:
It would be nice if CustomScriptTest told you which script failed.
I know which one, this time.

test-attack-element-proxy-fields.txt uses the "anglerbush" as a monster with no attack element.
I could have chosen one of many others, but I happened to choose one that now DOES have an EA in this PR.

Veracity0 (approved):

PR #806, opened 2022-06-09
Recognize bad record indexing without throwing exception

heeheehee-kolmafia:
This also updates AshRuntime to print all error messages, not just the
first one.


Veracity0:
This was digging into the weeds of unfamiliar code. I had never looked at all the "Location" stuff and propagating BadTypes and stuff. Was this from Language Server? A project I paid no attention to?

In any case, I really like collecting and displaying multiple errors.

I think the root case was that RecordType should simply return a BadType rather than throwing a ScriptException.

This ASH script:
```
record rec {
  int x;
};
rec a;
int ax1 = a.x;
int ax2 = a[0];
int ax3 = a[ax1];
rec b = new rec;
rec c = new rec(b.x);
rec d = new rec[1];
rec e = new rec[a.x];
```
Now does this:
```
> badrec

Variable 'a' cannot be indexed (badrec.ash, line 6, char 11 to char 13)
Variable 'a' cannot be indexed (badrec.ash, line 7, char 11 to char 13)
Variable 'null' cannot be indexed (badrec.ash, line 10, char 9 to char 17)
Variable 'null' cannot be indexed (badrec.ash, line 11, char 9 to char 17)
```
Now, I wish that "new rec" displayed as "anonymous record" rather than "Variable 'null'".
I may look further at that.

Perhaps putting that script and results into the test/root/scripts and test/root/expected directories will work for CustomScriptTest. I know it will run scripts and display the output. Will it work if the script fails compilation and what we want to see is error messages? Having multiple error messages collected is super convenient.

heeheehee-kolmafia:
> Was this from Language Server?

Yes, this was one of the pieces of Language Server wherein we needed to complete parsing the file in order to highlight all the places where we detected errors, in, say, an editor.

Of course, there's no reason why we couldn't also spit out the errors when trying to actually run the script.

> Will it work if the script fails compilation and what we want to see is error messages?

As you've no doubt discovered, yes -- CustomScriptTest works by injecting a custom RequestLogger stream (which will contain anything a session log would).

Veracity0 (approved):

PR #807, opened 2022-06-09
June Cleaver - Choice 1475 Correction

JamesDowney:
Choice 1475, option 2 was described as Mysticality substats, it should be Muscle substats.

For reference:
https://kol.coldfront.net/thekolwiki/index.php/Hypnotic_Master
From session log:
```
Took choice 1475/2: Mysticality substats
choice.php?whichchoice=1475&option=2&pwd
You gain 400 Strongness```


horrible-little-slime:
Thanks for catching this!

Veracity0 (approved):

PR #808, opened 2022-06-10
When checking price to make, do not assume quantityPossible > 0

Veracity0:
When recursively calculating priceToMake - taking into account the price to buy the ingredients - acquire was assuming that the concoction's quantityPossible was greater than 0. I.e. that you had enough "accessible" ingredients to do the creation.

Considering that we are explicitly assuming we are willing to buy ingredients, this is less than useful.

Especially since this was the place where we compared autoBuyPriceLimit against the total cost of buying all the ingredients.

Removing the quantityPossible > 0 check means we can now get this:

```
> acquire? etched hourglass
...
The average amount of meat spent on components (6,496,048) for one etched hourglass exceeds autoBuyPriceLimit (20,000)
etched hourglass: fail
```

Which is more desirable than buying more than 10,000 grains of sand and making the item, autoBuyPriceLimit be damned.

I also made Long.MAXIMUM_VALUE print as &infin; for debugBuy logging.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/InventoryManager.java
+++ b/src/net/sourceforge/kolmafia/session/InventoryManager.java
@@ -925,7 +925,7 @@ public abstract class InventoryManager {
 
     if (creator != null && mixingMethod != CraftingType.NOCREATE && !scriptSaysBuy) {
       boolean makeFromComponents = true;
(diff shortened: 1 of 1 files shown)


jaadams5 (approved):
I will add "learn how to grep for a unicode character as entered from the keyboard" to my list of things to learn :-)

PR #812, opened 2022-06-11
Check wiki monster ids and pages

midgleyc:
Add a `checkwikimonsters` command that checks pages on the wiki exist and have the same ids as Mafia thinks they should. Old ids presumably came from the Twitch streams when the monster ids were visible, player dev communication, or still being able to fight an old monster after the ids were added to the HTML.

Some monsters changed name as part of this. Generally, I changed Mafia's name only when I was pretty sure Mafia's name was wrong:
* QuickBASIC Elemental -> QuickBASIC elemental
* snakefire in the grass -> snakefire in the grassfire
* Tio Cadáver -> Tío Cadáver


Veracity0 (approved):
Exciting to get all those monster ids!

PR #815, opened 2022-06-12
Initial support for a "journeyman" CLI command.

Veracity0:
This data will be used by a "journeyman" CLI command, whose initial implementation I will supply tomorrow, after sleep.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia;
 
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -82,6 +83,14 @@ public class KoLAdventure implements Comparable<KoLAdventure>, Runnable {
   private static final Pattern ADVENTURE_AGAIN =
       Pattern.compile("<a href=\"([^\"]*)\">Adventure Again \\((.*?)\\)</a>");
   private static final HashSet<String> unknownAdventures = new HashSet<>();
+
+  public static final Comparator<KoLAdventure> NameComparator =
+      new Comparator<KoLAdventure>() {
+        @Override
+        public int compare(KoLAdventure v1, KoLAdventure v2) {
+          return v1.adventureName.compareTo(v2.adventureName);
+        }
+      };
 
   /**
    * Constructs a new <code>KoLAdventure</code> with the given specifications.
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -678,6 +709,11 @@ private void validate1() {
     this.isValidAdventure = true;
   }
 
+  public boolean isAccessible() {
+    this.validate1();
```
midgleyc, line 713: could `isAccessible` contain the logic that `validate1` currently contains (but instead of setting `this.isValidAdventure`, just returning a boolean)? And then `validate1` calls `this.isValidAdventure = this.isAccessible()`?

As-is there's a potential race here, and I think the other way is neater anyway.

src/net/sourceforge/kolmafia/textui/command/JourneyCommand.java
```diff
@@ -0,0 +1,267 @@
...
+    RequestLogger.printLine("Usage: journey" + this.usage);
+  }
+
+  private AscensionClass parseClass(String[] params) {
+    switch (params[1].toLowerCase().trim()) {
+      case "sc":
+        return AscensionClass.SEAL_CLUBBER;
+      case "tt":
+        return AscensionClass.TURTLE_TAMER;
+      case "pa":
```
midgleyc, line 63: while this is what the usage says, I think the standard abbreviation for pastamancer is "P" or "PM".

similarly I think S or SR is more common than SA for sauceror. I would recognise SA though.

midgleyc (changes requested):
Any chance of a test covering the command? I think the lines you pasted in the first post would give a decent amount of coverage.

Veracity0:
> Any chance of a test covering the command? I think the lines you pasted in the first post would give a decent amount of coverage.

I will do that.

src/net/sourceforge/kolmafia/textui/command/JourneyCommand.java
Veracity0: You know, I intended to use PM. So you spotted a bug, not just an unconventional decision. :)
I have seen "S" but not "SR". I think I have seen "SA", but, yeah - "S" is more common.
I will fix both of those. both in the usage string and the code.

src/net/sourceforge/kolmafia/KoLAdventure.java
Veracity0: I will look into this. It does seem cleaner.

Veracity0: I did this. It also makes the code more compact.
There are a couple of places I am suspicious of the current behavior.
Why not simply say that the Airship is accessible if you have given given the quest?
Yes, you have to plant a beanstalk first, but the code is intentionally "giving the between battle script the opportunity to get a bean".

I don't think this PR is the place to think about changing the validation. Just refactoring to be cleaner.

Veracity0:
Test Coverage for the "journey find" command is now comprehensive.
None for "journey zones"; not so easy to check content of massive HTML tables.

midgleyc (approved):
I'll see what I can do after, but I think we can cover:

* the error cases
* for a given character with a given sign: is a signed zone they can reach present (somewhere), a non-signed zone present, and a zone for a different sign not present?

midgleyc:
yes, good job!

PR #816, opened 2022-06-12
feat: take elemental attacks from the wiki

midgleyc:
Make Mafia aware of the elements "shadow" and "bad spelling".

Parse wiki pages for elemental attacks, and update Mafia accordingly.

The parsing picks up "X damage", and so will pick up elemental attacks, auras, thorns etc. Not sure if this is what we want long-term but an elemental aura means you want to defend against that element, so it seems reasonable.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
@@ -66,7 +66,17 @@ public class MonsterDatabase {
         "supercold",
         "ice.gif",
         "#ADD8E6", // lightblue
-        "is Supercold");
+        "is Supercold."),
+    BADSPELLING(
+        "bad spelling",
+        "cookbook.gif",
+        "#800080", // purple
+        "is Bad Spelling. Bad Spelling is weak against dictionaries."),
+    SHADOW(
+        "shadow",
+        "ice.gif",
+        "#808080", // gray
+        "is Shadow.");
 
     private final String name;
     private final String image;
--- a/src/data/monsters.txt
+++ b/src/data/monsters.txt
@@ -12,9 +12,9 @@
 
 # Regular monsters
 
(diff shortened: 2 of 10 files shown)


src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
```diff
@@ -66,7 +66,17 @@ public enum Element {
         "supercold",
         "ice.gif",
         "#ADD8E6", // lightblue
-        "is Supercold");
+        "is Supercold."),
+    BADSPELLING(
+        "bad spelling",
+        "cookbook.gif",
+        "#800080", // purple
+        "is Bad Spelling. Bad Spelling is weak against dictionaries."),
```
Veracity0, line 74: If the monsters with Bad Spelling attacks are, actually, vulnerable against dictionaries, does that mean they are also ED: "bad spelling" - or actually "E: bad spelling"?

That might be a can of worms; Monster Manuel claims they all have no elemental alignment.
I don't think we need to go there. :)

Veracity0 (approved):
I had a comment in the code, but I don't think it blocks merging.

src/net/sourceforge/kolmafia/persistence/MonsterDatabase.java
midgleyc: As-is, it was a bit of a joke with some in-game relation: monsters that attack with bad spelling come from The Valley of Rof L'm Fao, and monsters from the valley are vulnerable to a dictionary. Outside the game, a dictionary might fix bad spelling ;)

I don't actually know what the point of bad spelling attacks are, because you can't resist "bad spelling". The monsters with bad spelling elemental attacks are vulnerable to dictionaries, so possibly this was the game's way of indicating that? This was far before manuel. Bad ASCII Art doesn't have any "bad spelling" attacks but is still vulnerable to the dictionary, but again that one came far after the others.

There are various elements that don't show in Manuel, two of which already existed in Mafia -- supercold and slime. Again, before Manuel, we said that slimes (in the Slime Tube) were of the "slime" element. This didn't really mean anything though.

PR #823, opened 2022-06-18
Add `get_fishing_locations` function

midgleyc:
As requested in https://kolmafia.us/threads/track-floundry-fishing-locations.27641/


test/net/sourceforge/kolmafia/request/ClanLoungeRequestTest.java
```diff
@@ -0,0 +1,53 @@
...
+  @BeforeEach
+  public void init() {
+    Preferences.saveSettingsToFile = false;
+    KoLCharacter.reset("ClanLoungeRequestTest");
+    Preferences.reset("ClanLoungeRequestTest");
+    // don't try to visit the fireworks shop
+    Preferences.setBoolean("_fireworksShop", true);
+  }
+
+  static String html(String path) {
```
Veracity0, line 26: This is a nice method. It could be used all over the place. I have added a version which does the same but doesn't catch the IOException in many test files.

Perhaps you could put it somewhere in internal/helpers. Perhaps even in Player, although that is almost exclusively Cleanups. Almost; somebody but two variants for countItem there, as wrappers for InventoryManager.getCount.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -86,4 +88,27 @@ void zapWandAvailable() {
       assertThat(output, containsString("name => marble wand"));
     }
   }
+
+  @Test
+  void floundryLocations() throws IOException {
```
Veracity0, line 93: For example, this new test right here could use your html() function that catches IOException.

Veracity0 (approved):
I think adding the html() method to a helper library would be helpful all over the place, but I'm not going to block approval. There are a ton of existing tests that could use that, and updating them is way out of scope, but considering you are providing tests in two different files, it seems like your two tests could both use your new helper method.

test/net/sourceforge/kolmafia/request/ClanLoungeRequestTest.java
midgleyc: This reminds me that a while back I added some extra request / response asserts to `AbstractCommandTestBase`, but only used them in one class. These too could do with being used elsewhere.

PR #825, opened 2022-06-19
refactor: use shared test methods in more places

midgleyc:
Follow-up to #823: use `html` and `assertPostRequest` where interior code was used.


Veracity0 (approved):
Love it.

PR #826, opened 2022-06-20
refactor: remove Object[][]

midgleyc:
Many classes in one in this PR. Can split if necessary, but I think they're all quite simple. Simple test for `EdServantCommand` to cover `EdServantData`.


Veracity0 (approved):
I get the feeling that that some previously Object[][] conversions that became classes might have become records, had our source code version been higher.

This is nice and concise.

PR #835, opened 2022-06-29
Organ size should all be zero in Grey You

gausie:
This does that.

I also had problems getting spotless to run, so attempted fixing everything in AdventureResult in an attempt to get my problem to go away. Turns out upgrading spotless sorted it for me, but I kept the improvements that I made to AdventureResult!


src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -857,20 +851,18 @@ public String toConditionString() {
...
       return o.equals(this);
     }
 
-    AdventureResult ar = (AdventureResult) o;
-
     return this.priority == ar.priority
         && this.id == ar.id
         && (this.name == null || ar.name == null
-            ? this.name == ar.name
+            ? Objects.equals(this.name, ar.name)
```
Veracity0, line 865: We know that least one name is null. Therefore,  "==" is a perfectly good check that both of them are null.

Why add a method call?

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -857,20 +851,18 @@ public String toConditionString() {
    */
   @Override
   public boolean equals(final Object o) {
-    if (!(o instanceof AdventureResult)) {
+    if (!(o instanceof AdventureResult ar)) {
```
Veracity0, line 854: I am not familiar with this. Which Java version did this come in?

Veracity0 (changes requested):
How about some tests?

src/net/sourceforge/kolmafia/AdventureResult.java
midgleyc: This is [pattern matching for instanceof](https://docs.oracle.com/en/java/javase/17/language/pattern-matching-instanceof-operator.html): previewed in 14, released properly in 16.

Avoids a manual cast, so very nice in my book.

jaadams5:
Was the spotless upgrade required because of the instance pattern matching or was there some other trigger?  If the former can we think of other tools that may need to be upgraded as we actually start using Java 17 features?

gausie:
> Was the spotless upgrade required because of the instance pattern matching or was there some other trigger? If the former can we think of other tools that may need to be upgraded as we actually start using Java 17 features?

No, spotless just wouldn't run for me at all! I added the pattern matching after the spotless failure (by making changes IntelliJ told me to make hoping I'd fix some underlying spotless issue).

jaadams5:
> > Was the spotless upgrade required because of the instance pattern matching or was there some other trigger? If the former can we think of other tools that may need to be upgraded as we actually start using Java 17 features?
> 
> No, spotless just wouldn't run for me at all! I added the pattern matching after the spotless failure (by making changes IntelliJ told me to make hoping I'd fix some underlying spotless issue).

Thanks.  I had not seen evidence of spotless failures and so wondered if I wasn't paying attention or whether there were environmental differences or what.

gausie:
This is ready for re review!

Veracity0 (approved):

PR #836, opened 2022-06-30
equipped survival knife while Ultrahydrated gives +2 desert exploration

Veracity0:
I decided to give Quantum Terrarium a go.

Today, the May-Day package had a survival knife and a Melodramedary appeared at my side after I had unlocked the desert and bought a UV-resistant compass. :)

I got a response text at the end of a battle where all three of those kicked in, giving me +5 Desert exploration.


gausie (approved):
Looks great! It's a little weird to test all the different combinations with a test file that says "Desert Exploration increased by 5%" in it. Makes me think that we want to add a warning if our calculation doesn't match the number we get from KoL.

Veracity0:
Whoah. Hardcore auto pulls all of your IoTMs when you are out of Standard. Powdered gold for adventures via spleen and the Melodramedary were lucky, but other than that, I mostly ignored my familiar. I freed the king with full organs and 10 turns left on day 2 - including opening the volcanic island for the nemesis quest.

When doing this calculation, the only thing we look at the response text for is to see if it mentions "Desert Exploration", since, apparently, clingy monsters do not advance exploration.

Printing a warning would let us detect new things that we don't support yet. Like the survival knife, before this PR.

Perhaps next run I will collect more response texts for the various other options and we can use them in these tests - and additional tests for the warning.

PR #838, opened 2022-06-30
tests: make hp tests run

midgleyc:
Fix some tests not tagged with `@Test`.


Veracity0 (approved):
No wonder I didn't see any failing tests when I added those. :)

PR #844, opened 2022-07-02
refactor: remove spotless workaround

midgleyc:
Now that we're on at least 6.7.2, we should be able to remove the
`add-exports` calls.

See

https://github.com/diffplug/spotless/blob/d71bc1812f9fbcabb5cda3e82dac11d7da45e1d5/plugin-gradle/CHANGES.md#671---2022-06-10


gausie (approved):
Good spot!

jaadams5 (approved):

PR #845, opened 2022-07-02
Maintain a homebrew tap and cask

gausie:
Since we're already making dmg files, I've made [a Homebrew Tap](https://github.com/kolmafia/homebrew-kolmafia) that will eventually simplify installation for macOS users. Using [Homebrew](https://brew.sh) (the standard dependency manager for macOS), users can just install just by running `brew tap kolmafia/homebrew-kolmafia` and then `brew install --cask kolmafia` and enjoy easy updates on KoLmafia forever!

If this is popular I'll look into doing the same on [Chocolatey](https://chocolatey.org/), which I believe is the equivalent for Windows.

This PR just pokes the tap repo when a new release is completed so that a GitHub action over there can update the version number.


jaadams5 (approved):

jaadams5:
I'm clearly not the target audience for this...

PR #846, opened 2022-07-02
quantum terrarium fixes

Veracity0:
- When parse quantum terrarium, get familiar name, owner name, and owner player id for both current and next familiars.
- Add familiar proxy record field: familiar.owner and familiar.owner_id
- Add new properties: nextQuantumFamiliarName, nextQuantumFamiliarOwner, nextQuantumFamiliarOwnerId
- In CompactSidePane, include owner in hover text for both current and next familiar
- Fix a bug in parsing familiar stuff from charpane/api: when checking the "well fed" attribute of a familiar, when comparing KoL's idea of buffed familiar weight to KoLmafia's do NOT modify the familiar's "weight". That is derived (correctly) exclusively from the familiar experience. The incorrect code had me starting with a 47 pound familiar, rather than a 9 pound one, which was corrected on the next api or charpane refresh.


src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -199,6 +202,8 @@ public FamiliarData(
       final int id, final String name, final int weight, final AdventureResult item) {
     this.id = id;
     this.name = name;
+    this.owner = KoLCharacter.getUserName();
+    this.ownerId = KoLCharacter.getUserId();
```
midgleyc, line 206: Minor, but worth calling `setOwner` here? The method already exists, it more clearly shows these two fields are related.

src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -225,6 +230,8 @@ private FamiliarData(final Matcher dataMatcher, boolean idFirst) {
     FamiliarDatabase.registerFamiliar(this.id, this.race, image);
 
     this.update(dataMatcher);
+    this.owner = KoLCharacter.getUserName();
+    this.ownerId = KoLCharacter.getUserId();
```
midgleyc, line 234: Same as other comment.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/FamiliarData.java
Veracity0: OK.

src/net/sourceforge/kolmafia/FamiliarData.java
Veracity0: OK.

midgleyc (approved):

PR #848, opened 2022-07-03
relay warning to use survival knife.

Veracity0:
- Trying to adventure in the Relay Browser with an unequipped survival knife asks if you are sure, just like UV-resistant compass
- (Note that it does not warn you to get Ultrahydrated in that warning; there is another warning already for that.)
- I also collected HTML and wrote individual tests for each of the following:

1) No bonuses
2) compass
3) knife (hydrated)
4) compass + knife (hydrated)
5) Melodramedary
6) camel + compass
7) camel + knife (hydrated)
8) camel + compass + knife (hydrated)
9) camel + knife (not hydrated)

Interestingly, although the response texts tells you exactly how much desert exploration you did - and the tests verify that that text is present and that we come up with the same increment - we don't actually process that, depending instead on knowing which things give you which bonuses.

If we decide to change that someday, these tests will verfiy that.


midgleyc (approved):
very nice

PR #850, opened 2022-07-03
Actively include Fidoxene when calculating familiar weight

gausie:
Hopefully fixes https://kolmafia.us/threads/26526-breaks-fidoxene-tracking.27683

This also makes me think: the two places I had to make this change are trying to achieve the same thing (I think) - should I look at combining them?

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
@@ -320,9 +320,9 @@ public class EffectPool {
   public static final int ENHANCED_INTERROGATION = 2415;
   public static final int BOXING_DAY_BREAKFAST = 2429;
   public static final int WOLF_FORM = 2449;
-  public static final int FIDOXENE = 2520;
   public static final int MIST_FORM = 2450;
   public static final int BATS_FORM = 2451;
+  public static final int FIDOXENE = 2520;
   public static final int BLESSING_OF_THE_BIRD = 2551;
   public static final int BLESSING_OF_YOUR_FAVORITE_BIRD = 2552;
   public static final int FIZZY_FIZZY = 2561;
--- a/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
@@ -320,6 +320,7 @@ public class EffectPool {
   public static final int ENHANCED_INTERROGATION = 2415;
   public static final int BOXING_DAY_BREAKFAST = 2429;
   public static final int WOLF_FORM = 2449;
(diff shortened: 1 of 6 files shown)


src/net/sourceforge/kolmafia/objectpool/EffectPool.java
```diff
@@ -320,6 +320,7 @@ public class EffectPool {
   public static final int ENHANCED_INTERROGATION = 2415;
   public static final int BOXING_DAY_BREAKFAST = 2429;
   public static final int WOLF_FORM = 2449;
+  public static final int FIDOXENE = 2520;
```
jaadams5, line 323: If we are trying to sort effects by ID this is in the wrong place.

gausie: Thanks! Brain fart

test/net/sourceforge/kolmafia/FamiliarDataTest.java
```diff
@@ -38,6 +41,28 @@ public void canTellIfFamiliarIsNotTrainable() {
...
+      assertThat(familiar.getModifiedWeight(), equalTo(20));
+    }
+  }
+
+  @Test
+  void familiarReportsModifiedWeightCorrectlyDespiteFidoxene() {
+    var cleanups = addEffect("Fidoxene");
+
+    try (cleanups) {
+      var familiar = FamiliarData.registerFamiliar(FamiliarPool.ALIEN, 4000);
```
jaadams5, line 60: I don't know what base weight FamiliarData.registerFamiliar(FamiliarPool.ALIEN, 4000) gives but if it is less than 20 this test seems redundant and if it is greater than 20 shouldn't the modified weight also be greater than 20?

jaadams5:
I can make this an approving review if necessary but I am a little confused about the tests.

test/net/sourceforge/kolmafia/FamiliarDataTest.java
gausie: This is meant to be 400, again sorry. The 400 there is experience not weight, and thus 20 lbs

gausie: Ultimately this is the same as 400 because experience above that number doesn't count towards weight.

midgleyc (changes requested):
Can we have a test checking a homemade robot or similar to confirm that Fidoxene doesn't limit the weight to 20?

gausie:
> Can we have a test checking a homemade robot or similar to confirm that Fidoxene doesn't limit the weight to 20?

Done!

jaadams5 (approved):

midgleyc (approved):

PR #853, opened 2022-07-03
 Generic handling for items configurable via preference: Modeables

gausie:
There are a number of items that are configured via a preference. They all have identical implementation over several files and are generally problematic to implement (see: umbrella, implementation of which was non-obvious to anyone but seasoned contributors).

This is some code I worked on a few months ago to bring that into a new class type. I need to write tests for it but I present it here for feedback and thoughts on my implementation

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/AbstractModeCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/AbstractModeCommand.java
@@ -1,0 +1,25 @@
+package net.sourceforge.kolmafia.textui.command;
+
+public abstract class AbstractModeCommand extends AbstractCommand {
+    /**
+     * Basic side-effect-less command validation
+     * @param command Command string
+     * @param parameters Parameters as a string
+     * @return Whether command is valid
+     */
+    public boolean validate(final String command, final String parameters) {
+        return true;
+    }
+
+    /**
+     * Normalise parameters to the value expected in modifiers.txt
+     *
+     * For example, one might specify "ml" to the "umbrella" command, but the value to compare in
+     * state would be "broken".
+     * @param parameters Input parameters
+     * @return Appropriate state value
+     */
+    public String normalize(final String parameters) {
+        return parameters;
+    }
+}
--- a/src/net/sourceforge/kolmafia/textui/command/AbstractModeCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/AbstractModeCommand.java
@@ -1,25 +1,27 @@
 package net.sourceforge.kolmafia.textui.command;
 
 public abstract class AbstractModeCommand extends AbstractCommand {
-    /**
-     * Basic side-effect-less command validation
-     * @param command Command string
-     * @param parameters Parameters as a string
-     * @return Whether command is valid
-     */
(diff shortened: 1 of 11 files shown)


gausie:
I'm pretty happy that this works now and am taking it out of draft. It came along with a bunch of extra test coverage and improvements elsewhere. I'd appreciate ideas for stress-testing-tests I could add to `MaximizerTest`!

gausie:
I'm finding some issues with this running it so will write some more tests.

src/net/sourceforge/kolmafia/textui/command/AbstractModeCommand.java
```diff
@@ -0,0 +1,32 @@
...
+public abstract class AbstractModeCommand extends AbstractCommand {
+  /**
+   * Basic side-effect-less command validation
+   *
+   * @param command Command string
+   * @param parameters Parameters as a string
+   * @return Whether command is valid
+   */
+  public boolean validate(final String command, final String parameters) {
+    return true;
```
midgleyc, line 14: if this is never called at all and always overridden, `AbstractModeCommand` might be better as an interface.

test/net/sourceforge/kolmafia/textui/command/BackupCameraCommandTest.java
```diff
@@ -0,0 +1,125 @@
...
+      assertThat(requests, hasSize(2));
+      assertPostRequest(requests.get(0), "/inventory.php", "action=bcmode");
+      assertPostRequest(requests.get(1), "/choice.php", "whichchoice=1449&option=3");
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"off", "disable"})
+  public void canDisableReverser(String disable) {
+    // But... why?
```
midgleyc, line 96: why what?

midgleyc (changes requested):
First, I really like how this reveals a good abstraction and removes a bunch of code.

Can we have some more tests, please? I would like to see:
- [x] SnowsuitCommandTests
- [x] RetroCapeCommandTests
* Maximizer tests:
  - [ ] backup camera where it's not the best accessory, but should still be equipped (or where you forcibly discount the other slots, "-acc1 -acc2")
  - [x] something to check the umbrella can go on the left-hand man
  - [ ] something that equips the snowsuit
  - [ ] something that equips the retrocape
  - [x] something where it should leave a slot blank instead of equipping a modeable (e.g. "-hp" for the umbrella)
  - [x] something where it should prefer a different item to a modeable

Consider putting the modeable tests in MaximizerTests (crown, umbrella, backup camera) in a `@Nested` class for organisation purposes.

I can help write the tests if you'd like.

test/net/sourceforge/kolmafia/request/CharPaneRequestTest.java
```diff
@@ -24,6 +24,16 @@ public void beforeEach() {
     CharPaneRequest.reset();
   }
 
+  @Test
+  void canParseSnowsuit() {
```
midgleyc, line 28: nice pick for a test, hadn't thought of this :+1:

midgleyc (approved):

PR #854, opened 2022-07-04
log if KoL's modified familiar weight differs from KoLmafia's

Veracity0:
When I worked on improving the compact sidepane for Quantum Terrarium, I noticed it initially showed the incorrect base weight of the familiar. This fixed itself with the next api request.

I discovered that ApiRequest (and CharpaneRequest) were calling a FamiliarData method to check - and "fix" - if KoL's idea of modified familiar weight disagreed with KoLmafia's. It "fixed" it by changing the familiar's base weight. That is incorrect. The base weight is based on experience - and we track that and adjust as necessary.

If we don't do that, that is a KoLmafia bug.

Similarly, if KoL displays a modified weight which disagrees with our calculation, that is a KoLmafia bug.

I removed the adjustment to the base familiar weight - and, sure enough, a user noticed that we did not support Fidoxene.

The code I removed had a commented out log message. The comment said "The following is informational, not an error, but it confuses people, so don't print it."

I would prefer to have an "infrmational" log message reporting a KoLmafia bug which "confuses" people and convinces them to report the bug than no message - and then they are "confused" when the compact side pane and the browser charpane disagree.

So I put back a log message - and still do not incorrectly adjust the base familiar weight.

### DIFF
--- a/test/net/sourceforge/kolmafia/FamiliarDataTest.java
+++ b/test/net/sourceforge/kolmafia/FamiliarDataTest.java
@@ -23,7 +23,6 @@ import org.json.JSONObject;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
@@ -170,16 +169,16 @@ public class FamiliarDataTest {
 
   @Nested
   class API {
-    @AfterEach
-    private void beforeEach() {
+    @BeforeAll
+    private static void beforeAll() {
       // Other tests add familiars to the character
       // Start clean.
       KoLCharacter.reset("");
       KoLCharacter.reset("familiar data test");
     }
 
-    @BeforeEach
-    private void afterAll() {
+    @AfterEach
+    private void afterEach() {
       // ApiRequest.parseStatus() sets all sorts of stuff
       // Reset the character to eliminate leaks to other tests
       KoLCharacter.reset("");
--- a/test/net/sourceforge/kolmafia/FamiliarDataTest.java
+++ b/test/net/sourceforge/kolmafia/FamiliarDataTest.java
@@ -1,17 +1,30 @@
 package net.sourceforge.kolmafia;
 
+import static internal.helpers.Networking.html;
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/FamiliarDataTest.java
```diff
@@ -154,4 +167,56 @@ public void canHandleGreyGooseCombatSkills() {
     assertFalse(KoLCharacter.hasCombatSkill(SkillPool.CONVERT_MATTER_TO_ENERGY));
     assertFalse(KoLCharacter.hasCombatSkill(SkillPool.CONVERT_MATTER_TO_POMADE));
   }
+
+  @Nested
+  class API {
+    @AfterEach
```
midgleyc, line 173: `@BeforeEach`, or possibly `@BeforeAll`.

test/net/sourceforge/kolmafia/FamiliarDataTest.java
```diff
@@ -154,4 +167,56 @@ public void canHandleGreyGooseCombatSkills() {
...
+    private void beforeEach() {
+      // Other tests add familiars to the character
+      // Start clean.
+      KoLCharacter.reset("");
+      KoLCharacter.reset("familiar data test");
+    }
+
+    @BeforeEach
```
midgleyc, line 181: `@AfterAll`

test/net/sourceforge/kolmafia/FamiliarDataTest.java
Veracity0: BeforeAll. My previous experience was that you couldn't use BeforeAll and AfterAll in Nested test classes.
I was wrong. It does work.

There are other tests in this package which add familiars to KoLCharacter - such as the Fidoxene tests - and before I made this API test clean up after itself, they started failing - because of leakage.

test/net/sourceforge/kolmafia/FamiliarDataTest.java
Veracity0: I want AfterEach - in case we ever put another api.php test in this nested class.
Surely AfterEach will run if there is only one test in the Nested class?
Thus also serving as AfterAll?

midgleyc: I think it will, yes

midgleyc (approved):

PR #856, opened 2022-07-05
Designer Sweatpants

gausie:
Support for July 2022 IotM


src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1016,6 +1018,7 @@ public static final void resetAfterLimitmode() {
     KoLCharacter.resetSkills();
     RequestThread.postRequest(new CharSheetRequest());
     InventoryManager.checkPowerfulGlove();
+    InventoryManager.checkDesignerSweatpants();
```
midgleyc, line 1021: Is this pair of lines worth spinning out into its own method? I presume you've looked through the codebase for wherever powerful glove was called and made sure to call this after -- having a single location would make the next equipment piece like this need adding in only one place.

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -798,6 +798,17 @@ public static final void removeConditionalSkills(final int slot, AdventureResult
         KoLCharacter.removeAvailableSkill("Fire Extinguisher: Blast the Area");
         KoLCharacter.removeAvailableSkill("Fire Extinguisher: Zone Specific");
         break;
+      case ItemPool.DESIGNER_SWEATPANTS:
+        // *** Special case: the buffs are always available
+        // KoLCharacter.removeAvailableSkill("Make Sweat-Ade");
```
midgleyc, line 803: "// Remove combat skills" instead of commented out code?

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -1131,6 +1142,17 @@ public static final void addConditionalSkills(final int slot, AdventureResult it
         KoLCharacter.addAvailableSkill("Fire Extinguisher: Blast the Area");
         KoLCharacter.addAvailableSkill("Fire Extinguisher: Zone Specific");
         break;
+      case ItemPool.DESIGNER_SWEATPANTS:
+        // *** Special case: the buffs are always available
+        // KoLCharacter.addAvailableSkill("Make Sweat-Ade");
```
midgleyc, line 1147: "// Add combat skills" instead of commented out code?

src/net/sourceforge/kolmafia/KoLmafia.java
gausie: I considered this but I don't think it's worth doing until we get another item of this type - I don't want to abstract too early.

gausie: Basically I think we could have a data file that says what equipment provides which skills, and do this all automatically based on whether the skill is a combat skill or not. But I don't want to do that right now.

src/net/sourceforge/kolmafia/objectpool/SkillPool.java
```diff
@@ -534,6 +534,10 @@ public class SkillPool {
   public static final int SPOOKY_VEINS = 27060;
   public static final int SEVEN_FOOT_FEELINGS = 27061;
   public static final int SELF_ACTUALIZED = 27062;
+  public static final int SWEAT_OUT_BOOZE = 7414;
```
jaadams5, line 537: It seems like this was once sorted by ID.  Should these be moved?

jaadams5 (dismissed):

midgleyc (dismissed):

src/net/sourceforge/kolmafia/objectpool/SkillPool.java
gausie: Ha I forget newest != highest id

PR #858, opened 2022-07-06
make getEffectDuration function properly for bind skills

horrible-little-slime:
Right now, getEffectDuration is functioning improperly for Bind X stills. It looks like what's happening is it's seeing a 0 and returning that before it can hit the logic to not return 0.

I've set the value in classskills.txt as -1, because that feels like an appropriate "always treat this as a special case" value, and made it return 0 for pastamancers as well as returning 10 for non-pastamancers. -1 doesn't quite feel right, but sometimes we do things that feel wrong.


gausie:
Please add tests. I think -1 is fine, though you could just remove the early return when a zero is found since its clearly just causing issues instead of reducing unnecessary computation.

gausie (approved):

PR #859, opened 2022-07-06
Track sweat changes in a designer sweatpants fight, and track by item description.

libraryaddict:
I'm unaware of any tests for consequences.txt, though the line appears to work fine.

Marked this PR as a draft until the following two have been checked.

- [x] Parses correctly at 100% sweat
- [x] What happens when we're at full sweaty, or close enough: And the sweat gained from the fight would take us over 100%. Does kol lower the sweat gained, or do we need to cap it ourselves?

The fight request, there are different messages that can be displayed, but we have a handy dandy percentage incremented which we can grab instead of catching every single message.

Though I am unsure here if its possible for other messages to be in the same html tag group. Given we only care about the directly enclosing html tag, I'm reasonably confident this will not be an issue.

One potential issue that does come to mind, is that the fight parsing will load after the charpane refreshes, which would obviously result in an incorrect increment. I'm not sure what scenarios that can occur, other than the obvious lag spike. I suppose it depends on when the charpane is refreshed, if it doesn't refresh naturally and requires the fight to progress to refresh, then this is a non issue.

Though if that is an issue, I'm not sure if there's any good solutions there, other than the obvious of disabling charpane parsing. Though given that fight requests could themselves result in incorrect values, this is also a bad idea.

I'm probably being too dramatic here regardless. But it's an interesting problem if it does exist.


src/data/consequences.txt
```diff
@@ -80,6 +80,7 @@ DESC_ITEM	The Landscaper's leafblower	\+(\d+) Monster Level	_leafblowerML=$1
 DESC_ITEM	Yearbook Club Camera	asked you to photograph.*?<b>(.*?)</b>	_yearbookCameraTarget=$1
 DESC_ITEM	familiar scrapbook	Scraps Collected: <b>([\d,]+)</b>	scrapbookCharges=[stripcommas($1)]
 DESC_ITEM	industrial fire extinguisher	The extinguisher's tank is currently <b>([\d]+)% full</b>	_fireExtinguisherCharge=$1
+DESC_ITEM	designer sweatpants	Your sweatpants are currently (\d+)% sweat-logged.	sweat=$1
```
gausie, line 83: Having no message here means that your sweat is at 0% - want to cover that too?

gausie: Also add a lil test here

gausie (changes requested):

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -671,4 +673,20 @@ public void canTrackBellydancerPickpocket() {
...
     assertEquals(3, Preferences.getInteger("_bellydancerPickpockets"));
   }
+
+  @Test
+  public void canTrackDesignerSweatpants() {
+    var cleanups =
+        new Cleanups(equip(EquipmentManager.PANTS, "designer sweatpants"), setProperty("sweat", 0));
+
+    try (cleanups) {
+      assertEquals(0, Preferences.getInteger("sweat"));
```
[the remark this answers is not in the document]
libraryaddict: Get rid of the cleanup, or the useless assertion that sweat is 0?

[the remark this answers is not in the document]
gausie: Assertion

src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
```diff
@@ -2205,6 +2205,17 @@ public static void parseCleaver(final String desc) {
     }
   }
 
+  public static void parseDesignerSweatpants(final String desc) {
+    Pattern sweatPattern = Pattern.compile("Your sweatpants are currently (\\d+)% sweat-logged.");
```
gausie, line 2209: Ah no this is not exactly what I meant - keep the consequences.txt bit and just make this bit say -> if we can't find "Your sweatpants are currently" set it to zero.

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -671,4 +673,17 @@ public void canTrackBellydancerPickpocket() {
...
+
+  @ParameterizedTest
+  @ValueSource(ints = {7, 12})
+  public void canTrackDesignerSweatpants(int expectedSweat) {
+    var cleanups =
+        new Cleanups(
+            equip(EquipmentManager.PANTS, "designer sweatpants"), setProperty("sweat", 10));
+
+    try (cleanups) {
+      parseCombatData("request/test_fight_designer_sweatpants_" + expectedSweat + ".html");
```
gausie, line 685: These files are now named based on a base sweat of 10. You'd be better here to make a CsvSource of the HTML name (which is renamed to be like "test_fight_gain_x_sweat" etc) and the expectedSweat

test/net/sourceforge/kolmafia/request/GenericRequestTest.java
```diff
@@ -60,4 +60,15 @@ public void learnLocketPhylumFromLocketDescription(String phylum) {
 
     assertThat("locketPhylum", isSetTo(phylum));
   }
+
+  @ParameterizedTest
+  @ValueSource(ints = {100, 0})
```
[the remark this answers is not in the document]
libraryaddict: Changed file names from "_0.html" to "_0_sweat.html"

test/net/sourceforge/kolmafia/request/FightRequestTest.java
libraryaddict: Implemented

src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
gausie: My suggestion is slightly better in terms of pattern compilation I think.

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -671,4 +674,22 @@ public void canTrackBellydancerPickpocket() {
     parseCombatData("request/test_fight_bellydancing_pickpocket_3.html");
     assertEquals(3, Preferences.getInteger("_bellydancerPickpockets"));
   }
+
+  @ParameterizedTest
+  @CsvSource({"gain, 2", "lose, -3"})
```
gausie, line 679: `@CsvSource({"gain_2, 2", "lose_3", -3})` would avoid you having to have lose_-3 in your filename...

PR #862, opened 2022-07-07
snow snow suit shrinks, recalculate adjustments

Veracity0:
2 little fixes:

1) When you log in, KoLmafia refreshes your session. The very first thing it does is to fetch api.php to see if it redirects to valhalla. If it does not so-redirect, it parses the status and learns all sorts of stuff about equipment, effects, current familiar, and so on. Note that it has not looked at your char sheet or terrarium yet. Therefore, if you have Amphibian Sympathy, we don't know it, and cannot accurately calculate the modified weight of your familiar.

Therefore, don't check modified weight if you are currently logging in. Subsequent session refreshes work fine.

2) If a familiar is wearing a Snow Suit, it has Familiar Weight +20. Every five combats (up to 75), that decrements by 1, to a minimum of Familiar Weight +5. We track that in a property, and the modifier uses that property, but we weren't noticing it in the compact side pane because we were not recalculating adjustments.

Now we do so when incrementing that property would actually change the Familiar Weight modifier of the Snow Suit.

### DIFF
--- a/src/net/sourceforge/kolmafia/FamiliarData.java
+++ b/src/net/sourceforge/kolmafia/FamiliarData.java
@@ -23,6 +23,7 @@ import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.EquipmentRequest;
 import net.sourceforge.kolmafia.request.GenericRequest;
+import net.sourceforge.kolmafia.request.LoginRequest;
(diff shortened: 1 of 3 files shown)


gausie (approved):
Looks great. Do you think we could add a test for the FightRequest bit? Shouldn't need any source HTML in theory

gausie:
> Well, "in theory" - except the method takes a responseText. Hmm. I'lll see what I can do with an empty string. :)

Or just pass it any old combat we already have in the fixtures folder!

gausie (approved):

PR #865, opened 2022-07-08
Retire StringArray and IntegerArray

gausie:
I want to turn some of our constants into enums but this was the first thing in my way. There is a note to deprecate these classes anyway so I'm happy to have done so. There's very little of anything "clever" done here, though there are a few. I've mostly just converted these custom arrays over to Arrays (simple) or to Maps keyed by integers (thus LinkedHashMap) depending on how they're used.

I considered moving a lot of the int[] usages over to ArrayList but that snowballs into a huge set of changes.


midgleyc:
Two further comments:

First, I think that `adventureTable` would be better as an object than a list of lists, but that refactor can wait.

Second, there are some small behavioural changes with this PR, but I think they're more in keeping with what the author meant.

For example, in `EquipmentDatabase.java`:
```java
public static final boolean contains(final int itemId) {
  return itemId > 0 && EquipmentDatabase.statRequirements.get(itemId) != null;
}
```

I think this would always return 'true' because `StringArray.get` returns `""` if given an absent key, instead of `null`. But I think the author _intended_ it to check whether or not the value was present.

src/net/sourceforge/kolmafia/persistence/AdventureDatabase.java
```diff
@@ -45,7 +44,8 @@ public class AdventureDatabase {
   public static final Map<String, String> PARENT_ZONES = new HashMap<>();
   public static final Map<String, String> ZONE_DESCRIPTIONS = new HashMap<>();
 
-  private static final StringArray[] adventureTable = new StringArray[4];
+  private static final List<ArrayList<String>> adventureTable =
+      List.of(new ArrayList<>(), new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
```
Veracity0, line 48: If you don't like StringArray[4] (cool) then why not ArrayList<String>[4]?

The code, as exists, KNOWS there are indices 0-3, and you replaced adventureTable[0] with adventureTable.get(0). Why is that an improvement?

If you want to replace the fixed-size (4) array with something, why not make an object with 4 fields, or something, and replace the magic indices with fields names?

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -139,7 +144,7 @@ private static boolean decodedNamesEqual(String name1, String name2) {
   private static final String ITEM_HTML = "itemhtml.txt";
 
   private static final String ITEM_DATA = "itemdata.txt";
-  private static final StringArray rawItems = new StringArray();
+  private static final Map<Integer, String> rawItems = new LinkedHashMap<>();
```
Veracity0, line 147: Why can't this be an ArrayList? Looking at how this is used, we don't iterate over it (although we could). We just get and set by itemId - and ArrayList works fine for that. Or HashMap.

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -1558,7 +1563,7 @@ private static Modifier makeModifier(final String mod) {
 
   private static final String OUTFIT_HTML = "outfithtml.txt";
   private static final String OUTFIT_DATA = "outfitdata.txt";
-  private static final StringArray rawOutfits = new StringArray();
+  private static final Map<Integer, String> rawOutfits = new LinkedHashMap<>();
```
Veracity0, line 1566: Why a Map? Why a LinkedHashMap?

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2250,9 +2255,8 @@ private static void loadScrapeData(final StringArray array, final String fileNam
             currentHTML.append(currentLine);
             currentHTML.append(KoLConstants.LINE_BREAK);
           } while (!currentLine.equals("</html>"));
-
-          if (array.get(currentId).equals("")) {
-            array.set(currentId, currentHTML.toString());
+          if (stringMap.get(currentId).isEmpty() || stringMap.get(currentId) == null) {
```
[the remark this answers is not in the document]
gausie: Oh yeah!

> One comment on nulls.
> 
> A separate comment on the behaviour of the original arrays: if you `set` beyond the bounds of the array, it would populate the intermediate values with the default (either 0 or ""). If you then iterated over or converted to an array, it would keep the zeroes (I believe). I can't find anything that did this, but it's something to watch out for.

Yeah I have tried to keep an eye on this - its why I use Map with getOrDefault where that feature may have been used, otherwise using a List

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -1735,7 +1740,7 @@ public static final String parseOutfitEnchantments(
 
   private static final String EFFECT_HTML = "effecthtml.txt";
   private static final String EFFECT_DATA = "effectdata.txt";
-  private static final StringArray rawEffects = new StringArray();
+  private static final Map<Integer, String> rawEffects = new LinkedHashMap<>();
```
Veracity0, line 1743: Why a Map? Why a LinkedHashMap?

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -1977,7 +1982,7 @@ private static void checkEffectModifierDatum(
 
   private static final String SKILL_HTML = "skillhtml.txt";
   private static final String SKILL_DATA = "skilldata.txt";
-  private static final StringArray rawSkills = new StringArray();
+  private static final Map<Integer, String> rawSkills = new LinkedHashMap<>();
```
Veracity0, line 1985: Why a Map? Why a LinkedHashMap?

gausie:
> Since IntegerArray and StringArray are build around an internal ArrayList, seems like they could be replaced by exactly an ArrayList.

I don't think so and will explain in the next para.

> I'm going to look at the cases where you replaced by a Map. And why a LinkedHashMap rather than a HashMap? You are saying that all the usages depend on a fixed iteration order?

They behave differently when you try to get or set out of bounds. A Map better reflects that behaviour as I can `getOrDefault` in much the same way. I've used LinkedHashMap because it was modelled on an internal ArrayList and I figured that implementation would result in the least change. I think most (/all) situations could be a regular HashMap. We can validate that and move them over afterwards in my mind, though I'm happy to do so now.
 
> What is the benefit of replacing int[] by ArrayList for arrays that are not variable in size?

No reason at all, I just wanted to stop having to do the conversions if it were easy, as I imagine they are more costly than whatever processing power we'd save using an int[]. But as I said, decided that wasn't a good idea to do.

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2264,7 +2268,7 @@ private static void loadScrapeData(final StringArray array, final String fileNam
   }
 
   private static void saveScrapeData(
-      final Iterator<Integer> it, final StringArray array, final String fileName) {
+      final Iterator<Integer> it, final Map<Integer, String> stringMap, final String fileName) {
```
Veracity0, line 2271: And this method right here is the demonstration that you don't need a LInkedHashMap. It does not iterate over the map. It passed in an Iterator over the IDs of a data set - items, effects, outfits, whatever - and fetches  using those indices.

You could pass in an ArrayList<String> or a HashMap<String> and it would generate the output file in exactly the same order, whether or not there are gaps in the data.

src/net/sourceforge/kolmafia/persistence/EquipmentDatabase.java
```diff
@@ -31,23 +32,21 @@
...
 
 public class EquipmentDatabase {
-  private static final IntegerArray power = new IntegerArray();
-  private static final IntegerArray hands = new IntegerArray();
-  private static final StringArray itemTypes = new StringArray();
-  private static final StringArray statRequirements = new StringArray();
+  private static final Map<Integer, Integer> power = new LinkedHashMap<>();
+  private static final Map<Integer, Integer> hands = new LinkedHashMap<>();
+  private static final Map<Integer, String> itemTypes = new LinkedHashMap<>();
+  private static final Map<Integer, String> statRequirements = new LinkedHashMap<>();
```
Veracity0, line 42: Why are these not ArrayLists or HashMaps? They are accessed via .get() and .set(). They are never iterated.

src/net/sourceforge/kolmafia/persistence/ItemDatabase.java
```diff
@@ -52,12 +51,12 @@ public class ItemDatabase {
...
-  private static final IntegerArray priceById = new IntegerArray();
-  private static final IntegerArray nameLength = new IntegerArray();
-  private static final StringArray pluralById = new StringArray();
-  private static final StringArray imageById = new StringArray();
+  private static final Map<Integer, Integer> useTypeById = new LinkedHashMap<>();
+  private static final Map<Integer, Integer> attributesById = new LinkedHashMap<>();
+  private static final Map<Integer, Integer> priceById = new LinkedHashMap<>();
+  private static final Map<Integer, Integer> nameLength = new LinkedHashMap<>();
+  private static final Map<Integer, String> pluralById = new LinkedHashMap<>();
+  private static final Map<Integer, String> imageById = new LinkedHashMap<>();
```
Veracity0, line 59: Even the names of these - xxxById - make you think they are maps. But, again, are they ever iterated? If not, why not just a HashMap?

src/net/sourceforge/kolmafia/persistence/EquipmentDatabase.java
```diff
@@ -31,23 +32,21 @@
...
 
   private static final HashMap<Integer, Integer> outfitPieces = new HashMap<Integer, Integer>();
   public static final SpecialOutfitArray normalOutfits = new SpecialOutfitArray();
   private static final Map<Integer, String> outfitById = new TreeMap<Integer, String>();
   public static final SpecialOutfitArray weirdOutfits = new SpecialOutfitArray();
 
-  private static final IntegerArray pulverize = new IntegerArray();
+  private static final Map<Integer, Integer> pulverize = new LinkedHashMap<>();
```
Veracity0, line 49: HashMap?

Veracity0 (changes requested):
If you want things in a particular order - like in StoreManager - ArrayList is fine.
If you want to add and lookup things based on an id, a Map is fine.
But unless you intend to iterate over the Map expect a particular iteration order, a HashMap is just fine.
Otherwise, LinkedHashMap if they are always inserted in id order. Or TreeMap if you can add items later.
I saw no examples where any of the LinkedHashMaps you created are iterated.

I do see examples where they are given an iterator of keys to use for looking up data - and looking at where those come from, they are keySets of TreeMaps.

Veracity0:
> > I'm going to look at the cases where you replaced by a Map. And why a LinkedHashMap rather than a HashMap? You are saying that all the usages depend on a fixed iteration order?
> 
> They behave differently when you try to get or set out of bounds. A Map better reflects that behaviour as I can `getOrDefault` in much the same way. I've used LinkedHashMap because it was modelled on an internal ArrayList and I figured that implementation would result in the least change. I think most (/all) situations could be a regular HashMap. We can validate that and move them over afterwards in my mind, though I'm happy to do so now.

OK, this is fine. Maps with getOrDefault() are good.
I think I verified that they could all be HashMaps, though. :)

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2250,9 +2255,8 @@ private static void loadScrapeData(final StringArray array, final String fileNam
             currentHTML.append(currentLine);
             currentHTML.append(KoLConstants.LINE_BREAK);
           } while (!currentLine.equals("</html>"));
-
-          if (array.get(currentId).equals("")) {
-            array.set(currentId, currentHTML.toString());
+          if (stringMap.get(currentId) == null || stringMap.get(currentId).isEmpty()) {
```
Veracity0, line 2258: You could reduce this to a single hash table lookup via ```stringMap.getOrDefault(currentId, "").isEmpty()```

Veracity0 (changes requested):
So close. :)

midgleyc (approved):

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
gausie: I'm not convinced the latter case actually happens anymore! But yes I will do this.

Veracity0 (approved):

PR #873, opened 2022-07-09
Add the <main> tag inside <body>, help HtmlDocument recognize the root

libraryaddict:
Java swingx render sometimes can't recognize the true root element, which we fetch here

https://github.com/libraryaddict/kolmafia/blob/82bc912ef32175281cec391678c5f05fe0ac9b8c/lib/net/java/dev/spellcast/utilities/ChatBuffer.java#L481

Adding debugging, it uses the "content" tag until the document is trimmed to size, then it seems to bug out and declares the <table> tag as the root.

The issue might not be fixed with this, however I can't replicate it anymore with this change with what I was previously doing to trigger the issue, namely "prefref" as tables are obvious when they break.

A better fix likely exists.

### DIFF
--- a/lib/net/java/dev/spellcast/utilities/ChatBuffer.java
+++ b/lib/net/java/dev/spellcast/utilities/ChatBuffer.java
@@ -307,11 +307,11 @@ public class ChatBuffer
 
 		htmlContent.append( "<html><head><style>" );
 		htmlContent.append( this.getStyle() );
-		htmlContent.append( "</style></head><body>" );
+		htmlContent.append( "</style></head><body><main>" );
 
 		htmlContent.append( this.content.toString() );
 
-		htmlContent.append( "</body></html>" );
+		htmlContent.append( "</main></body></html>" );
 
 		return htmlContent.toString();
 	}


gausie:
Could you write a bit explaining what the actual issue is, or link to a thread in the forums

libraryaddict:
> Could you write a bit explaining what the actual issue is, or link to a thread in the forums

Did you mean this?

https://i.imgur.com/SNe9xEX.png

gausie:
I'd just like an explanation of the bug that this change fixes. A screenshot with zero context is not that.

gausie:
> I'd just like an explanation of the bug that this change fixes. A screenshot with zero context is not that.

Oh I see, that screenshot suggests that a prefref table at some point in your repeat commanf has been busted . Why does adding a main tag fix this

midgleyc (approved):
I like this, fixes a problem I have when I forget to `ashq` a `visit_url` in the GCLI.

To reproduce the issue:

1. run `ash visit_url("main")`. Observe that the page is dumped into the CLI
2. run `prefref alpha`. Observe you get an empty misaligned table.
3. do some other stuff (e.g. `adv`). Observe that the CLI jumps a lot, canceling out the previous output in favour of showing the main page output again.

libraryaddict:
> > I'd just like an explanation of the bug that this change fixes. A screenshot with zero context is not that.
> 
> Oh I see, that screenshot suggests that a prefref table at some point in your repeat commanf has been busted . Why does adding a main tag fix this

I assume because <main> can only be drawn once, and this adds a root tag. I'm not too sure honestly. As I said, there's probably a better fix out there.

PR #876, opened 2022-07-11
Correctly report last location if you multifight from an itemMonster

gausie:
I was tired of seeing `[1234] null` when I run a relativity fight from a Combat Lover's Locket fight - i.e. a fight without a location.

We have an existing infrastructure for these fights that are triggered without visiting a location. `itemMonster` might not be the best name for it really its there for handling locationless fights, so let's make sure we use it for some of the newer sources of those.

### DIFF
--- a/src/net/sourceforge/kolmafia/RequestLogger.java
+++ b/src/net/sourceforge/kolmafia/RequestLogger.java
@@ -1594,7 +1594,14 @@ public class RequestLogger extends NullStream {
     RequestLogger.updateSessionLog(message);
   }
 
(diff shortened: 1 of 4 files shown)


gausie:
> I didn't see this until it was already approved and merged, but I will point out that all of your new tests are named something like this: `canReportPreviousLocationInMutliFightInLocation()`
> 
> A "mutli-fight" is a new concept. :)

What do you mean? I named it after `inMultiFight` which was definitely not new to this PR

Veracity0:
Look more closely at the spelling.
"Mutli" != "Multi".

PR #877, opened 2022-07-11
feat: write_ccs and read_ccs functions

midgleyc:
Introduce two new functions:
```
buffer read_ccs(string name)
boolean write_ccs(buffer data, string name)
```

Like `buffer_to_file` and `file_to_buffer` can be used to write to the `data` directory, these can be used to write to the `ccs` directory.

The intended users are script writers who want more control over combats. The main ways of controlling combat at the moment are:

1. ship a CCS file with your script, acknowledging that users can change or delete it, and have it handle all combats (or ship many such scripts)
2. use consult scripts
3. use combat filter functions

The first is unpopular because script writers want to do something very dynamic -- different things depending on which IoTMs, skills, and items the user has available. The second is comprehensive, but can be slow -- each combat, the consult script is checked, and the consult script has to handle all potential monsters -- even if the calling script knows exactly which monster is being fought, which items are available, and what it wants to do (e.g. something more complicated than winning the fight, like polar vortex farming smoke patches).

The third is nice, but sometimes unwieldly. Some functions like `adventure` or `adv1` are convenient and take it as a parameter directly, but others like `reminisce` do not. As such, library authors write code using `visit_url`, `run_choice` and `run_combat`, avoiding the convenience methods to gain full control over the combat.

`write_ccs` is intended as a friendlier way of managing this last option: script authors can set a CCS that runs their macro, call the `reminisce` command, then set everything back (or not, as desired).

As an aside: the [wiki article on buffer_to_file](https://wiki.kolmafia.us/index.php/Buffer_to_file) states that strings should be automatically converted to buffers, but this doesn't seem to happen, and scripts using it assume it can't: for example [excavator](https://github.com/gausie/excavator/blob/1d9e96fb7c34b99dc8d61ea30e3708a1cf2b5b0c/RELEASE/scripts/excavator/x_utils.ash#L268) calls `page.replace_string( "", "" )` to get a buffer from a string.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/CcsFileManager.java
+++ b/src/net/sourceforge/kolmafia/textui/CcsFileManager.java
@@ -1,0 +1,58 @@
+package net.sourceforge.kolmafia.textui;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import net.sourceforge.kolmafia.KoLConstants;
+
+public class CcsFileManager {
+
+  private CcsFileManager() {}
+
+  public static byte[] getBytes(String name) {
+    Path path = getPath(name);
+
+    if (path == null) {
+      return new byte[0];
+    }
+
+    if (!Files.exists(path)) {
+      return new byte[0];
+    }
+
+    try {
+      return Files.readAllBytes(path);
+    } catch (IOException e) {
+      return new byte[0];
+    }
+  }
+
+  public static boolean printBytes(String name, byte[] bytes) {
+    Path path = getPath(name);
+
+    if (path == null) {
+      return false;
+    }
+
+    if (!KoLConstants.CCS_LOCATION.exists() && !KoLConstants.CCS_LOCATION.mkdir()) {
+      return false;
+    }
+
+    try {
+      Files.write(path, bytes);
+    } catch (IOException e) {
+      return false;
+    }
+
+    return true;
+  }
+
+  private static Path getPath(String name) {
+    if (name.contains("..") || name.contains("/") || name.contains("\\")) {
+      // forbid both .. and folders
+      return null;
+    }
+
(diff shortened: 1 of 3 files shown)


Veracity0:
Regarding combat filter functions - it is a feature of those that once they have done as much as they intend to, they can just return and let FightRequest continue from the user's CCS. consult scripts can also return before finishing a combat.

Perhaps we should look at (in a different PR) why strings don't auto-convert to buffers?

jaadams5:
Not the solution I would have chosen, but a good compromise means no one is happy, so....

In general, do changes to RuntimeLibrary also require changes to LibraryFunctionStub or is this a special case?  How did you know?  (Asking to increase the chances of me doing the right thing in the future).

If you have reason to edit CssFileManager you might consider moving getPath higher in the file or renaming it something like getAllowedCssPath.  As a reader that would have clued me in earlier that you really were enforcing the path restrictions and that there wasn't some NIO getPath I had forgotten about.

Tests would be appropriate but the absence is not a blocker.  Since Copy/Paste is an overused tool in my toolbox, I wonder if the test for Moods might be a place to start?

Thank you.

jaadams5 (approved):

midgleyc:
I would personally have preferred a "do nothing" battle action, but I couldn't get it to work. This is an incremental improvement that could help and was easy to implement ;)

> In general, do changes to RuntimeLibrary also require changes to LibraryFunctionStub or is this a special case? How did you know? (Asking to increase the chances of me doing the right thing in the future).

I believe it's just functions that take a "buffer" parameter, and want to be able to accept a string. I will look into this as I look into why buffer isn't autoconverting to a string on the ASH side.

> Tests would be appropriate but the absence is not a blocker. Since Copy/Paste is an overused tool in my toolbox, I wonder if the test for Moods might be a place to start?

I agree tests would be good but it was all filesystem based which is annoying to write for. If moods has filesystem-based tests already I'll look into this after work.

PR #878, opened 2022-07-12
Have ChatBuffer write to file with utf-8

libraryaddict:
Noticed it wasn't writing the files with the utf-8 charset when playing with the class, with symbols missing when trying to view the html file.

TCRSDatabase also doesn't write in UTF-8 as seen [here](https://github.com/kolmafia/kolmafia/blob/e0a9e8d358bde820c963b1339fb2ec0932ecda45/src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java#L849). But there's no way to be sure if it should be written in UTF-8 or not as its not ensured to be [grabbed as UTF-8](https://github.com/kolmafia/kolmafia/blob/292b3d1524d4b87c5cd28c520b1be61ff00466c2/lib/net/java/dev/spellcast/utilities/DataUtilities.java#L181), so I've left that alone.

### DIFF
--- a/lib/net/java/dev/spellcast/utilities/ChatBuffer.java
+++ b/lib/net/java/dev/spellcast/utilities/ChatBuffer.java
@@ -41,6 +41,7 @@ package net.java.dev.spellcast.utilities;
 import java.io.File;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -155,7 +156,7 @@ public class ChatBuffer
 		else
 		{
 			boolean shouldAppend = f.exists();
-			this.logWriter = new PrintWriter( DataUtilities.getOutputStream( f, shouldAppend ), true );
+			this.logWriter = new PrintWriter( DataUtilities.getOutputStream( f, shouldAppend ), true , StandardCharsets.UTF_8 );
 
 			ChatBuffer.ACTIVE_LOG_FILES.put( filename, this.logWriter );
 


midgleyc:
`GenericRequest` has a bunch of code where if something is a chat request it encodes / decodes it as latin-1. Is this still true? Is this unrelated?

libraryaddict:
> `GenericRequest` has a bunch of code where if something is a chat request it encodes / decodes it as latin-1. Is this still true? Is this unrelated?

Not entirely sure about that, given that when I checked the chat log, it seems like special characters are encoded.

Eg, https://www.w3.org/MarkUp/html3/latin1.html

What this change would really cover are those messages that were not sent by KoL, such as usage of [Chat_Notify()](https://wiki.kolmafia.us/index.php/Chat_notify)

gausie:
@libraryaddict without a clear understanding of what this is solving I'm going to close the PR. Please feel free to open again with a supporting test that would fail without your change

libraryaddict:
> @libraryaddict without a clear understanding of what this is solving I'm going to close the PR. Please feel free to open again with a supporting test that would fail without your change

@gausie checked through GenericRequest, only the form fields are being encoded & decoded to latin1

And of course, the chat messages are not received in the request fields, but as the request body data.

The chat messages are sent in latin1 to be clear, but that's not what the chat buffer is parsing.

To rephrase.
GenericRequest only has fields encoded/decoded. The response text is UTF-8. Chat only looks at response text to parse the messages. So there was a few misunderstandings. The server may receive the messages as latin1, but it sends messages as UTF-8.

https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/request/GenericRequest.java#L2061

gausie:
Ok! If you can add a failing test that this code fixes I'd be happy to
reopen the PR

libraryaddict:
Added a test https://github.com/libraryaddict/kolmafia/tree/write_utf8

However, this PR is closed and you can't see the changes.

PR #879, opened 2022-07-12
do not auto focus gCLI when execute a script from a menu

Veracity0:
This is that part of PR 759 which reverted the unpopular behavior from PR 751


jaadams5 (approved):
I'm not certain I understand the context, unless is intended for submission before #759 but it looks good otherwise.

Veracity0:
> I'm not certain I understand the context, unless is intended for submission before #759 but it looks good otherwise.

That is the sole point.

759 will come soonish, but this will let people - like you - see if solves the windows bouncing around issue, which you suspect was caused by this.

jaadams5:
OK.  Built, checked and no focus changes.  Thank you.

PR #880, opened 2022-07-12
tests: add tests for ccs file manager

midgleyc:
Follow-up to #877: add tests for CcsFileManager.

Used the existing file system, couldn't find existing file system-related tests (read / write).

Didn't move `getPath`: I prefer to have private methods below public ones, so it being where it is is about the only place it can go.


jaadams5 (approved):

jaadams5:
Thank you.  Moving or renaming was a readability preference so no concerns about that.

I seem to remember doing a lot of work that built a faux file system for testing but maybe it didn't get submitted or wasn't where I thought it was.  In any event, thank you.

PR #881, opened 2022-07-12
haveEquipped Alice Army Card Failing Test

Tokoeka:
Noticed that have_equipped() counts the Alice Army Card in the card sleeve as equipped even when the card sleeve itself is not equipped. 

Haing done some digging around, this behaviour also seems to be the case for the stickers on the scratch 'n' sniff sticker sword, and the folders in the folder holder. These are all $slots that are technically within equippable items. 

However, in my testing, it does not appear to be the case with regards the skin & spur that are equipped to the "your cowboy boots" relevant slots - these appeared to never count as equipped, even when the "your cowboy boots" were equipped.

Not sure whether this is expected or wanted behaviour, but figured I'd query it, as it feels like it should be made consistent in some fashion at least. I would definitely prefer things in these equippable slots do not count as equipped when the equipment holding the slot is not equipped, but *shrugs*

Only managed to do a failing test as my Java is rudimentary at best and I am still trying to get some way to understanding the mafia codebase


Veracity0:
Nice start. I assume you have a development environment set up and can actually build.

Your code is failing code formatting standards.
Fix it like this:

./gradlew spotlessapply

Run just your (file of) tests:

./gradlew test —tests=EauipmentManagerTest

gausie:
> Also, I like assertTrue and assertFalse as appropriate.

Haha I prefer assertThat! Though I do wish there was a `isTrue` matcher

jaadams5:
> > Also, I like assertTrue and assertFalse as appropriate.
> 
> Haha I prefer assertThat! Though I do wish there were an `isTrue` matcher

Any test that increases coverage, logically or as measured, is fine whether it uses my preferred idiom or not :-)

That said, I tend to shy away from "==" because I have dealt with buggy code with a typo that used "=".  Similarly I think about x.equals() because I have tried to use it when x is null.  So assertEquals and assertNotEquals are my preference if equal is the comparison.  If I already have a boolean I will use assertTrue or assertFalse.  If I need anything else the complexity of the test is likely to be pushing the bounds of my expertise :-)

gausie:
In any case what do we think about this @kolmafia/staff. Is this intended behaviour? I can see both sides to be honest.

Veracity0:
Me too. It is the only way to tell which card is in your card sleeve. If you don’t have the card sleeve equipped, we already do not count the card in your active modifiers.

If a script is considering using a card sleeve, it already has to know that you have one and be willing and able to equip and configure it.

The maximizer can do that. But if it doesn’t know what card is currently in the sleeve, how will it deal with it?

What benefit is had by declaring a card to be unequipped unless the sleeve itself is equipped?

If you want consider which cards are available, in order to see which one is in the sleeve, you’d  have to equip it in order to look.

I don’t think that is an improvement.

jaadams5:
If I had concept of something being equipped I would expect it to mean that the thing was equipped and that is was providing "benefits".  So I am inclined towards

> things in these equippable slots do not count as equipped when the equipment holding the slot is not equipped

understanding that

>What benefit is had by declaring a card to be unequipped unless the sleeve itself is equipped?

is answered with some vague statement about consistency and meeting my expectations given the name of the function.

If there is some idiomatic use where equipment is checked in order to determine some kind of configuration I am not in favor of breaking that but would be interested in a FR that provided a more direct answer.

Veracity0:
Additional thought.

This is from api.php:
```
"equipment":{"hat":"10804","shirt":"10702","pants":"10596","weapon":"10920","offhand":"8437","acc2":"5460","acc3":"1795","container":"10242","familiarequip":"9138","fakehands":0,"cardsleeve":0},
"stickers":[0,0,0],
"folder_holder":["22","26","20","00","00"],
```
I have no sticker sword, my card sleeve is empty and in storage, and my folder holder is in storage.

I pulled the card sleeve, slid a card into it.
I pulled the folder holder.
I made a sticker sword.

```
"equipment":{"hat":"10804","shirt":"10702","pants":"10596","weapon":"10920","offhand":"8437","acc2":"5460","acc3":"1795","container":"10242","familiarequip":"9138","fakehands":0,"cardsleeve":"4993"},
"stickers":["3514","3509","3511"]
"folder_holder":["22","26","20","00","00"],
```
Notice that KoL tells us what is in the card sleeve, folder holder, and sticker sword - even though none of them are equipped.

gausie:
Great! Thanks for the PR @Tokoeka but we're closing as Not a Bug

PR #882, opened 2022-07-13
use correct modifiers parsing method for voting

Veracity0:
Local Voting has Experience (familiar) and Experience (Muscle), at least, on occasion.
Since they are not variable, we can use Modifiers.parseModifiers, rather than Modifiers.evaluateModifiers.
This will look up modifiers by tag, rather than name.

Now that I think about it, why does Modifiers.evaluateModifiers not do that, too?
It splits a String into a ModifierList. The modifiers can be either names or tags.
But when we turn a ModifierList into a Modifiers object, we look up the modifier index by name.
It's easy enough to add the tag to the same nameToIndex hash table and be able to look up the index by whichever one.

I also added a catch for a bad URI string in RelayLoader. This is defensive; we supposedly always give a good input, but recent in-development code gave a bad input and an exception was thrown. In the Swing event thread. Oops.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -5123,9 +5123,7 @@ public abstract class KoLCharacter {
 
     // Mummery
     newModifiers.add(
-        new Modifiers(
-            "Mummery",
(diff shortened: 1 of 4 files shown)


midgleyc (approved):

gausie (approved):
This is much better thank you

PR #883, opened 2022-07-14
Track Homemade robot weight

gausie:
Homemade Robot has weird weight tracking that was not captured by Mafia but papered over by our automatic familiar weight adjustment. The experience is not considered (it doesn't gain it from combat but even experience gained outside of combat is ignored).

If anyone complains of a mismatch reported in their CLI after this is merged, they just need to `set homemadeRobotUpgrades = <however many they've used, probably 9>`. Unfortunately it's not easy to reliably infer what a missing value should be - we could detect familiar weight not accounted for but it could be from anywhere.

I also refactored the test file for this (I think it was amongst our first unit tests for mafia!) so that I could more reliably use BeforeEach/BeforeAll.


test/net/sourceforge/kolmafia/request/UseItemRequestTest.java
```diff
@@ -1,123 +1,179 @@
...
-    // Simulate logging out and back in again.
-    KoLCharacter.reset("");
-    KoLCharacter.reset("milk user");
-    // Reset preferences to defaults.
-    KoLCharacter.reset(true);
-    // This shouldn't be necessary if reset does what is expected but....
-    Preferences.setBoolean("_milkOfMagnesiumUsed", false);
-    AdventureResult.addResultToList(
-        KoLConstants.inventory, ItemPool.get(ItemPool.MILK_OF_MAGNESIUM));
+  @BeforeEach
```
midgleyc, line 28: using cleanups, I think this should be fine as `BeforeAll`.

midgleyc (approved):

midgleyc (approved):

PR #889, opened 2022-07-15
Script cascading menus

Veracity0:


src/data/defaults.txt
```diff
@@ -1222,6 +1222,7 @@ user	schoolOfHardKnocksDiplomaCost	1000
 user	schoolSwagger	0
 user	scrapbookCharges	0
 user	screencappedMonster	
+user	scriptCascadingMenus	false
```
jaadams5, line 1225: The previous paradigm was that the default was cascading menus built from the file system and all of the control was opting in or out of scriptMRU.  A new preference is fine since if I understand the Control-K changes (which I have no personal interest in and so have not tried yet) they are now the new default.  But we should check to make sure that if a user sets both the menus and the MRU they get a consistent experience - one or the other - and not an unintended hybrid.  My preference GUI work, when finished, could probably be extended or repurposed to handle either or but not both.  We may have to go to a preference whatGoesBelowTheLoadScript option that has three possible values - CK, Cascade or MRU and make scriptMRULength strictly the length and not the control.  Anyway thanks for doing this.

Veracity0: I thought about having one control for "what goes below Load Script" but I didn't want to interfere with your pending Preference work. :)

As coded,

if ScriptMRULength <= 0 && scriptCascadingMenus is false, you get nothing
if scriptMRULength > 0, the list of files shown is ONLY from the MRU list
otherwise (scriptCascadingMenus is true), the list of files is from the "scripts" directory

midgleyc, line 1225: I think defaulting this to true would be better -- I think it's more user friendly in two ways:
* showing all the scripts instead of nothing gives users another way to interact with scripts
* showing the cascading menu was the default before the bookmark change went in, so users don't have to change a setting to get back to the behaviour they were used to.

As there's no GUI for this at the moment (but perhaps there will be 😉), I think it's nicer for this to default to on. Your choice, though.

midgleyc (approved):

src/data/defaults.txt
jaadams5: User friendly depends upon which users you listen to :-)

Do we set the default for the existing users who are openly hostile to Bookmarks or do we set it for the new user who presumably should be encouraged to use Bookmarks?

How much does this default matter if we are likely to come up with something that lets a user select which one of the three options they want and easily switch between them?

I'm Ok with false.

jaadams5 (dismissed):

src/data/defaults.txt
Veracity0: What the heck. I defaulted it to 2. It will shut up the openly hostile users, at least. :)

Veracity0:
I just fixed an existing bug with our handling of scriptMRUList

We have three modes:

scriptMRULength == 0 and useCascadingMenus == false
scriptMRULength > 0
useCascadingMenus == true

You can now switch among any of those options and we will rebuild the menu to agree with the new configuration.
This means changing scriptMRULength from 0 to 10 will actually show you your save MRU list now. :)

midgleyc (approved):

PR #892, opened 2022-07-16
feat: add `git_` equivalents to `svn_` ASH functions

midgleyc:
Follow-up to #849.

There are various `svn_` ASH functions, of which `svn_exists_at_head` is currently know to be used. Add `git_` equivalents doing similar things.

I thought about adding a `script_` equivalent that would look over both folders, but for most of these functions you need to know the exact project name, and it's likely to be different for svn or git installations, so this wouldn't be terribly useful.


jaadams5 (approved):
No glaring issues ;-)  Thanks.

PR #893, opened 2022-07-16
Preference GUI for scriptMRULength

jaadams5:
Goal is to add a GUI element that lets the user set scriptMRULength.

I'm stuck.  My ability to write Swing is only marginally better than that of a monkey randomly pounding a keyboard.  I tend to copy and paste from a working example until what I have works.

My immediate problem is that I am expecting actionConfirmed() to be called.  Since some of the other preference screens don't have an explicit button to save the change.  I was hoping extending OptionsPanel would give me that when focus was shifted or the window was closed.  But that doesn't seem to be happening.

Do I need to do something with focus, perhaps overriding GenericFrame focusGained() and focusLost()?

Something else?

Of lesser importance, but on my system there is a scroll bar that isn't really functional and the "help" text gets cut off at the bottom.  I may have completely fouled up sizing or it may be that addStatusBarToFrames = true is interacting in a way I don't understand.

At present when this code actually sets the preference a user will still have to restart KoLmafia to see any change.  I expect to add a listener and trigger a refresh to address that but I'm not even looking there until I get this to work.

Thanks for the help.


Veracity0:
I munged it by looking at SVNPanel (which has JTextArea and PreferenceCheckBox elements) and MaximizerOptions (which has JTextField elements) and ActionVerifyPanel (which has VerifiableElements which combine a JLabel and another component ).

I think it is simpler, being based only on JPanel.
Something I did not figure out is why the vertical spacing only works if the checkbox is above input field.
Maybe you can treat that as a Swing learning experience. :)

I also made it support both scriptMRULength and scriptCascadingMenus.

Yes, you needed to implement FocusListener.

Veracity0:
I figured out how to make a PreferenceIntegerTextField widget which will listen for and update a single preference that is associated with a text field.

Veracity0:
By the way - I also learned (some) Swing by copying snd modifying existing code which did much (or some) of what I wanted.

I respectfully suggest that you needed to look harder and find some better models to copy and modify.

Veracity0 (approved):
Looks good to me. ;)

PR #897, opened 2022-07-17
feat: un-unspade some spleen items

midgleyc:
Consumables tend to stay `Unspaded`. Go through the spleen items and update the ones we know about.

### DIFF
--- a/src/data/spleenhit.txt
+++ b/src/data/spleenhit.txt
@@ -4,8 +4,8 @@
 
 # Item	Spleenhit	Level Req	Quality	Adv	Musc	Myst	Moxie
 
-[experimental crimbo spleen]	2	1		0	0	0	0	Unspaded
-a bug's lymph	1	1	awesome	0	0	0	0	Unspaded 60 A Bug's Life (+100% HP, +8-10 HP Regen)
+[experimental crimbo spleen]	2	1		0	0	0	0	50 Festively Experimented-Upon (+200% HP/MP), one random effect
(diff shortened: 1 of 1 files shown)


src/data/spleenhit.txt
```diff
@@ -60,7 +60,7 @@ cold jelly	1	1	crappy	0	0	50-60	0	50 Cold Jellied (+3 mys stats/fight, +5 cold d
 cold wad	1	6	good	0	50-60	30-40	10-20	20 Cold Blooded (+30% mus, +20% mys, +10% mox, cold dmg/res)
 concentrated magicalness pill	1	2	decent	0	0	5-7	0
 cosmic paste	4	4	good	5-10	18-22	18-22	18-22	10 Cosmic Flavor (+10% critical spell)
-Crimbeau de toilette	1	1	awesome	0	0	0	0	Unspaded, 50 Crimbeau'd (+10 familiar weight, +25 item drop, +50 meat drop)
+Crimbeau de toilette	1	1	awesome	0	20	20	20	Unspaded, 50 Crimbeau'd (+10 familiar weight, +25 item drop, +50 meat drop)
```
Veracity0, line 63: I like the numbers, but we can remove Unspaded from the comment, no?

Veracity0 (changes requested):

src/data/spleenhit.txt
midgleyc: Unspaded was sticking around here because this was off a single data point. It's more useful than zeroes, but there might be a range in reality.

Alternatively it's a limited 2018 Crimbo item, so nobody's realistically using it anyway.

Veracity0 (approved):

PR #899, opened 2022-07-18
Fix hobo songs showing as castable when not an AT

gausie:


gausie:
Fixes https://kolmafia.us/threads/maximizer-recommending-i-cast-skills-i-cant-actually-cast.27757/

PR #900, opened 2022-07-18
More dailylimits work

gausie:
Some cleaning up, more test coverage, more cases handled, more edge cases handled

### DIFF
--- a/src/data/dailylimits.txt
+++ b/src/data/dailylimits.txt
@@ -4,6 +4,7 @@
 Cast	Acquire Rhinestones	_rhinestonesAcquired
 Cast	Advanced Cocktailcrafting	cocktailSummons
 Cast	Advanced Saucecrafting	reagentSummons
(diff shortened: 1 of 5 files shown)


src/data/dailylimits.txt
```diff
@@ -1,6 +1,124 @@
 1
 # Limiter: daily limit
 # Type	Item name	Tracking property	[Max = 1]
+Cast	Acquire Rhinestones	_rhinestonesAcquired
```
jaadams5, line 4: Sorting is better than not sorting.  Are we pretty much convinced that we want case sensitive sorts in code where a string sort is used because there is no associated numeric?  Just asking for future reference.

jaadams5 (approved):

src/data/dailylimits.txt
gausie: I just used IntelliJ's sort function - I have no real opinion on that just yet

PR #902, opened 2022-07-18
Fix a typo

jaadams5:
Fix a typo in displayed text.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
+++ b/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
@@ -380,7 +380,7 @@ public class OptionsFrame extends GenericFrame {
               """
 	      Configure the behavior of Mafia's Script menu.
 
-	      If you set the script MRU length to a value greater than zero, that many of your most recently run scripts will be displayed.
+	      If you set the script MRU length to a value greater than zero then many of your most recently run scripts will be displayed.
 
 	      If you select cascading script menus, all the scripts in your 'scripts' folder will be displayed.
 


gausie:
How about

> If you set the script MRU length to a quantity greater than zero, that quantity of your most recently run scripts will be displayed.

jaadams5:
Ah ha.  I did not understand "that" as a placeholder for a numeric value , "that many".  I also read "many" as some kind of conditional.  If MRU is 3, I explicitly run a script and it implicitly runs three others, the actual script I ran would not be on the list.  So many as a conditional prepares a literal minded user for the possibility that a script could not appear but the absence would not be a bug.

If it confused me it might confuse others but I am not going to insist on a change because now I understand the intent.

Tangentially, opening the Preferences at less than full screen and changing to full screen does not behave as I expect.  Item Acquisition truncates the line beginning with "The following",  Main  Interface no longer displays the three drag and drop lists sized to fit the window.  Relay Browser truncates the option starting with "Warn if".  Daily Deeds does not render everything in the smaller window.  I might expect the text in the SVN and Git windows to wrap differently with resizing.  In several cases the text size, font or boldness does not seem to be the same with the object on the RHS (Frame?)

All of the above were done without being logged in so the behavior may be a consequence of default or unset preferences.

Veracity0:
1) "that many" was referring to the previously seen "a value greater than zero".
I really don't think that is obscure English usage. :)

2) You wrote:

"If it confused me it might confuse others".

Surely you MEANT to write:

"If it confused me, then it might confuse others".

;)

3) If you set MRU size to three and run a script which runs three other scripts, those three scripts are your three "most recently run" scripts.

4) Concerning text wrapping and resizing windows, you are opening many cans of worms with  how various panels in the GUI are set to show text. JLabels truncate and do not wrap. Other text fields have fixed width and behave like that too. Others will re-wrap if the width increases but not if it decreases. There are probably window resizing events which can be used to change preferred sizes and such, but apparently our widgets are not using them.

Feel free to dive into Swing and figure out what needs to be done. :)

gausie:
Closing this one lest we discuss it more!

PR #903, opened 2022-07-18
move more stuff into ConfigQueuingPanel

Veracity0:
Make ConfigQueuingPanel more functional.

1) JTextFields need special handling regarding their size. Add newTextField(String content) to do that.
2) I created PreferenceButtonGroup for handling a group of Radio Buttons controlled by a single Preference.
It uses the Maximizer's SmartButtonGroup - which I extended to allow setting an ActionListener for each button.
There are unit tests for the widget.
3) MaximizerOptionsPanel is now based on ConfigQueuingPanel and uses PreferenceCheckBox, PreferenceIntegerTextField, and PreferenceButtonGroup.


gausie (approved):
Awesome

PR #904, opened 2022-07-18
Some bits of code from the removal of decorated tabs was still hanging around

gausie:


jaadams5 (approved):

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -897,35 +881,12 @@ public void actionConfirmed() {
     @Override
     public void actionCancelled() {
       super.actionCancelled();
-
-      this.innerGradient.setBackground(
-          DataUtilities.toColor(Preferences.getString("innerChatColor")));
-      this.outerGradient.setBackground(
-          DataUtilities.toColor(Preferences.getString("outerChatColor")));
+      super.actionCancelled();
```
midgleyc, line 884: don't think this needs calling twice.

gausie: But it's optimal

midgleyc (approved):

PR #906, opened 2022-07-19
When your carnivorous plant swallows your foe whole, log it

Veracity0:


Veracity0:
I just pushed a change intended for a different branch. Oops.
Well, it didn't depend on changed code from that branch...

midgleyc (approved):

PR #907, opened 2022-07-19
DailyLimits applies to combat skills!

gausie:
This PR extends the daily limit database to combat skills. There are lots of improvements that can be made iteratively here but the key ideas (as with all the dailylimits code) are to a) make adding new daily limited skills easier and b) reduce amount of code overall.

So far this PR:

- [x] Describes all daily limited skills in datafile format
- [x] Handles tracking once-per-fight skills better. This could be extended though as it's very repetitive and I imagine 
we're missing many. I'm going to look at turning it into a boolean modifier.
- [x] ~Handle skills that are tracked with a pref that counts down instead of up. There are a few that count the other way (burrowgrubSummonsRemaining, _fireExtinguisherCharge).~ I'm not going to do this this round
- [x] Tests

This also makes me think... should we store the strings we check for to confirm success in a datafile somewhere? That would be a much wider-reaching goal but would certainly be quite easy to do for skill casting (for example).


src/data/dailylimits.txt
```diff
@@ -40,24 +62,50 @@ Cast	Feel Nostalgic	_feelNostalgicUsed	3
...
+Cast	KGB tranquilizer dart	_kgbTranquilizerDartUses	3
+Cast	Lash of the Cobra	_edLashCount	30
 Cast	Lock Picking	lockPicked
+Cast	Long Con	_longConUsed	5
 Cast	Lunch Break	_lunchBreak
+Cast	Macrometeorite	_macrometeoriteUses	10
 Cast	Managerial Manipulation	_managerialManipulationUsed
 Cast	Map the Monsters	_monstersMapped	3
 Cast	Meatify Matter	_meatifyMatterUsed
+Cast	Meatify Matter 	_meatifyMatterUsed
```
midgleyc, line 89: dupe

gausie: Thank youpe

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9373,19 +9166,15 @@ private static void payActionCost(final String responseText) {
         }
         break;
 
-      case SkillPool.MICROMETEOR:
-        Preferences.increment("_micrometeoriteUses");
```
midgleyc, line 9377: this has disappeared with no apparent replacement. While you can use it an unlimited number of times per day, the % delevel decreases with use.

gausie: You know what, for years I had thought this pref was added by mistake and had always meant to remove it. TIL!

midgleyc (dismissed):

PR #911, opened 2022-07-20
feat: support installing git scripts from non-root folders

midgleyc:
Certain existing scripts, such as [ChIT](https://github.com/Loathing-Associates-Scripting-Society/ChIT) install from a folder one level lower than the base directory when using SVN. It is desirable to support this from git as well, to avoid breaking existing installations.

Introduce a file `manifest.json` with a key `root_directory` to specify the directory that contains the other files Mafia is interested in: those under the permissible folders (e.g. `scripts`), and the `dependencies.txt` file.

Removed comment about not supporting local changes because local changes are supported implicitly without me having to write any code, which is nice. It seems that `git update` will handle unstaged changes -- I thought asking people to commit changes was a good halfway house, but even this isn't necessary. Merge conflicts will probably error, though.

To install ChIT locally, a user could:
* run `git checkout https://github.com/Loathing-Associates-Scripting-Society/ChIT.git` to install ChIT, but not copy any files
* create a `manifest.json` file under `MAFIA_ROOT/git/Loathing-Associates-Scripting-Society-ChIT` containing
```
{
  "root_directory": "src"
}
```
* run `git sync`

This file could be added to the remote git repo, after this feature is released, to avoid the need for local changes.

I'm not certain I want the configuration to be json or to be named as it is -- arguments for / against welcomed.

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
@@ -234,17 +234,18 @@ public class GitManager extends ScriptManager {
     }
     var folder = folderOpt.get();
     var projectPath = KoLConstants.GIT_LOCATION.toPath().resolve(folder);
+    var root = getRoot(projectPath);
     KoLmafia.updateDisplay("Removing project " + folder);
     List<Path> toDelete;
     try {
-      toDelete = getPermissibleFiles(projectPath, true);
+      toDelete = getPermissibleFiles(root, true);
     } catch (IOException e) {
       KoLmafia.updateDisplay(MafiaState.ERROR, "Failed to remove project " + folder + ": " + e);
       return;
     }
     var errored = false;
     for (var absPath : toDelete) {
-      var shortPath = projectPath.relativize(absPath);
+      var shortPath = root.relativize(absPath);
       var relPath = KoLConstants.ROOT_LOCATION.toPath().resolve(shortPath);
       try {
         Files.deleteIfExists(relPath);
--- a/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
@@ -551,6 +551,8 @@ public class GitManager extends ScriptManager {
     var manifest = json.get();
     var root = manifest.optString(MANIFEST_ROOTDIR, "");
     if (root.length() == 0) return projectPath;
+    // deny absolute paths or folder escapes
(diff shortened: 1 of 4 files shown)


gausie:
👍 for manifest.json

gausie:
Would you be up for adding more tests to GitManager?

midgleyc:
I think it's feasible to add some, with downside that:
* it requires an external git repo (I've been using https://github.com/midgleyc/mafia-script-install-test), which requires a network connection. I might be able to use a created local repo somehow?
* it uses the real filesystem (have to be careful to clean up after)
* you can't test (or at least I can't think of a way to test) the really complicated bits (e.g. that `git update` can delete files successfully. This is easy manually.)
* it's way easier to test manually than it is to write the unit tests

I would be looking to go through the CLI and ASH commands, instead of testing the manager, because I don't want to be tied to the current public interface.

gausie:
> I think it's feasible to add some, with downside that:
> 
>     * it requires an external git repo (I've been using https://github.com/midgleyc/mafia-script-install-test), which requires a network connection. I might be able to use a created local repo somehow?

I wouldn't mind keeping an example git repo in the tests directory

>     * it uses the real filesystem (have to be careful to clean up after)

Yeah I guess alternatively we mock the filesystem layer but I appreciate that's tedious.

>     * you can't test (or at least I can't think of a way to test) the really complicated bits (e.g. that `git update` can delete files successfully. This is easy manually.)

Mocking the fs I suppose this would be doable.

midgleyc:
If I were to avoid the remote network, I think the way would be:
* in the test, use JGit to create a local repo on the filesystem
* figure out what URI to pass JGit to clone that repo (`file://<path>`?)
* test accordingly

As an increment, though, I think I can create some branches on my test repo and use that.

jaadams5:
FWIW, DebugDatabaseTest mocks a "simple" and "complex" file system.  I'm not sure how relevant it is because the focus is in files and attributes.  To test git commands you'd also need file contents, but maybe this would be extendable?  MoodManagerTest mocks reading a text file.  Probably irrelevant but sometimes even a bad example can be inspirational.

midgleyc:
I saw that: that creates a fake `File` and passes it to a few functions. I don't think mocking the filesystem the same way in GitManager is feasible because of the entrypoints. For example:
* checkout takes a URI
* list looks in a directory for filenames recursively
* update takes a path to a full git repository

Mocking any of these would be extremely fragile and tightly coupled to the implementation, and wouldn't give high confidence the code actually worked. I feel like it's a non-starter.

gausie (approved):
This is great! Thank you

PR #913, opened 2022-07-21
Vintner charge tracking

gausie:
This PR adds tracking for the "charge" of the vampire vintner. A future PR should contain tracking for the types of damage dealt, though that's much trickier.


midgleyc (approved):
I assume moving the Robortender case block is because the cases are sorted by ID or something like that?

gausie:
> I assume moving the Robortender case block is because the cases are sorted by ID or something like that?

Yes. It was just added to the very end in a recent PR and I thought I might as well fix it

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4063,6 +4065,27 @@ else if (responseText.contains("a horrible grinning clown head emerges")) {
           Preferences.increment("garbageFireProgress");
           break;
 
+        case FamiliarPool.ROBORTENDER:
+          String[] roboDropMessages =
```
Veracity0, line 4069: Can this be static?

Veracity0 (dismissed):

PR #914, opened 2022-07-21
Environment tracking for CMC

gausie:
This is the first step providing basic environment tracking for the CMC. Future PRs will add more functionality and information to the `cmc` function.


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4387,6 +4386,28 @@ else if (responseText.contains("a horrible grinning clown head emerges")) {
...
+  private static void trackEnvironment(final KoLAdventure location) {
+    var environment = location != null ? location.getEnvironment() : "none";
+
+    var symbol =
+        switch (environment) {
+          case "outdoor" -> "o";
+          case "indoor" -> "i";
+          case "underground" -> "u";
+          case "underwater" -> "x";
+          default -> "?";
```
midgleyc, line 4398: I think default (e.g. as will appear for "none") should be "x" same as underwater is "x".

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4387,6 +4386,28 @@ else if (responseText.contains("a horrible grinning clown head emerges")) {
...
+        };
+
+    var environments = new StringBuilder(Preferences.getString("lastCombatEnvironments"));
+    environments.append(symbol);
+    while (environments.length() < 20) {
+      environments.insert(0, "x");
+    }
```
midgleyc, line 4405: Feel like `environments = "x".repeat(20 - environments.length()) + environments` or something like that is nicer.

midgleyc (changes requested):

midgleyc:
sorry, didn't recognise this was still draft

src/net/sourceforge/kolmafia/request/FightRequest.java
gausie: I want to track default separately for when new locations are added. I'm adding some functions to predict your current pill which take the ? into account smartly.

src/net/sourceforge/kolmafia/request/FightRequest.java
gausie: Agreed!

gausie: I was even cheekier

src/net/sourceforge/kolmafia/request/FightRequest.java
midgleyc: looks like I was wrong about this anyway -- by the forum thread, zoneless combats don't count towards Fleshazole, but I thought they did

midgleyc (changes requested):
I think the code is fine but I'm going to query a couple of the names.

`next_cold_medicine_cabinet_pill` -- this is the estimated next pill based on what Mafia has tracked by zones. To find the actual next pill we can check in the workshed (assuming we've a consult open). Hopefully these are the same, but they're not certain to be. `expected_next_cmc_pill`? We might even want to do something fancy and use excavator to tell us how often we're wrong.

`cmc` command -- I'd expect this to let you interact with the cmc, or at least give you info on what the CMC currently offers. Instead it just returns the expected next pill. For extensibility reasons, I'd rather this not be the base command -- `cmc estimate`? `cmc expect`? Then `cmc info` (or `cmc`) could visit the workshed and tell you everything you've got available and `cmc pill` (etc.) could get it. This would be nice for me because I'm currently doing `campground.php?action=workshed; choice 5`.

gausie:
> `next_cold_medicine_cabinet_pill` -- this is the estimated next pill based on what Mafia has tracked by zones. To find the actual next pill we can check in the workshed (assuming we've a consult open). Hopefully these are the same, but they're not certain to be. `expected_next_cmc_pill`? We might even want to do something fancy and use excavator to tell us how often we're wrong.

I can do this.

> `cmc` command -- I'd expect this to let you interact with the cmc, or at least give you info on what the CMC currently offers. Instead it just returns the expected next pill. For extensibility reasons, I'd rather this not be the base command -- `cmc estimate`? `cmc expect`? Then `cmc info` (or `cmc`) could visit the workshed and tell you everything you've got available and `cmc pill` (etc.) could get it. This would be nice for me because I'm currently doing `campground.php?action=workshed; choice 5`.

I think its quite normal for commands in mafia to print a status when there's no parameter and I would like to keep doing that here. c.f. AutoAttackCommand, CargoCultCommand, ChoiceCommand, ConditionsCommand and many more (just from searching `parameters.equals("")` on the codebase!).

I agree though that this command should be extended. As I said in the PR description, I'll do so in future branches.

gausie:
To elaborate, in a future branch I want the cmc function with no params to not estimate and instead visit the cmc and tell you whats on offer if your consults are ready. And tell you how many turns to next consult etc etc. @aenimuskol is working on a function to tell you how many adventures to spend in certain environments to tip the scales.

midgleyc (approved):
I agree it's fine to have something print a status with no parameters -- the different I see here is that the plan is to have "no parameters" go from being an estimated pill to being the real thing, whereas I think you could set up the parameters so that you wouldn't need this to change.

But I'm not blocking on it.

gausie:
Legit. I feel like I want cmc to just output all the info someone could want to know on the state of the item, rather than make them go digging through different params to see everything. Appreciate the approval.

midgleyc:
Yeah so I have nothing against "all the info", just that currently it makes no network requests and just uses internal state, and later it would change to visit the page if you had a consult available, and I feel "tell me what Mafia thinks" and "visit KoL to see reality" are conceptually two different commands.

PR #918, opened 2022-07-22
More CMC command work

gausie:
Fleshing out the function so you can do things like this:

image

image

image


src/net/sourceforge/kolmafia/textui/command/ColdMedicineCabinetCommand.java
```diff
@@ -56,17 +81,211 @@ public static AdventureResult nextPill(Map<Character, Integer> counts) {
...
+
+    switch (parameter) {
+      case "" -> status();
+      case "equipment", "equip" -> collect(1);
+      case "food" -> collect(2);
+      case "booze", "wine" -> collect(3);
+      case "potion" -> collect(4);
+      case "pill" -> collect(5);
+      default -> {
+        KoLmafia.updateDisplay(KoLConstants.MafiaState.ERROR, "Paramter not recognised");
```
midgleyc, line 287:
```suggestion
        KoLmafia.updateDisplay(KoLConstants.MafiaState.ERROR, "Parameter not recognised");
```

test/internal/helpers/Player.java
```diff
@@ -192,6 +192,42 @@ public static Cleanups setStats(int muscle, int mysticality, int moxie) {
     return new Cleanups(() -> setStats(0, 0, 0));
   }
 
+  public static Cleanups setMuscle(int muscle, int buffedMuscle) {
```
midgleyc, line 195: I feel like KoLCharacter should have methods for setting each of Muscle, Myst and Moxie that get called here, instead of calling these methods that use the existing stat values.

src/net/sourceforge/kolmafia/textui/command/ColdMedicineCabinetCommand.java
gausie: I spotted this and forgot to fix it - thank you

test/internal/helpers/Player.java
gausie: I don't know if it makes any real sense to do this other than for tests - we never see isolated stats in KoL - they always come as one big dump of info.

midgleyc: yes, but tests are also code, and this has the opportunity to simplify a bit of test logic (and make tests that only want to set one stat) at the cost of three extra method calls each time `KoLCharacter.setStatPoints` gets called. I think it's worthwhile.

gausie: Ok!

midgleyc (dismissed):
You could also update the ModifierExpressionTests to use the new methods, but not blocking on that.

PR #921, opened 2022-07-23
Remove "Using cached search results" print

Rinn:
This print is very spammy and doesn't really provide any actionable information

https://kolmafia.us/threads/retrieve_price-is-too-noisy.27523/

### DIFF
--- a/src/net/sourceforge/kolmafia/session/MallPriceManager.java
+++ b/src/net/sourceforge/kolmafia/session/MallPriceManager.java
@@ -297,7 +297,6 @@ public abstract class MallPriceManager {
 
     List<PurchaseRequest> results = MallPriceManager.getSavedSearch(id, needed);
     if (results != null) {
-      KoLmafia.updateDisplay("Using cached search results for " + name + "...");
       return results;
     }
 


Veracity0:
In February, I did a lot of work on MallPriceManager.

I moved it out of StoreManager in https://github.com/kolmafia/kolmafia/pull/598
I provided unit tests in https://github.com/kolmafia/kolmafia/pull/607
I further whacked the internals in https://github.com/kolmafia/kolmafia/pull/620

This message was extremely valuable in understanding how mall price caching was working - or not working.

I read the original PR where the OP admitted it wasn't a "bug per se".
I also saw that he was going to "keep looking into this". So I waited for his followup.
Which did not happen.

"Too noisy" is a matter of opinion. "Spammy" is a more judgemental way to say the same thing.

I'm done modifying MallPriceManager - unless an actual issue is eventually reported. The message would be a useful diagnostic - or evidence of an issue. Without it, I expect we'll see reports of "KoLmafia said it was searching for the price for such-and-such an item but it looks like it didn't really do a search."

I've not seen any such reports since I finished my work on MallPriceManager.
I am perfectly comfortable simply ignoring the message, but if others can't (or won't) do it, whatever.

I am neither going to approve or disapprove this PR.

I encourage other devs to make their own decisions about it.

Peace. :)

gausie:
I agree with Veracity. I think rather than removing the message entirely I think it would be a good idea to log this message once per function call from the user layer.

Veracity0:
> I think it would be a good idea to log this message once per function call from the user layer.

Rinn did a followup post to the Bug Report with output that seems to demonstrate that the user script called retrieve_price 110 times for "sweet tooth".

gausie:
Well...

PR #922, opened 2022-07-24
Start tracking ascensions, reset potato alarm clock only on rollover

libraryaddict:
[Requested here](https://kolmafia.us/threads/count-ascensions-since-rollover.27790/)

This adds _ascensionsToday which is a daily flag, but is only reset on rollover.
This also changes _potatoAlarmClockUsed to be reset on rollover only, as it only triggers once per real life day.


src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -1133,6 +1136,16 @@ public static void resetPerAscension() {
 
     // Some need special treatment
     MonorailManager.resetMuffinOrder();
+
+    // Increment the amount of times we've ascended today
+    Preferences.increment("_ascensionsToday");
```
gausie, line 1141: This will work here but it's probably put it somewhere more meaningful like in the function that initially tracks an ascension event. Do we want to track it on entering or on leaving Valhalla? What happens if you sit in Valhalla over rollover?

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -55,6 +55,8 @@ public class Preferences {
 
   private static final Set<String> defaultsSet = new HashSet<>();
   private static final Set<String> perUserGlobalSet = new HashSet<>();
+  private static final Set<String> resetOnRollover =
```
gausie, line 58: I think this would be better named doesntResetOnAscension or onlyResetOnRollover since everything with a leading _ sounds like it should be here

gausie (changes requested):
Stone feedback but also I'd like some tests integration tests for this - mock a Valhalla event (once we've resolved when it should happen in that cycle) and mock a login with a new day

gausie:
Musing on this, I think we should restrict _ to being only reset on ascension AND rollover. So for now I think change these (including the old one imo!) to have no leading underscore.

libraryaddict:
Asked by Gausie in a resolved message: Do we want to track it on entering or on leaving Valhalla? What happens if you sit in Valhalla over rollover?

Needs spading, though it can't really be spaded by a multi given that the account needs to be able to loop.

gausie:
> Asked by Gausie in a resolved message: `Do we want to track it on entering or on leaving Valhalla? What happens if you sit in Valhalla over rollover?`
> 
> Needs spading, though it can't really be spaded by a multi given that the account needs to be able to loop.

If it needs spading why mark the comment as resolved?

gausie:
> Some feedback but also I'd like some integration tests for this - mock a Valhalla event (once we've resolved when it should happen in that cycle) and mock a login with a new day

Other than resolving the mechanical question, this is all we need now! Thanks irrat

Veracity0:
So close.

Are we going to see those additional tests?

gausie:
My understanding is that @libraryaddict is still trying to spade or get someone to spade how exactly the daily ascension limit is tracked - if you start the day in Valhalla can you still enter it twice?

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,66 @@
...
+  @AfterEach
+  public void resetCharAndPrefs() {
+    KoLCharacter.reset("");
+    KoLCharacter.reset(true);
+    KoLCharacter.setUserId(0);
+    Preferences.saveSettingsToFile = false;
+  }
+
+  @Test
+  void testOnAscension() {
```
gausie: What is this? Is it possible to split all these requests into different tests? If not why not?

libraryaddict: Tests has been seperated

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -440,6 +449,20 @@ void resetAscensionProperties() {
     assertNotEquals(val, Preferences.getInteger(name));
   }
 
+  void resetRolloverProperties() {
```
gausie, line 452: This needs an @Test annotation, and could use a better name

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,143 @@
...
+    // Set breakfast and ascension prefs
+    Preferences.setInteger(ascensionsPref, ascensionBefore);
+    Preferences.setInteger(breakfastPref, breakfastBefore);
+
+    // This does a 302 redirect to afterlife.php
+    var jumpGash = new GenericRequest("ascend.php?action=ascend&pwd&confirm=on&confirm2=on");
+    jumpGash.execute();
+
+    // Confirm our last run breakfast is now 0
+    assertEquals(breakfastAfter, Preferences.getInteger(breakfastPref));
```
gausie, line 53: `assertThat(breakfastPref, isSetTo(breakfastAfter));`

gausie: (and the same for allt hese other tests

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,143 @@
...
+    int ascensionBefore = 1;
+    int ascensionAfter = 2;
+
+    // Set breakfast and ascension prefs
+    Preferences.setInteger(ascensionsPref, ascensionBefore);
+    Preferences.setInteger(breakfastPref, breakfastBefore);
+
+    // This does a 302 redirect to afterlife.php
+    var jumpGash = new GenericRequest("ascend.php?action=ascend&pwd&confirm=on&confirm2=on");
+    jumpGash.execute();
```
gausie, line 50: Why is this not a .run()?

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,143 @@
...
+    String breakfastPref = "lastBreakfast";
+    int breakfastBefore = 5;
+    int breakfastAfter = 0;
+    int breakfastValhalla = -1;
+    String ascensionsPref = "ascensionsToday";
+    int ascensionBefore = 1;
+    int ascensionAfter = 2;
+
+    // Set breakfast and ascension prefs
+    Preferences.setInteger(ascensionsPref, ascensionBefore);
```
gausie, line 45: These should all be withProperty cleanups

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,143 @@
...
+    var jumpGash = new GenericRequest("ascend.php?action=ascend&pwd&confirm=on&confirm2=on");
+    jumpGash.execute();
+
+    // Confirm our last run breakfast is now 0
+    assertEquals(breakfastAfter, Preferences.getInteger(breakfastPref));
+    // Confirm our ascensions today has not changed
+    assertEquals(ascensionBefore, Preferences.getInteger(ascensionsPref));
+
+    // Execute the result of the 302 redirect and get the output which puts us in valhalla
+    var afterlife = new GenericRequest("afterlife.php");
```
gausie, line 58: Please use fakeHttpClient and `run` here

libraryaddict: Done

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
libraryaddict: Changed

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
libraryaddict: Changed to use FakeHttpClient and to use run()

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
libraryaddict: Performed

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
libraryaddict: Renamed to testRolloverPropertiesAreReset and annotated

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,160 @@
...
+    String ascensionsPref = "ascensionsToday";
+    int ascensionBefore = 1;
+    int ascensionAfter = 2;
+    var cleanups =
+        new Cleanups(
+            withProperty(breakfastPref, breakfastBefore),
+            withProperty(ascensionsPref, ascensionBefore));
+
+    try (cleanups) {
+      var gashCleanups = withNextResponse(302, Map.of("location", List.of("afterlife.php")), "");
```
Veracity0, line 57: I want this so bad. I'm currently working on code that does a redirect and was pondering how to test it. I was about to look into modifying our fake client - and you have done it for me!

Want want want.

Veracity0 (dismissed):

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -440,6 +451,21 @@ void resetAscensionProperties() {
...
   }
 
+  @Test
+  void testRolloverPropertiesAreReset() {
+    String name = "ascensionsToday";
+    int beforeRollover = 1;
+    int afterRollover = 0;
+
+    // Confirm it was set
+    Preferences.setInteger(name, beforeRollover);
```
gausie, line 461: Please use cleanups withProperty here

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,160 @@
...
+    KoLCharacter.reset("fakePrefUser");
+    KoLCharacter.reset(true);
+    Preferences.saveSettingsToFile = false;
+    // Fix another test not cleaning up
+    FightRequest.currentRound = 0;
+  }
+
+  @AfterEach
+  public void resetCharAndPrefs() {
+    KoLCharacter.reset("");
```
gausie, line 35: You shouldn't need to do this - make better use of cleanups

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,160 @@
...
+import org.junit.jupiter.api.Test;
+
+public class AscensionRequestTest {
+  @BeforeEach
+  public void initializeCharPrefs() {
+    KoLCharacter.reset("fakePrefUser");
+    KoLCharacter.reset(true);
+    Preferences.saveSettingsToFile = false;
+    // Fix another test not cleaning up
+    FightRequest.currentRound = 0;
```
gausie, line 30: At best, remove this. At worst, make it a before all rather than beforeEach

test/internal/helpers/Player.java
```diff
@@ -1005,21 +1006,50 @@ public static Cleanups withQuestProgress(final QuestDatabase.Quest quest, final
...
+  }
+
   /**
    * Sets next response to a GenericRequest Note that this uses its own FakeHttpClientBuilder so
    * getRequests() will not work on one set separately
    *
    * @param code Status code to fake
    * @param response Response text to fake
-   * @return Cleans up so this response is not given again
+   * @return Cleans up so the responses won't be given
```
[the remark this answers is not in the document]
libraryaddict: Leftover code, definitely my bad

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  @Test
+  void testSelectPathConfirmedProcessed() {
+    // Set last breakfast to -1 to mark that we've ascended as done in testAscensionsTodayTracked
+    var cleanups =
+        new Cleanups(
+            withProperty("lastBreakfast", -1),
+            withProperty("bankedKarma", 150),
+            withNextResponse(302, Map.of("location", List.of("main.php")), ""));
```
libraryaddict, line 129: Need to create another test where ascending does not place you into choice.php, and fix this test to place you into choice.php

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -440,6 +452,20 @@ void resetAscensionProperties() {
     assertNotEquals(val, Preferences.getInteger(name));
   }
 
+  @Test
+  void testRolloverPropertiesAreReset() {
+    String name = "ascensionsToday";
+    int beforeRollover = 1;
+    int afterRollover = 0;
+
+    try (var cleanups = withProperty(name, beforeRollover)) {
```
gausie, line 461: This will cause a warning for unused variable in intellij

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -0,0 +1,151 @@
...
+    String pref = "bankedKarma";
+    int karmaBeforehand = 100;
+    int karmaAfter = 211;
+
+    // Set our karma beforehand to 100
+    var cleanups = withProperty(pref, karmaBeforehand);
+
+    try (cleanups) {
+      // Confirm our karma is 100
+      assertThat(pref, isSetTo(karmaBeforehand));
```
gausie, line 83: A bit redundant since you've set this two lines prior, but that's fine

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
gausie: 👍

libraryaddict:
The forced path choice throws an unhandled redirect error in the log, this is the same as the actual situation.
I'm not sure if this would be a bug or not.

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
libraryaddict: Done, although I'm not too happy with the way I went about things.
I'm reusing Grey You, but although it should be fine; The path ultimately does resolve into a choice which could in theory, play up at a later date.

Veracity0:
> The forced path choice throws an unhandled redirect error in the log, this is the same as the actual situation.
> I'm not sure if this would be a bug or not.

This means that nothing within KoLmafia itself is following that redirect.
It could be a bug. What Request class is being executed?
What URL was submitted, followed by a redirect to what location?

Typically, GenericRequest itself will follow a redirect, but other classes that extend it have to either handle the redirect on their own or setup allowing GenericRequest to follow it.

libraryaddict:
> > The forced path choice throws an unhandled redirect error in the log, this is the same as the actual situation.
> > I'm not sure if this would be a bug or not.
> 
> This means that nothing within KoLmafia itself is following that redirect. It could be a bug. What Request class is being executed? What URL was submitted, followed by a redirect to what location?
> 
> Typically, GenericRequest itself will follow a redirect, but other classes that extend it have to either handle the redirect on their own or setup allowing GenericRequest to follow it.

https://gist.github.com/9beae6375e8d3de7037ce138ec98dc32

I think I'm wrong about it throwing a script error, and instead the error is that you're not in a choice which you actually are, because we didn't follow the redirect.

```
      visitUrl(
          "afterlife.php?action=ascend&confirmascend=1&whichsign=3&gender=2&whichclass=27&whichpath=44&asctype=2&nopetok=1&noskillsok=1"
        );
        visitUrl("main.php");
```

But you can see in the log that I've ascended, told to redirect to a choice.php. It noticably does not redirect, instead throwing an error. 

For some reason, it wants the description of the industrial fire extinguisher. Then I call main.php which sends it to the choice properly. Then I use runChoice(1) later.

Veracity0:
GenericRequest.handleServerRedirect:

```
    // If this is a redirect from valhalla, we are reincarnating
    if (this.formURLString.startsWith("afterlife.php")) {
      // Reset all per-ascension counters
      KoLmafia.resetCounters();

      // Certain paths send you into a choice adventure.
      // Defer new-ascension processing until that is done.
      if (this.redirectLocation.startsWith("choice.php")) {
        ChoiceManager.ascendAfterChoice();
      }

      // Otherwise, do post-ascension processing immediately.
      else {
        ValhallaManager.postAscension();
      }

      return true;
    }
```

It is redirecting to the Welcome to Grey You choice.
If you did this in the browser, it would follow the redirect and you would see it.
You are scripting this with visitUrl - which uses a GenericRequest.
That normally follows redirects, but not in this case; it does "return true".

I wonder what it redirects to when it is not choice.php?
It calls ValhallaManager.postAscension in that case.
The only other place it calls that method is in ChoiceManager.postChoice2.
So it must be normal.

This might be correct in that code snippet:

```
      if (this.redirectLocation.startsWith("choice.php")) {
        ChoiceManager.ascendAfterChoice();
        ChoiceManager.preChoice(this);
      }
```

which will set handlingChoice to true.
In which case, I think your script should do something like "if (handlingChoice) runChoice(1)". following your visitUrl.

I will ponder further. Including why it is looking at your industrial fire extinguisher.

Veracity0:
I really need the work you did to the FakeHttp support.

Only sending the same response once - without this, I had two tests fail with a Stack Overflow. Not a KoLmafia issue. A test framework issue.

And setting up for redirects. *chef's kiss*

I'm going to extract those improvements into a standalone PR, rather than wait more weeks before I can use them. 🙂

libraryaddict:
> I really need the work you did to the FakeHttp support.
> 
> Only sending the same response once - without this, I had two tests fail with a Stack Overflow. Not a KoLmafia issue. A test framework issue.
> 
> And setting up for redirects. _chef's kiss_
> 
> I'm going to extract those improvements into a standalone PR, rather than wait more weeks before I can use them. 🙂

To be fair, I think this could be merged in as a feature and the bugs mentioned above are not new, and can be addressed separately

Veracity0:
OK. I will approve. You fix the merge conflicts.

Veracity0 (approved):

Veracity0 (approved):
I see you modified some tests I wrote which wanted your new features. I will adjust them to use your new technology. Thanks!

PR #927, opened 2022-07-25
Allow well_stocked to process amounts less than 6

libraryaddict:
Previously if you provide an amount less than 6, the function would always return false.

Now this will work similar to mall_price, only returning true once it's found the 5th item that's below the price point.

The line with the `new Value` was deleted as it was both redundant in the previous code, and would provide a workaround for the 5th cheapest in the new code.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -5235,7 +5235,7 @@ public abstract class RuntimeLibrary {
     // check parameters and return false if problems
     if (item == null) return DataTypes.FALSE_VALUE;
     if (itemID < 1) return DataTypes.FALSE_VALUE;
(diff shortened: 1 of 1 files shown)


jaadams5:
As per the discussion at https://kolmafia.us/threads/well_stocked-ash-function.26815 the decision to return false for inputs of five was deliberate.  I hope no one approves this change without continuing the discussion and arriving at a revised consensus.

gausie:
> As per the discussion at https://kolmafia.us/threads/well_stocked-ash-function.26815 the decision to return false for inputs less than 6 was deliberate. I hope no one approves this change without continuing the discussion and arriving at a revised consensus.

Of course, although reading this code it still respects the 5 limit, it just does it later.

I think I just need an explanation of why you want to make this change, not just how it works. You should be able to find a mall search results HTML adnd construct a test that fails before but passes after your changes.

libraryaddict:
Ah, wasn't aware it was intentional.

PR #929, opened 2022-07-26
feat: mark more monsters as NOCOPY

midgleyc:
The tower monsters are interesting, in that they are wishable (and I think Rain Man-able) but not copyable. Apparently they get marked as nocopy after the fight starts.


gausie:
> The tower monsters are interesting, in that they are wishable (and I think Rain Man-able) but not copyable. Apparently they get marked as nocopy after the fight starts.

I think they're just on an explicit allow list in the genie, for factoid purposes. The genie function should just (in/ex)clude them in a special case

gausie (approved):

PR #933, opened 2022-07-28
Fix You, Robot tests. Looking at CPU upgrades counts as a visit.

Veracity0:
Bug Report says that looking at CPU Upgrades in Reassembly Station in a You, Robot run does not learn which ones you have installed.

I could have sworn that worked for me - and I even had a working unit test that showed it.

Well, yes. If YouRobotManager.visitChoice is called with a responseText for choice 1445, it will parse the upgrades.
Unfortunately, the normal choice processing path never called visitChoice in that way for CPU upgrades. It just called postChoice1 with a URL that indicated no change was being made.

The tests for YouRobotManager all called the specific method they thought was needed - visitChoice or postChoice1.

I changed them all to call the processResponse method of the request, which does all the steps of choice processing - and that showed the error I mentioned about visitChoice not being called.

I added one line to YouRobotManager to call visitChoice from postChoice1, if appropriate.

The test now passes.  One hopes the Bug Report will be fixed.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/YouRobotManager.java
+++ b/src/net/sourceforge/kolmafia/session/YouRobotManager.java
@@ -647,6 +647,7 @@ public abstract class YouRobotManager {
       boolean changed = false;
 
       if (upgrade == null) {
+        visitChoice(request);
         return;
       }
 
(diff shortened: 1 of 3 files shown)


jaadams5 (approved):
Pending successful completion of tests, of course.

gausie:
Just as a note I think I made a cleanup that will restore handling choice without setting it (if not you can use withChoice(false)) I suppose

PR #934, opened 2022-07-29
Log can of mixed everything

libraryaddict:
Noticed this wasn't logged, despite the other three items that drop items being logged. Could be more items than this not being logged?


libraryaddict:
Actually, should the message `You think you see a weird thing out of the corner of your eye, but it turns out to be nothing. Which is actually pretty weird, if you think about it, man.` be logged as well?

gausie:
> Noticed this wasn't logged

Thanks!

> Could be more items than this not being logged?

Probably!

> Actually, should the message ... be logged as well?

Don't think so - why would it?

gausie (approved):

jaadams5 (approved):

PR #936, opened 2022-07-29
Do not process choices when we get a "whoops" response

gausie:
Do folks reckon I should add a 

```java
 KoLmafia.updateDisplay(MafiaState.ERROR, "Whoops! You're not actually in a choice adventure");
 ```

or something?

### CODE REVIEW

File: src/net/sourceforge/kolmafia/request/GenericRequest.java [hunk 1]
```diff
@@ -2158,6 +2158,12 @@ public void processResponse() {
       GenericRequest.isRatQuest = false;
     }
 
+    // Check that we are actually handling a choice
+    if (ChoiceManager.handlingChoice
+        && this.responseText.contains("Whoops!  You're not actually in a choice adventure.")) {
+      return;
```
Veracity0, line 2164: OK. I guess it is fair to short-circuit the processing; there is nothing more of interest to process in the response.

However, by doing so, you are not letting postChoice2 set ChoiceManager.handlingChoice to false.

You could do that here, although every place that sets that variable is currently in ChoiceManager.

I dunno. Perhaps:
```
    
    if (ChoiceManager.bogusChoice(this)) {
      return;
    }
```
And something like:
```
  public static final boolean bogusChoice(final GenericRequest request) {
    if (!ChoiceManager.handlingChoice) {
      return false;
    }
    if (request.responseText.contains("Whoops!  You're not actually in a choice adventure.") {
      ChoiceManager.handlingChoice = false;
      return true;
    }
    return false;
  }
```

test/internal/helpers/Player.java
```diff
@@ -846,6 +846,17 @@ public static Cleanups withRange() {
         });
   }
 
+  /**
+   * Does nothing, but ensures the given property is reverted as part of cleanup
+   *
+   * @param key Key of property
+   * @return Restores the previous value of the property
+   */
+  public static Cleanups withProperty(final String key) {
```
Veracity0, line 855: I like this. I have all sorts of tests that call methods which set a property as a side effect.
Could list the properties I expect to be modified, rather than cleaning them up in an afterEach method, say.

test/net/sourceforge/kolmafia/request/GenericRequestTest.java
```diff
@@ -79,4 +79,20 @@ public void parseDesignerSweatpants(int expectedSweat) {
 
     assertThat("sweat", isSetTo(expectedSweat));
   }
+
+  @Test
+  public void doesNotProcessPostChoiceWhenNotInChoice() {
+    var cleanup =
+        new Cleanups(
+            withNextResponse(200, html("request/test_choice_whoops.html")),
```
Veracity0, line 87: This is clearly a subset of a responseText. I guess that is OK.
I wonder if KoL can include other stuff - like events - which could affect where we want to look for this response during response processing.

Veracity0 (changes requested):

Veracity0:
> Do folks reckon I should add a
> 
> ```java
>  KoLmafia.updateDisplay(MafiaState.ERROR, "Whoops! You're not actually in a choice adventure");
> ```
> 
> or something?

Considering that we should only see this as a result of an error in a script, displaying the message seems useful.
Especially if it stops the script, if the request is not caught.

src/net/sourceforge/kolmafia/request/GenericRequest.java
gausie: Ah I was previously doing that and removed it without thinking

test/internal/helpers/Player.java
gausie: Yeah. We can't necessarily judge every changed pref but I think this should be the way forwards

test/net/sourceforge/kolmafia/request/GenericRequestTest.java
gausie: What do you mean?

gausie: Oh you mean the test fixture? It's just what the DEBUG log reported for the request that I made

```
Retrieved: https://www.kingdomofloathing.com/choice.php?whichchoice=999&pwd&option=1&topper=3&lights=5&garland=1&gift=2
11 header fields
Field: :status = [200]
Field: cache-control = [no-store, no-cache, must-revalidate, post-check=0, pre-check=0]
Field: content-encoding = [gzip]
Field: content-type = [text/html; charset=UTF-8]
Field: date = [Fri, 29 Jul 2022 13:44:42 GMT]
Field: expires = [Thu, 19 Nov 1981 08:52:00 GMT]
Field: pragma = [no-cache]
Field: server = [nginx/1.8.1]
Field: set-cookie = [AWSALB=VmQejbPpsWYA/jkaikO/7b/JH+LVG6USP/AR5CIWzNTCLZCAitEP4gmcUNIvyBgFc/FrHW5mD4kEBySFjobtayus73VCQqsGtVjhe0N9j5IayY9H12rH4bBYX+v+; Expires=Fri, 05 Aug 2022 13:44:42 GMT; Path=/, AWSALBCORS=VmQejbPpsWYA/jkaikO/7b/JH+LVG6USP/AR5CIWzNTCLZCAitEP4gmcUNIvyBgFc/FrHW5mD4kEBySFjobtayus73VCQqsGtVjhe0N9j5IayY9H12rH4bBYX+v+; Expires=Fri, 05 Aug 2022 13:44:42 GMT; Path=/; SameSite=None; Secure]
Field: vary = [Accept-Encoding]
Field: x-powered-by = [PHP/5.3.29]

Retrieving server reply
ResponseText has 438 characters.
<center><table  width=95%  cellspacing=0 cellpadding=0><tr><td style="color: white;" align=center bgcolor=blue><b>Choice Adventure</b></td></tr><tr><td style="padding: 5px; border: 1px solid blue;"><center><table><tr><td><font face=Arial>Whoops!  You're not actually in a choice adventure.<p><Center><a href=main.php>Back to the Main Map</a></center></font></td></tr></table></center></td></tr><tr><td height=4></td></tr></table></center>
Processing results...
Returned: <center><table  width=95%  cellspacing=0 cellpadding=0><tr><td style="color: white;" align=center bgcolor=blue><b>Choice Adventure</b></td></tr><tr><td style="padding: 5px; border: 1px solid blue;"><center><table><tr><td><font face=Arial>Whoops!  You're not actually in a choice adventure.<p><Center><a href=main.php>Back to the Main Map</a></center></font></td></tr></table></center></td></tr><tr><td height=4></td></tr></table></center>

```

src/net/sourceforge/kolmafia/session/ChoiceManager.java
```diff
@@ -2137,6 +2137,28 @@ public static void postChoice0(final String urlString, final GenericRequest requ
...
 
+  /**
+   * Determine if a request to choice.php showed that we weren't actually in a choice adventure
+   *
+   * @param request Completed request to check
+   * @return Whether KoL indicated that the player was not actually in a choice adventure
+   */
+  public static boolean bogusChoice(final GenericRequest request) {
+    if (!ChoiceManager.handlingChoice
+        || !request.getURLString().startsWith("choice.php")
```
gausie, line 2148: I'm defensively coding here against someone using this function unexpectedly in the future

test/net/sourceforge/kolmafia/request/GenericRequestTest.java
Veracity0: So that WAS the entire HTML returned by KoL.
OK, cool!
I am used to seeing whole pages with "html" and "body" and so on.

src/net/sourceforge/kolmafia/session/ChoiceManager.java
Veracity0: Fair enough.

Note that where this is called, GenericRequest has already extracted the urlString out of the request, so you could just pass it in, as we do for ```public static void postChoice0(final String urlString, final GenericRequest request)``` et. al. but I'm not going to require that.

Veracity0 (approved):

PR #937, opened 2022-07-30
feat: provide preference for not installing git dependencies

midgleyc:
Could be useful for debugging broken installs.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
+++ b/src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
@@ -370,13 +370,13 @@ public class OptionsFrame extends GenericFrame {
       this.queue(
           this.newTextArea(
               """
-	      Configure the behavior of Mafia's Script menu.
+        Configure the behavior of Mafia's Script menu.
 
-	      If you set the script MRU length to a value greater than zero, that many of your most recently run scripts will be displayed.
+        If you set the script MRU length to a value greater than zero, that many of your most recently run scripts will be displayed.
 
-	      If you select cascading script menus, all the scripts in your 'scripts' folder will be displayed.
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
```diff
@@ -1187,9 +1187,9 @@ public SVNPanel() {
       this.queue(
           this.newTextArea(
               """
-	      Configure the behavior of Mafia's built-in SVN client here.
+        Configure the behavior of Mafia's built-in SVN client here.
```
jaadams5, line 1190: These look to be spotless adjustments.  But since this file was modified recently IIRC how come there are spotless changes?  Is spotless not consistent across all environments?  Does this mean there is an IDE somewhere that is not set properly for tabs vs spaces but spotless isn't always catching it?  Observation, not blocker.

jaadams5 (approved):
If this is a good option for install/update via git, is it also a good idea to be retrofitted to install/update via SVN or is there an expectation that SVN will soon be deprecated and only minimally maintained?

src/net/sourceforge/kolmafia/swingui/OptionsFrame.java
midgleyc: These weren't spotless adjustments -- while editing the file I got some warnings about tabs / spaces in the same line, which I fixed.

midgleyc:
There already exists a "don't install dependencies for SVN scripts" option. This one was made in part to be similar to that one.

There is no expectation that SVN will be deprecated -- there are a lot of existing SVN scripts, all dependencies are SVN, etc. I think SVN has been stable without needing maintenance, but I might want to refactor it a bit if I have to make changes.

PR #940, opened 2022-07-30
feat: allow accessing exactly choice.php

midgleyc:
Related to #936.

https://github.com/Ezandora/Far-Future/blob/Release/scripts/FarFuture.ash#L1794 contains the code:
```
//Are we in a game?
string page_text = visit_url("choice.php");
if (!page_text.contains_text("Starship ") && !page_text.contains_text("blue><b>The Far Future")) //hack
{
```

Keep the code working by allowing accessing exactly "choice.php" (with no arguments).

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ChoiceManager.java
+++ b/src/net/sourceforge/kolmafia/session/ChoiceManager.java
@@ -2146,6 +2146,7 @@ public abstract class ChoiceManager {
   public static boolean bogusChoice(final GenericRequest request) {
     if (!ChoiceManager.handlingChoice
         || !request.getURLString().startsWith("choice.php")
+        || "choice.php".equals(request.getURLString())
         || request.responseText == null) {
       return false;
     }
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/session/ChoiceManager.java
```diff
@@ -2146,6 +2146,7 @@ public static void postChoice0(final String urlString, final GenericRequest requ
   public static boolean bogusChoice(final GenericRequest request) {
     if (!ChoiceManager.handlingChoice
         || !request.getURLString().startsWith("choice.php")
+        || "choice.php".equals(request.getURLString())
```
Veracity0, line 2149: This will leave ChoiceManager.handlingChoice as true.
Sometimes that is correct.
But in the Whoops case, it is not correct; we'd like to exit choice processing immediately.

Veracity0 (commented):
I am not approving, but am not requesting changes, because I think this approach is incorrect.

src/net/sourceforge/kolmafia/session/ChoiceManager.java
midgleyc: If you copy the newly added test to your PR I am happy to close this one

Veracity0: OK. I'm adding two tests:

Bogus message and raw choice.php: returns true but no abort
Bogus message and choice.php with arguments: returns true and abort state

midgleyc: Perfect 👍

PR #942, opened 2022-07-31
Change the "mirror" command to completely mirror gCLI

libraryaddict:
Marked as draft, pending resolution of [conversation](https://kolmafia.us/threads/add-option-to-log-cli-output-to-file.27723/#post-168974).

This change replaces the previous usage of mirror which is not well understood, and doesn't seem of interest to most users.

This change allows you to log all gCLI output to a file, allowing you to open it in a browser and view the full html of your session.


src/net/sourceforge/kolmafia/textui/command/MirrorLogCommand.java
```diff
@@ -20,8 +20,8 @@ public void run(final String command, String parameters) {
       RequestLogger.closeMirror();
       KoLmafia.updateDisplay("Mirror stream closed.");
     } else {
-      if (!parameters.endsWith(".txt")) {
-        parameters += ".txt";
+      if (!parameters.endsWith(".txt") && !parameters.endsWith(".html")) {
+        parameters += ".html";
```
gausie, line 24: Keep it as a txt

libraryaddict: Mind explaining why? The log is basically written as html and allowing it to be appended with the .html tag means it can be easily viewed as such.

jaadams5: The tools I use to process mirror files will select the files of interest based upon a .txt ext.  This is all about whose workflow has to be changed and I am quite willing to oppose making me change mine trivial as the adjustments may be :-)

lib/net/java/dev/spellcast/utilities/ChatBuffer.java
```diff
@@ -136,8 +136,16 @@ public JScrollPane addDisplay( final JEditorPane displayPane )
 	 * Sets the log file used to actively record messages that are being stored in the buffer.
 	 */
 
-	public void setLogFile( final File f )
-	{
+	public void setLogFile( final File f ) {
```
jaadams5, line 139: As a general rule we treat Java source in /lib/net as being someone else's code and we do not modify it in case they ever update.  On that alone I would oppose theses changes.  If there is a reason for them and, better yet, a failing test I'd be more willing to relax that rule or at least see if the changes could be done in the calling code instead.

src/net/sourceforge/kolmafia/RequestLogger.java
```diff
@@ -21,7 +22,6 @@ public class RequestLogger extends NullStream {
   public static final RequestLogger INSTANCE = new RequestLogger();
 
   private static PrintStream outputStream = KoLmafiaTUI.outputStream;
-  private static PrintStream mirrorStream = NullStream.INSTANCE;
```
jaadams5, line 24: This is not what I was expecting.  If the problem is something written to the gCLI that is not mirrored and should be then I would expect a line like RequestLogger.mirrorStream.println(message); just needs to be every place something is written to the gCLI.  A discussion of your approach might help me.

src/net/sourceforge/kolmafia/textui/command/MirrorLogCommand.java
libraryaddict: Does the new change of the default being .txt satisfy?

lib/net/java/dev/spellcast/utilities/ChatBuffer.java
libraryaddict: Reverted the changes to using ChatBuffer

src/net/sourceforge/kolmafia/RequestLogger.java
libraryaddict: Reverted the changes to using ChatBuffer

src/net/sourceforge/kolmafia/textui/command/MirrorLogCommand.java
gausie: The file produced will not be a valid HTML file

libraryaddict: It wouldn't be no, but most modern browsers will still render it correctly.

jaadams5:
I still don't have a failing test.

I believe the hypothesis that the presence of absence of HTML influences the logging is incorrect.  I mirror'd prefref and got raw html.  I renamed the file to be .html and it rendered as expected.

I'm willing to force the extension to be txt unless the USER explicitly typed .htm or .html

libraryaddict:
> I still don't have a failing test.
> 
> I believe the hypothesis that the presence of absence of HTML influences the logging is incorrect. I mirror'd prefref and got raw html. I renamed the file to be .html and it rendered as expected.
> 
> I'm willing to force the extension to be txt unless the USER explicitly typed .htm or .html

The test will fail if you revert the changes in request logger

jaadams5:
OK your test is failing for me but the problem is the test.  getMirrorLog is not actually reading the file.  More later since I am waiting for a meeting that just started.

jaadams5:
Bolloxed up my changes to RequestLogger...

jaadams5:
I think my change to RequestLogger is equivalent to yours but it took me a while to see that.  I didn't change printList similarly because I wasn't certain things weren't being written twice.  The test should be restored so it allows htm and html if the user specified the ext. but I haven't done that yet.  I'd like some real world testing but I think we are closer than we were before.

src/net/sourceforge/kolmafia/textui/command/MirrorLogCommand.java
gausie: I feel like this is resolved now, we're making it txt unless explicitly set to html

jaadams5:
I'd like a test for printList and then I think I'd approve this, or more likely, open it for review first :-)

jaadams5:
I'm not happy.

mirroring ashq print("Mirror test."); now adds <br> in the mirror which AFAIK is wrong.

libraryaddict:
> I'm not happy.
> 
> mirroring ashq print("Mirror test."); now adds in the mirror which AFAIK is wrong.

Actually, that's unchanged behavior.

jaadams5:
> > I'm not happy.
> > mirroring ashq print("Mirror test."); now adds in the mirror which AFAIK is wrong.
> 
> Actually, that's unchanged behavior.

Actually GitHub ate my "less than b r greater than".

I was trying to say

mirroring ashq print("Mirror test."); now adds <br>  in the mirror.

There's probably a test here.

libraryaddict:
> > > I'm not happy.
> > > mirroring ashq print("Mirror test."); now adds in the mirror which AFAIK is wrong.
> > 
> > 
> > Actually, that's unchanged behavior.
> 
> Actually GitHub ate my "less than b r greater than".
> 
> I was trying to say
> 
> mirroring ashq print("Mirror test."); now adds an HTML break in the mirror.
> 
> There's probably a test here.

Yes, which is what the html test is? That test would previously fail.

jaadams5:
I wonder if we are at cross-purposes?

My goals:

Everything that is written to the gCLI when mirror is enabled is also written to the mirror file.

Nothing is duplicated in the mirror unless it was duplicated in the gCLI.

Nothing is in the mirror unless it was also in the gCLI.

gCLI text with embedded HTML in the gCLI will be written as text with HTML tags to the mirror.

gCLI text with no embedded HTML will not have embedded HTML in the mirror.

In addition to automated testing a human using a text editor and with the gCLI on screen should be able to confirm that the gCLI and the mirror have the same content.

libraryaddict:
> I wonder if we are at cross-purposes?
> 
> My goals:
> 
> Everything that is written to the gCLI when mirror is enabled is also written to the mirror file.
> 
> Nothing is duplicated in the mirror unless it was duplicated in the gCLI.
> 
> Nothing is in the mirror unless it was also in the gCLI.
> 
> gCLI text with embedded HTML in the gCLI will be written as text with HTML tags to the mirror.
> 
> gCLI text with no embedded HTML will not have embedded HTML in the mirror.
> 
> In addition to automated testing a human using a text editor and with the gCLI on screen should be able to confirm that the gCLI and the mirror have the same content.

That sounds right.

I understand that your problem is the line break? That's a part of what's written to gcli which confuses me on what you're expecting

jaadams5:
I made a version of MirrorTest that executes against main, without these changes.  

https://github.com/kolmafia/kolmafia/pull/951

At this point I'm not certain our tests for this change are correct but step 1 is to understand the existing behaviour.

jaadams5:
Closing in favor of #981

PR #943, opened 2022-07-31
feat: more support for The Big Book of Every Skill

midgleyc:
Add a preference for The Big Book of Every Skill use (once per day), note that it's not consumed on use, and add the skill obtained from it.


Veracity0:
Note that marking it "reusable" (rather than "usable") in items.txt also marks it as not going away on use.
And that means that $item[The Big Book of Every Skill].reusable will (correctly) return true.

Doing a return rather than a break in the mammoth switch statement in UseItemRequest short circuits the check to see if it should be reused, so that is correct, regardless.

Veracity0 (approved):
I put mine in my display case, since I have all skills HC permed. But if people have the item and don't have all skills, it would be nice to have it in breakfast. Preferably configurable.

Heck, I am considering having my five test multis each ascend once through Journeyman to (eventually) make their lives way easier. :)

I'm not insisting you do this in this PR.

midgleyc:
I agree on configurable breakfast, but won't do it in this PR.

PR #947, opened 2022-08-03
Add checking for availability of more Journeyman zones

Veracity0:


gausie:
@Veracity0 what do you think about exposing isCurrentlyAccessible to the script runtime?

gausie (dismissed):

Veracity0:
I think that is a fine idea, especially considering that there are scripts which try to reproduce this functionality. And CanAdv has disappeared off sorceforge?

I'll make a "Feature" request so that people can tell us what we're missing. :)

I already know that we are missing the Cola Battlefield zones - level 4 or 5 and have returned Fernswarthy's Key the first time.

Considering that I can envision a potentially heft test suite for this functionality, it deserves another PR.

gausie:
> I think that is a fine idea, especially considering that there are scripts which try to reproduce this functionality. And CanAdv has disappeared off sorceforge?
> 
> I'll make a "Feature" request so that people can tell us what we're missing. :)
> 
> I already know that we are missing the Cola Battlefield zones - level 4 or 5, have ascended at least once, and have returned Fernswarthy's Key for the first time.
> 
> I'll add the function, but I can envision a potentially heft test suite - especially when people start reporting issues - and that should be another PR.

A perfect approach IMO

jaadams5 (approved):

jaadams5:
Is there an historical reason why AdventurePool has int X, String X_ID and sometimes both?

Veracity0:
Yes. The id field of a KoLAdventure is a String. If it is adventure.php, it is the string representation of the snarfblat parameter - an integer - but not all KoLAdventures are adventure.php

PR #949, opened 2022-08-03
Add runtimelibrary function to equip an item to a specific familiar (even in terrarium)

gausie:


horrible-little-slime:
Just to clarify, this is desirable because the new stillsuit functions even when in the terrarium. It is also useful for equipping the blue plate to the shorter-order cook.

gausie:
Yeah that. It really all exists already so why not.

midgleyc (approved):
Might be nice to assert on requests sent, but this is fine.

PR #950, opened 2022-08-04
Refactor KoLAdventure validation.

Veracity0:
The goal is for KoLAdventure.validate1 (exposed via ASH ```can_adventure(location)```) to tell you if a zone is mechanically available to you right now.

- true if your level/path/quest status/path/etc. means you COULD adventure in that location right now.
Some zones require an outfit. We sometimes check that you own such - the Pirate Ship, for example - but mostly not; we figure your script will know what you need.
Some zones may not be accessible right this second, but if you have items in inventory which will make them available - sonars, enchanted bean - we'll give you a true.
KoLAdventure.validate2 will put on outfits and use such (simple) items to finish giving you access.
- false if there is literally nothing you can do to go to the zone.
For example,
-- wrong zodiac sign or class or path
--  IOTM supplied content, but you have Standard restrictions and the item is out of Standard
- false if there is nothing you can do to go to a zone right now without spending expensive resources.
For example,
-- Adventuring to gain levels or fulfill quests.
-- Using a (possibly expensive) day pass to access IOTM adventuring content.

Note that this is still used in the "automated adventuring" workflow:

1) validate1 to decide if an adventuring zone is available
2) (User's betweenBattle script to, perhaps, acquire outfit pieces)
3) validate2 to do simple things that validate1 saw were already possible

This is ready to be released into the wild for testing. It is almost complete (additional work described at end), but should be eminently usable as is.

And although I wrote a bunch of tests - KoLAdventureValidationTest has over 1000 lines - I did not test anything close to all adventuring zones. I expect to add many more in future PRs. :)

User visible changes:

- I added some adventures into one-adventure zones so I could attach an IOTM-related item. This means that, for example, The Neverending Party doesn't sort into the "Town" zone in the GUI - although its parent zone is "Town". And there is not a single "Psychoses" zone; each of the seven jars has its own zone - whose parent zone is "Psychoses" - so I can tag each zone with the appropriate jar.

If that is too annoying, perhaps we'll need to revamp how the GUI sorts zones with parent zones. Perhaps it should group together every zone whose parent (or parent of parent ...) is the same "top level" zone?

- The betweenBattle script can no longer acquire items that would have made a less-strict validate1 return true and allow validate2 to return true. Like day-passes, which are probably too expensive. :)

Tasks:

- AdventurePool.XXX -> an int for adventure.php
AdventurePool.XXX_ID -> a non-int for another formSource
- Renamed Postwar Junkyard to The Junkyard - which is what KoL calls it.
- Renamed Hippy Camp Disguised to Hippy Camp (Hippy Disguise)
Inspired by Wartime Hippy Camp (Frat Disguise)
- Renamed Frat House Disguised to Frat House (Frat Disguise)
Inspired by Wartime Frat House (Hippy Disguise)
- Put Pirate ship areas into a Pirate zone.
- Put all retired zones underneath Retired and when AdventureDatabase reads all the areas, if they have Retired anywhere in their zone's parents, add them to a set of retired areas
- Reorganized zonelist.txt and adventures.txt into a more logical order
- Reorganized KoLAdventure.validate1 to check by zones as much as possible, and in order of the Council quests.
- Added KoLAdventureValidationTest.
Added tests for the 6 "Island" and 4 "IsleWar" zones for the Hippy Camp and the Frat House
Added tests for Mothership adventures available only in Bugbear Invasion
Added tests for all Cobb's Knob areas
Added tests for all Cyrpt areas
Added tests for all Pirate areas
Added tests for all airport charters - no access, all access, daily access
- zones can now depend on an Ascension Path or upon an item
Nested zones can choose their own path (unlikely) or item (reasonable) or inherit from their parent.
- Early on in validation, if a zone requires a specific path, check and reject if not on it
- Early on in validation, check if required item is disallowed to you by Standard restrictions and reject if so.
- I added paths to four zones and items to many zones.

Observations:

HALF-ASTRAL should be a LimitMode; you must adventure for 5 turns in Astral zone
SHAPE_OF_MOLE should be a LimitMode; you must adventure for 12 turns on Mt. Molehill
FORM_OF_BIRD should be a LimitMode; replaces all combat skills and cannot use another llama lama gong
(So should FORM_OF_ROACH, except that is only in effect within a choice chain,, so it is pointless)

Do LimitModes work correctly in Validation?
Write some tests!

The "LOV Entrance Pass" is the day pass to The Tunnel of L.O.V.E.
The "Deep Machine Tunnels snowglobe"  is the day pass to The Deep Machine Tunnels
The "inflatable LT&T telegraph office" is the day pass to the LT&T office
KoLmafia does not support any of those yet.

Not yet accurately supported:

- El Vibrato Island
- Orchard Sidequest - progress depends on status effects and you can "finish" it.
- Farm Sidequest - exactly 3 duck zones chosen and you can "finish" it.
- Gingerbread City - limited daiiy turns
- Spacegate - limited daiiy turns
- FantasyRealm - limited daily turns
- PirateRealm - limited daily turns


gausie (approved):
Truly fantastic

PR #951, opened 2022-08-05
MirrorTest

jaadams5:
Test from https://github.com/kolmafia/kolmafia/pull/942 tweaked to work without any of the code changes from 942.

Need a test that writes something to the GCLI but is not written to the mirror file.

Need to explore the assertion that adding > will trigger colorization and how that relates to something not being mirrored.


jaadams5:
@libraryaddict 

Comments?

libraryaddict:
On phone and will have to look at it tomorrow, but initial thought is that it does not contain a super minor change that switches it from contains text, to assert equals.

The original contains check was due to chat buffer being the original change which really didn't play well with multiple distros. Could also be the case here, not sure.

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
```diff
@@ -0,0 +1,163 @@
...
+    execute("");
+    // Read the log
+    String mirrorOutput = getMirrorLog("test_mirror_list_strings.txt");
+    // Test that the output matches as expected
+    assertEquals("Line 1\n" + "Line 2", mirrorOutput);
+  }
+
+  @Test
+  public void testMirrorListSkills() {
+    try (var cleanups = new Cleanups(withSkill(1), withSkill(2))) {
```
gausie, line 113: This doesn't work on all OS I think? @midgleyc couldn't you not do this

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
```diff
@@ -0,0 +1,163 @@
...
+    execute("chats/test_writes_html.txt");
+    // When `> ` is used as a prefix, the RequestLogger will colorize it assuming its a command
+    // input.
+    RequestLogger.printLine("> Fake command input");
+    // Log another line, this should be raw.
+    RequestLogger.printLine("Raw Line");
+    // Close the mirror
+    execute("");
+    String mirrorOutput = getMirrorLog("test_writes_html.txt");
+    assertTrue(mirrorOutput.contains("<font color=olive>> Fake command input</font><br>"));
```
gausie, line 65: These would all have nicer output on failure if they were made with
```
assertThat(mirrorOutput, containsString(...));
```

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
midgleyc: couldn't I not do what?

This is fine, it just makes the IDE complain that `cleanups` is unused (which it is, but you need to assign to a variable in a try-with-resources), hence why I normally go for `var cleanups = ...; try (cleanups)`.

jaadams5: IntelliJ isn't flagging var as unused for me.

midgleyc: does it flag any variables as unused? If it does there might be a setting I'm missing.

jaadams5: Yes.  I normally use Analyze->Inspect.  I introduced an unused variable and it was recognized as such (by being gray in the IDE text display) and flagged by the inspection.

jaadams5:
I am somewhat comfortable with these tests.  I would love a failing test that showed something written to the gCLI but not the mirror but it may be that the unmodified mirror command really does mirror everything so no such test is possible.

The tests assume two things.  First, a mirror file will have a .txt extension regardless of what the user typed.  Second, if HTML is written to the gCLI HTML will be written to the mirror.  

I'll let this sit for a day but if there is no advice to the contrary I am going to undraft it and commit it.  After that we can revisit https://github.com/kolmafia/kolmafia/pull/942 and decide what should happen.

@gausie @libraryaddict

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
jaadams5: So I did a recompile from within the IDE and the complier complained about unused.  The manually run analysis did not.  So I am seeing what you did

gausie: Ah it was an IntelliJ thing. So it sounds like you need to change this frono?

gausie (approved):
Seems good to me!

libraryaddict:
I have a review pending, asking about a test that is labeled incorrectly and made to deliberately successful using the incorrect output.

midgleyc:
I think you need to submit the review before anybody except you can see it.

libraryaddict (changes requested):

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
```diff
@@ -0,0 +1,177 @@
...
+    execute("chats/test_writes_html.txt");
+    // When `> ` is used as a prefix, the RequestLogger will colorize it assuming its a command
+    // input but that is not true at the moment, hence plain text assertion.
+    RequestLogger.printLine("> Fake command input");
+    // Log another line, this should be raw.
+    RequestLogger.printLine("Raw Line");
+    // Close the mirror
+    execute("");
+    String mirrorOutput = getMirrorLog("test_writes_html.txt");
+    assertThat(mirrorOutput, containsString("> Fake command input"));
```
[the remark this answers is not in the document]
jaadams5: Specifically which test?

I broke this out because the other code and our discussions there make me think we are not talking about the same thing.

The alleged bug, to be fixed, is that a message appears in the gCLI but does not appear in the mirror.  I have been unable to write a failing test for this and as a result I am inclined to believe there is no such bug.

There is a feature request that, if the text written to the gCLI contains HTML (and is rendered as such in the gCLI) then the mirror should contain the HTML as well.  Tests to exercise changes to implement that feature have been preserved in this code but they have been converted to passing tests rather than commented out.  It could have been clearer, I suppose, if they had been Disabled.  In any event once it is established that the mirror command is mirroring everything, then the HTML inclusion can be implemented and my hope is that these tests will fail as a result and need to be changed.  The tests are documenting current behavior with the expectation that changed behavior will be detected.  If that is too confusing tests can be disabled.

The big thing here is the inability to find a test that demonstrates something written to the gCLI but not the mirror.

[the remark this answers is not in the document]
jaadams5: > It's a bug that the html is not being printed as it does in gCLI, not a feature.

Ignoring the name, the test shows that the gCLI output and the mirror output are identical and that neither contains HTML.

Since running mafia, I could not get ">" to trigger html to the gCLI I assumed your expectation otherwise was incorrect.  Maybe I was mistaken?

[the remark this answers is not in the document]
libraryaddict: Ah, I see what you mean.

I believe it's a bug in that what's written to GCLI isn't exactly what's written to mirror. 

Mirror writes exactly what the logger was given, but the logger applies html to the input based on rules which is what's displayed in cli.

So it's arguable if it's a bug fix or a feature request. I was talking under the assumption that it's a bug, because I expect to see contents identical to what's written in cgli in the mirror.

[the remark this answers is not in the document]
jaadams5: > I believe it's a bug in that what's written to GCLI isn't exactly what's written to mirror.

We're getting closer, I think.

The original assertion was that things were being written to the gCLI but not to the mirror and in the context of the original "complaint" the possibility that the missing information was merely HTML formatting was not really considered/expected.  

My thinking right now is that I want a test that doesn't use a mirror and shows HTML written to to the gCLI.  That's probably in RequestLoggerTest.  Once we have that then we can wrap a mirror around it and see the existing behavior as a test.  I'm headed down that rabbit hole but might not come out today.

jaadams5:
Reverted to draft since there are a couple of things that can be resolved.

jaadams5:
Closing in favor of #981

PR #952, opened 2022-08-05
Added output to cmc command that tells the user how many turns are re…

Aenimus:
…quired for a certain pill.


src/net/sourceforge/kolmafia/textui/command/ColdMedicineCabinetCommand.java
```diff
@@ -1,12 +1,10 @@
 package net.sourceforge.kolmafia.textui.command;
 
-import java.util.List;
-import java.util.Map;
-import java.util.SortedMap;
-import java.util.TreeMap;
+import java.util.*;
```
jaadams5, line 3: Would you please expand the imports instead of collapsing them?  I thought expanding them was a code convention enforced by spotless but apparently not.

jaadams5, line 3: Would you please expand the imports instead of collapsing them? I thought expanding them was a code convention enforced by spotless but apparently not.

jaadams5 (changes requested):

src/net/sourceforge/kolmafia/textui/command/ColdMedicineCabinetCommand.java
gausie: Yeah I don't think it's something we can check with spotless at the moment (https://github.com/diffplug/spotless/issues/649)

Aenimus: It's a setting in my IntelliJ preferences I need to disable. I'll post a fix later today.

jaadams5: Ditto and over the years the preference has not been as sticky or global as I assumed.  At least with reviews it gets caught before commit :-)

jaadams5 (approved):

jaadams5 (approved):

PR #955, opened 2022-08-06
feat: allow installing git dependencies from svn project

midgleyc:
Allow projects checked out using SVN to contain dependencies that should be checked out using git.

This changes the behaviour of SVN dependency installs in a few ways:
* installs are now depth-first instead of breadth-first
* there is no dependency recursion limit
* instead of skipping the entire file if any dep is invalid, only the invalid dep is skipped

I think these are all reasonable.


jaadams5 (approved):

jaadams5:
> * there is no dependency recursion limit

I think as implemented one is no longer needed but it will be interesting to see if a user maliciously or inadvertently constructs a case where it is ;-)

PR #958, opened 2022-08-08
Extended Stillsuit support

gausie:
Outstanding questions:

- What do we do with the current Buzzed on Distillate effect (easy enough to parse modifiers and store in a pref)
- What do we do with the next Buzzed on Distillate effect (at the moment I am parsing the modifiers but not sure what to do with them)

EDIT: I've tied this up to work as is, better support can (and probably will!) come down the line


src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -1393,15 +1393,7 @@ public void externalExecute() {
...
-    return requestURL == null
-        ||
-        // Disallow mall searches
-        requestURL.contains("mall.php")
-        || requestURL.contains("manageprices.php")
-        || requestURL.contains("backoffice.php")
-        ||
-        // Disallow anything to do with chat
-        request.isChatRequest;
+    return false;
```
midgleyc, line 1396: is this change related, or has it snuck in?

gausie: One day I'll stop accidentally adding this... test... code

src/net/sourceforge/kolmafia/session/StillSuitManager.java
```diff
@@ -0,0 +1,57 @@
...
+
+    var familiar =
+        KoLCharacter.familiars.stream()
+            .filter(f -> f.getItem() != null && f.getItem().getItemId() == ItemPool.STILLSUIT)
+            .findFirst()
+            .orElse(null);
+
+    int drams = 0;
+    if (familiar == KoLCharacter.getFamiliar()) {
+      drams = 2;
```
midgleyc, line 32:
```suggestion
      drams = 3;
```

midgleyc: unless I am wrong about what this is, equipped familiars give 3 drams of sweat?

gausie: You're right! I forgot

midgleyc (approved):

PR #962, opened 2022-08-09
Additions to KoLAdventure validation

gausie:
Adding a cases to @Veracity0's new body of work

So far this adds support and testing for:

* Down the Rabbit Hole
* Spacegate
* Hippy Orchard
  * Do we want to use the glands in prepareForAdventure?
* FantasyRealm
  * Do we want to pick up the G.E.M. if the user already has their choice selected for the Welcome center? Probably not, but maybe.


src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -959,14 +967,23 @@ public boolean canAdventure() {
...
-      return true;
+      // Once Filthworm Queen is defeated none of the zones can be accessed
+      if (InventoryManager.hasItem(ItemPool.FILTHWORM_QUEEN_HEART)
+          || !Preferences.getString("sidequestOrchardCompleted").equals("none")) {
+        return false;
+      }
+
+      return switch (this.adventureNumber) {
+        case AdventurePool.FILTHWORM_HATCHING_CHAMBER -> true;
+        case AdventurePool.FILTHWORM_FEEDING_CHAMBER -> KoLConstants.activeEffects.contains(
```
Veracity0, line 979: I think requiring the effect OR having the appropriate gland would be good.

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -1816,6 +1855,26 @@ public boolean prepareForAdventure() {
       return true;
     }
 
+    if (this.zone.equals("FantasyRealm")) {
+      if (!InventoryManager.hasItem(FANTASY_REALM_GEM)) {
```
Veracity0, line 1859: This is reasonable, but a counterargument would be how the Astral adventures are intended to work:

If you ask to take such-and-such a trip, if you have not already consumed the mushroom and chosen, we will use the mushroom and trigger the initial choice adventure to choose the trip. At that point, you are all set up and can go right into the chosen zone.

Here, the "get a GEM (while choosing a hat) and equip it" is, sort of, the equivalent of "use a mushroom and make a request to choose the trip". The difference is, the "type of hat" is not implied by the adventure ID.

In PrepareForAdventure, the user is saying "suit me up! I am ready to go in!"

Now, perhaps if they wanted a particular hat to actually wear in this zone, presumably they would have picked it first - and got the GEM. If they didn't do that, if they have set a "default" for the hat picking choice, perhaps we should automate it for them and get the GEM as well. Is the hat choice GUI-configurable?

I'm leaning towards that, but don't feel strongly about it.

Veracity0 (changes requested):
Very nice!

KoLAdventureValidationTest is going to end up as our largest collection of tests. :)

src/net/sourceforge/kolmafia/KoLAdventure.java
gausie: I started doing this but it would actually make it quite complicated - now our canAdventure function needs to know what locations *would* become available if you were to complete the Welcome choice and return true for them, but that pref is only set after visiting. It's possible, I'd split out the function that currently works that out and put it in FantasyRealmRequest or something. But I think it's probably beyond the scope of this change, and I feel like doing a bit more breadth before this depth. What do you think?

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -1368,7 +1356,33 @@ void cannotAdventureWhenQueenHeartIsHandedIn() {
               withProperty("sidequestOrchardCompleted", "frat"));
```
Veracity0, line 1356: Technically, IslandManager sets this to "hippy" or "fratboy", not "frat".
In retrospect, "frat" would have been better.

But since you correctly check for "sidequestOrchardCompleted" != "none", it doesn't really matter. :)

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -1383,6 +1397,36 @@ void cannotAdventureInUnknownNewOrchardZone() {
         assertThat(UNKNOWN.canAdventure(), is(false));
       }
     }
+
+    @CartesianTest
+    void preparingForChamberUsesGland(
+        @CartesianTest.Enum Chambers chamber,
+        @Values(booleans = {true, false}) final boolean hasGland,
```
Veracity0, line 1404: Consistency: previous tests are "haveEffect" and "haveGland".
This one is "hasEffect" and "hasGland".

Which is it: "I have the effect" or "the character has the effect"?

src/net/sourceforge/kolmafia/KoLAdventure.java
Veracity0: I didn't know that. I had the impression that the unlocked zones persisted, not that we learned them by looking at the map.

Yeah, this is fine. Once we have canAdventure and prepareForAdventure  in-synch - the former checks for everything the latter will provide - then we can consider going deeper into individual zones.

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
gausie: I will rename this one :)

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
gausie: Fixed anyway

PR #963, opened 2022-08-09
Allow deprecated adventure names to be used as synonyms

Veracity0:
Rather than outright breaking existing ASH scripts, give them "friendly warnings"

```
> ash $location[Hippy Camp (Hippy Disguise)].id

Returned: 65

> ash $location[Hippy Camp in Disguise].id

Changing "Hippy Camp in Disguise" to "Hippy Camp (Hippy Disguise)" would get rid of this message. (char 11 to char 33)
Returned: 65

> ash $location[Frat House (Frat Disguise)].id

Returned: 29

> ash $location[Frat House in Disguise].id

Changing "Frat House in Disguise" to "Frat House (Frat Disguise)" would get rid of this message. (char 11 to char 33)
Returned: 29

> ash $location[The Junkyard].id

Returned: 154

> ash $location[Post-War Junkyard].id

Changing "Post-War Junkyard" to "The Junkyard" would get rid of this message. (char 11 to char 28)
Returned: 154
```


gausie (approved):
This is reasonable - there are also some breakages from zone name changes but they're harder to change since they're just pulled as strings and compared

Very friebdly indeed!

PR #964, opened 2022-08-09
feat: allow setting a property to not abort on choice

midgleyc:
As suggested in https://kolmafia.us/threads/recent-problems-with-batfellow-ash.27847/#post-169118

There exist old scripts that have worked up to now and probably won't receive updates. Allow an out for users of those scripts to keep them going.

I considered whether the preference should be global or user. I figured global, as you're using this because you're using some script that needs it, and the script is shared across characters.

Abort by default, because we consider choice access when not in a choice as a bug that should be fixed.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1,4 +1,5 @@
 1
+global	abortOnChoiceWhenNotInChoice	true
 global	addChatCommandLine	false
 global	addCreationQueue	true
 global	addStatusBarToFrames	false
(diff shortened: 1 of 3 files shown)


Veracity0 (approved):

gausie:
I'm inclined to think we should think harder about this. I would want to make it harder for people to just set this to true and never think about it again. Can we let people add scripts to a set of script paths that are allowed to error instead of aborting. I dunno

midgleyc:
The original issue was caused by someone running a script that hit a choice URL without being in the choice, and they complained about the error they got.

I figure if somebody explicitly sets this pref to ignore errors in scripts, then gets a script behaving a way they don't want, they have literally requested that behaviour.

Contrariwise, if somebody wants scripts to work and sometimes do things Mafia can get confused by, they can set the pref and forget it. I think that's still a reasonable way to want Mafia to work.

PR #965, opened 2022-08-09
Update paths for Theraze scripts

Veracity0:
Theraze restored his sourceforge directory but changed the paths.

And....maybe he is struggling to NOT change his paths.

If he fixes that, I will delete this. Stay tuned.


midgleyc (approved):
I think the idea is good, but also those URLs aren't working for me at the moment.

After this becomes stable, I am happy for it to be merged.

Veracity0:
Theraze has restored his SF repositories into the same place as before.
I just checkedout out CanAdv and The Sea, no problem.

Closing this as no longer necessary.

PR #966, opened 2022-08-09
chore: update junit, svnkit

midgleyc:
JUnit has [changed](https://junit.org/junit5/docs/5.9.0/release-notes/) behaviour to be consistent with their docs and not accept private lifecycle methods. This leads to a lot of churn.

SVNKit has some [bug fixes](https://svn.svnkit.com/repos/svnkit/trunk/CHANGES.txt) that look worth having.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -56,13 +56,13 @@ repositories {
 
 dependencies {
 	// Use JUnit Jupiter for running JUnit5 tests.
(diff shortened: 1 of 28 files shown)


jaadams5 (approved):
I assume you made the build changes and then made the others until everything compiled and tested?

midgleyc:
Yeah, I just committed in reverse order.

I used `id 'com.github.ben-manes.versions' version '0.42.0'` to check the versions but didn't commit it.

PR #968, opened 2022-08-10
Market quests

Veracity0:
Two big things:

1) Reported on forum that we pick wrong outfit for disguised adventuring in verge-of-war hippy camp
Yes. We had the wrong adventure IDs for both verge-of-war disguised camps
2) Skeleton Store, Madness Bakery, and Overgrown Lot can all be unlocked immediately.
Two ways 
-- Accept quest from appropriate shop
-- Use dropped item from one of the three starter zones. This does not start the quest.
- canAdventure() returns true for all three zones
- prepareForAdventure will unlock the area by either using the item or accepting the quest.
- added "skeletonStoreUnlocked", "madnessBakeryUnlocked", overgrownLotUnlocked"
(distinct from the three quests, since you unlock without starting quest)
- set those when accept quest, use item, or see the zone in town.

Things to do before this is done:

- I have outfit tests for Frat House (since that was reported on forum) but not yet Hippy Camp. Easily done by copy/paste/modify from the Frat House tests.
- I need tests for canAdventure() and prepareToAdventure() for Skeleton Store, Madness Bakery, and Overgrown Lot, opened either via item or shopkeeper.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia;
 
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


gausie:
@Veracity0 I also picked this up from the forum and have some tests that you could take from my branch. Or merge that and resolve here

Veracity0:
> @Veracity0 I also picked this up from the forum and have some tests that you could take from my branch. Or merge that and resolve here

I'll pick your tests, but not merge your PR - since I think you made the wrong fix. The adventure numbers in adventures.txt were correct (per the Wiki) - but the constants in AdventurePool were incorrect.

Veracity0:
I appear to have somehow pushed what I had right into main?
Well, sorry for the lack of review.

I'm closing this and will reopen a PR with my new branch.

PR #975, opened 2022-08-12
Fix up some missing path points trackers and incorrect path images

gausie:
- Nuclear autumn can be tracked like any other path, rather than just looking at the item
- 2CRS is worth tracking because you get that number of extra enchants on your ring
- Some path images were just wrong... including wosp!


jaadams5 (approved):

Veracity0:
Is this supposed to update the counters from your Ascension History?

I have 19 Hardcore TCRS runs - twocrazydice.gif - and twoCRSPoints = 0.
I have 9 Hardcore and 4 Normal Nuclear Autumn runs - radiation.gif - and nuclearAutumnPoints = 22

PR #976, opened 2022-08-12
Apply Meteor Showered adjustments as soon as we see it so that it can apply to later steps in a macro FightRequest

gausie:
Do we need to do this for more combat skills that give effects?

- Turbo
- Giant Growth
- Fire extinguisher: Foam Yourself
- Fire Extinguisher: Blast the Area
- Fire Extinguisher: Take a Drink
- Camel, spit on me
- Form of {Bat,Wolf,Mist Cloud}
- Spew Poison
- Anything that gives Everything Looks {Yellow, Red, Blue}
- Disco rave skills

In theory we could even need to check for The 17 Cuts - something could happen later in a macro combat that is based on total number of active effects or something


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9468,6 +9471,9 @@ private static void payActionCost(final String responseText) {
 
       case SkillPool.METEOR_SHOWER:
         if (responseText.contains("consult your mental index") || skillSuccess) {
+          KoLConstants.activeEffects.remove(METEOR_SHOWERED);
+          KoLConstants.activeEffects.add(METEOR_SHOWERED);
+          KoLCharacter.recalculateAdjustments();
```
jaadams5, line 9476: I don't understand the mechanics.  Why are we removing then adding?  Why would recalculation not suffice?  Is this somehow defensive in case the effect had not "expired" after one turn and was then cast again?

Just curious, not blocking.

jaadams5 (approved):

src/net/sourceforge/kolmafia/request/FightRequest.java
gausie: Yeah this is to stop it being added twice. I could also do addToList I suppose. It needs to be added because it's not in the list yet (it only gets added to the list by the charpane refresh

PR #978, opened 2022-08-12
Validation Tests for more zones

Veracity0:
More canAdventure/prepareForAdventure validation tests.

1) The Bat Hole areas may require stench resistance or sonar-in-a-biscuits to open.
You can have either used sufficient sonars or have them in inventory ready to use.
2) withEquipped needs to call KoLCharacter.recalculateAdjustments to account for item enchantments.
3) Tests for Pirate Ship areas with Swashbuckling Getup or pirate fledges
4) Tests for Beanstalk areas (including planting enchanted bean, if necessary)
5) Tests for Hidden City areas
6) Tests for Cobb's Knob Throne room.
harem girl outfit (equipped or unequipped) and Knob Goblin perfume effect (including using potion if not in Beecore)
elite guard uniform (equipped or unequipped) and Knob cake (including creating from ingredients)
7) Validate Farm areas based on which were selected/cleared
8) Detect "there are no more ducks here" message and mark the area cleared.
*** No detection yet of defeating the last duck and not visiting the area again to see that message.
Not sure how to do that, since there might be either 5 or 10 ducks in an area.
9) Tests for McLargeHuge areas, including Shrouded Peak and Icy Peak
10) Support for Drunken Stupor and St. Sneaky Pete's Day Stupor.
11) Put Grimstone areas into a Grimstone zone.
12) Use parentZone as appropriate during validation
13) Add rootZone to KoLAdventure.
Add $location.root
Use rootZone to check for "Removed" areas
14) The Drip requires a drip harness, which it will equip. Add tests.

Note that three areas require elemental resistance:
1) Guano Junction - stench >= 1
2) Shrouded Peak - cold >= 5
3) Icy Peak - cold >= 1

As of this PR, these are enforced in canAdventure.
If we decide to allow the betweenBattleScript to "fix" things by getting elemental resistance:
The same elemental resistance checks are done in prepareForAdventure.

We should get community feedback on what, exactly, betweenAdventure scripts are allowed to do.
As of now, they cannot get required outfits or resistances, since canAdventure checks for those.

I also notice we have five item-driven zones where canAdventure will return true if the item is available, but will not actually use it in prepareForAdventure. This is probably a good thing, since the said items can very expensive.
But should we even return true if we will not use them?

This is a matter for discussion, but I moved the InventoryManager.hasItem to AFTER the KoLConstants.activeEffects check, to perhaps, make it more obvious that we want to think about this and maybe remove. Also, the "***" comments I added in front of possibly dubious hasItem calls might signal that. :)


Veracity0:
I've been waiting for 48 hours for somebody to review this. I have added on with this and that, at least, but it has been good to go every time I submitted.

I have another project that builds on this that I will be starting tomorrow. I need a clean start.

If there is no review when I look again tomorrow morning, I'm going to merge it without one,

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -454,7 +450,7 @@ public boolean canAdventure() {
     // There are lots of zones from past events (like Crimbos) that are no
     // longer available. AdventureDatabase maintains a handy Set of all such,
```
midgleyc, line 451: this comment is no longer accurate

Veracity0: Fixed.

jaadams5 (approved):

PR #979, opened 2022-08-13
Remove unused preferences.

heeheehee-kolmafia:
These were identified to have no references elsewhere in the codebase.

I manually took a pass to further filter this set down (removing things like HC vs SC breakfast, choice adventures, PvP season prefs, etc).

Would be good to have a second set of eyes to double-check that I didn't make any mistakes here.


Veracity0 (approved):
It would be nice if there was a technique such that removing a property from defaults.txt would also remove it from preferences.

I wrote a script which would generate a list of all properties not in default.txt and let me decide which ones were from scripts and which from dead properties.

If you can come up with a method to facilitate this, cool.

I am not expecting not.

gausie (changes requested):
https://github.com/kolmafia/kolmafia/blob/e58cf2bcc7fae7c5d34675ae2ed27ad676a0e017/src/net/sourceforge/kolmafia/session/ResultProcessor.java#L320

Keep the bird a day mod prefs please!

gausie:
Also umiversalSeasoningAvailable is composed

https://github.com/kolmafia/kolmafia/blob/fbfd2252d01c76a8b52d01d14a8f23573646e162/src/net/sourceforge/kolmafia/request/SwaggerShopRequest.java#L70

Veracity0:
> Also universalSeasoningAvailable is composed

Seems to be true.

jaadams5:
> I wrote a script which would generate a list of all properties not in default.txt and let me decide which ones were from scripts and which from dead properties.

I seem to recall doing something similar but I think I used awk and access to source code so perhaps yours is the better candidate for publication than mine :-)

zarqon has some similar capability in PrefRefPlus.

heeheehee-kolmafia:
> https://github.com/kolmafia/kolmafia/blob/e58cf2bcc7fae7c5d34675ae2ed27ad676a0e017/src/net/sourceforge/kolmafia/session/ResultProcessor.java#L320
> 
> Keep the bird a day mod prefs please!

Done. Same with universalSeasoningAvailable.

src/data/defaults.txt
```diff
@@ -819,7 +811,6 @@ user	lastDwarfHopper3
 user	lastDwarfHopper4
 user	lastEVHelmetValue	0
 user	lastEVHelmetReset	-1
-user	lastEasterEggBalloon	-1
```
Veracity0, line 822: This KoL change is a tragedy. :-(

Veracity0 (approved):

PR #980, opened 2022-08-13
Add absorbed_monsters() to expose absorbed Grey You monsters

libraryaddict:
Could also be renamed to `grey_absorbs`  `grey_absorbed` and possibly expose & track reprocessed monsters beyond just preferences.


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -9698,4 +9704,17 @@ public static Value expected_cold_medicine_cabinet(ScriptRuntime controller) {
 
     return value;
   }
+
+  public static Value absorbed_monsters(ScriptRuntime controller) {
+    AggregateType type =
+        new AggregateType(DataTypes.MONSTER_TYPE, GreyYouManager.absorbedMonsters.size());
```
Veracity0, line 9710: You made this an array rather than a set.
Why?

As an array:

foreach index, mon in xxx

As a set:

foreach mon in XXX

And

XXX contains mon

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
libraryaddict: By set, I assume you mean Map? Changes has been made.

Veracity0: Yes.

Array:  monster []
Sat: boolean [monster]

Veracity0 (approved):

PR #981, opened 2022-08-13
Mirror related feature requests

jaadams5:
1) Allow the mirror file name extension to be htm or html is the user specifies eother of those.
2) If something is written to the gCLI with HTML then write the HTML to the mirror as well.

Also some lint.

Includes a test of dubious value but demonstrates that the request logger does not routinely add HTML.


jaadams5:
Finally.  @libraryaddict

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
```diff
@@ -0,0 +1,176 @@
...
+    execute("chats/test_writes_html.txt");
+    // When `> ` is used as a prefix, the RequestLogger will colorize it assuming its a command
+    RequestLogger.printLine("> Fake command input");
+    // Log another line, this should be raw.
+    RequestLogger.printLine("Raw Line");
+    // Close the mirror
+    execute("");
+    String mirrorOutput = getMirrorLog("test_writes_html.txt");
+    assertThat(mirrorOutput, containsString("<font color=olive>> Fake command input</font><br>"));
+    assertThat(mirrorOutput, containsString("Raw Line"));
```
libraryaddict, line 65: Without checking the surrounding text, the "Raw Line" is meaningless as it could've been wrapped inside a block of html, and this would still pass the test.

This test is basically a meaningless repeat of "testMirrorDoesNotAddHTML" afaik

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
```diff
@@ -0,0 +1,176 @@
...
+  @Test
+  public void testMirrorDoesNotAddHTML() {
+    // Open the mirror
+    execute("chats/test_writes.txt");
+    // Log a raw line, this should be raw.
+    RequestLogger.printLine("Raw Line");
+    // Close the mirror
+    execute("");
+    String mirrorOutput = getMirrorLog("test_writes.txt");
+    assertThat(mirrorOutput, containsString("Raw Line"));
```
libraryaddict, line 77: This test does not actually check for the existence of HTML. Only that the string contains the text we're looking for.

Either check the whole file contents match as expected, or perhaps check for enclosing tags < and >

libraryaddict (changes requested):

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
jaadams5: Tweaked some things.  Note there are now failing tests since HTML is added to the mirror when there was none in the gCLI.  I'm thinking what we really need is to grab some ideas from RequestLoggerTest and have a test that compares the gCLI and the mirror contents.  Not sure I'm ready to implement that.

test/net/sourceforge/kolmafia/textui/command/MirrorCommandTest.java
jaadams5: Noted and tweaked.

gausie (approved):

jaadams5:
Reverted because gCLI missing line endings.

PR #984, opened 2022-08-16
Unlock Adventure Zones if necessary

Veracity0:
I've noticed for a long time that when I dance with Lady Spookyraven in the Ballroom, if I immediately jump to some other zone to start adventuring, next time I get a Lights Out adventure directing me to a third floor location, KoL gives me a "You can't get there" response.

Turns out I'd diagnosed the root cause of this a long time ago when I wanted  to add a "Adventure Again" link to that page. From RequestEditorKit.java:
```
    if (buffer.indexOf("Having a Ball in the Ballroom") != -1) {
      // Give the player a link to talk to Lady Spookyraven again (on the third floor)
      //
      // Unfortunately, place.php?whichplace=manor3&action=manor3_ladys does not work until you
      // visit the third floor map
      // link = "<p><a href=\"place.php?whichplace=manor3&action=manor3_ladys\">Talk to Lady
      // Spookyraven on the Third Floor</a>";
      //
      // Unfortunately, place.php?whichplace=manor3 doesn't work until you visit the second floor
      // map again.
      // link = "<p><a href=\"place.php?whichplace=manor3\">Go to the Third Floor</a>";
      //
      // Which makes this whole function useless, unless we
      // can figure out a way - via a sidepane command, say -
      // to unlock the third floor and then talk to Lady S on
      // the third floor
    }
```

The solution is to simply run a request immediately, just as we do for reading the MacGuffin diary or looking at the Volcano map.

It also occurred to me that KoL USED to (still does?) required that you look at the beanstalk and climb it before you could go to the PFA.

And, for some reason, there was code in the old KoLAdventure validation to go to woods.php to look to see if the Hidden Temple was there. I don't know why. Did using the map or turning the items into Prof. Fanning similarly not unlock the area for you?

I will test all of these things in my multi's new ascension, which I am about to start right now.

I have premptively coded up some "visits" for those cases and will get Debug logs for all these cases.

I am sure of the bug with the Ballroom. I will probably comment out the other two automatic "visits" and will do a visit_url() to the PFA or Hidden Temple before leaving the response page to test whether they are necessary.

1) GenericRequest has a static global "suppressUpdate" which means "while this is set, do not call api.php to update status after processing the response". Its purpose is specifically for "auto visits" done in response to a request that already will make such an update. Why do the update from the auto visit and then do the update from the triggering request?

I added a "shouldSuppressUpdate" method which defaults to the value of that static global.
I created UpdateSuppressedRequest which extends GenericRequest and overrides that method to return true.
I made RelayRequest override that method to return true.
I made UseItemRequest use that class when reading diary or volcano map. (One of those previously used a RelayRequest - "because it wouldn't update" and the other had a try/finally block to set the global static to achieve the same end.

I have tests for that functionality.

The only other place the global static is used is in RelayRequest.submitCommand. I need to study that and figure out why that sometimes suppresses update and sometimes does not.

2) I verified that opening the Hidden Temple via Professor Fanning does not require you to visit the Woods to see the Hidden Temple. In fact, the map of the Woods is part of the responseText for handing over the items he wanted.

3) I verified that looking at the second floor after dancing with Lady Spookyraven will let you adventure on the third floor.
Unfortunately, my "Talk to Lady Spookyraven on the the Third Floor" link did not appear - even though the HTML looked to be formatted correctly.

4) That's because the method to generate the link was there, but the call to it was commented out. Fix.

5) Test for obtaining & reading diary when you get it from The Shore

6) Test for reading volcano map when you the final Nemesis assassin drops it.

I did not have the following things, yet. Will get when next character ascends.

1) Open Hidden Temple by using the Spooky Temple map.

2) Verify my "talk to Lady Spookyraven" link works.
Heck, maybe add two links: One to talk to her, another to simply visit the 3rd floor.

3) Did not correctly test that planting the bean programattically successfuly unlocks the beanstalk.
I see "Climb the Beanstalk" in the map - which is part of the responseText - but I active code to go visit it.
That line is commented out now, so next ascnsion, I will do the test again.

4) We didn't follow the redirect when reading the volcano map for the first time. That time - and that time only - it redirects to inventory.php to give you a message. Subsequently, it redirects to volcanoisland.php.
I made it follow the redirect and, once I can do that with our fake HTTP client, I'll adjust the text - and hopefully have HTML to use with that.


gausie (approved):
Looks great. So we should use suppressed requests for hitting a page that we want to have been visited without performing any processing in the results?

Veracity0:
GenericRequest will look at the responseText and see if it references charpane.php.

Unless it is a request which is suppressing updates - globally, or locally (by overriding suppressUpdate() - which includes RelayRequest), it will make a call to api.php.

RelayRequests will let the browser call charpane.php, as directed by KoL.

The point of suppressing the update is if you're making an extra request as part of the result processing of another request, which is, itself, asking for a charpane (api) update. 

In other words, let the outer request do the update after processing results, rather than the inner one make an update and the the outer one also makes an update.

gausie:
Ok. We can use this if we need to check the stillsuit effect after combat -
neat.

PR #985, opened 2022-08-17
After a fight test, reset current round

libraryaddict:
Noticed this was failing a test I was creating because the test assumed it was still in a fight.


gausie:
It would be more fitting with the way we do tests to have a `withCurrentRound()` Cleanup that, with no arg, just restores the round to whatever it was before the test

PR #990, opened 2022-08-18
Fix failure to add BR

jaadams5:
Previous PR failed to notice that BR was being added to gCLI input.


jaadams5:
Most of this was in a previous PR that was reverted.

PR #993, opened 2022-08-18
Fix Concoction.compareTo

jaadams5:
https://kolmafia.us/threads/r26661-attempt-to-permanently-meet-otherwise-stochastic-coverage-changes-by-g.27892/

This fixes the violation of the compareTo contract discovered by running checkconcoctions (from the gCLI).  checkconcoctions was moved from test code to debug code because of the resources it required as a test and that decision was not revisited.


gausie:
Let me look into this before merging. They shouldn't have the same sort order because they're all different types of consumable and different consumables have different sort orders. I removed this line after finding it impossible to test because of this! I'll run checkconcoctions with the debugger on

gausie:
@jaadams5 I found the actual issue and have a testable reproduction now. Will close this in favor https://github.com/kolmafia/kolmafia/pull/995

PR #995, opened 2022-08-19
Fix comparing two Concoctions when they are == equal

gausie:


src/net/sourceforge/kolmafia/objectpool/Concoction.java
```diff
@@ -414,6 +414,10 @@ public int compareTo(final Concoction o) {
       throw new NullPointerException();
     }
 
+    if (this == o) {
```
jaadams5, line 417: Somewhere I picked up the idea that == in compareTo was not a best practice so I avoided this.  I don't think this is what we want and will try and get a test.  I think if I clone a steel item and then compare the original to the clone this compareTo will not return 0 and that is wrong.  Stand by, please.

gausie:
Reopening this until you have an alternative

src/net/sourceforge/kolmafia/objectpool/Concoction.java
gausie: Do we not just want to use `.equals()` then instead? Seems like the right move

src/net/sourceforge/kolmafia/objectpool/Concoction.java
```diff
@@ -182,6 +182,49 @@ public Concoction(final String name, final String property) {
     this.setEffectName();
   }
 
+  // Copy constructor primarily for testing.  Shallow copy.
+  public Concoction(Concoction source) {
```
gausie, line 186: Is this code you were writing to test if cloning breaks it? Did you mean to push it?

jaadams5: WIP.  Give me 15 minutes.

jaadams5: OK.  I implemented a copy constructor because that seemed more robust than making Concoction cloneable.  I wrote a new test comparing the clone and original.  compareTo failed when comparing.  They should be equal and were not.  I then put my previous fix in (but left yours there) uncommented the failing tests, saw they passed, delinted the tests and confirmed checkconcotion works.

test/net/sourceforge/kolmafia/objectpool/ConcoctionTest.java
```diff
@@ -37,6 +38,26 @@ public void steelOrgansAreSortedToTheTop() {
...
+  @Test
+  public void copyComparesAsEqual() {
+    Concoction original = ConcoctionPool.get(ItemPool.STEEL_LIVER);
+    Concoction copy = new Concoction(original);
+    // Not same object
+    assertNotSame(original, copy, "Different objects not supposed to be ==");
+    // But are equals
+    assertEquals(original, copy, "Original and copy not equals.");
+    assertEquals(copy, original, "Copy and original.");
+    // Sort the same via compareTo
```
jaadams5, line 56: The above were different tests when they were explicitly a.equals(b) and b.equals(a).  Maybe I should have ignored the suggestion to let the test framework do the compare?  Otherwise line 55 can and should be deleted.

src/net/sourceforge/kolmafia/objectpool/Concoction.java
gausie: @jaadams5 I've again removed the line below - I told you it doesn't make sense! I've replaced the == with a .equals() as I suggested many hours ago and this contrived test now passes. Once you've seen that it does, I'd like to remove this function and the associated test as they're not meaningful

gausie: Also please don't add commits to a branch someone else is working on - you can just make a branch from this branch and add them there.

jaadams5: First an apology for committing.  My model of how to collaborate on GitHub continues to evolve.

As for the rest, you are right and I figured it out.  First I had missed the fact that Concoctions overrides equals and so the comparison was looking at name and an id and not just asking "are they the same object?".  Second I did have a failing test with only your change.  It was contrived under an incorrect assumption about equals but it failed.  When I tried to reproduce it after your correct assertion that I was missing something, I failed to do so.  That was operator error, presumably due to multi-tasking :-)  In any event the next step is to back out what I did, confirm checkconcoctions and tests work and approve it.

jaadams5:
Could not figure out how to uncommit resolving conflicts.  I believe I have undone my commits.  You might want to look at resolved conversations but you may push when you are ready.

jaadams5 (approved):

PR #996, opened 2022-08-19
Refresh coat of paint after freeing the king in an avatar path

libraryaddict:
After freeing the king in an avatar path, our coat of paint modifiers will change as it is a seeded item.

To my knowledge, this may be the only item that does so.


libraryaddict:
Tests need to be rerun, looks to have failed from a non-test?

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -939,6 +939,7 @@ public static final void resetAfterAvatar() {
     RequestThread.postRequest(new CharSheetRequest());
     InventoryManager.checkPowerfulGlove();
     InventoryManager.checkDesignerSweatpants();
+    InventoryManager.checkCoatOfPaint();
```
Veracity0, line 942: Why? KoLmafia.refreshSessionData has:

```
    // Check items that vary per person
    // These won't actually generate a server hit if the item
    // has been seen at its current modifiers
    InventoryManager.checkNoHat();
    InventoryManager.checkJickSword();
    InventoryManager.checkPantogram();
    InventoryManager.checkLatte();
    InventoryManager.checkSaber();
    InventoryManager.checkCoatOfPaint();
    InventoryManager.checkUmbrella();

    // Items that conditionally grant skills
    InventoryManager.checkPowerfulGlove();
    InventoryManager.checkDesignerSweatpants();
```
KoLMafia.resetAfterAvatars has:
```
    // Clear skills first, since we no longer know Avatar skills
    KoLCharacter.resetSkills();
    RequestThread.postRequest(new CharSheetRequest());
    InventoryManager.checkPowerfulGlove();
    InventoryManager.checkDesignerSweatpants();
```
The fresh coat of paint does not offer conditional skills, does it?
It provides four randomly generated modifiers.
Do those change when you cease to be an avatar?

libraryaddict: When I exit grey you, I no longer have +5 fam weight and other enchants on my shirt.
But mafia doesn't recognize this, and until I manually visit the page, it constantly tells me that fam weight is wrong.

Just occurs to me that Voting Booth may also be effected, and operation patriot shield could also be effected.
I can't test voting booth.

Veracity0: Ok.

Veracity0 (approved):

PR #997, opened 2022-08-19
Make HolidayDatabase testable

gausie:
This branch started as an attempt to fix issues with running tests in a significant time zone difference to the USA (@libraryaddict is in New Zealand). It became clear that using Date was not going to sufficiently handle our requirements around timezones, so I've moved HolidayDatabase to use ZonedDateTime instead. This is actually really nice since we frequently use two different "timezones" in mafia - "Rollover" time and Arizona time.

Calendar and Date, while not officially deprecated, are declared as legacy - and in writing this PR I can see why! We're in a much better place now with our handling of time. I've left local reporting of dates and times (like for logging) alone at the moment, but everything else is moved over. I also used our mockable "now" getter in more places, which will make that code more testable in future.


src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
```diff
@@ -294,31 +257,29 @@ public static final void setMoonPhases(final int ronaldPhase, final int grimaceP
 
       // Adjust the new year by the appropriate number of days.
 
-      HolidayDatabase.NEWYEAR += phaseError * MS_PER_DAY;
-      HolidayDatabase.BOUNDARY += phaseError * MS_PER_DAY;
-      HolidayDatabase.COLLISION += phaseError * MS_PER_DAY;
+      HolidayDatabase.NEWYEAR.plusDays(phaseError);
```
gausie, line 260: Oh this is non mutating so I need to fix this tomorrow. Though I don't quite understand what's going on here anyway

gausie: I do understand what's going on here now (there is a very helpful comment above! It seems that my changes make the first two points handled already. So now this code only exists for handling an incorrect system clock - with that in mind, am of a mind to remove this code. I'd be interested in hearing from other @kolmafia/staff, in particular @Veracity0

gausie: I would be tempted to add a message to the console if the HolidayDatabase's calculated day number doesn't match with the daynumber reported by api.php and leave it at that. Or simply warn that you must have your system clock wrong if the moon phases are not what we expected!

midgleyc (approved):
I think it's fine.

The comment also mentions that time changes at rollover, not Arizona time, but that the code assumes Arizona. Now that we have separate timezones for rollover and Arizona could (does?) the code use rollover instead?

gausie:
> I think it's fine.
> 
> The comment also mentions that time changes at rollover, not Arizona time, but that the code assumes Arizona. Now that we have separate timezones for rollover and Arizona could (does?) the code use rollover instead?

It does already. Very little uses arizona time now - UMD, December, Monday (for gar-ish), Tuesday Ruby

PR #1001, opened 2022-08-20
feat: warn with dinodollars at Ralph

midgleyc:
Not sure whether the dinodollars go away, or the shop goes away, or the shop contents don't go away, but I think warning is a good idea if that's right.

The shop is `whichshop=dino`, though.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/RelayRequest.java
+++ b/src/net/sourceforge/kolmafia/request/RelayRequest.java
@@ -923,6 +923,31 @@ public class RelayRequest extends PasswordHashRequest {
       return true;
     }
 
+    // In Fall of the Dinosaurs, you will lose access to the Dinostaur and can therefore not spend
+    // your dinodollars
+
+    if (KoLCharacter.inDinocore()) {
+      if (InventoryManager.getCount(ItemPool.DINODOLLAR) <= 0) {
+        return false;
+      }
+
+      String warning =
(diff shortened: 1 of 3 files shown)


gausie:
The problem with this is that there's nothing really to spend them on that isn't also a quest item, unless you have hundreds

midgleyc:
You could grab some DNAde for the basement, or the shirts for a low-shiny account combat rate adjustment?

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -923,6 +923,31 @@ public boolean sendBreakPrismWarning(final String urlString) {
       return true;
     }
 
+    // In Fall of the Dinosaurs, you will lose access to the Dinostaur and can therefore not spend
+    // your dinodollars
+
+    if (KoLCharacter.inDinocore()) {
```
Veracity0, line 929: How about some tests in RelayRequestWarningsRequestTest?
RelayRequest.sendBreakPrismWarning has a warning for leaving Robocore for a similar reason.
No tests for Kingdom of Exploathing and Plumber paths, alas. for 

That whole file could used some @Nested classes - and Cleanups.

Veracity0 (changes requested):

Veracity0 (approved):

PR #1004, opened 2022-08-21
fix: mistranslated method in r26667

midgleyc:
There are bug reports that something maximizer-related has started breaking on https://github.com/kolmafia/kolmafia/commit/f9a268c62529d240a27c701cd3f27740adc5506f, so I took a closer look.

All I could find that appeared to have changed is folder-related: this method checked the currently equipped item instead of the passed-in item, and ValhallaManager checks all folders instead of the first 3 (though the latter two should be null, so it should wash out).


gausie (approved):
Didn't even notice it said FOLDER3!!

PR #1005, opened 2022-08-21
feat: more dino stuff

midgleyc:
More dino stuff: bosses (Frat route), last item, last effect

### DIFF
--- a/src/data/monsters.txt
+++ b/src/data/monsters.txt
@@ -1360,7 +1360,7 @@ glass-shelled archelon	2233	dino_turtle.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: 
 supersonic velociraptor	2234	dino_raptor.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 spikolodon	2235	dino_spike.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 ghostasaurus	2236	dino_ghost.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: undead EA: spooky Article: a
-flatusaurus	2237	dino_flat.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
+flatusaurus	2237	dino_flatus.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 primitive chicken	2238	dino_chicken.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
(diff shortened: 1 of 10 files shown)


src/data/monsters.txt
```diff
@@ -1355,13 +1355,22 @@ The Naughty Scorcheress	2220	sorcform1.gif	BOSS NOCOPY Atk: 190 Def: 171 HP: 400
 
 kachungasaur	2230	dino_kachung.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 dilophosaur	2231	dino_diplo.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
-high-altitude pterodactyl	2232	dino_ptero.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
+high-altitude pterodactyl	2232	dino_ptero.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast EA: stench Article: a
 glass-shelled archelon	2233	dino_turtle.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 supersonic velociraptor	2234	dino_raptor.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 spikolodon	2235	dino_spike.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
 ghostasaurus	2236	dino_ghost.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: undead EA: spooky Article: a
 flatusaurus	2237	dino_flat.gif	Atk: 0 Def: 0 HP: 0 Init: -10000 P: beast Article: a
```
Veracity0, line 1363: Manuel tells me image is dino_flatus.gif

src/net/sourceforge/kolmafia/request/NPCPurchaseRequest.java
```diff
@@ -910,6 +910,11 @@ public static final void parseShopResponse(final String urlString, final String
       }
     }
 
+    if (shopId.equals("dino")) {
```
Veracity0, line 913: Right. I was going to say we alphabetize shopids, but, no, we don't. :)

However, you also need to add this shopid in registerShopRequest farther down in this same file.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/NPCPurchaseRequest.java
midgleyc: okay, hope I got that right.

I checked "marioitems" above and that one wasn't further down. I'm not sure whether it should be, or how many are missing.

Veracity0 (approved):

PR #1006, opened 2022-08-21
Fix outfit swapping with empty accessory slots.

heeheehee-kolmafia:
The trigger seems to be having an accessory in acc1 and an empty slot
in some higher value, then donning an outfit that fills that empty
slot.

Note: not ready for commit as-is, as I haven't written any tests
yet. I've only confirmed that the fix works :)

### DIFF
--- a/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
+++ b/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
@@ -1826,10 +1826,11 @@ public class EquipmentRequest extends PasswordHashRequest {
     // Calculate accessory fill order
     int accessoryIndex = 0;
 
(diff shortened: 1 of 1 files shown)


midgleyc:
I think this is also the cause of the issues people have been having with maximizer and accessories: some combination of outfit changes leads `EquipmentManager.ACCESSORY_SLOTS` to not contain all the accessories (i.e. it contains duplicates). I can't replicate this, though.

midgleyc:
Ah, I see you knew about this already.

gausie:
OHHHHHHH reading your change this was a dumb thing for me to have done. Sorry everyone. Thanks for the fix.

gausie:
I want to just merge this since it's an important hot fix that will be hard to test. If tests are forthcoming later that would be fantastic

gausie (approved):

PR #1008, opened 2022-08-21
More day passes.

Veracity0:
1) Tunnel of L.O.V.E. and LT&T Telegraph Office have one-day-use items.

2) QuestManager (or PlaceRequest) will now detect permanent IOTMs in Town Wrong, Town RIght, Elemental Airport, or the Mountains.
For items which have a day-pass, it will set the "today" property if it sees them on the map and you do not have (known) "always" access.
I uploaded HTML maps with all IOTMs on them. I use those responseTexts to write tests for permanent or daily access.

3) canAdventure now looks for daily use items, not for original IOTM, when appropriate.

4) canAdventure/prepareForAdventure re willing to use more items to access zones via a status effect

- devilish folio
- tiny bottle of absinthe
- transporter transponder
- Deep Machine Tunnels snowglobe

I have tests for those.

5) withFamiliar caused test leakage: it creates a new familiar and makes it current, but when it cleans up, it just puts it back in the terrarium. Subsequent tests which assume you don't have the familiar fail.


gausie (approved):
Looks good and thanks for resolving the withFamiliar issue!

PR #1010, opened 2022-08-22
fix: avoid debug log on moons command

midgleyc:
No test for the mini-moon because it looks like it's wrong. Looks like an off-by-one error was introduced in the refactor: it tells you the position the mini-moon was in yesterday.


jaadams5 (approved):
You got yours published before mine, and your test is slightly more comprehensive.

jaadams5:
You are just a couple of minutes faster than me.  Your PR was up while I was running local tests on mine and you pushed while I was confirming no NPE from the app :-)

PR #1011, opened 2022-08-22
fix: hamburglar position

midgleyc:
Add tests for moon positions, reading off https://noblesse-oblige.org/calendar/

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
@@ -3,6 +3,7 @@ package net.sourceforge.kolmafia.persistence;
 import static net.sourceforge.kolmafia.KoLConstants.DAILY_DATETIME_FORMAT;
 
 import java.time.DayOfWeek;
+import java.time.Duration;
 import java.time.Month;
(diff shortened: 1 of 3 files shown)


Veracity0 (approved):

jaadams5 (approved):

src/net/sourceforge/kolmafia/persistence/HolidayDatabase.java
```diff
@@ -278,7 +279,7 @@ public static int getHamburglarPosition(final ZonedDateTime dateTime) {
       return -1;
     }
 
-    var days = ChronoUnit.DAYS.between(dateTime, HolidayDatabase.COLLISION);
+    var days = Duration.between(HolidayDatabase.COLLISION, dateTime).toDays();
```
gausie, line 282: @midgleyc what was the difference here?

midgleyc: I read a stackoverflow post saying that ChronoUnit.DAYS.between is calendar days, but Duration.between is 24-hour days. We want the latter.

However that didn't matter for this case, what mattered is that the earlier date should come first, but it didn't here.

PR #1012, opened 2022-08-22
Give Ed stages the same monsterId when creating a MonsterData

Veracity0:
1) In monsters.txt, bas Ed the Undying is monsterId = 473, but Ed the Undying (1) - Ed the Undying (7) all have monsterId = 0. This is because we can't have a map keyed by monsterId with 8 different key = 473.

However, when processing a fight, when we look at MONSTERID of 473, but we've disambiguated the monster by image to one of the (1) - (7) versions with monsterId = 0, that doesn't mtch, so we look up the monster by ID using the aforementioned map.

When creating the MonsterData associated with the (1) - (7) names, we now set the id to 473. FIghtRequest is happy.

2) I noticed that consequences.txt tried to figure out the Ultrarare pooltergeist by image, rather than use monsterId = -94. Since we know the actual monsterId  - 425 - no need to do that any more.

### DIFF
--- a/src/data/consequences.txt
+++ b/src/data/consequences.txt
@@ -124,9 +124,6 @@ DESC_EFFECT	Wine-Friendly	\+(\d+) .*? Damage	vintnerWineLevel=[$1/3]
 
 # Monster disambiguation:
 
-# Monster ID -94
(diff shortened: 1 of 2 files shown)


gausie (approved):
Congratulations on the ultrarare!

jaadams5 (approved):

PR #1016, opened 2022-08-23
feat: more dino stuff from Yendor

midgleyc:
Source: https://kolmafia.us/threads/fall-2022-challenge-path-fall-of-the-dinosaurs.27887/page-2#post-169386


Veracity0 (approved):
Oh, ha ha. I also have all of this in my PR which deals with dinosaurs having multiple ways to gobble down their prey.

You can have this. I will merge with you. :)

PR #1018, opened 2022-08-24
feat: mark glitch item as edible

midgleyc:
Mark glitch item as edible, so "eat glitch season" works.

Unfortunately, this means that "use glitch season" does not work, even though the reward is marked as reusable. I don't know how to fix this, but would appreciate suggestions.


libraryaddict:
First thing that pops to mind is adding `usable`, which to my mind at least; Should allow it to be `use`'d

horrible-little-slime:
Would `eat` as a CLI command and an ash function need extra support to handle the incoming combat?

midgleyc:
I don't think so, because it already works for black pudding (I'm pretty sure).

midgleyc:
Gausie has fixed the issue. Thanks!

gausie:
@midgleyc I think it's fair for you to review this since I've fiddled with it so much

test/net/sourceforge/kolmafia/textui/command/EatCommandTest.java
```diff
@@ -0,0 +1,48 @@
...
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class EatCommandTest extends AbstractCommandTestBase {
+  @BeforeEach
+  void beforeEach() {
```
midgleyc, line 21: There's only one test so doesn't matter at the moment, but I think this should be able to be `BeforeAll` (with cleanups).

test/net/sourceforge/kolmafia/textui/command/UseCommandTest.java
```diff
@@ -0,0 +1,48 @@
...
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class UseCommandTest extends AbstractCommandTestBase {
+  @BeforeEach
+  void beforeEach() {
```
midgleyc, line 21: same `BeforeAll` comment here

midgleyc (commented):
I can't approve or request changes on my own PR, but here are some changes.

Approve and merge when done ;)

gausie (approved):

PR #1019, opened 2022-08-24
More stillsuit distillate support

gausie:
* Allows `drink stillsuit distillate` (tested)
* Shows stillsuit distillate in the Booze panel with accurate figures for adventure gain and effect turns (tested)
* Adds support for reading the number of drams you have if below 10 (tested)
* Tested for letting you distill when your stillsuit is in the terrarium
* Adds a `stillsuitFamiliar` pref, which will automatically throw your stillsuit on the specified familiar when you are putting away a familiar wearing the suit (tested). If we want I suppose we could respect this in ValhallaManager.postAscension too, that can come later.


gausie:
Ok I'm just going to merge this one

PR #1020, opened 2022-08-24
Track Powerful Glove's charge from description

libraryaddict:
Noticed the charge wasn't being tracked when an issue popped up with "Use the Force" being used in a macro which changed the final output of the resulting page to being a choice, meaning mafia never saw the combat page, meaning mafia never realized I used a powerful glove charge in fight to switch the monster.

Or basically.
If a charge is used without mafia knowing, mafia will never adjust for it. Feels strange this wasn't tracked previously.

Tracking when it is fully charged feels meaningless as it should only be inaccurate if someone was messing with the preference, but included for the full coverage.


Veracity0:
You could make this even better.

Consider this:

When you enter a fight, unless your macro is executed as an autoattack, you will be shown an initial fight page, complete with the "skills" dropdown.

adventure.php?snarfblat=149
302 Field: location = [fight.php?ireallymeanit=1661372296]
fight.php?ireallymeanit=1661372296

```
<option value="4012" picurl="saucegeyser" >Saucegeyser (24 Mojo Points)</option>
<option value="7327" picurl="shrinkenemy" >CHEAT CODE: Shrink Enemy (5 of today's remaining 100%)</option>
<option value="7326" picurl="replaceenemy" >CHEAT CODE: Replace Enemy (10 of today's remaining 100%)</option>
```

FightRequest.parseAvailableCombatSkills(String responseText):

```
    Matcher m =
        FightRequest.AVAILABLE_COMBATSKILL_PATTERN.matcher(
            responseText.substring(startIndex, endIndex));
    while (m.find()) {
      int skillId = StringUtilities.parseInt(m.group(1));
      String skillName = SkillDatabase.getSkillName(skillId);
      if (skillName == null) {
        skillName = m.group(2);
        SkillDatabase.registerSkill(skillId, skillName);
      }
```

The pattern:

```
  private static final Pattern AVAILABLE_COMBATSKILL_PATTERN =
      Pattern.compile("<option[^>]*?value=\"(\\d+)[^>]*?>(.*?) \\((\\d+)[^<]*</option>");
```

notice that:

group(1) -> the skillId
group(2) -> the skillName
group(3) -> the number after the "(" - which is usually the MP cost.
That last group is never used.

I think you could capture the entire phrase inside the () and determine right from there the charge on your powerful glove.

Here are all the ones on MY skill dropdown that are not simply (XX Mojo Points):

```
<option value="188" picurl="light" >Implode Universe (13 times today, 50 MP)</option>
<option value="19" picurl="snout" >Transcendent Olfaction (3 uses left today)</option>
<option value="7327" picurl="shrinkenemy" >CHEAT CODE: Shrink Enemy (5 of today's remaining 100%)</option>
<option value="7326" picurl="replaceenemy" >CHEAT CODE: Replace Enemy (10 of today's remaining 100%)</option>
<option value="7310" picurl="bat" >Become a Bat (10 time(s) remaining today))</option>
<option value="7309" picurl="puff" >Become a Cloud of Mist (10 time(s) remaining today))</option>
<option value="7308" picurl="wolfmask" >Become a Wolf (10 time(s) remaining today))</option>
```

libraryaddict:
Don't really plan to add more skill parsing to this PR, added cosplay only cos it occured to me. Probably should just share the same html file tbh

libraryaddict:
Although it occurs to me that if we were going to track all this stuff, could add it into Consequences as something like `COMBAT_SKILL` given how many different formats there are for counters in skill names.

Veracity0:
Interesting. Notice FightRequest.parseAvailableCombatSkills:

```
    int startIndex = responseText.indexOf("<select name=whichskill>");
    if (startIndex == -1) {
      return;
    }
    int endIndex = responseText.indexOf("</select>", startIndex);
    if (endIndex == -1) {
      return;
    }

    Matcher m =
        FightRequest.AVAILABLE_COMBATSKILL_PATTERN.matcher(
            responseText.substring(startIndex, endIndex));
```

Notice that ``` responseText.substring(startIndex, endIndex))``` is just the Dropdown with the skills.

If we have ConsequenceManager do COMBAT_SKILLS, it is exactly this spot in FightRequest that would call it, using exactly that substring of the responseText.

Veracity0 (approved):
Nicely done!

Veracity0:
> Noticed the charge wasn't being tracked when an issue popped up with "Use the Force" being used in a macro which changed the final output of the resulting page to being a choice, meaning mafia never saw the combat page, meaning mafia never realized I used a powerful glove charge in fight to switch the monster.

I'm curious about this. Did KoL actually show the output if the previous rounds, but because Use The Force redirected to a choice, we didn't process that output?

What did the session log say?

I can try this out myself with debug log on and look at the requests submitted and the responses - and redirects - received. That might very well be a bug we need to deal with in another way.

But fixing that does not invalidate anything about this PR.

Should we open a new Bug to discuss it?
And how about receiving a redirect after ascending? I have thoughts about that, too. Another Bug?

Having conversations in the comments if approved/merged PRs doesn't seem the right way to do that. 😀

libraryaddict:
> > Noticed the charge wasn't being tracked when an issue popped up with "Use the Force" being used in a macro which changed the final output of the resulting page to being a choice, meaning mafia never saw the combat page, meaning mafia never realized I used a powerful glove charge in fight to switch the monster.
> 
> I'm curious about this. Did KoL actually show the output of the previous rounds, but because Use The Force redirected to a choice, we didn't process that output?
> 
> What did the session log say?
> 
> I can try this out myself with debug log on and look at the requests submitted and the responses - and redirects - received. That might very well be a bug we need to deal with in another way.
> 
> But fixing that does not invalidate anything about this PR.
> 
> Should we open a new Bug to discuss it? And how about receiving a redirect after ascending? I have thoughts about that, too. Another Bug?
> 
> Having conversations in the comments of approved/merged PRs doesn't seem the right way proceed. 😀

The TL;DR

I executed a macro with replace then use the force. I had debug logs turned on for that test, no combat logs were returned. Just a redirect to saber. I'm not sure there's much you can do.

PR #1022, opened 2022-08-25
Track uses of a skill by their combat skill name

libraryaddict:
The main dubious part of this PR is the test file.

It's been copied from other test files and merged into a single file which technically, isn't possible as I'm in an avatar path.
This was done by simply copying the combat skill option and pasting it inside.

That said, it should be a non-issue.

Note that this does not modify `Skill.timescast`


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4603,6 +4606,15 @@ public static final void parseAvailableCombatSkills(String responseText) {
           Preferences.setInteger("gladiatorBallMovesKnown", skillId - 7090);
         }
       }
+      if (skillId == SkillPool.BACK_UP) {
+        Matcher matcher = FightRequest.BACKUP_CAMERA_USES.matcher(m.group(2));
+
+        if (matcher.find()) {
+          int limit = KoLCharacter.inRobocore() ? 16 : 11;
+
+          Preferences.setInteger("_backUpUses", limit - StringUtilities.parseInt(matcher.group(1)));
```
gausie, line 4615: Could you not achieve this with:

```
COMBAT_SKILL	Back-Up to your Last Enemy	Back-Up to your Last Enemy \((\d+) uses? today\)	_backUpUses=[(11+path(You Robot)*5)-$1]
```

If not, we just need to make sure `path` is available to that sort of Expression

gausie:
This is awesome! I reckon you can update `timescast` pretty easily here, by stepping in when `COMBAT_SKILL` is being parsed by the consequences manager

src/net/sourceforge/kolmafia/request/FightRequest.java
libraryaddict: Thanks for the suggestion!

libraryaddict:
> This is awesome! I reckon you can update `timescast` pretty easily here, by stepping in when `COMBAT_SKILL` is being parsed by the consequences manager

Just checked, and I think that would go against the implementation.

Times Cast looks to be reset on session, while this change would definitely violate that more than it'd fix anything.

gausie (approved):

PR #1027, opened 2022-08-26
Start of preValidateAdventure

Veracity0:
1) Since the portable spacegate (open) is its own zone, if we see the Spacegate in the mountains, we have permanent access.

2) Seeing the Barrel Shrine in the Dungeoneer's Association is permanent access

3) Seeing the Getaway Campsite in the woods is permanent access.

4) Adventure Validation part 0 - preValidateAdventure.
For things which have an expensive day pass (which we will not check/use in canAdventure/prepareForAdventure), if we don't have permanent access or know that we have daily access, take a peek at the place where the zone would appear, thereby setting daily access if appropriate.
- Spring Break Beach
- Conspiracy Island
- Dinsey Land Fill
- That 70's Volcano
- The Glaciest
- Gingerbread City
- Neverending Party
- Tunnel of L.O.V.E.
- The Spacegate

5) Unit tests for all of the above.

6) ASH function: boolean pre_validate_adventure(location)

7) My Sauceror got the Color Wheel of Yuck trophy and discovered that the vial of brown slime is multiusable.


src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -454,6 +481,143 @@ private void validate1() {
...
+      // Unlimited adventuring if available
+      if (Preferences.getBoolean("sleazeAirportAlways")) {
+        return true;
+      }
+      if (!Preferences.getBoolean("_sleazeAirportToday")) {
+        // Take a look at the airport.
+        var request = new PlaceRequest("airport");
+        RequestThread.postRequest(request);
+      }
+      return Preferences.getBoolean("_sleazeAirportToday");
```
midgleyc, line 498: All the airport versions are mostly the same, and could be refactored into `boolean checkAirportZone(String element)` to make the code read more easily.

You could go further and refactor everything but spacegate (and love tunnel, with its different comment) into `boolean checkZone(String alwaysPref, String todayPref, String place)`. I'm not sure whether this is better overall.

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -71,6 +73,213 @@ public static void afterAll() {
...
+      KoLAdventure adventure = AdventureDatabase.getAdventureByName(adventureName);
+      var html = html("request/test_visit_town_wrong.html");
+      // If we have always access, we don't have today access
+      checkDayPasses(adventure, "town_wrong", html, true, false, always, today);
+      // If we don't have always access, we might today access
+      checkDayPasses(adventure, "town_wrong", html, false, true, always, today);
+      // If we don't have always or today access, we might still have today access
+      checkDayPasses(adventure, "town_wrong", html, false, false, always, today);
+      // If we don't have always or today access, we might really not have today access
+      checkDayPasses(adventure, "town_wrong", null, false, false, always, today);
```
midgleyc, line 138: This block of code is the same bar strings and could be refactored into `void checkPassCombinations(String place)`.

midgleyc (changes requested):
Two small function extractions I think are beneficial.

src/net/sourceforge/kolmafia/KoLAdventure.java
Veracity0: checkZone works well and is clean.

Veracity0: Done.

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
Veracity0: Done.

midgleyc (approved):

PR #1028, opened 2022-08-26
Strings with html removed, should be trimmed

libraryaddict:
When a string such as

```
<table border="2" cols="4">
      <tr>
        <td>
          <p>
            roninStoragePulls
          </p>
        </td>
      </tr>
    </table>
```

has the html removed, you're left with a bunch of spaces surrounding the string itself.
You will get this, note the spaces around the string.

```
                                    copyAsHTML                            
```

This could also be seen as another problem where the html shouldn't be formatted, but that's outside the scope of this small change.


gausie:
I know its a small change but... you know what I'm gonna ask for

midgleyc (approved):

PR #1030, opened 2022-08-28
fix: don't suggest horsery if not allowed in standard

midgleyc:
actual reality goggles and fifteen minutes of flame were listed, despite the comment that they weren't.

Mocking `isAllowedInStandard` makes that test take 11 seconds, which is not good. Might refactor the class to make it more easily configurable for tests.

It looks like `ClearSharedStateBefore` runs before all nested classes, which breaks setting properties. Will look into fixing outside this PR.


Veracity0 (approved):
I will be curious to see the alternative to static mocking the method, if you do happen to do that. Looks like the maximizer calls that method a lot.

midgleyc:
The options I see are:
* make a fake `standard.php` response
* expose the maps and modify them
* entirely fake maps created by a factory, like `HttpUtilities.setClientBuilder`

Of those, just exposing the maps for modification is simplest.

PR #1032, opened 2022-08-28
Adds $path

gausie:
This purposely does not add `my_path()` (and remove `my_path_name()` and `my_path_id()`) just yet as I'd like to warn for that breaking change on the forums. However, those will be easy to do once this is merged.


midgleyc:
How does this handle variables called "path"?

gausie:
Oh of course. I was so careful not to mess with those functions and forgot this will need warning ahead of time anyway. Well at least that means I can do those other deprecations in this pass!

Veracity0:
> How does this handle variables called "path"?

It appears that he did not add "path" to Parser.reservedWords.

```
    // Data types
    reservedWords.add("void");
    reservedWords.add("boolean");
    reservedWords.add("int");
    reservedWords.add("float");
    reservedWords.add("string");
    reservedWords.add("buffer");
    reservedWords.add("matcher");
    reservedWords.add("aggregate");

    reservedWords.add("item");
    reservedWords.add("location");
    reservedWords.add("class");
    reservedWords.add("stat");
    reservedWords.add("skill");
    reservedWords.add("effect");
    reservedWords.add("familiar");
    reservedWords.add("slot");
    reservedWords.add("monster");
    reservedWords.add("element");
    reservedWords.add("coinmaster");
```

Looking at DataTypes, it appears that other data types have been added without doing that.

```
  public static final int TYPE_PHYLUM = 111;
  public static final int TYPE_THRALL = 112;
  public static final int TYPE_BOUNTY = 113;
  public static final int TYPE_SERVANT = 114;
  public static final int TYPE_VYKEA = 115;
```

That means you can say:

```
phylum phylum = $phylum[beast];
```
but not:
```
element element = $element[hot]
```

gausie:
> > How does this handle variables called "path"?
> 
> It appears that he did not add "path" to Parser.reservedWords.
> 
> ```
>     // Data types
>     reservedWords.add("void");
>     reservedWords.add("boolean");
>     reservedWords.add("int");
>     reservedWords.add("float");
>     reservedWords.add("string");
>     reservedWords.add("buffer");
>     reservedWords.add("matcher");
>     reservedWords.add("aggregate");
> 
>     reservedWords.add("item");
>     reservedWords.add("location");
>     reservedWords.add("class");
>     reservedWords.add("stat");
>     reservedWords.add("skill");
>     reservedWords.add("effect");
>     reservedWords.add("familiar");
>     reservedWords.add("slot");
>     reservedWords.add("monster");
>     reservedWords.add("element");
>     reservedWords.add("coinmaster");
> ```
> 
> Looking at DataTypes, it appears that other data types have been added without doing that.
> 
> ```
>   public static final int TYPE_PHYLUM = 111;
>   public static final int TYPE_THRALL = 112;
>   public static final int TYPE_BOUNTY = 113;
>   public static final int TYPE_SERVANT = 114;
>   public static final int TYPE_VYKEA = 115;
> ```
> 
> That means you can say:
> 
> ```
> phylum phylum = $phylum[beast];
> ```
> 
> but not:
> 
> ```
> element element = $element[hot]
> ```

What do we think about this? Do we want to add those to reserved words or challenge why the others are in that list? Or just keep on keepin' on?

midgleyc:
Adding it to the reserved words will break existing scripts (e.g. philter has a variable called 'path').

If it's not necessary for this to work, what would be the benefit?

gausie:
No I agree! But it encourages the question: should we unreserve our other
enumerated types?

gausie:
Ok I originally said I would not make this PR do any breaking RTL changes just yet but I've decided it might as well all land together when it does. So, this PR has:

### Added

- `int to_int(path)`
- `path to_path(string)`
- `path to_path(int)`
- `boolean [string] voting_booth_initiatives(class, path, int)`

### Removed

- `int path_name_to_id(string)`
- `string path_id_to_name(int)`
- `int my_path_id()`

### Changed

- `string my_path()` -> `path my_path()`

What sort of timeframe do people think is appropriate to do this? I'll make a second PR that adds a deprecation warning to the removed and changed functions along with the date of deprecation once we've decided

Veracity0:
> No I agree! But it encourages the question: should we unreserve our other enumerated types?

```
phylum bounty = $phylum[beast];
bounty phylum = $bounty[triffid bark];

bounty[phylum] phylumTobountyMap;
phylum[bounty] bountyToPhylumMap;

phylumToBountyMap[bounty] = phylum;
bountyToPhylumMap[phylum] = bounty;

print("phylumToBountyMap[" + bounty + "] = " + phylumToBountyMap[bounty]);
print("bountyToPhylumMap[" + phylum + "] = " + bountyToPhylumMap[phylum]);

int servant = 5;
servant thrall = $servant[Belly-Dancer];
print(++servant);
```

yields 

```
> bogo

phylumToBountyMap[beast] = triffid bark
bountyToPhylumMap[triffid bark] = beast
6
```

You can certainly obfusticate your code without reserved words, but ASH seems to be able to handle it.
I don't think it's our job to prevent scripters from writing confusing code.

We COULD un-reserve some type names...

src/net/sourceforge/kolmafia/textui/Parser.java
```diff
@@ -1062,6 +1063,13 @@ private Evaluable autoCoerceValue(final Type ltype, final Evaluable rhs, final B
       }
     }
 
+    if (ltype.equals(PATH_TYPE)) {
```
gausie, line 1066: @Veracity0 as the original author of the coercion code what do you think of this? Why do we only do this for typedefs and not for our own types? Should I extend this to include all of our enumerated types?

midgleyc:
A couple of currently non-coercing cases (both based on things seen in the wild):

> ($strings[Trendy] contains my_path())

Gives "Cannot apply operator contains to aggregate boolean [string] (boolean [string]) and my_path() (path)"

> boolean test(string path) { switch (my_path()) { case "Trendy":  return true; break; } return false; } test($path[Trendy])

Gives "Switch conditional has type path but label expression has type string"

I think in general the == case will cover 90% of what's done with `my_path()`. I like the solution.

midgleyc (dismissed):
I think this can go in as-is if the removed functions are added back. The coercion seems to work well.

I don't think we should unreserve the existing reserved words. Not being able to call a variable something is not a big detraction, and there is a gain in e.g. syntax highlighting simplicity.

gausie:
> I don't think we should unreserve the existing reserved words. Not being able to call a variable something is not a big detraction, and there is a gain in e.g. syntax highlighting simplicity.

I don't really write scripts in ASH anymore so my opinion probably doesn't matter all too much now but when I did I would have loved to use the name item instead of `it` all the time. But yeah looking at the most popular ash syntax highlighter it does just regex for the constant names to mark types.

midgleyc (dismissed):

midgleyc (approved):

gausie:
I'm wanting to merge this but I think I'll give it a bit more time.

PR #1033, opened 2022-08-28
refactor: standard request

midgleyc:
Refactor StandardRequest (and TrendyRequest) to use enum values over Strings.

As mentioned in #1030, don't static mock StandardRequest.


gausie (approved):
LGTM! I love enums and faster tests

PR #1036, opened 2022-08-29
Create deprecation warnings, add them for path functions with example date

gausie:
We can change this date but we can now give users advanced warning of deprecation of functions (either for removal or signature/behaviour change)


midgleyc:
Removing old functions is going to break a lot of old scripts for seemingly no benefit (or at least lead to a lot of churn). Can we please not do that?

Even the type change is going to break scripts. Can it be done in a way such that `my_path() == "Bees Hate You"` continues to work?

Also, there is no way for script authors to fix their scripts ahead of time, even knowing about the coming deprecations?

midgleyc:
If the type of `my_path` has to change, I would rather this be opt-in. If it stays a string, existing scripts will continue to work, and people who want to use the `path` type can do `my_path().to_path()`.

Not solutionizing on how opt-in could work, but something like JavaScript's `'use strict'` or Python's `from __future__ import` (or Rust's editions) were my initial thoughts.

gausie:
> If the type of `my_path` has to change, I would rather this be opt-in. If it stays a string, existing scripts will continue to work, and people who want to use the `path` type can do `my_path().to_path()`.
> 
> Not solutionizing on how opt-in could work, but something like JavaScript's `'use strict'` or Python's `from __future__ import` (or Rust's editions) were my initial thoughts.

In general (and in theory) I am happy to break scripts with changes to ASH, as long as they are well communicated and well designed changes. I think people who script with mafia should have a good and consistent runtime library that nudges you into writing good code. I do agree that it would be nice to have some sort of opt-in syntax. Though I'm not sure how we maintain that sanely on our side. I suppose in many ways we want to version ash separately from the rest of mafia. I think it's a discussion for another time.

With regards to this specific change, I'd happily compromise on keeping the old functions that are now just redundant (though they were always marked as experimental with a mind to move to a path type at some point, not sure if you had seen that at all).

I am very strongly pro `my_path()` changing to return a `path` type, though I agree it should coerce to a string where appropriate (and I think it will do that in my implementation).

Maybe I should go through scripts posted on the forums and drop the author a note that they'll need to make a change? Though that will surface naturally with these deprecation notices.

Maybe we mark the experimental functions with an open-ended deprecation notice but don't remove them any time soon.

Veracity0:
I'd like to see this:

path my_path()
int to_int(path)
string to_string(path)
a "path" auto-coerces to int or string if used in a context that expects an int or a string.

I (mostly) added the ability for typedefs and records to auto-coerce a while ago: if the user provided a to_string method, we'd use it. This is from BeachComber.ash

```
typedef string [int] beach_layout;
...
beach_layout to_beach_layout( string squares )
...
string to_string( beach_layout layout )
...
string get_layout()
{
    return get_property( "_beachLayout" );
}

string set_squares( string squares )
{
    return set_property( "_beachLayout", squares );
}
...

beach_layout get_beach_layout()
{
    return get_layout().to_beach_layout();
}

void set_beach_layout( beach_layout layout )
{
    set_property( "_beachLayout", layout );
}
```
My original hope was that I could just use ```return get_layout();``` to look up ```to_beach_layout(string)``` and auto-coerce - just like it looks up ```to_string(beach_layout)``` and auto-coerces for set_property - but that proved to have issues.

But, why not auto-coerce path to int or string, as needed? get_path() returns a path, existing scripts that expect a string continue to work and don't even need a warning.

Then, "friebdly warnings" for deprecated functions?

my_path_id is deprecated. Changing 'my_path_id()' to 'my_path.id' will remove this warning.
path_name_to_id is deprecated. Changing 'path_name_to_id(XXX)' to 'to_path(XXX).id will remove this warning.
path_id_to_name is deprecated. Changing 'path_id_to_name(XXX)' to 'to_path(XXX).name will remove this warning.

midgleyc:
> In general (and in theory) I am happy to break scripts with changes to ASH, as long as they are well communicated and well designed changes.

Broken scripts affect two groups of people: authors and users. Authors can update their scripts. Users may be using scripts written years (decades?) ago, by people who have since left the game, who will never update their scripts.

We introduce warning messages that affect these users a fair amount, but outright breaking things to improve the runtime library for current scripters would be new.

gausie:
> > In general (and in theory) I am happy to break scripts with changes to ASH, as long as they are well communicated and well designed changes.
> 
> Broken scripts affect two groups of people: authors and users. Authors can update their scripts. Users may be using scripts written years (decades?) ago, by people who have since left the game, who will never update their scripts.
> 
> We introduce warning messages that affect these users a fair amount, but outright breaking things to improve the runtime library for current scripters would be new.

We've renamed prefs a few times which introduce script issues in far less obvious ways than this one. I'm very keen to change the signature of `my_path()` and rely on coercion to handle most (all?) of the issues. I'll write up some tests to make sure that it works. Let me know if you think of any weird cases that I might miss.

gausie:
Oh we've been having these conversations on the wrong PR

midgleyc:
> We've renamed prefs a few times which introduce script issues in far less obvious ways than this one.

Yes but I think there is a big difference between "some scripts may break in non-obvious ways" and "almost all scripts will definitely break".

But I think coercion should cover most breaking cases very well. I've found a couple of failing cases, which I'll post on the other PR.

gausie:
I'm merging this into #1032 since the messages are now reliant on $path being available. Discussion can continue there!

PR #1037, opened 2022-08-30
Simplify operator comparsions when coercing values

gausie:
This seemed like an unnecessary piece of duplicated code, so here we go.


midgleyc:
Some behaviour has changed, though I can't say I understand how.

When running:
```
ash ($monster[spooky vampire] > $monster[dodecapede])
ash ($monster[apathetic lizardman] > $monster[dodecapede])
ash ($monster[apathetic lizardman] > $monster[gnollish plungermaster])
```

Before this change, the result was `true`, `false`, `true`. After this change, it's `false`, `true`, `false`: the exact opposite. I think it's comparing monster ids both times, but I don't know why they're inverted the first time.

While this is a stupid command that nobody should ever run, this is enough to indicate to me that this isn't purely a refactoring (i.e. it contains an accidental behaviour change), and also that we probably want more tests around this area 😉.

gausie:
Happy to put more tests in and thank you for finding this. I just saw that
the code looked silly and wanted to change it.

gausie:
> vampire

Right 

> Some behaviour has changed, though I can't say I understand how.
> 
> When running:
> 
> ```
> ash ($monster[spooky vampire] > $monster[dodecapede])
> ash ($monster[apathetic lizardman] > $monster[dodecapede])
> ash ($monster[apathetic lizardman] > $monster[gnollish plungermaster])
> ```
> 
> Before this change, the result was `true`, `false`, `true`. After this change, it's `false`, `true`, `false`: the exact opposite. I think it's comparing monster ids both times, but I don't know why they're inverted the first time.
> 
> While this is a stupid command that nobody should ever run, this is enough to indicate to me that this isn't purely a refactoring (i.e. it contains an accidental behaviour change), and also that we probably want more tests around this area 😉.

This was just implemented incorrectly beforehand. The fact that noone ever noticed that greater than is actually less than for monsters suggests to me that noone ever noticed this issue. So I'm inclined to keep it how it is now.

jaadams5:
> This was just implemented incorrectly beforehand. The fact that noone ever noticed that greater than is actually less than for monsters suggests to me that noone ever noticed this issue. So I'm inclined to keep it how it is now.

In this case if it was previously wrong then I would advocate for fixing it.  The fact that no one noticed doesn't really seem important and suggests that fixing it would not be especially disruptive.

If you choose not to fix it perhaps a comment in the code and/or a test with comments to indicate that the incorrect behaviour is deliberate would be appropriate.

midgleyc:
I think it's fine, and even good, to fix; I just want to know why it worked the way it did before.

I also think we should have some more tests around operator comparisons and operator comparison coercion.

gausie:
@midgleyc This is now fully tested. I think we now have known behavior here and the best way forward is to merge and see if anyone's scripts get broken somehow. This part of the code was quite brittle as I can see (for example) that Vykea companions would NPE if compared to a different data type.

src/net/sourceforge/kolmafia/textui/parsetree/Value.java
```diff
@@ -271,6 +271,20 @@ public int compareToIgnoreCase(final Value o) {
     return this.compareTo(o, true);
   }
 
+  private static final List<Type> COMPARE_BY_LONG =
```
midgleyc, line 274: you're only checking containment, so Set is better (assuming types are hashable).

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
```diff
@@ -0,0 +1,341 @@
...
+import net.sourceforge.kolmafia.persistence.AdventureDatabase;
+import net.sourceforge.kolmafia.persistence.MonsterDatabase;
+import net.sourceforge.kolmafia.persistence.MonsterDatabase.Element;
+import net.sourceforge.kolmafia.request.DiscoGiftCoRequest;
+import net.sourceforge.kolmafia.request.NinjaStoreRequest;
+import net.sourceforge.kolmafia.textui.DataTypes;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class ValueTest {
```
midgleyc, line 21: Looking at the test in general:

When running comparisons, there are three options: x less than y, x equal to y, x greater than y. Respectively, this corresponds to a result of -1, 0 or 1. In `comparePaths`, you have `greaterThanOrEqualTo(0)` which means that you think the comparison is either equal or greater. I would hope instead you knew enough to know which one.

Somewhat similar: for `greaterThanOrEqualTo(1)`, is there any chance it is not exactly 1? Could these be "is(-1)", "is(1)", and maybe even refactored out: "isBefore()", "isAfter()"?

In general, I'd prefer the test describes what's going on: e.g. "known monsters compare by id / unknown monsters compare by string". For example, "compare effect to other effect": the assertion written presumably covers the current behaviour, but I think it should be recorded what the current behaviour is (presumably id?), to check that's being asserted against instead of the assertion just asserting that the current behaviour is right.

Great coverage though!

midgleyc (changes requested):

midgleyc:
We can merge and take improving it forward (taking other people's scripts as canaries), but I would still like to fix up the tests / docs.

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
gausie: > When running comparisons, there are three options: x less than y, x equal to y, x greater than y. Respectively, this corresponds to a result of -1, 0 or 1. In `comparePaths`, you have `greaterThanOrEqualTo(0)` which means that you think the comparison is either equal or greater. I would hope instead you knew enough to know which one.
> 
> Somewhat similar: for `greaterThanOrEqualTo(1)`, is there any chance it is not exactly 1? Could these be "is(-1)", "is(1)", and maybe even refactored out: "isBefore()", "isAfter()"?

It's not always -1,0 or 1, it can be any magnitude. It's the direction that matters. I agree a sign check would be preferable, but I couldn't find one and thought this was fine

jaadams5: Math.signum() returns -1, 0 or 1.

Sounds like wrapping things with signum would work in this case.

PR #1038, opened 2022-08-30
No longer Failing Test to illustrate a concern with acquire

jaadams5:
I had a script fail because mafia thought "1 7-ball" was a 17-ball.

`boolean ok = cli_execute("acquire 1 7-ball");`

is a line in an ash script that also illustrates the problem.

I added a test that I think captures the problem and fails.

Feel free to commit to this branch if you agree this is an error and have a fix before I do and please comment if you think this is operator error.


Veracity0:
Well, you COULD stayed in ASH and done things the ASH way via:

```boolean ok = retrieve_item(1, $item[7-ball]);```

Without depending on how the CLI parses item-names-with-optional counts.

That parsing is in ItemFinder, by the way. I will be interested in seeing what you find out is the root cause of your observed behavior.

**Comment by jaadams5** [comment]
Failing test now works as do a couple of new tests.

I am not proud of the working code but it was the easiest way to isolate what I was changing without having to deal with if else nesting.  It is definitely inefficient in terms of doing the same thing twice.  We can leave it that way and plan on a refactor task or tackle things now.  The IntelliJ analyze command suggests a lot of lint in ItemFinder but I will do that in another PR rather than add it to this one and obscure the changes.

ItemFinderTest also has lint that will be addressed later.  If assertEquals fails, the error message generated calls the first parameter what was expected and the second what was received.  For equals that is a bad message since order doesn't matter but if we don't have a convention I would propose assertEquals(a,b) where a is something that is fixed and was known when the test was written and b is something that was calculated as part of the test activity.  Regardless as implemented at least one assertEquals doesn't follow the same convention as the others.

File: test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java [hunk 1]
```diff
@@ -900,4 +903,35 @@ public void itShouldParseMultipleItems() {
     assertEquals(results[0].getCount(), 1);
     StaticEntity.setContinuationState(MafiaState.CONTINUE);
   }
+
+  @Test
+  public void itShouldFindSomeEdgeCasesSatisfactorily() {
```
gausie, line 908: Can you split this into multiple tests?

jaadams5: Sure.  Since I could not completely figure out the logic for grouping the existing tests together I just bundled mine all together :-)

jaadams5: Better?  You may resolve if so.

gausie:
Other than the test parameterization, looking good

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -900,4 +905,50 @@ public void itShouldParseMultipleItems() {
...
+    assertTrue(item != null);
+    assertEquals(ItemPool.SEVEN_BALL, item.getItemId());
+    assertEquals(item.getCount(), 1);
+  }
+
+  @Test
+  public void itShouldHandleCountForItemBeginningWithDigitAndContainingSpace() {
+    AdventureResult item;
+    // Meat, 1 Meat, 1 1 Meat, 123 1 Meat
+    item = ItemFinder.getFirstMatchingItem("Meat", false, null, Match.ANY);
```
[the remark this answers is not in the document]
jaadams5: OK.  I'll try.  Stand by.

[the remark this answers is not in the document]
jaadams5: How's this?  Resolve if OK.

gausie (approved):
Nice!

PR #1039, opened 2022-08-31
Add parameter to print_html to log to sessions log

libraryaddict:
My main concern about this PR is that I can't actually test multiline support properly.

The way the runtime functions are executed currently results in a malformed command if we attempt to add a newline.

Try run for example, `js cliExecute("ash print('Test\nMessage');")`

Session stream testing can also be cleaner if we expose the stream setter.


test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -123,6 +128,51 @@ void floundryLocations() {
...
+                        <td>
+                          <p>
+                            roninStoragePulls
+                          </p>
+                        </td>
+                      </tr>
+                    </table>"""
+              .replaceAll("\r|\n", "<br>"); // The command tests don't handle newlines very well
+
+      // Confirm that print_html doesn't log to session
```
gausie, line 161: Split these tests

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -2959,6 +2966,18 @@ public static Value print(ScriptRuntime controller, final Value string, final Va
     return DataTypes.VOID_VALUE;
   }
 
+  public static Value print_html(
+      ScriptRuntime controller, final Value string, final Value logToSession) {
+    if (logToSession.intValue() == 1) {
+      String parameters = StringUtilities.stripHtml(string.toString());
```
gausie, line 2972: Parameters makes less sense as a variable name here

gausie (commented):
Looks good to me, pending these small changes

libraryaddict:
Actually, that's a thought.

I'm stripping away newlines as done in print. But I'm not sure that makes as much sense here.
One alternative, for example. Would be to split each line and print them individually.

jaadams5 (approved):

jaadams5:
I made a bunch of local modifications to the mirror tests and they passed or failed as I expected.  No real benefit to keeping them but that addressed my concerns.

PR #1042, opened 2022-09-01
CMC tracking fixes: better handling of null-zones

horrible-little-slime:
https://kolmafia.us/threads/tracking-zone-types-for-the-cold-medicine-cabinet.27777/#post-169497


gausie:
Thanks! This isn't quite right, I'll do a PR

PR #1048, opened 2022-09-02
Produce TypeScript definitions from mafia

gausie:
This is a work in progress but I'd like eventually to be able to produce a set of TypeScript definitions for our JS runtime library entirely from within mafia. After that process works well, I'll set up 
some infrastructure such that these are created and pushed to the TypeScript package manager automatically


gausie:
This definitely now produces good TS definitions from within the mafia codebase. Whether it actually publishes to npm will have to be seen after merging

gausie:
This definitely now produces good TS definitions from within the mafia codebase. Whether it actually publishes to npm will have to be seen after merging

.github/workflows/npm.yml
```diff
@@ -0,0 +1,46 @@
...
+  push:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  npm:
+    if: github.repository == 'kolmafia/kolmafia'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
midgleyc, line 13:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/npm.yml
```diff
@@ -0,0 +1,46 @@
...
+          ORG_GRADLE_PROJECT_commit: ${{ github.sha }}
+        run: ./gradlew --no-daemon tsDefs
+
+      - name: Prepare package
+        run: |
+          mv index.d.ts .github/npm/index.d.ts
+          KOLMAFIA_VERSION=$(git rev-list --count $GITHUB_SHA)
+          sed -i 's/{{version}}/$KOLMAFIA_VERSION/g' .github/npm/package.json
+
+      - uses: actions/setup-node@v2
```
midgleyc, line 36:
```suggestion
      - uses: actions/setup-node@v3
```

midgleyc (approved):
Idea seems reasonable but, as node publishes are immutable, maybe you don't want to publish without checking what gets produced?

gausie:
> Idea seems reasonable but, as node publishes are immutable, maybe you don't want to publish without checking what gets produced?

I think it's probably just fine. It's not so high volume that I can't just push a new patch version by hand to fix an issue

PR #1049, opened 2022-09-02
Do not mistake other PvP gains for a goth kid charge

gausie:
https://kolmafia.us/threads/getting-pvp-fights-from-feel-superior-advances-mafias-goth-kid-pvp-tracker.27983/


midgleyc (approved):
Thanks!

PR #1050, opened 2022-09-02
Fix spikolodon spikes use not being tracked

gausie:
Worth remembering that daily limits still need something to say they were successful. This being forgotten is just a side-effect of not having actually implemented the functionality yes though.


jaadams5 (dismissed):
Pending spotless

midgleyc (approved):

PR #1051, opened 2022-09-02
Updating 2CRS seeds

chunkinaround:
Someone let me know there was a reshuffle in the 2CRS seeds so I regenerated them on my multis.


midgleyc (approved):
Thanks!

PR #1053, opened 2022-09-05
Don't sort mall results by shop name

libraryaddict:
First commit is a failing test


jaadams5 (approved):
I'll hold off pushing this in case someone else wants to chime in/

Veracity0:
I'm fine with this.
As the comment on the removed code said, sort alphabetically because ... whatever.

PR #1055, opened 2022-09-06
Add limit of 1 implements a day to glitch season reward

libraryaddict:
As part of breakfast, mafia will implement the glitch season reward if enabled.

However, the implementation can only be incremented once a day. Beyond that doesn't appear to do anything.


gausie:
If we're sure that this is the behaviour, I'm happy. Can you change the verb from implemented to `used`? There's a vague standard for `used` for binary daily limit prefs and `uses` for integer daily limit prefs.

libraryaddict:
> If we're sure that this is the behaviour, I'm happy. Can you change the verb from implemented to `used`? There's a vague standard for `used` for binary daily limit prefs and `uses` for integer daily limit prefs.

Are you sure? This could potentially break expectations and possibly existing scripts. You're referencing to the pref name correct?

gausie:
Oh I see this is not a new pref. Whoops! You're fine

gausie (approved):

PR #1059, opened 2022-09-06
fix: can equip / unequip items to unusable familiars

midgleyc:
Fixes https://kolmafia.us/threads/npe-when-equipping-unequipping-familiar-equipment-to-currently-restricted-familiar.27988/

You can equip items (e.g. tiny stillsuit) to familiars you can't use (e.g. because they are out of standard and you are in standard, or because you are in BHY and they contain 'b's). Before the stillsuit, I don't think you would have ever wanted to do this, but now we have the stillsuit you do.

Avoid crashing when this happens.


gausie (approved):
LGTM!

gausie:
I fear that this will get patched soon, but for now let's merge

PR #1060, opened 2022-09-06
Buzzed on Distillate

gausie:
This tracks the current and next Buzzed on Distillate effect mods, as well as making it possible to track whole effect mods through consequences.txt

There is a bigger refactor to consequences.txt that should be done in my mind - make it o that each consequence knows how it can be checked and if it should be checked (effects that you have, items that you have or have equipped, skills you know etc) and then a single fn to check them all if necessary. Instead we have loads of repeated "check" functions. Anyway, that will come another time.


midgleyc (dismissed):
To confirm: entauntauned is correctly updated by the ConsequenceManager?

gausie:
> To confirm: entauntauned is correctly updated by the ConsequenceManager?

Yes but I'll add a test for that now

PR #1063, opened 2022-09-09
tests: have reasons describe tests, use hasSign everywhere

midgleyc:
Follow-up to #1037.

Use `hasSign` everywhere: couple of places were missed.

Use the reason field to describe what the test is trying to show, in pseudocode (kind of like ASH).

Add contentStrings to the buffers equal to the value of the buffers, because I don't think it's possible for it to be blank for a real buffer.

Add a element - string comparison, because by the looks of it it's case-sensitive, so any element always appears before any location (as all locations start with a capital or number).

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
+++ b/test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
@@ -339,11 +339,11 @@ class ValueTest {
           hasSign(ZERO));
 
       assertThat(
-          "$vykea[level 1 blood dresser] == \"la\"",
+          "$vykea[level 1 blood dresser] > \"la\"",
           value.compareTo(new Value("la")),
           hasSign(POSITIVE));
       assertThat(
-          "$vykea[level 1 blood dresser] == \"lz\"",
+          "$vykea[level 1 blood dresser] < \"lz\"",
           value.compareTo(new Value("lz")),
           hasSign(NEGATIVE));
     }
--- a/test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
+++ b/test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
@@ -33,23 +33,12 @@ class ValueTest {
     void compareBooleans() {
       // booleans compare as ints, except to strings
       var value = new Value(true);
-      assertThat(
-          "true > false",
-          value.compareTo(new Value(false)),
-          hasSign(POSITIVE));
-      assertThat(
-          "true == true",
-          value.compareTo(new Value(true)),
-          hasSign(ZERO));
+      assertThat("true > false", value.compareTo(new Value(false)), hasSign(POSITIVE));
+      assertThat("true == true", value.compareTo(new Value(true)), hasSign(ZERO));
       assertThat("true < 5", value.compareTo(new Value(5)), hasSign(NEGATIVE));
       assertThat(
-          "true > \"non numeric\"",
-          value.compareTo(new Value("non numeric")),
-          hasSign(POSITIVE));
-      assertThat(
-          "true < 1.4",
-          value.compareTo(new Value(1.4)),
-          hasSign(NEGATIVE));
+          "true > \"non numeric\"", value.compareTo(new Value("non numeric")), hasSign(POSITIVE));
+      assertThat("true < 1.4", value.compareTo(new Value(1.4)), hasSign(NEGATIVE));
     }
 
     @Test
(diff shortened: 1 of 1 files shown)


gausie:
> Use the reason field to describe what the test is trying to show, in pseudocode (kind of like ASH).

Oh that is better! Except this is not exactly correct - we're not quite checking operators here. Maybe I should be moving all operator code into `compareTo` as the takeaway from this

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
```diff
@@ -33,167 +31,143 @@ void compareToNull() {
...
-          hasSign(NEGATIVE));
+          "$coinmaster[Disco GiftCo] < 2.4", value.compareTo(new Value(2.4)), hasSign(NEGATIVE));
+      assertThat("$coinmaster[Disco GiftCo] < 2", value.compareTo(new Value(2)), hasSign(NEGATIVE));
       assertThat(
-          "coinmasters compare to ints by int value",
-          value.compareTo(new Value(2)),
-          hasSign(NEGATIVE));
-      assertThat(
-          "coinmasters compare to monsters by int value",
+          "$coinmaster[Disco GiftCo] < $monster[453]",
```
gausie, line 128: Honestly this needs to just be not allowed. It needs to throw an error or something

gausie: Seeing it in pseudocode makes me realise.

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
```diff
@@ -338,65 +293,64 @@ void comparePaths() {
...
           hasSign(ZERO));
 
       assertThat(
-          "vykeas compare to strings by string value (higher value)",
+          "$vykea[level 1 blood dresser] == \"la\"",
           value.compareTo(new Value("la")),
           hasSign(POSITIVE));
       assertThat(
-          "vykeas compare to strings by strings value (lower value)",
+          "$vykea[level 1 blood dresser] == \"lz\"",
```
gausie, line 346: $vykea[level 1 blood dresser] < \"lz\"

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
```diff
@@ -338,65 +293,64 @@ void comparePaths() {
...
-          "vykeas compare to vykeas by vykea value (same type, same rune, same level)",
+          "$vykea[level 1 blood dresser] == $vykea[level 1 blood dresser]",
           value.compareTo(
               DataTypes.makeVykeaValue(
                   VYKEACompanionData.fromString("level 1 blood dresser"), true)),
           hasSign(ZERO));
 
       assertThat(
-          "vykeas compare to strings by string value (higher value)",
+          "$vykea[level 1 blood dresser] == \"la\"",
```
gausie, line 342: `$vykea[level 1 blood dresser] > \"la\"`

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
midgleyc: In Mafia (ASH), most of these aren't allowed -- you get an error about not being able to compare types.

midgleyc:
> this is not exactly correct

It's not supposed to be correct (e.g. $monster[6] is not the monster with id 6, most of the comparisons don't work because they don't auto-cast). It's supposed to be indicative of what's being tested: NEGATIVE -> "<", ZERO -> "==", POSITIVE -> ">".

I think given the helper strings we can hopefully make some simple rules that define the current behaviour, so we know how we expect it to work. Something like:
* if comparing to same object, maybe custom rules (e.g. vykea). Else compare as id if exists, else compare as string.
* if comparing to number-like, cast to float (using id) or use 0 if no id
* if comparing to string-like, cast to string (using name)

Number-like things are things that have an id field. String-like things are things that have a name, or strings, or buffers.

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
midgleyc: thanks, got sloppy right at the end

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
midgleyc: thanks

test/net/sourceforge/kolmafia/textui/parsetree/ValueTest.java
gausie: Right ok

gausie (approved):

PR #1065, opened 2022-09-10
fix: check permitted concoctions in standard

midgleyc:
Revert some of #1064.

This issue was introduced in 228c600a384f99c60028ce38f413e91969fe7108, where the `ConcoctionDatabase.isPermittedMethod` check was moved out of the middle of the function, and so wasn't checked for a variety of createables. This meant that things like the barrel shrine, floundry etc. were considered to be available in standard when they aren't.

Add tests for these things.


gausie (approved):
Ah I see. Man this code is hard to change. But these tests will make it much easier next time!

jaadams5 (approved):

PR #1067, opened 2022-09-11
feat: reserve extra types

midgleyc:
Follow-up to #1032.

By https://kolmafia.us/threads/witchess-puzzle-solver.19964/page-3#post-169591, ASH scripts that used "path" as an array variable broke anyway.

Reserve all the unreserved types to give a better error message.


jaadams5 (approved):

midgleyc:
I expect this may break other scripts, but hopefully the error message will make it obvious what needs to be done.

I had a quick run-through those I had installed and it seemed okay.

PR #1068, opened 2022-09-12
Fix coat of paint not updating modifiers after avatar reset

libraryaddict:


gausie:
This needs a test otherwise it could just be broken again in the future

PR #1075, opened 2022-09-12
More canAdventure and prepareForAdventure

Veracity0:
1) prepareForAdventure will make the Talisman o' Namsilat if you have the components.
canAdventure understands that
test suite for this: talisman equipped, in inventory, components in inventory, plunger available, meat paste needed, ...
2) Investigating a Plaintive Telegraph is only available if you actually have  started a quest.
preValidateAdventure -> you have permanent or daily access to the LT&T Telegraph Office
canAdventure -> you have accepted a quest
3) preValidateAdventure checks perm/daily access to PirateRealm & FantasyRealm
visiting monorail will set daily access, if appropriate
4) Put all Manor validation under parentZone.equals("Manor")
5) Just as can_adventure() ASH function returns false for $location[none], so too do pre_validate_adventure() and prepare_for_adventure()
6) Cleanups withMeat and withSign need to refresh concoctions.
7) The various floors of Spookyraven Manor open up when you receive kmail from Lady Spookyraven, either immediately after interacting with her on the previous floor or at levels 5/7/9 for unascended characters. WIth the exception of dancing with her in the ballroom (which takes a turn), we can do those interactions.
8) Provide actual HTML for reading the telegram
9) We can acquire the continuum transfunctioner from the crackpot mystic.
10) Test with actual HTML from talking with Lady Spookyraven.


gausie:
Welcome back!

Veracity0:
While working on this, I realized that we SET The isValidAdventure field of a KoLAdventure object, but we never look at it anymore.

Seems obsolete and ripe for removal.

Veracity0:
I also just realized that the Pixel coinmaster is not open unless you have a transfunctioner.

prepareForAdventure will reliable get you one, so could be used to make that coinmaster available.

gausie (approved):
This is great work, well needed, well commented, well tested. Thank you!

PR #1076, opened 2022-09-12
Move inebriety checks into adventure validation, allow explicit stupor adventuring

gausie:
This is ready to review!!


Veracity0:
I would consider looking at preValidateAdventure.

We don't call that internally yet, but will, soon.

You absolutely cannot adventure in a drunken stupor unless you are overdrunk. Cut it off at step 0.

Note that the wineglass lets you AVOID drunkenstupor. It does not allow it.

gausie:
> I would consider looking at preValidateAdventure.
> 
> We don't call that internally yet, but will, soon.
> 
> You absolutely cannot adventure in a drunken stupor unless you are overdrunk. Cut it off at step 0.
> 
> Note that the wineglass lets you AVOID drunkenstupor. It does not allow it.

Ok! I'll mark this as draft, and come back to it when we're using preValidateAdventure. Might as well get this all in one go

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1363,14 +1364,15 @@ private static void executeAdventureOnce(
 
       if (KoLCharacter.hasEquipped(ItemPool.get(ItemPool.DRUNKULA_WINEGLASS, 1))) {
         // The wine glass allows you to adventure while falling down drunk
-      } else if (KoLCharacter.getLimitmode() == Limitmode.SPELUNKY) {
+      } else if (Objects.equals(KoLCharacter.getLimitmode(), Limitmode.SPELUNKY)) {
```
Veracity0, line 1367: Why? Look at KoLCharacter.setLimitMode(String limitmode):

```
    } else if (limitmode.equals(Limitmode.SPELUNKY)) {
      KoLCharacter.limitmode = Limitmode.SPELUNKY;
    } else if (limitmode.equals(Limitmode.BATMAN)) {
      KoLCharacter.limitmode = Limitmode.BATMAN;
      BatManager.setCombatSkills();
    } else if (limitmode.equals(Limitmode.ED)) {
      KoLCharacter.limitmode = Limitmode.ED;
    } else {
      KoLCharacter.limitmode = limitmode;
    }
```

Here is Limitmode.java:
```
  // Limitmode
  public static final String SPELUNKY = "spelunky";
  public static final String BATMAN = "batman";
  public static final String ED = "edunder";
```

There are currently only 3 limitmodes. I suppose it would be nice if Limitmode was an enum, but as coded,  those final strings are effectively the equivalent. Specifically so you can use "==" with them, as we do, all over the place.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -1363,14 +1364,15 @@ private static void executeAdventureOnce(
...
         // You're allowed to Spelunk even while falling down drunk
-      } else if (KoLCharacter.getLimitmode() == Limitmode.BATMAN) {
+      } else if (Objects.equals(KoLCharacter.getLimitmode(), Limitmode.BATMAN)) {
```
Veracity0, line 1369: Again. Using Objects.equals just obfuscates the code, to my eye.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/KoLmafia.java
gausie: Agreed. A limitmode enum refactor sounds nice for the future.

gausie:
Just need to test this myself and I'll take it out of draft. I'm cautious that perhaps there are some situations where the drunkenness check no longer applies since it was so low-level before. But anywhere where it once applied, we should be applying these validation checks anyway.

gausie:
Ok I've tested this now and am happy with it.

gausie:
> Note that the wineglass lets you AVOID drunkenstupor. It does not allow it.

Haha interestingly if you go straight to the snarfblat for drunken stupor, this is not the case!

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -2863,12 +2878,8 @@ private void prepareToAdventure(final String urlString) {
...
     // steps have been executed.
 
-    // If we are in a drunken stupor, return now.
-    if (KoLCharacter.isFallingDown()
-        && !urlString.startsWith("trickortreat")
-        && !KoLCharacter.hasEquipped(ItemPool.get(ItemPool.DRUNKULA_WINEGLASS, 1))) {
-      return;
-    }
+    // If we are too drunk adventure, return now.
+    if (tooDrunkToAdventure()) return;
```
Veracity0, line 2882: OK.

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -1498,8 +1513,8 @@ public boolean canAdventure() {
       return switch (this.adventureNumber) {
         case AdventurePool.DRUNKEN_STUPOR -> KoLCharacter.isFallingDown();
         case AdventurePool.SSPD_STUPOR -> today.contains(holiday)
-            && KoLCharacter.getInebriety() >= 26;
-        default -> (holiday == null) ? true : today.contains(holiday);
+            || today.equals("Drunksgiving") && KoLCharacter.getInebriety() >= 26;
```
Veracity0, line 1516: We used to have (in KoLmafia.java):

```
      } else if (!holiday.contains("St. Sneaky Pete's Day") && !holiday.contains("Drunksgiving")) {
        KoLmafia.updateDisplay(MafiaState.ERROR, "You are too drunk to continue.");
        return;
      } else if (KoLCharacter.getInebriety() <= 25) {
        KoLmafia.updateDisplay(MafiaState.ERROR, "You are not drunk enough to continue.");
        return;
      }
```

You do the first check and print the error message in preValidateAdventure.
You are doing the second check here. I think you should print the error message.

Veracity0 (changes requested):
What "HTML fixtures" do you need?
My main will be on day 3 of a Dinocore run tomorrow.
She has not accepted the quest from Meatsmith, Armorer, or Doc Galaktik.
She has a bone with a price tag on it.
She does not have hypnotic breadcrumbs (but could get them)
She does not have a map to a hidden booze cache (but could get it).

I could easily get HTML for prepareForAdventure for each quest using one or the other method - using item for starting quest. Perhaps I should let you submit this, and then un-disable some tests.

src/net/sourceforge/kolmafia/KoLAdventure.java
gausie: > Perhaps I should let you submit this, and then un-disable some tests.

Yeah that was my plan. I didn't write these tests, I just observef that they were implicitly disables and instead made them explicitly disabled so we're less likely to forget about them!

gausie:
> Perhaps I should let you submit this, and then un-disable some tests.

Yeah that was my plan. I didn't write these tests, I just observef that they were implicitly disables and instead made them explicitly disabled so we're less likely to forget about them!

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -1504,9 +1500,16 @@ public boolean canAdventure() {
       String today = HolidayDatabase.getHoliday();
       return switch (this.adventureNumber) {
         case AdventurePool.DRUNKEN_STUPOR -> KoLCharacter.isFallingDown();
-        case AdventurePool.SSPD_STUPOR -> today.contains(holiday)
-            && KoLCharacter.getInebriety() >= 26;
-        default -> (holiday == null) ? true : today.contains(holiday);
+        case AdventurePool.SSPD_STUPOR -> {
+          if (today.contains(holiday)
+              || today.equals("Drunksgiving") && KoLCharacter.getInebriety() >= 26) {
+            yield true;
```
Veracity0, line 1506: Is this correct? && has a higher operator precedence than ||.
I think you mean
```(today.contains(holiday) || today.equals("Drunksgiving) && KoLCharacter.getInebriety() >= 16```

Or maybe even ```today.contains("Drunksgiving")``` although when is that going to next happen? :)

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/KoLAdventure.java
gausie: Good spot!

Veracity0 (approved):

PR #1081, opened 2022-09-14
feat: split owned familiars from usable familiars

midgleyc:
Split the concept of "owning" a familiar (it would be in your terrarium in aftercore) from the concept of "being able to use" a familiar (you can equip it).

There are several familiars / items that can do things on familiars you can't use
* you can get a Grim Brother buff if you own the familiar, even if you can't use it
* Unconscious Collective offers extra free rests per day, as long as you aren't in Standard (e.g. if you're in G-Lover)
* Shorter-Order Cook gives extra EXP to your familiars, and even more with the blue plate
* tiny stillsuit accumulates sweat even on familiars you can't use
* items you equipped on familiars you couldn't use can still be unequipped

In standard, familiars you can't use don't show in the dropdown to equip, but you can still URL submit them in. In paths like G-Lover / Zombie Slayer, unusable familiars still show up and can be equipped.

Add an ash function `own_familiar` that can check whether a familiar is owned, instead of whether it is usable.


gausie:
This is great. Maybe `in_terrarium` or something is a better name?

Also, do you think we should use Optional instead of null returns moving forward? I like doing monadish sort of things elsewhere but just don't in mafia and I'm not sure why. It feels great ergonomically here.

midgleyc:
`in_terrarium` sounds nicer, but maybe isn't the best -- in paths like Ed you can still "own" a familiar even if you don't have a terrarium.

I think Optional is good for clearly saying "this method might return null, and you should handle that". I think it's slightly less ergonomic than null (and in languages with `?.` and `??` operators, much less ergonomic), but still worthwhile for the clarity of intent. Something like C#'s nullable reference types or Rust's `match` would be nice but I don't expect to ever make their way to Java.

gausie:
> `in_terrarium` sounds nicer, but maybe isn't the best -- in paths like Ed you can still "own" a familiar even if you don't have a terrarium.

Legit, though I feel like `own` to me perhaps implies that you might have the hatchling? Maybe it makes even more sense in Ed in that explains where your familiar is - it's left in your otherworldly terrarium!

> I think Optional is good for clearly saying "this method might return null, and you should handle that". I think it's slightly less ergonomic than null (and in languages with `?.` and `??` operators, much less ergonomic), but still worthwhile for the clarity of intent.

Super agreed

>Something like C#'s nullable reference types or Rust's `match` would be nice but I don't expect to ever make their way to Java.

https://openjdk.org/jeps/406!

midgleyc:
Given it a whirl -- fixed registering familiars in standard. Owning and using functions work. Available items works for those equipped on unusable familiars. Stillsuit works. Familiar training pane works.

Good to go I think :)

gausie (approved):

PR #1084, opened 2022-09-15
Move limitmode to enum

gausie:
This PR:

- Turns the Limitmode class into LimitMode enum
- Flattens out the dependency tree - instead of asking LimitMode's statics questions that then asks KoLCharacter for the current limit mode and assesses that, LimitModes have instance methods and we query against KoLCharacter.getLimitMode() directly when appropriate
- Adds tests for all those functions in LimitMode


gausie:
Why has this seemingly changed the HTTP method that api.php requests are made with? Shouldn't they be GET anyway?? What is going on!

Veracity0:
I love this.

I will be adding a few more Limitmodes - ASTRAL and MOLE, both of which limit which zones you can adventure in.  Maybe BIRD, since that replaces your combat skills with bird skills. MOLE and BIRD are mutually exclusive. The question is whether ASTRAL and BIRD are.

In any case, this is really good work.

gausie:
That sounds good! Though from this code it seems that limit modes are defined kol-side and indicated via the API. We'll need to keep that in mind

Veracity0:
Well, poo. I may need to rethink that, then. I need the equivalent, for limiting where you can adventure when you are Half-Astral or Shape of Mole.

gausie:
> Well, poo. I may need to rethink that, then. I need the equivalent, for limiting where you can adventure when you are Half-Astral or Shape of Mole.

I think there's a way we can do it, by looking at certain prefs/effects when we get a "0" back from that API key...

midgleyc (approved):
in the area, might be worth a comment that limit mode "edunder" is Ed in the Underworld, because it's fairly confusing if you don't know that.

Spelunky and Batfellow are self-explanatory.

PR #1086, opened 2022-09-16
Replace Violet Fog mapping tool

gausie:
I was doing some astral mushrooming recently and wanted to explore the Violet Fog, but the tooling that once existed doesn't really anymore. So I made my own, this time built into the relay browser.

* Remove dead links for and code to support Violet Fog mapper
* Add local mapper that should never go away
* Add tests for existing code and new code

This new mapper is a little different to the old one and perhaps not as pretty but it is a lot easier to code 😅. More importantly, it is more interactive! You can:

* Hover over the nodes to see what the name of the choice is
* Click on the nodes to open the wiki page for that choice
* Hover over the arrows to see what choice they refer to

Here's what it looks like:

Screenshot 2022-09-16 at 21 03 26


gausie:
Ok I've added a legend to the bottom of the map and I think that makes it ready to roll
Screenshot 2022-09-16 at 21 53 31

midgleyc (approved):
Much better than a broken map!

Our other external dependency is the Wumpinator, which is fortunately still up for now.

Veracity0:
Wumpus was down for a while. Gemelli has not been on KoL for a while and feesher.com apparently was no longer online.

I could swear I've exchanged email with Gemelli before, but could not find it.

It struck me that I bet fewyn would be willing to host it, if we could get the code, so I reached out to CDM to ask if he knew how to contact Gemelli. I mentioned the issue and gave him my (real) name and email and asked if he could forward them along.

Never got a response from either of those two, but the Wumpinator is back online. For now.

gausie:
I'd be happy to improve this to being as good as the Gemelli map to be honest. I can basically make an SVG in the right shape and just get mafia to change the colors of specific objects inside it. I just didn't want to make the SVG!

I think in reality despite being less pretty it is just as if not more easy to use.

gausie:
It occurs to me that I could do that for the wumpinator and bring it in house, which I am also tempted to do

PR #1089, opened 2022-09-17
fix: limit_mode for none must be empty

midgleyc:
#1084 broke ChIT, which assumes the result of `limit_mode` for no limit mode must be "", but it is instead "0".


gausie:
I really think this isn't right. We should keep it as 0 - the official key we get from KoL servers - and add a getASHName instance function to LimitMode that returns "" when the enum is LimitMode.NONE, and use that in limit_mode().

midgleyc:
"0" isn't the key, though. The key is an integer -- 0.

I'd rather change the API code to something like

```
    Object lmo = JSON.get("limitmode");
    if (lmo instanceof Integer && lmo.equals(0)) {
      KoLCharacter.setLimitMode(LimitMode.NONE);
    } else {
      KoLCharacter.setLimitMode(lmo.toString());
    }
```

and add UNKNOWN for an unknown string.

gausie:
> "0" isn't the key, though. The key is an integer -- 0.
> 
> I'd rather change the API code to something like
> 
> ```
>     Object lmo = JSON.get("limitmode");
>     if (lmo instanceof Integer && lmo.equals(0)) {
>       KoLCharacter.setLimitMode(LimitMode.NONE);
>     } else {
>       KoLCharacter.setLimitMode(lmo.toString());
>     }
> ```
> 
> and add UNKNOWN for an unknown string.

This works with me

src/net/sourceforge/kolmafia/request/CharPaneRequest.java
```diff
@@ -1559,7 +1559,11 @@ public static final void parseStatus(final JSONObject JSON) throws JSONException
     }
 
     Object lmo = JSON.get("limitmode");
-    KoLCharacter.setLimitMode(lmo.toString());
+    if (lmo instanceof Integer && lmo.equals(0)) {
+      KoLCharacter.setLimitMode(LimitMode.NONE);
+    } else {
+      KoLCharacter.setLimitMode(lmo.toString());
```
gausie, line 1565: We should probably instanceof string check here and error out otherwise. And add a coverage test to this slice of the code too, why not

gausie (approved):

src/net/sourceforge/kolmafia/request/CharPaneRequest.java
midgleyc: aye, will merge here and follow-up.

PR #1091, opened 2022-09-17
Fix transitivity of Value#compareTo.

heeheehee-kolmafia:
This is breaking certain map operations.

https://kolmafia.us/threads/6445/page-115#post-169713

For instance:

- the darkness (blind) > fired-up Knob Goblin King (string comparison)
- the darkness (blind) < zombie waltzers (string comparison)
- zombie waltzers < fired-up Knob Goblin King (id comparison)

There is still... special breakage with how we handle Ed the Undying (which has some special fix-up in MonsterDatabase), although the fix isn't superficial.

Note that I removed various comparisons of mixed Types. I do not think these are exercised in real code because Operator will perform coercions such that the resulting types are sufficiently similar.


midgleyc:
This change means that `"150" == 150` is no longer true, which I think at least one script was relying on.

Instead, can we order monsters by id, then string? I don't think anything relies on the string ordering (though I may be proven wrong).

And also:

```
> ash (150 == "150")

Cannot apply operator == to 150 (int) and 150 (string) (char 2 to char 14)
Internal error: left hand side and right hand side do not correspond ()
Returned: void

> ash ("150" == 150)

Returned: true
```

This is plain odd! Equality isn't symmetric.

heeheehee-kolmafia:
> Instead, can we order monsters by id, then string? I don't think anything relies on the string ordering (though I may be proven wrong).

By that, do you mean add a special case to bypass all of the isStringLike() logic, as we do for VYKEA_TYPE? We currently only use the string ordering as a fallback if monster id == 0. The problem arises when we compare a monster with id == 0 with one with id != 0.

> This is plain odd! Equality isn't symmetric. It wasn't before this PR, either (but the latter comparison was true).

This reads to me like an issue with how we do coercion.

Anyways, I'll think about this a little more.

midgleyc (approved):

PR #1094, opened 2022-09-18
Fix serialization of Ed the Undying's forms.

heeheehee-kolmafia:
MonsterDatabase has special handling for Ed's forms such that they all have the same nonzero id. This has two consequences in the current implementation:

1. Ed's forms are all == to each other. This has unexpected results when inserting multiple forms into the same map.

2. Attempting to write any of Ed's forms via map_to_file will look up the Monster with id=473, namely $monster[Ed the Undying].

This commit addresses both of those cases.


gausie (approved):
Thanks h³, looks great

PR #1095, opened 2022-09-18
feat: add a link to the chaosthetician

midgleyc:
Provide a link to go directly to the Chaosthetician when you get the message indicating you're full up on research.


jaadams5 (approved):

jaadams5:
Do we have any support for this for automated adventuring?

In my case if I need the Dino Dollars enough to wear the harness it would be nice to have "research completed" abort adventuring or, better yet, just turn in the research and continue adventuring.

midgleyc:
At the moment I don't think we even track how much research you need or are accumulating.

Turning in the research is an interesting idea: I don't think we do anything similar anywhere else. It could be analogous to automatically making certain concoctions (e.g. McClusky file, reassembled blackbird) after the items drop in a fight.

jaadams5:
Using an enchanted bean might be another example.

A post-adventure script could do it but mafia support would still be helpful.  The maintenance issue when several post adventure script from different authors are desired would be avoided.

PR #1096, opened 2022-09-18
Update build.yml

jaadams5:
Don't build exe

The exe build always does a new install trashing what was there before.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -104,7 +104,6 @@ jobs:
           files: |
             dist/*.jar
             releases/*.deb
-            releases/*.exe
             releases/*.dmg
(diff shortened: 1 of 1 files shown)


gausie:
This is worth doing though it's probably possible to fix the exes too

jaadams5:
> This is worth doing though it's probably possible to fix the exes too

I did it now because using the exe really did trash a lot of files that the user wanted.  If we choose to continue to distribute the exe we should try and fix it.  However I cannot build a KoLmafia exe on my system from a command line, and I can't make any progress until someone helps me understand why GitHub can build it but I cannot?  Windows Home vs. Professional?  Dunno but my efforts have been stalled for a while.

midgleyc (approved):

PR #1097, opened 2022-09-18
Lint and more tests (coverage) for ItemFinder

jaadams5:


src/net/sourceforge/kolmafia/persistence/ItemFinder.java
```diff
@@ -39,15 +39,15 @@ public enum Match {
...
 
-  public static final String getFirstMatchingItemName(List<String> nameList, String searchString) {
+  public static String getFirstMatchingItemName(List<String> nameList, String searchString) {
     return ItemFinder.getFirstMatchingItemName(nameList, searchString, Match.ANY);
   }
 
-  public static final String getFirstMatchingItemName(
+  public static String getFirstMatchingItemName(
       List<String> nameList, String searchString, Match filterType) {
     if (nameList == null || nameList.isEmpty()) {
```
jaadams5, line 52: I have a failing test and I don't understand whether it is my misunderstanding or whether the underlying code is wrong.
The parameter searchString is not used which seems to me to be an error.

jaadams5:
I started to some test driven by coverage and found several things I did not understand.  What is there about the contract/usage of filterNameList that causes it to ignore everything that is not a restore?  Why do candy hearts etc. take precedence?  Why doesn't getFirstMatchingItemName need a search string to check?  First, do no harm suggested I might find better uses of my time so I stopped, pretty much, with just the lint.

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -906,20 +902,68 @@ public void itShouldParseMultipleItems() {
...
+  public void itShouldMatchANYWhenNotExplicitlySpecified() {
+    List<String> nameList = new ArrayList<>();
+    String item;
+    nameList.add("7-ball");
+    nameList.add("17-ball");
+    item = ItemFinder.getFirstMatchingItemName(nameList, "7-ball");
+    assertEquals(StaticEntity.getContinuationState(), MafiaState.CONTINUE);
+    assertNotNull(item);
+    assertEquals(item, "7-ball");
+    nameList.clear();
```
gausie, line 937: Please don't do this with tests - make a second test or parameterise this one!

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -906,20 +902,68 @@ public void itShouldParseMultipleItems() {
...
+    assertEquals(StaticEntity.getContinuationState(), MafiaState.CONTINUE);
+    assertNotNull(item);
+    assertEquals(item, "7-ball");
+  }
+
+  @Test
+  public void itShouldReturnNullWhenPassedNullOrEmptyList() {
+    List<String> nameList = new ArrayList<>();
+    String item;
+    item = ItemFinder.getFirstMatchingItemName(null, "7-ball");
```
gausie, line 950: Again, this could be parameterised rather than repeating a test interally (and thus obfuscating later failures)

gausie (changes requested):

test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
jaadams5: Someday I'll remember that this is a better practice and remember it.  Will address.

jaadams5:
I broke up a couple of tests but did not parameterize them.  I deleted one test because as I tried to figure out what to name it and what I was testing, besides coverage, I realized I didn't really understand what was going on.  There's also unused code in ItemFinder that is being tested.  Decided to defer all of that and basically just commit lint.

gausie (approved):

PR #1099, opened 2022-09-19
Add nc alias for the parka command

s-k-z:
Parka in pterodactyl mode has -5% combat frequency. This additional alias will make it easier to select, and be more consistent with the umbrella command which also has an 'nc' option.


midgleyc:
I deliberately didn't add this because to me "nc" means "select the spikeolodon, which has a skill you can use to force an NC".

Interestingly everybody else who has asked for this feature is sure it should mean the pterodactyl. I could be convinced.

gausie:
I think perhaps we do this with a message to the GCLI that this means the -combat modifier and if they want to force noncombats to run `parka whatever`

PR #1100, opened 2022-09-19
refactor: make KoLCharacter.getSign return a ZodiacSign instead of a String

midgleyc:
Inspired by https://kolmafia.us/threads/unable-to-fetch-tcrs-data-fetches-empty-files.28055/.

Also make TCRSCommand convert the received string into a ZodiacSign.


gausie (approved):
Looks great. $zodiac when? 😉

Veracity0:
> Looks great. $zodiac when? 😉

Ha ha ha. I wanted to do that when the signs were first introduced, but hola vetoed the idea, since he couldn't imagine why a script would need it.

gausie:
> > Looks great. $zodiac when? 😉
> 
> Ha ha ha. I wanted to do that when the signs were first introduced, but hola vetoed the idea, since he couldn't imagine why a script would need it.

Now seeding is so frequently done by sign, there are lots of situations in which sign strings are thrown around in scriptland. Maybe something to genuinely consider!

PR #1102, opened 2022-09-20
Add coverage for GenericRequest protections

gausie:
Occasionally we (I) push code that was not intended and this is to ensure that never gets missed and accidentally merged.


midgleyc (approved):
uh, k

Veracity0:
> woah look at that coverage change

0.40% was from my PR, which for reasons I do not understand, was called out when I looked at the (1) file which was updated in this PR.

https://github.com/kolmafia/kolmafia/pull/1087

PR #1106, opened 2022-09-21
Fix detecting start of Untinker quest

Veracity0:
I observed that accepting the Untinker's quest doesn't actually set the quest property.

1) I reorganized and categorized QuestManagerTest:
- Council quests
- Other Quests
- Guild Quests
- Path Quests
- Zodiac Quests
- Non-Quest Related
- Kingdom-Wide Unlocks
2) I made liberal use of @Nested to group tests related to the same quest
3) KoL requests frequently have "action=WORD". Sometimes they also have "preaction=WORD".
GenericRequest.getAction() is intended to return the first - but guess what you get if the second appears first in the URL?
That method now uses a Pattern with zero-width negative lookbehind. (!)
4) PlaceRequest only asked UntinkerRequest to parse responses where the URL equalled "fv_untinker".
That's a double whammy:
- fv_untinker_quest is the action before you accept the quest
- preaction=screwquest is used to actually accept it
5) UntinkerRequest.parseResponse was coded to work only if "screwquest" was in the URL.
So, all the other code - like removing the rusty screwdriver when you turn it in - was not executed.
In fact, that code also handled removing things you untinkered from inventory. How did that ever work?
6) canAdventure in hostile Knoll zones depends on either starting Paco's Meatcar quest or starting the Untinker's quest
prepareForAdventure will accept the Untinker's quest, if necessary


gausie (dismissed):
Awesome. I love the reorganisation of QuestManagerTest too - I put most of those tests together but it was before I really discovered and embraced the @Nested annotation

I'm surprised this isn't increasing coverage - was there any coverage lost from elsewhere?

Veracity0:
It's confusing.

It says my diff coverage for UntinkerRequest is 25% - but every single line I changed or added was covered by a test, as looking at the Code diff in the PR tells you.

It also says coverage went down for files I did not change?
I don't THINK I removed any tests.

gausie (approved):

PR #1107, opened 2022-09-21
Change big red button from nopickpocket to conditional

horrible-little-slime:
https://cdn.discordapp.com/attachments/473525651043909642/1022198810874757240/Screenshot_20220921-113240_Brave.jpg

Per spading, fails to drop when YRed, so it is not "just" nopp.


gausie (approved):

Veracity0:
What  is the necessary condition?

PR #1108, opened 2022-09-21
Include mode in names of Jurassic Parka and backup camera

Veracity0:
Just as we do for the unbreakable umbrella.


gausie:
Nice. We could take this one step further and automatically apply these aliases for Modeables. No need to do it on this PR but I may look into it if you don't

gausie (approved):

PR #1109, opened 2022-09-22
Fix action/preaction issue

gausie:
This is mostly here for the failing test, though I have also added some band-aid fixes for matching preactions with actions where necessary.


midgleyc (approved):

Veracity0:
Nice.

I think this tells me that when I do the Azazel quest next run, I need to get some DEBUG logs for when I interact with the various NPCs. :)

Veracity0 (approved):

PR #1112, opened 2022-09-22
Relay Browser decorations for El Vibrato adventuring

Veracity0:
1) When your EL Vibrato Portal runs out of power, give up to three links:
- (If you have an overcharged El Vibrato power sphere)
"Insert an overcharged El Vibrato power sphere"
- (If you have an El Vibrato power sphere)
"Insert an El Vibrato power sphere"
- (always, although it's not really useful without one of the above)
"Go to your El Vibrato portal"
2) When you are in battle with an El Vibrato construct, add a little table with 5 columns per row:
COMMAND! (this will be a button)
[verb card] (count)
[object card] (count)
[required item, if any] (count)
description
3) Now there is a button.
It is disabled if you do not have both cards - and the item, if one is required
It submits a macro to use the two cards in order.
It will use Funkslinging, if you know it.

Tested:

- "Out of power" with a sphere
- "Out of power" with an overcharged sphere
- Constructs untranslated
- Constructs translated

I suppose I could automate testing the El VibratoManager.decorate method by passing in a responseText with the expected insertion point and verifying that various table lines were generated as expected.


gausie:
Just tested this and it works great!

While getting my multi set up I note that it would be nice to be able to "use" a sphere or overcharged sphere and mafia put it in the portal

gausie (approved):

Veracity0:
Thanks!

FWIW, I added unit tests for the decorator

- lonely construct and lonely construct (translated)
- with and without funkslinging
- with two cards and object
- with two cards missing object
- with 2 cards no object needed
- with 1 card
- lonely construct (translated)
I'm not going to wait for another review, since it was all added testing.

PR #1115, opened 2022-09-23
Fix more action/preaction stuff

gausie:
- Asdon fueling does use action not preaction
- Make sure we handle subaction too...


gausie:
Just going to merge this since its a bug fix

PR #1116, opened 2022-09-23
Add test for chips being tracked in Clan Rumpus Room

gausie:
I had a report that this wasn't working. My test here shows that it is, so I think they must have been using an outdated mafia, but here is a test anyway.


test/internal/helpers/Player.java
```diff
@@ -326,6 +327,58 @@ public static Cleanups withClanLoungeItem(final int itemId) {
...
+  }
+
+  /**
+   * Adds or removes the given furniture to the player's Clan Rumpus Room
+   *
+   * @param furniture Furniture to install
+   * @param shouldHave Whether the player should have the furniture
+   * @return Resets to previous value
+   */
+  public static Cleanups withClanFurniture(final String furniture, final boolean shouldHave) {
```
midgleyc, line 364: this function is essentially two different functions based on the boolean parameter -- they only share the first line. I think it'd be clearer as `withClanFurniture` and `withoutClanFurniture`.

test/internal/helpers/Player.java
```diff
@@ -326,6 +327,58 @@ public static Cleanups withClanLoungeItem(final int itemId) {
     return new Cleanups(() -> ClanLoungeRequest.setClanLoungeItem(itemId, 0));
   }
 
+  /**
+   * Restores Clan Furniture after cleanup
+   *
+   * @return Resets to previous state
+   */
+  public static Cleanups withClanFurniture() {
```
midgleyc, line 335: this function removes added items but doesn't add items that were removed. Also with `withClanFurniture(final String furniture)` cleaning up after itself I'm not sure when this would be used?

test/net/sourceforge/kolmafia/request/ClanRumpusRequestTest.java
```diff
@@ -25,11 +30,34 @@ void ballpitRequestParsesAction() {
...
+  void canTrackChipMachineFinished() {
+    var builder = new FakeHttpClientBuilder();
+
+    builder.client.addResponse(200, html("request/test_clan_rumpus_chips_used.html"));
+
+    var cleanups =
+        new Cleanups(
+            withHttpClientBuilder(builder),
+            withClanFurniture(),
+            withClanFurniture("Snack Machine"),
```
midgleyc, line 51: I feel like the first line is redundant with the second.

test/internal/helpers/Player.java
gausie: When you interact with your clan lounge it populates your whole clan rumpus room by reading the entire page. I can just be explicit about each one though. I'll refactor these functions

test/internal/helpers/Player.java
gausie: I've never liked withoutX functions but I can change it if you want.

jaadams5:
> I had a report that this wasn't working. My test here shows that it is, so I think they must have been using an outdated mafia, but here is a test anyway.

It was not working for me yesterday and worked for me this morning.  I think some of the action/preaction changes broke it and it was quickly fixed.

test/internal/helpers/Player.java
midgleyc: I feel like `withoutX` reads much more easily as a condition than `withX(false)`.

In this particular case, I think this function gives a great example of the [flag argument](https://martinfowler.com/bliki/FlagArgument.html) and should be split up for readability.

Veracity0:
Technically out of scope for this test, but the Klaw is not working because it does not use a preaction.
```
clan_rumpus.php?action=click&spot=3&furni=3
```
contains:

```
<td>You carefully guide the claw over the prize that looks the easiest to grab.  You press the button and the claw slowly descends. <p>Unfortunately, the claw slips off of the prize without grabbing it.<p>The machine makes a horrible clanking noise, and a wisp of smoke pours out of the prize chute.</td>
```

and the code to handle it:

```
    if (preaction == null) {
      return;
    }
...
    switch (preaction) {
...
      default:
        if (urlString.contains("spot=3") && urlString.contains("furni=3")) {
          // You carefully guide the claw over the prize that
          // looks the easiest to grab. You press the button and
          // the claw slowly descends.
          if (responseText.contains("slowly descends")) {
            Preferences.increment("_klawSummons", 1);
          }
          // The machine makes a horrible clanking noise, and a
          // wisp of smoke pours out of the prize chute.
          //
          // The crane machine seems to be broken down. Oh
          // well. Maybe they'll fix it by tomorrow.
          else if (responseText.contains("seems to be broken down")) {
            Preferences.setInteger("_klawSummons", 3);
          }

          return;
        }
        break;
    }
```

test/internal/helpers/Player.java
midgleyc: ah, if it's doing that I think it's worth having one to reset everything (like the one that resets fights). I just wasn't clear that was what it was supposed to be doing. `resetLoungeVisit()`?

jaadams5 (approved):

PR #1118, opened 2022-09-24
Break Prism Warnings should return false if previously confirmed

Veracity0:
I was irked when I got a warning that suggested I spend my DInodollars, clicked the icon to say I was fine with not doing this, and was immediately given the same warning.

The "hand" icon add a confirmation field to the URL.
The warning failed to check the confirmation.

This same issue was present in Kingdom of Explothing and Robocore.

1) I fixed all three paths.
2) I split the Robocore test into multiple single tests
I put them into a Nested test class
I made them use Cleanups
3) I added a test for Robocore
4) I added a test for Dinocore


midgleyc (approved):
Thanks!

PR #1120, opened 2022-09-24
Remove unbreakable umbrella from left-hand man if it's in the terrarium before attempting to fold

horrible-little-slime:
To properly fold the umbrella, it must be in your inventory, equipped, or equipped on your active familiar. When it gets trapped in the terrarium, we run into issues. This (hopefully) resolves those issues!


src/net/sourceforge/kolmafia/textui/command/UmbrellaCommand.java
```diff
@@ -47,6 +49,13 @@ public void run(final String cmd, String parameter) {
       return;
     }
 
+    var lefty = KoLCharacter.ownedFamiliar(FamiliarPool.LEFT_HAND);
```
gausie, line 52: Throw this all in lefty.isPresent to avoid the repeated gets

gausie:
Does this work under standard restrictions?

horrible-little-slime:
I'm not sure what you mean by that; Lefty and Umbrella are currently both standard. Lefty is leaving this December, and when it leaves you presumably won't be able to get an umbrella onto it in standard. Or, if you can, presumably the same method would work to remove it.

jaadams5:
I don't have a LHM but if I read the wiki it can wield any accessory.  So can this occur with more things than just the Umbrella?  Stinky cheese?  Legion Loathing Knife?  Would we be better off with a more general solution?

horrible-little-slime:
It can only equip off-hands, not accessories.

A more general solution could be helpful, but one distinction worth keeping in mind is that stinky cheese and LLK are both foldables and thus pass through the `fold` command; this is a modeable (that is, there's only one itemID) and has its own unique command. And the `fold` command has a `retrieveItem` call already:

```
> fold loathing legion moondi

You acquire an item: Loathing Legion moondial

> familiar left;equip loathing legion moondi

Putting Card the Baby Sandworm back into terrarium...
Taking Nick Cave the Left-Hand Man out of terrarium...
Holding Loathing Legion moondial...
Equipment changed.

> familiar stooper

Putting Nick Cave the Left-Hand Man back into terrarium...
Taking dEnIsSsS the Stooper out of terrarium...
Requests complete.

> fold loathing legion helicopter

Taking off Loathing Legion moondial...
Equipment changed.
You acquire an item: Loathing Legion helicopter
```

jaadams5:
OK.  Accessory vs. off hand was Brain Flatulence on my part.  I wasn't making a distinction between foldables and modeables but I should.  I'll ask if there are any other modeables that need the same treatment because they can be used offhand.  I'll also ask if there are any other special treatments for modeables that should be extended to the umbrella.  But I won't ask that this be held until those issues are answered or implemented :-)

jaadams5 (approved):

PR #1121, opened 2022-09-25
Track time of last friars NCs

horrible-little-slime:
In each of the three level 6 zones, there's a "pity" version of each NC, that will show up 5 turns after the last one. That is, in the heart, if you encounter ` "Moon Over the Dark Heart"` on your 7th adventure in the zone, you'll encounter `"Running the Lode"` on the 12th at the latest.

This adds tracking for when we last hit a relevant NC in each of these zones. 

Scripters or users can track their actual quest progress by checking the noncombatQueue of each zone (although a future PR could add some prefs for quest steps in each of these zones, and would piggyback nicely off of this), but this PR now gives them the ability to predict when their "pity" NC will show up.

Creating this as a draft because I am probably not going to write tests for this until late tomorrow, but I'd love to get some eyes on it in case this implementation is extremely silly.


src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -633,6 +639,42 @@ private static void handleFarmChange(final String location, final String respons
     }
   }
 
+  private static void handleFriarsChange(final String location, final String responseText) {
+    String[] elbowNCs = {
```
gausie, line 643: Bring these out to statics probably

src/data/defaults.txt
```diff
@@ -835,6 +835,9 @@ user	lastGuildStoreOpen	-1
 user	lastGuyMadeOfBeesReset	-1
 user	lastFratboyCall	-1
 user	lastFriarCeremonyAscension	-1
+user	lastFriarElbowNC	0
```
gausie, line 838: Why not -1?

src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -633,6 +639,42 @@ private static void handleFarmChange(final String location, final String respons
...
+      "Dodecahedrariffic!"
+    };
+    String[] heartNCs = {
+      "Moon Over the Dark Heart", "Running the Lode", "I, Martin", "Imp Be Nimble, Imp Be Quick"
+    };
+
+    switch (location) {
+      case "The Dark Elbow of the Woods":
+        if (Arrays.stream(elbowNCs).anyMatch((nc) -> responseText.contains(nc))) {
+          Preferences.setInteger("lastFriarsElbowNC", AdventureSpentDatabase.getTurns(location));
```
gausie, line 662: This is the encounter name right? I imagine we have a function somewhere to extract it from response text

src/data/defaults.txt
horrible-little-slime: I feel like that's a less informative number than 0; under normal circumstances, if `Location.get("the dark elbow of the woods").turnsSpent - get("lastFriarEblowNc") === 5`, you're due for a pity NC. Setting the default to -1 doesn't accomplish that.

src/net/sourceforge/kolmafia/session/QuestManager.java
horrible-little-slime: That was based on other code I saw in the area, but I'll look around

src/data/defaults.txt
gausie: The current default informs me that the last friar elbow NC was on turn zero, which is incorrect...

gausie: Your code snippet can just be

```ts
Location.get("the dark elbow of the woods").turnsSpent - Math.max(get("lastFriarEblowNc"), 0) === 5
```

I realise this is more awkward to code but I'd rather not make confusing prefs

src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -633,6 +655,27 @@ private static void handleFarmChange(final String location, final String respons
...
+  private static void handleFriarsCopseChange(final String location, final String responseText) {
+    String encounterName = AdventureRequest.parseEncounter(responseText);
+    switch (location) {
+      case "The Dark Elbow of the Woods":
+        if (Arrays.asList(friarElbowNCs).contains(encounterName)) {
+          Preferences.setInteger("lastFriarsElbowNC", AdventureSpentDatabase.getTurns(location));
+        }
+        break;
+      case "The Dark Heart of the Woods":
+        if (Arrays.asList(friarHeartNCs).contains(encounterName)) {
```
gausie, line 667: If you're always using Arrays.asList, just make it a List (or better, a Set) in the first place

horrible-little-slime: Updated them to lists using `List.of`; if there's a nicer setup, let me know, but I am easily bamboozled by the number of distinct types of iterable

gausie: Set.of would be even better because it's a non ordered lost of unique elements

gausie:
@horrible-little-slime What's the situation here?

horrible-little-slime:
Alright this is failing tests and it seems to be that my cleanups doesn't work because `AdventureSpentDatabase.setTurns` fails if you don't currently have the turns set for that particular zone. I feel like we can nix that behavior and the world will be okay? But I'd also be very happy with another solution.

src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -64,6 +62,25 @@ public class QuestManager {
...
+          "A Secret, But Not the Secret You're Looking For",
+          "Butter Knife?  I'll Take the Knife");
+  private static List<String> friarNeckNCs =
+      List.of(
+          "How Do We Do It? Quaint and Curious Volume!",
+          "Strike One!",
+          "Olive My Love To You, Oh.",
+          "Dodecahedrariffic!");
+  private static List<String> friarHeartNCs =
+      List.of(
```
gausie, line 78: One last thing - since we're very close and you have it loaded now I think these should be Sets not Lists if there's nothing major stopping you doing so

test/net/sourceforge/kolmafia/session/QuestManagerTest.java
```diff
@@ -4,19 +4,7 @@
...
-import static internal.helpers.Player.withGender;
-import static internal.helpers.Player.withHandlingChoice;
-import static internal.helpers.Player.withHttpClientBuilder;
-import static internal.helpers.Player.withItem;
-import static internal.helpers.Player.withLastLocation;
-import static internal.helpers.Player.withNextResponse;
-import static internal.helpers.Player.withPasswordHash;
-import static internal.helpers.Player.withProperty;
-import static internal.helpers.Player.withQuestProgress;
+import static internal.helpers.Player.*;
```
gausie, line 7: Don't do this please - your IntelliJ is probably set to do this automatically, but luckily it can also fix it automatically!

gausie (approved):

PR #1127, opened 2022-09-25
Fixing "Refresh menu" menu item.

Linknoid:
The "Refresh menu" menu item under scripts doesn't do anything anymore.  It only refreshes the backing list, not the actual UI.  This change makes it refresh the UI as well.


jaadams5 (approved):
Have not tested but...

PR #1128, opened 2022-09-25
Enable sorting candies by name in Sweet Synthesis grid.

Linknoid:
The Name column in the Sweet Synthesis grid is unsortable.  I assumed that was an oversight in the code, but it looks like it was intentional.  I can only assume this was done because the original author couldn't imagine anyone ever wanting to sort it by name.

But I have had multiple occasions where I wanted to sort it by name.  Usually for searching for a specific candy in the list.

This change simply removes the setSortable(0, false) from the Sweet Synthesis grid.


Veracity0:
They are sorted by price, low to high. The way it is intended to work:

Select a desired buff.
Column A fills with all candies which can contribute to it (low price to high price).
Select a candy in column A.
Column B fills with all candies which can combine with the selected candy to get the buff (low price to high price).

So, OK. You want to be able to find a candy in Column A, regardless of its price. Given that, sorting by name would do it.

The thing is, the actual expectation of how this will be used is this:

Select a desired buff.
Click the Check Prices button to make sure prices are up-to-data
Click the Automatic button, which will find the lowest price combination of one from Column A and one from column B, highlighting those two candies.
Click the Synthesize button which will get those two candies and combine them.

Is your intent to preferably use items you have in inventory?
Check the "available" checkbox and you will see only candies you have.

Is your intent to show all candies in Column A but manually select one you own and be willing to buy candies from Column B that combine with it, even though you may end up spending a lot more than if you are willing to buy either or both candies?

Well, I guess being able to search for a specific candy you own would be useful for that use case.

But you are certainly not using the frame like I expected it to be used!

Linknoid:
Sometimes it can be useful to determine which candy you need to buy to get a certain buff.  With a gnome moon sign, you can buy gum for sweet synthesis.  And if you have a stocking mimic, you can get a free candy on the first adventure, so it's worth checking if bag of many confections is relevant before wasting a familiar adventure.

jaadams5:
> Sometimes it can be useful to determine which candy you need to buy to get a certain buff. With a gnome moon sign, you can buy gum for sweet synthesis. And if you have a stocking mimic, you can get a free candy on the first adventure, so it's worth checking if bag of many confections is relevant before wasting a familiar adventure.

I'm not convinced.  This undoes something that was deliberate and discussed prior to implementation.  If @Veracity0 says No then I'll support that.  Otherwise maybe pause this to see if anyone else at kolmafia.us cares enough to chime in there?

Linknoid:
I forgot what I think is the primary use: choosing a trick-or-treat outfit and then finding out the role of the candy it generates.  Basically all the use cases boil down to determining the role of a known candy, and being able to find that candy in the list quickly.

gausie:
I don't see why not, it's a minor change and doesn't make sense with how I
use the tool but  if it works for you I'm happy to merge it

Veracity0:
Yeah. I can't imagine ever wanting to do this myself. Disabling sorting was a conscious design decision, as you can tell by the comment and the fact that it was done at all, but, whatever. If people don't want to do it, they can simply not click on that header.

As long as they can UN-do it if they did it accidentally.

PR #1129, opened 2022-09-25
No Relay Boss Warnings in Dinocore

Veracity0:
1) 'cause regular bosses are replaced with special monsters.
2) Add "turn in" link in relay browser to turn in dinosaur droppings to Dino World owner


midgleyc (approved):
Very nice!

PR #1134, opened 2022-09-30
Create per-character locks on login/logout.

heeheehee-kolmafia:
This prevents two sessions of Mafia trying to log into the same account, while also lifting the arbitrary session limit of 2.

Users who run many parallel sessions should be warned that this does not add any synchronization / safety primitives, in part because said changes would be impossible to implement without significant performance implications.

Very lightly tested manually.

I invite others to try out this change.


jaadams5:
Only skimmed.  More later.  Does Preferences.baseUserName always return something that would be legal when used in a file name?  I know there are raw usernames that KoL allows that have to be massaged to work as portions of a file name on all OSes.

Also I would be more comfortable keeping the arbitrary limit at 2.  We know that 2 can break the GLOBAL files and we are not doing anything to make that more robust.  My intuition from experiencing the problem is that it is most likely to occur when two accounts are starting up at the same time.  My choice would be to keep this at 2, find time to solve the file access issues and then raise it or just make it go away.

gausie:
I don't see a difference between 2 and uncapped personally, and would like to see the cap lifted in this PR

jaadams5:
I've done some research and experimentation.  Java supports "file locking" via Channel.  However what actually happens is OS dependent so that is probably not a robust way to do things.  I did implement a reader for GLOBAL_prefs.txt that did obtain a lock but I couldn't set up a manual test that actually showed a failure getting a lock because some other process had a lock on the file.  To continue in that direction would require making things more robust - primarily waiting and retying to get locks and changing mafia to handle failures in a more nuanced fashion.  I considered making the GLOBAL files a database or having some kind of inter-process communication so mafia instances on the same host could "coordinate" file access but ultimately I convinced myself that doing it right was a lot harder than I thought.

TLDR: Making access to GLOBAL files robust across instances is harder than I thought so it is no longer a justifiable reason to oppose raising or eliminating the session limit.

gausie:
Has anyone been trying this out? I've used it a few times but I don't often run two accounts at once.

jaadams5:
> Has anyone been trying this out? I've used it a few times but I don't often run two accounts at once.

Haven't tried lately.  I was thinking that as a Draft there was some uncertainty about whether it remained A Good Idea.  I play three characters daily so I'll try and run with this tomorrow.

gausie:
> A Good Idea

Well it definitely seems like a good idea, the question is whether it causes more issues than our current concurrent use setup.

jaadams5:
My usual routine is to run three characters with two simultaneously logged in.  When one of those finishes I log out and log in with the other.  I did all three simultaneously and didn't see any issues.  That may or may not mean much but at least there is one instance where the increased concurrency was not an issue.  It occurs to me I did not test the motivational case - trying to log in with a character who is already logged in.  Tomorrow?

jaadams5:
No problems today and the feature preventing a second log in was tested and worked.

If we want to increase the simultaneous limit then I think this is fine as is.  There is almost certainly an integer N such that things work for N simultaneous logins but break at N+1 but when a user finds it and tells us we just say Don't Do That and possibly embed the limit in the code.  Repeatable instances of file corruption when N was 2 are much rarer now and we fixed what we can so I am not especially concerned.

jaadams5:
I've been running with this and I both like it and have seen no problems.  It is hard for me to switch it from draft and eventually commit it because I do remember the days when the Dev team consistently rejected requests to change the login limit for fear of bot abuse.  If that is still a valid fear for others then maybe we should just close this.  If it is no longer (or never was) a rational or reasonable fear then let's commit.  I just don't want to make this decision by myslef.

heeheehee-kolmafia:
>  It is hard for me to switch it from draft and eventually commit it because I do remember the days when the Dev team consistently rejected requests to change the login limit for fear of bot abuse. If that is still a valid fear for others then maybe we should just close this. If it is no longer (or never was) a rational or reasonable fear then let's commit. I just don't want to make this decision by myslef.

Old thread with some discussion: https://kolmafia.us/threads/mafia-limits.2955/#post-20079

There's not much context in the original commit: https://github.com/kolmafia/kolmafia/commit/285ea9ba19329aa1dc91b9b83a23d34be6e802fc

Over time, we have relaxed a number of these stances (e.g. the ban on supporting Crimbo content to reduce load on KoL servers). I don't know that the conditions exist that people would actually want to run hundreds of meatfarming accounts (in such a way that they wouldn't get banned for multi abuse).

I don't see as much value in this specific stance vs, say, restricting mall API access for scripts (even if it's leaky, or can be patched out with a small code change). If someone really wanted to bot hundreds of accounts, they could just put the same jar into a bunch of folders... which would probably be safer anyways.

And again, this does nothing to address the preferences corruption issue.

I'll flip the bit from draft --> review to reflect that I'm in support, if nothing else.

jaadams5:
I have been running for a while and have not seen any file corruption.  It may be that my processes for running multis already avoid corruption as much as possible but if corruption is the blocker then I'm Ok with releasing this into the wild and seeing if someone can come with a repeatable use case.  If it does happen, we can suggest to people that they run no more than two characters which gets us back to the previous status quo.

jaadams5 (approved):

Veracity0:
> Over time, we have relaxed a number of these stances (e.g. the ban on supporting Crimbo content to reduce load on KoL servers).

FWIW, we did this when I reached out to Jick to get his thoughts and he actually responded. He said he appreciated our concern, but he didn't think there was any need for that any more. And that was that; we now treat Crimbo like any other "new content" now. Or  perhaps with even more urgency, since it is time sensitive, and we-all want to benefit from KoLmafia supporting it.

PR #1135, opened 2022-09-30
Convert PluralValues to arrays in Javascript

oxc:
The Effect.all value cannot be accessed in Javascript because it is a PluralValue, and that is not suppored by the ValueConverter:
```
> js Effect.get("Mallowed Out").all

Script    exception: ASH function returned native Java object.
```

Here is my try to map PluralValues to arrays in Javascript. It works as expected, however the Typescript definitions are now wrong. Since PluralValues use an AggregateType to specify their content, there is no good way to distinguish them from maps on a type level. 
Some advise here would be much appreciated.

I've started a discussion about this in summer: https://kolmafia.us/threads/pluralvalues-in-javascript-and-should-effect-all-be-one.27510/
@gausie seemed to support the idea of using Arrays in Javascript. I'm happy to try any other data model, if someone can think of something more appropriate (some frozen MultiValueSet or whatnot 😄).

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/JsRefCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/JsRefCommand.java
@@ -26,8 +26,11 @@ public class JsRefCommand extends AbstractCommand {
       return "any";
     }
 
+    if (type instanceof PluralValueType) {
+      return "readonly " + toJavascriptTypeName(((PluralValueType) type).getIndexType()) + "[]";
+    }
+
     if (type instanceof AggregateType) {
-      // FIXME: PluralValues are mapped to arrays, but there is no good way to distinguish them here
       if (((AggregateType) type).getSize() < 0) {
         return "{ ["
             + toObjectKeyType(((AggregateType) type).getIndexType())
--- a/src/net/sourceforge/kolmafia/textui/command/JsRefCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/JsRefCommand.java
@@ -27,6 +27,7 @@ public class JsRefCommand extends AbstractCommand {
     }
 
     if (type instanceof AggregateType) {
+      // FIXME: PluralValues are mapped to arrays, but there is no good way to distinguish them here
       if (((AggregateType) type).getSize() < 0) {
         return "{ ["
             + toObjectKeyType(((AggregateType) type).getIndexType())
--- a/src/net/sourceforge/kolmafia/textui/javascript/ProxyRecordMethodWrapper.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/ProxyRecordMethodWrapper.java
@@ -29,6 +29,13 @@ public class ProxyRecordMethodWrapper extends BaseFunction {
     try {
       Object returnValue = method.invoke(((EnumeratedWrapper) thisObj).getWrapped().asProxy());
 
+      // if the method returns a non-proxy Ash value (like Effect.all or Monster.attackElements),
+      // we need to convert it to a java object first
+      if (returnValue instanceof Value) {
+        ValueConverter coercer = new ValueConverter(cx, scope);
+        returnValue = coercer.asJava((Value) returnValue);
+      }
+
       if (returnValue instanceof Value
           && ((Value) returnValue).asProxy() instanceof ProxyRecordValue) {
         returnValue = EnumeratedWrapper.wrap(scope, returnValue.getClass(), (Value) returnValue);
--- a/src/net/sourceforge/kolmafia/textui/parsetree/AggregateType.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/AggregateType.java
@@ -62,9 +62,15 @@ public class AggregateType extends CompositeType {
   }
 
   // VarArg
-  public AggregateType(
+  protected AggregateType(
       final String name, final Type dataType, final int size, final Location location) {
     this(name, dataType, DataTypes.INT_TYPE, size, false, location);
+  }
+
+  // PluralValue
+  protected AggregateType(
+      final String name, final Type dataType, final Type indexType, final Location location) {
+    this(name, dataType, indexType, -1, false, location);
   }
 
   @Override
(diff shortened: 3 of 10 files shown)


gausie (changes requested):
Thank you! This will need tests to land.

If we can freeze the arrays coming out of `ProxyRecordMethodWrapper`, that would be ideal

src/net/sourceforge/kolmafia/textui/command/JsRefCommand.java
```diff
@@ -27,6 +27,7 @@ public String toJavascriptTypeName(final Type type) {
     }
 
     if (type instanceof AggregateType) {
+      // FIXME: PluralValues are mapped to arrays, but there is no good way to distinguish them here
```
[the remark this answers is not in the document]
oxc: Of course, that's why I mentioned it in the description. Can you give me a hint on how to recognize PluralValue from the AggregateType? 
I can't figure out a reliable way, that would not break other boolean maps.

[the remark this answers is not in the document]
oxc: I've tried to fix this with a separate PluralValueType that extends AggregateType.

src/net/sourceforge/kolmafia/textui/javascript/ProxyRecordMethodWrapper.java
```diff
@@ -29,6 +29,11 @@ public Object call(Context cx, Scriptable scope, Scriptable thisObj, Object[] ar
     try {
       Object returnValue = method.invoke(((EnumeratedWrapper) thisObj).getWrapped().asProxy());
 
+      if (returnValue instanceof Value) {
```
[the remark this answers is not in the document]
oxc: Done.

oxc:
I tried to freeze the returned arrays using what Rhino uses to implement Object.freeze. However, this seems to completely internal and cannot be used from the outside without going through the interpreter itself. 
Since the arrays are freshly created for each call anyway, I don't think it's worth the hustle. I have, however, marked them as ReadonlyArrays in the Typescript definitions.

oxc:
I've added a Test that spawns up a Javascript runtime and tests whether `Effect.get('Beaten up').all` is an array. It's not a very useful test, imho, and going through the whole Javascript engine does not feel appropriate.
However, there is no way to use the ValueConverter without a context, and generating possibly in a different way than the than the JavascriptRuntime does is not very useful either.

I've mentioned this once in the past: I don't feel it's appropriate to require contributors to write tests that should have been there in the beginning. I'm happy to extend any existing test to cover newly added code, and adjust breaking tests to reflect the new reality if a change is warranted. 
I am aware that this is a community project, of course. However, I don't feel equipped to basically spawn a whole new framework for testing things like ValueConverters, ProxyRecordMethodWrapper, and the likes. I don't have the necessary knowledge and experience about the inner workings of the kolmafia scripting system to create tests that test the prior state to ensure that my changes did not break anything.

Therefore, I would greatly appreciate some input, what you would like me to test here, and how I could set this up. From my POV, I would like some tests on how the ValueConverter handles certain inputs, but the JavascriptRuntime does not make it easy to create such a unit test.

midgleyc:
I would do these tests as e2e tests: you have some JS code, you have what you want that code to output: may as well test it at that level. I think that's the appropriate level of abstraction for this sort of thing: we don't care about precisely how the lower-level classes hook together: what we care about is the final output.

See e.g. `CustomScriptTest.java`, which has a bunch of ASH / JS scripts and what they output exactly, or `RuntimeLibraryTest.java`, which has a few places that assert like:
```
assertThat(output, endsWith("Returned: true\n"));
```

This is JS whereas most of "RuntimeLibraryTest" is ASH, so it's probably worth splitting out into either a nested class or a completely different class, but overall I think you want something like:

```
// set command to "js"
String output = execute("Monster.get('groar').attackElements");
assertThat(output, equalTo("""
Returned: aggregate string [1]
0 => cold
"""));
```

oxc:
Thanks for your feedback! I found that it made sense to add js tests to the existing `test_attack_element_proxy_fields.txt`. Does this sound reasonable?

gausie:
Sounds good to me

src/net/sourceforge/kolmafia/textui/parsetree/AggregateType.java
```diff
@@ -62,11 +62,17 @@ public AggregateType(final Type dataType, final int size) {
...
-  public AggregateType(
+  protected AggregateType(
       final String name, final Type dataType, final int size, final Location location) {
     this(name, dataType, DataTypes.INT_TYPE, size, false, location);
   }
 
+  // PluralValue
+  protected AggregateType(
+      final String name, final Type dataType, final Type indexType, final Location location) {
+    this(name, dataType, indexType, -1, false, location);
```
oxc, line 73: Just as some food for thought: I wonder if the PluralValueType should actually be more like an array, by knowing its size. 
I thought it would work out of the box, but of course the PluralValue has its dataType and indexType "mixed up".

oxc:
From my current POV, this PR is ready. Please let me know if you need any more changes.

gausie:
@oxc thanks! I want to add a quick step to the Typescript defs publishing that checks for valid TS generated. Once I've done that let's rebase this and then get it merged.

test/root/scripts/test_attack_element_proxy_fields.txt
```diff
@@ -4,3 +4,9 @@ ash $monster[Knott Yeti].attack_element
 ash $monster[Knott Yeti].attack_elements
 ash $monster[lihc].attack_element
 ash $monster[lihc].attack_elements
+js Monster.get('warwelf').attackElement
```
gausie, line 7: I'd like a test for each changed proxy record if possible. Also if you could instead make a new js file in this directory and add the tests to there that would be better I think.

Just print a bunch of stuff and make a .js.out file for what you expect to be printed. If you can make it human readable in some way, all the better (see test_path.ash for example)

gausie:
Ah I've just realised that my new workflow will only run on merge, not for PRs. I'll have to update it later and trust you that this makes a well-formed .d.ts file - did you check that?

test/root/scripts/test_attack_element_proxy_fields.txt
oxc: I can do that.

oxc:
I will generate the .d.ts again and run it through some kind of linter.

test/root/scripts/test_attack_element_proxy_fields.txt
oxc: I created a new test. I could not find a test that actually prints randomModifiers, I think those can happen only during gameplay, right?

oxc:
The generated .d.ts looks correct to me.

gausie (approved):

PR #1137, opened 2022-09-30
More work on Counter Warnings

Veracity0:
1) Test with counter = 7 for The Lower Chambers
2) Allow counter warning before fight in nemesis cave
3) Moved counter tests from RelayRequestWarningTest to TurnCounterTest, since they test both RelayRequest.sendCounterWarning() and GenericRequest.stopForCounters().
4) GenericRequest's getAdventuresUsed method can be overridden by subclasses. However, the Relay Browser uses RelayRequest objects and you can automate with GenericRequest objects. TurnCounter had code to look at URLs and call individual subclass methods to calculate the adventures used. I moved that code into GenericRequest's getAdventureUsed method and now TurnCounter can just call the request's method.
5) Have adventures Used tests for adventure.php and place.php cases.
6) Fixed Summoning Chamber available: only if MANOR quest is exactly "step3" - or "step2" with a wine bomb in inventory.
7) Support for:
DeckOfEveryCardRequest -> choice.php
LocketRequest -> choice.php
BeachCombRequest -> choice.php
CakeArenaRequest
CampgroundRequest (free rests)
CampAwayRequest -> place.php (free rests)
ChateauRequest -> place.php (free rests)
FalloutShelterRequest -> place.php (no free rests)
RichardRequest
SuburbanDisRequest
8) Barrel Full of Barrels is now a choice adventure, rather than being barrel.php.
Fixing the extensive existing code to adapt to that KoL change is a project beyond the scope of this.
However, we can handle adventure usage for choice adventures, so...
9) Support for:
volcanoisland.php
cellar.php
basement.php
mining.php
inv_use.php
inv_eat.php
runskillz.php
craft.php


libraryaddict:
The changes to chateau's free rests are not covered by tests.

```
Preference timesRested changed from 0 to 1
Preference timesRested changed from 1 to 15
```

PR #1138, opened 2022-10-01
fix: don't error on deleting shared folder for scripts

midgleyc:
Probably the issue on https://github.com/Loathing-Associates-Scripting-Society/ChIT/issues/24.

Suppose we have two scripts, both of which store something in the folder `images/relayimages`. Then deleting one script would fail, because the folder will fail to be deleted (as it's non-empty).

Check for a non-empty folder, and don't delete the folder in that case.

### DIFF
--- a/src/net/sourceforge/kolmafia/utilities/FileUtilities.java
+++ b/src/net/sourceforge/kolmafia/utilities/FileUtilities.java
@@ -16,6 +16,8 @@ import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
 import java.nio.channels.FileChannel;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.zip.GZIPInputStream;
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/utilities/FileUtilities.java
```diff
@@ -565,4 +567,14 @@ public static String getRelativePath(File home, File f) {
 
     return s;
   }
+
+  public static boolean isEmptyDirectory(Path path) throws IOException {
+    if (!Files.isDirectory(path)) return false;
+
+    var listFiles = Files.list(path);
```
jaadams5, line 574: Not certain of the syntax but I believe you.  Any particular benefit to this instead of checking the length of Files.list(path)?  I do note that Files.list() won't return the dot or dot dot references which is the only weird case I can think of.  Obviously not blocking.

jaadams5 (approved):

src/net/sourceforge/kolmafia/utilities/FileUtilities.java
midgleyc: `Files.list` returns a stream. Checking whether the stream has any values is faster than checking whether the length of the list is 0, because checking the length will enumerate every item in the list, which /could/ be slow (e.g. on HDD, or if a script somehow has hundreds or thousands of files).

PR #1141, opened 2022-10-03
The 6x6 Barrel Full of Barrels is dead. Long live the 3x3 replacement.

Veracity0:
As of August 25, 2015.


jaadams5 (approved):
Does the barrel sounds option work with the 3x3 or should we eventually track down and delete   barrel_sounds.js,  /images/barrel_smash.mp3 and related code?

Veracity0:
Barrel sounds still works fine. RequestEditorKit calls it for choice.php responses, not barrel.php

PR #1142, opened 2022-10-03
More fun fixes

Veracity0:
This is for things I discover/fix in my next Dinocore run

1) Log places we visit in Dino World rather than raw URLs
2) Can open Summoning Chamber with mortar dissolving solution.
Tests include using the HTML of doing that.
3) Can open Interior of Cobb's Knob with map and encryption key
Tests include using the HTML of doing that.
4) Can open Pandamonium by performing Friar ritual
I also verified that you don't even have to talk to the Friars first.
Tests include using the HTML of doing that.
5) I verified that you can adventure in the Fungal Nethers before looking at the Rubble in the Nemesis Cave.
Which is to say, Quest.NEMESIS at "step12" will do; no need to unlock via "step13".
6) On the Volcanic Island as an AT, you explore The Island Barracks. If you encounter a mariachi, you are given a link labeled "Fight!" with a peculiar URL. The page HTML has ```<form method="get" action="/fight.php?sinep=1664991973">```. But when the browser submits it, we get "fight.php?". That messes up request/encounter logging, which is unfortunate, since the response text actually has round 0 of the fight in it. Handle it. It now logs like this:
```
[743] The Island Barracks
Encounter: The Island Barracks
Took choice 409/1: Continue
choice.php?pwd&whichchoice=409&option=1
Encounter: A Short Hallway
Took choice 410/2: 3 keys or clues left, 0 keys located.
choice.php?pwd&whichchoice=410&option=2
Encounter: Hallway Right
Took choice 412/3: unknown / unknown / unknown
choice.php?pwd&whichchoice=412&option=3
Encounter: Parlour
Took choice 418/3: unknown
choice.php?pwd&whichchoice=418&option=3
Encounter: primitive chicken that consumed an alert mariachi
```
7) KoL no longer includes encounter name in Spooky Wheelbarrow. Bug reported, but we can work around it.
8) Flying in Circles does NOT take a turn; it is simply the transition between your Nemesis in the Volcano Cave and the demonic version.
9) Can talk to Trapper to open some zones.
Tests include using the HTML of doing that.
10) Fix KoLCharacter.freeRestsRemaining.


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -1668,7 +1668,7 @@ public static int freeRestsAvailable() {
   public static int freeRestsRemaining() {
     int restsUsed = Preferences.getInteger("timesRested");
     int restsAvailable = KoLCharacter.freeRestsAvailable();
-    return Math.min(0, restsAvailable - restsUsed);
+    return Math.max(0, restsAvailable - restsUsed);
```
jaadams5, line 1671: How long have we been missing this?

jaadams5 (approved):

src/net/sourceforge/kolmafia/KoLCharacter.java
Veracity0: Four days?

PR #1145, opened 2022-10-05
feat: allow specifying "github" dependencies

midgleyc:
`dependencies.txt` files can be used by scripts to specify necessary dependencies. At the moment, these dependencies are specified implicitly including the tool necessary to install them:
```
https://github.com/Loathing-Associates-Scripting-Society/combo/branches/release/ (SVN)
https://github.com/Loathing-Associates-Scripting-Society/combo.git release (git)
```

This PR introduces some new syntax
```
github Loathing-Associates-Scripting-Society/combo release
```

Dependencies specified with the new syntax will check the user's local install to see if the dependency is installed using either SVN or git and not install it if so (and otherwise install it using git). This means that it doesn't matter how the user initially installed the script, or what other scripts are using in their dependencies files.

The aim of this is to reduce the number of duplicate copies of scripts the user may install.

Scripts that use the folder install method are not supported. I figure that either:
* there is not a manifest file, so git install won't work, and SVN install should be used explicitly
* there is a manifest file, so the author intends the dep to be installed using git, and git install should be used explicitly

The main target of this PR is scripts that are on github, but still use the SVN install. There are competing philosophies for install:
* some users prefer to use git wherever possible
* some users prefer to use whatever the author recommends in the README

I hope that the new syntax will satisfy both.

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/ScriptManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/ScriptManager.java
@@ -88,26 +88,30 @@ public class ScriptManager {
       if (potential.startsWith("#")) continue;
       String[] args = potential.split("\\s+");
       if (args.length == 0 || args[0].length() == 0) continue;
-      var url = args[0];
-      if (args.length > 1 || url.endsWith(".git")) {
-        // git
-        String branch = args.length == 1 ? null : args[1];
-        var id = GitManager.getRepoId(url, branch);
-        if (!Files.exists(KoLConstants.GIT_LOCATION.toPath().resolve(id))) {
-          GitManager.clone(url, branch);
-        }
-      } else {
-        SVNURL repo;
-        try {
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
I didn't know we needed this ;-)

midgleyc:
When I initially designed the git scripts feature, I was expecting it to only be used for those github scripts which had issues with SVN, and authors with affected scripts would be motivated to change to recommending git for download to avoid support requests around failure to update (i.e. the very problem git was introduced to fix).

However, it's been adopted more generally. Script duplication has been an issue since the start, but "delete your SVN script and adopt the git one" is an easy fix for those scripts that adopt git. The issue that arises is that script preferences for direct downloads and script preferences for their dependencies may differ. I'm on the "whatever the author recommends in the README" side of the debate, so I didn't see this one coming ;)

PR #1146, opened 2022-10-05
chore: remove codeql

midgleyc:
The Java analysis has started failing with the error:

> rejecting SARIF, as there are more threadflow steps per result than allowed (45401 > 10000)

This looks like https://github.com/github/codeql-action/issues/1245. That ticket says that there should be a way to configure the action to eliminate the alerts that occur too much. However, I don't think we've ever got anything from this that wasn't a false positive (or a minor error that allowed a user (not a script) to create a file somewhere outside the Mafia directory on their own machine), so I'd be in favour of just removing it entirely.


jaadams5:
When we first stared using this we got several actionable warnings as well as a few that were deemed false positives or not going to fix.  I think we configured codeql to ignore those checks.  Rather than throwaway codeql could we tell it to ignore some more things?  "Polynomial regular expression used on uncontrolled data" is not an issue I am concerned about nor interested in fixing.

gausie (approved):
Let's remove this and replace with something that suits our needs if anyone finds such a thing. Anyone very interested can run these sort of analyses locally.

gausie:
> > Let's remove this and replace with something that suits our needs if anyone finds such a thing. Anyone very interested can run these sort of analyses locally.
> 
> I don't approve, but if you can provide instructions about running locally I can go grumble to myself in the corner. Making a branch, undoing the changes, publishing it to GitHub and then keeping it in synch with main does not fit my definition of "local".

You could run PMD locally

jaadams5:
> You could run PMD locally

Again, how?

PMD doesn't show up in a list of Gradle tasks.  Is it embedded in a non-obvious task?

If the instructions are to download something, install it and then run it perhaps you could commit to making running a gradle task?

I should also note that PMD might not be the right replacement.  PMD is a code quality tool.  CodeQL focuses on security issues.  

We could configure CodeQL to ignore some things.  See https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#using-a-custom-configuration-file  I thought @heeheehee-kolmafia had already done that for something else, but I can't find it quickly.

gausie:
You'd need to add PMD but it has a bunch of security checks. They're all just static analysis tools, codeql is just a weird one that GitHub is pushing. I don't think it's ever really helped us at all, other than give us busy work.

Veracity0:
I stopped looking at codeql output after seeing way too many "This Pattern has (\\d+) and you are parsing it into an int, but the data comes from a network read and since there are billions of web pages out there, you should be checking that it fits into an int. Otherwise, you can get a Exception". Considering that we are not reading from arbitrary web pages, just KoL pages, parsing itemIds out of URLs are just not going to give us overflows.

Making codeql happy about things like that is more than just busy work. It actively obfuscates our code for literally zero benefit.

jaadams5 (approved):
I give up ;-)

PR #1147, opened 2022-10-06
Allow automating one step at a time in the Volcano Maze in Relay Browser.

Veracity0:
> You wipe the sweat from your face as the tunnel ends at a massive lake of bubbling lava. In the distance, through the shimmering haze of the heat, you spot a small island. You can just make out the form of your Nemesis -- at this distance you can't tell what he's up to, but 'no good' is a safe bet. You'd better get there fast and put a stop to it, whatever it is.
>
> You survey the little stepping-stones that periodically rise and fall in the boiling magma, and realize that this is seriously going to suck. In fact, this is going to suck so much that I kind of feel like I ought to apologize in advance. Still, it wouldn't be a final battle without an especially fiendish final puzzle, now would it?

I remember when this came out. I'm not sure if Riff was expecting each person to manually solve it on graph paper each time, or something, but as expected, the community came to together, figured it out, and published the solutions.

The puzzle:

You have a 13 by 13 array of squares. Columns are numbered 0 to 12, from the left. Rows are numbered 0 to 12, from the top.
Only certain squares ("platforms") are available to stand on; the rest are bubbling lava.
Starting at 6,12 (the center square of the bottom row) you want to hop from platform to platform to reach 6,6 (the center square of the whole array).
The trick is that every time you hop to a new platform, all previously visible platforms sink into the lava and new platforms rise from the lava.

It turns out that the rising/falling of platforms is a cycle of 5.
It also turns out that Riff made six different sets of 5 maps.

Each of the 6 sets of maps has a single path the will take you from the start to the goal. If you stay on that path, you reach the goal. But if you step off that path, you are stranded. Therefore, KoL gives you chance to "jump" and swim through the lava back to the start square, where you can try again.

Some programmers (RoyalTonberry and me), coded up solvers, which would take a set of five maps and derive the (single) path from the starting spot to the end. Do that for all six sets of five maps and, voila! The puzzle is completely solved.

Immediately, people started publishing web pages showing the paths. RT was one such.
There were also one or two Greasemonkey scripts which loaded in all 30 maps and all 6 paths through groups of 5 and - somehow - decorated the map in the browser showing you where to hop.

My solution? I embedded my solver right into KoLmafia.

When you enter the Maze, we see Map #1 of a cycle and you see a Solve! button.
Click it and we  take 4 "random" steps into accessible square and see Maps #2-5
We then calculate the path from where you ended up to the center.
If we are on the main cycle, cool.
Otherwise we tell you to jump in and, when you are back at the starting square, we calculate the path from there.

There were complaints from users about having to jump in some times. "Why don't you just hardcode the 30 maps of data so you KNOW which path to use, just like the Greasemonkey script?" Why not? Because doing it my way future proofed the code against Riff adding sets of maps, which he was enthusiastically talking about at the time.

He soured on the idea because it was a LOT of work to come up with map-sets and people were quick to solve them. I asked him if KoLmafia's solver was a factor in this. He said "Not really; yes, it meant that if I introduced new maps, KoLmafia users would not even notice, but the the rest of the community solved and published solutions much more quickly than I'd expected. Too much work for me for something which remained a mystery for so little time."

-----------

Long intro done, what would I like to do?

I would really like the Solve! button to (optionally) actually show the animation:
1) make a hop
2) pass the response to the browser
3) wait a bit
4) If not done, back to step one
5) Otherwise, automation stops.

How to do this?

My first thought was that there is this META header stuff to ask the browser to "refresh" after an interval.
This is how the Solve! button works: it is attached to the following URL:

POST /KoLmafia/redirectedCommand?cmd=volcano+solve&pwd HTTP/1.1

What if it were connected to an internal "step" function of some sort which, when refreshed, submitted the next platform jump and send down the response?

Well ... the issue is that as coded, we submit the hops with &ajax=1 and the responses are JSON objects:

```
volcanomaze.php?move=5,12&ajax=1
{"won":false,"pos":"5,12","show":[3,6,10,14,18,23,26,30,32,41,43,50,52,57,59,60,64,82,84,94,97,102,106,109,111,114,115,117,119,129,136,145,148,153,154,157,164]}
```

Googling for how to get the browser to make requests came up with:

1) Use Ajax
2) Use META refresh and respond with updated data each time
3) Let the user control when refreshes happen by giving a button.

KoL is already using Ajax for this. Perhaps I could modify it to allow (optional) auto-refreshing?
I could stop letting KoL use ajax and just refresh the whole page. Ugly and slow.
I could just use a button.

For now, I am going for solution #3.

-----------

The plan:

1) I uploaded all the data for one "Solve" of the lava maze. HTML for looking at the maze, JSON for the hops, txt for the 5 properties that hold KoLmafia's representation of the maps, a list of all requests made during the solution, and so on.
2) Write a test for VolcanoManager.solve that uses all of those response texts and verifies that all the requests are submitted as expected.
3) Create VolcanoManager.step which will do the "next" step of the solution.
4) Refactor solve() to repeatedly call step(). My test from step 1 will be key. :)
5) Put two buttons on the Volcano Maze page:
Step - > calls a /KoLmafia sidecommand to connect to VolcanoManager.step() to make the next move and pass the JSON down to the browser.
Solve -> calls VolcanoManager.solve(), as always

Perhaps I need to look at the ajax code to make sure it doesn't depend on the "move" URL which was submitted. I suspect it does not, since the coordinates that are moved to are right there in the JSON.

-----------

The future:

There are other "Go to goal" or "Solve" buttons on puzzles.

- Dvorak's Revenge - the tile puzzle in the Hidden Temple
This, especially, would be fun to automate; I already log it with "Give me a B!",  "Give me an A!", etc. as if it you are a cheerleader at a pep rally - based on the "You put your pom-poms down" after the final step.
- The Chess puzzle in the Rabbit Hole
- The Louvre
- The Violet Fog

All of the above are HTML responses, so could be done via META refreshes, if we wished.

And the Volcano Maze... I don't know how Ajax works, but if it's all Javascript in the browser, all the code must have been downloaded. It would be interesting to study it (and Ajax) and figure out how it works. And maybe actually allow showing the automation, step by step, as KoLmafia does the platform hopping for you. My original goal.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
+++ b/src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
@@ -27,10 +27,10 @@ public abstract class VolcanoMazeManager {
   private static boolean loaded = false;
 
   // Java's Random takes a 48 bit seed.
-  public static long RANDOM_SEED = 0xe1d2c3b4a596L;
+  private static long RANDOM_SEED = 0xe1d2c3b4a596L;
 
   // We'll use our own RNG generator to make testing deterministic.
-  public static final Random RNG = new Random(RANDOM_SEED);
+  private static final Random RNG = new Random(RANDOM_SEED);
 
   public static void resetRNG() {
     RNG.setSeed(RANDOM_SEED);


Veracity0:
For amusement, here is the JS included in volcanoisland.php.
(It is all compact there, by removing spaces, but I let Emacs indent it the way it thinks JS should be.)
Isn't "$" a jquery thing?
I can see where it processes the JSON.
I imagine I could take the JSON KoL sends us and add an additional field, if the user wants to the browser to request updates and therefore show the animation, although I have no clue how to make the browser do that in this code; I know minimal Javascript and no jquery.

```
// <script type="text/javascript">

var uhohs = 0;
function setTitle(div, title) {
    var a = div.find('a');
    var cur = a.attr('title');
    a.attr('title', cur.replace(/- [^)]*\)/, '- '+title+')'));
}
$(document).ready(function () {
    $('#puzzle, .no').click(function (e) {
	if ($(e).hasClass('yes')) return false;
	if (confirm('Swim back to the start?')) {
	    document.location = '?jump=1';
	}
	return false;
    });
    $('.sq a').click(function () {
	var going = $(this).parent().attr('rel');
	$.getJSON($(this).attr('href') + '&ajax=1', function (res) {
	    if (res) {
		if (res.won) {
		    document.location = res.won;
		    return;
		}
		$('#you').remove();
		$('.lastyou').removeClass('lastyou');
		$('.you').removeClass('you').find('a').addClass('lastyou');
		$('.sq').each(function () {
		    t = $(this);
		    if (t.hasClass('goal')) return true;
		    var bg = t.find('a').css('background');
		    t.removeClass('yes');
		    if (bg && bg.match(/platformup[0-9]/) && t.attr('rel') != res.pos) {
			t.find('a').css('background','url("https://d2uyhvukfffg5a.cloudfront.net/itemimages/platformdown'+Math.floor((Math.random() * 4) +1)+'.gif?foo='+1+'")');
		    }
		    else t.find('a').css('background','');
		    if (!t.hasClass('no')) {
			t.addClass('no');
			setTitle(t, 'Lava');
		    }
		    if (t.attr('rel') == res.pos) {
			t.removeClass('no').addClass('you');
			setTitle(t, 'You');
		    }
		});
		for (i in res.show) {
		    if (!res.show.hasOwnProperty(i)) continue;
		    var sq = $('#sq'+res.show[i]);
		    if (sq.hasClass('goal')) continue;
		    sq.removeClass('no')if (!sq.hasClass('you')) {
			sq.addClass('yes').find('a').css('background','url("https://d2uyhvukfffg5a.cloudfront.net/itemimages/platformup'+Math.floor((Math.random() * 4) +1)+'.gif?foo='+1+'")');
			setTitle(sq, 'Platform');
		    }
		}
	    } else {
		uhohs++;
		if (uhohs > 3) {
		    document.location = '?';
		    return;
		}
		//alert('There doesn\'t seem to be a platform there.');
	    }
	});
	return false;
    });
});

// </script>
```

Veracity0:
The "Step" button (next to the Solve button) on the Volcano Maze page submits a "volcano step" comment.
KoLmafia calculates the next platform to jump to and returns a redirect to the appropriate volcanomaze.php?move=xxxx request.
The browser submits that, we relay the request/response to KoL and down to the browser,  and the browser updates the maze.

I could have done that by having KoLmafia submit the appropriate request, either with ajax (receiving JSON) or without (receiving HTML) and passing the response down to the browser.

My initial hope was I could ask KoL for the ajax JSON and send that down and the browser's ajax magic would animate the change in place.
That does not work; the browser displays the JSON.

My next hope was that the browser submitting the move request by itself - as it does when you step to a tile - would let it process a JSON response.

Nope.

I suppose I could have had KoL submit the non-ajax request and respond to the browser's request with that answer. THat might be worth trying. Same number of KoLmafia -> KoL request/responses (1) but only a single browser -> KoLmafia request/response.

In any case, I'm giving up on getting the ajax animation working. For now. That will require modifying the JS that I quoted above.

Veracity0:
Lastly, I did not refactor "solve" to call "step" X number of times, although it does the identical set of moves - assuming you don't take manual hops along the way.

"Solve" calculates the path of moves from where you are all the way to the goal and submits a sequence of requests to follow them.

"Step" calculates the path from where you are to the goal, takes the first one, and returns control to the user. If you hit "step" again, it calculates the path from your new location (which you could have changed by manually hopping platforms) and takes the first step.

Veracity0:
Huh. I may have to make Step let the browser submit the URL via redirection; as is, the other browser links which are relatve yo the page become relative to "/KoLmafia".

Veracity0:
I've been looking at the JS and I have things I can experiment with - in a followup PR.

The plaforms are like this:
```
<div id="sq7" class="sq yes  lv3" rel="7,0"><a href="?move=7,0" title="(7,0 - Platform)">&nbsp;</a></div>
```
And the thing that activates on a click is this:
```
$(document).ready(function () {
	$('#puzzle, .no').click(function (e) {
		if ($(e).hasClass('yes')) return false;
		if (confirm('Swim back to the start?')) {
			document.location = '?jump=1';
		}
		return false;
	});
	$('.sq a').click(function () {
		var going = $(this).parent().attr('rel');
		$.getJSON($(this).attr('href') + '&ajax=1', adjustPlatforms);
		return false;
	});
});
```
Note that I extracted the anonymous function called with a JSON result into 'adjustPlatforms'.

Seems like I could make my button - or equivalent - be like those divs, with a class with onclick handling, like those divs, to  submit "volcanomaze.php?step", which RelayAgent picks up just like "choice.php?auto", and dispatches to the volcanomaze step function to return the next JSON , which can be handled by that same Javascript function.

Or something.

As I said, in a followup PR.

Veracity0:
This should be good to go. The tests now call RelayRequest.reset, as appropriate, to avoid laking the static globals in that package.

src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
```diff
@@ -24,6 +26,16 @@
 public abstract class VolcanoMazeManager {
   private static boolean loaded = false;
 
+  // Java's Random takes a 48 bit seed.
+  public static long RANDOM_SEED = 0xe1d2c3b4a596L;
```
midgleyc, line 30: can be private?

src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
```diff
@@ -24,6 +26,16 @@
...
+
+  // We'll use our own RNG generator to make testing deterministic.
+  public static final Random RNG = new Random(RANDOM_SEED);
```
midgleyc, line 33: can be private?

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3194,7 +3198,7 @@ private void submitCommand(String command, boolean suppressUpdate, boolean waitF
 
   public void waitForCommandCompletion() {
     while (CommandDisplayFrame.hasQueuedCommands()) {
-      this.pauser.pause(500);
+      this.pauser.pause(50);
```
midgleyc, line 3201: hopefully this will improve relay perf elsewhere too :)

midgleyc (dismissed):
some comments, optional.

src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
Veracity0: Done.

src/net/sourceforge/kolmafia/session/VolcanoMazeManager.java
Veracity0: Done

src/net/sourceforge/kolmafia/request/RelayRequest.java
Veracity0: One hopes.

It means that any fast-executed side-command is detected more quickly.

The "step" test was painfully slow when simulating 67 side commands, with all the I/O going through our "fake" client. Essentially all the execution time was spent pausing. :)

Veracity0:
Since it was approved - with optional comments, which I did - I am not waiting for another approval.

PR #1149, opened 2022-10-06
Autumn-aton support

gausie:
Done

- Items added to data files
- Track ownership of autumnaton as it can fully leave your inventory
- Pref for upgrades
- Parse upgrades from visiting the choice
- Parse upgrades from installing new ones
- Track quests done today
- Track current quest location
- Track turn on which the autumnaton will return
- React to observed data to fix incorrect prefs
- Add location-specific +1 stat mod

Future work
- Refactor `withChoice()` as it is currently a bit of a mess. All tests should use either withChoice(choice, responseText) or withChoice(choice, decision, responseText), other choice cleanups should be made private, and split out to stop reusing code.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
@@ -224,16 +224,21 @@ public class JavascriptRuntime extends AbstractRuntime {
       returnValue = callback.get();
     } catch (WrappedException e) {
       Throwable unwrapped = e.getWrappedException();
-      if (stackOnAbort) {
-        StringBuilder message = new StringBuilder("Internal exception");
-        if (unwrapped.getMessage() != null) {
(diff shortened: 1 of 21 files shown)


src/data/spleenhit.txt
```diff
@@ -30,6 +30,7 @@ armored prawn	1	1	EPIC	0	30-50	30-50	30-50	50 Rushtacean' (+50 Init)
 [5140]astral energy drink	8	1	awesome	29-35	0	0	0
 [10882]astral energy drink	5	1		0	0	0	0	Lucky!
 astronaut ice-cream	1	7	EPIC	8-10	0	0	0
+autumn breeze	1	1		0	0	0	0	50 Delighted by Autumn (+100% all stats)
```
midgleyc, line 33: this gives stat points so should be Unspaded (until it gets spaded)

midgleyc (approved):

PR #1151, opened 2022-10-06
Allow Javascript to catch exceptions from ASH functions

oxc:
Exceptions thrown by a ASH library function cannot  be caught in Javascript. 

This PR changes the two function stubs to wrap any exception in Rhino's WrappedException (there is a helper function for that, `Context.throwAsScriptRuntimeEx(e)`).
The code that already catches the WrappedException is modified to also look at the unwrapped exception and handle ScriptExceptions the usual way.

I've added a Test as first commit before the change, to showcase the old behaviour, and make the change documented in code.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
@@ -224,16 +224,21 @@ public class JavascriptRuntime extends AbstractRuntime {
       returnValue = callback.get();
     } catch (WrappedException e) {
       Throwable unwrapped = e.getWrappedException();
-      if (stackOnAbort) {
-        StringBuilder message = new StringBuilder("Internal exception");
-        if (unwrapped.getMessage() != null) {
-          message.append(": ").append(e.getMessage());
-        }
-        message.append("\n").append(e.getScriptStackTrace());
-        String escapedMessage = escapeHtmlInMessage(message.toString());
+      if (unwrapped instanceof ScriptException) {
+        String escapedMessage = escapeHtmlInMessage("Script exception: " + unwrapped.getMessage());
         KoLmafia.updateDisplay(KoLConstants.MafiaState.ERROR, escapedMessage);
+      } else {
+        if (stackOnAbort) {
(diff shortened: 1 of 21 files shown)


gausie:
This awesome. Is it possible to cover more of the code routes for exceptions with tests? Looking at coverage there are a bunch that are not covered

oxc:
Yes, I noticed. I will try today 👍🏻

oxc:
I'm afraid I can test a user defined ASH function without a more elaborate setup. Any ideas?

oxc:
I've added several tests that document how certain exceptions are displayed when caught and when uncaught. As you can see, the changes also improve the display of ScriptExceptions that could have already been caught by Javascript (e.g. "bad effect value").

Is it possible to keep both commits in the history, and not squash them? I believe this would help to understand what has actually changed here.

oxc:
It seems I cannot cover the `throw` statements, because the `Context.throwAsScriptRuntimeEx` already throws. I will add istanbul comments to ignore them.

oxc:
Ok, so JaCoCo has no way of ignoring individual lines. Should I remove the comments again? At least they explain why the lines cannot be covered.

oxc:
Ok, I will remove them. The throw statement is not covered because it never executes. The function already throws an exception, so the throw in our java files only serves flow analysis, but is never run.

midgleyc:
Sorry, I don't quite understand: are you saying that `ashFunction.executeWithoutInterpreter` throws an exception which is not caught? What is the point of the catch block then?

Or are you saying that `Context.throwAsScriptRuntimeEx` itself throws, and does not return an error to be thrown?

oxc:
> Or are you saying that `Context.throwAsScriptRuntimeEx` itself throws, and does not return an error to be thrown?

This exactly. That's why the throw statement in _our_ code is not covered.

gausie:
Simulating a throw from an ash file shouldn't be too hard I think, just need to add an ash script that has a function that throws and require it.

oxc:
Yeah, you're right, I already figured that out in the meantime 😄 

This PR is ready from my POV, please let me know if you want anything changed.

gausie (approved):

PR #1152, opened 2022-10-06
fix: move index.d.ts to right place before validating

midgleyc:
Fix Typescript definition validation, take 2.


gausie (approved):
Nice, makes sense

PR #1155, opened 2022-10-07
feat: autumnaton command

midgleyc:
Doesn't check for valid locations: just sends you there, then checks for an error message.

Nice improvements would be to list the possible locations (and check against them), but that might be better suited to ASH.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/AutumnatonCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/AutumnatonCommand.java
@@ -39,14 +39,11 @@ public class AutumnatonCommand extends AbstractCommand {
     KoLmafia.updateDisplay("Sending autumn-aton to " + advName);
 
     GenericRequest request =
-        new GenericRequest("inv_use.php?pwd&which=3&whichitem=" + ItemPool.AUTUMNATON, false);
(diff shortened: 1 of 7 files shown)


horrible-little-slime:
I would love to have the option to install upgrades as we send the little fella off. Or to do it separately!

That being said, I assume default behavior for most users will be that they want to install any upgrades they have available. It's really only for spading reasons that you might not want to. That can be handled by running choice option 1 beforehand, if it is available.

jaadams5 (dismissed):

src/net/sourceforge/kolmafia/textui/command/AutumnatonCommand.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    var advName = adventure.getAdventureName();
+    KoLmafia.updateDisplay("Sending autumn-aton to " + advName);
+
+    GenericRequest request =
+        new GenericRequest("inv_use.php?which=3&whichitem=" + ItemPool.AUTUMNATON);
+    RequestThread.postRequest(request);
+
+    request.constructURLString(
```
gausie, line 45: You probably want to check hasSnarfblat too

PR #1156, opened 2022-10-07
Testing Dvorak's Revenge automation

Veracity0:
Executive summary:

I tried to automate the Hidden Temple puzzle-chain in Quantum Terrarium Relay Browser.
I had lots of glitches because KoLmafia kept trying to look at the Quantum Terrarium for some reason and you can't do that in the middle of choice.php or tiles.php, so it redirected and confused the automation.

I want to write tests to see if it is possible to automate it in a normal path.

To start, I have written tests for verifying automating the Dvorak's Revenge puzzle, both via the "Solve" button in the relay browser, and when encountered via Adventuring in the Hidden Temple - presumably because choices automated sufficiently to get you there?

I should have enough saved DEBUG log to verify that in a normal path - and also to debug the issues with QT.

----------------------------------
1) Using DEBUG logs from automating the tile puzzle, I created a test for running the "dvorak" via the "solve" button in the Relay Browser.
This required a minor adjustment to RelayRequest.submitCommand to extract the "wait until the command queue is empty" code so that I could use it when simulating the browser submitting a button which runs a command.
2) I created a test for adventure.php redirecting to tiles.php. I guess this could happen, somehow, if you did "adventure" (or "adv1") when your character just happened to be sitting at the end of the first puzzle in the temple? In any case, AdventureRequest handles that by calling its own method which submits the same requests as DvorakManager.solve().
Why the duplicated code? Why not simple call DvorakManager.solve()?
3) I started writing a test about why choice automation fails when started at the first puzzle. It's because taking a choice results in a response which which has a mandatory link to go to choice.php, but which is not a choice itself.
Taking that link will redirect to tiles.php, but ChoiceManager doesn't take it; it fails with an ERROR message.
4) I fixed choice automation in the Hidden Temple such that when it sees the mandatory link to choice.php after the first puzzle, it takes it.
It then calls DvorakManager to automate the tile puzzle.
And when that completes, it handles the choice for the final puzzle
5) I fixed the Quantum Terrarium issue: GenericRequest calls QuantumTerrariumRequest after every single request to see if it needs to look at the Quantum Terrarum. That deferred doing that for fights and choices - non-interruptable - but not for tiles.php - similarly non-interruptable. Now it defers in that case, too.
6) AdventureRequest now calls DvorakManager.solve() to automate, rather than duplicating code.
7) I got the HTML for using a wine bomb and added a validation test that prepareForAdventure will use it.


test/net/sourceforge/kolmafia/session/RelayAutomationTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+public class RelayAutomationTest {
+
+  @BeforeAll
+  public static void beforeAll() {
+    // Simulate logging out and back in again.
+    GenericRequest.passwordHash = "";
+    KoLCharacter.reset("");
+    KoLCharacter.reset("relay automation");
+    Preferences.saveSettingsToFile = false;
```
midgleyc, line 46: This is handled in `ClearSharedStateBefore` and shouldn't be necessary in each test.

midgleyc (dismissed):
One optional comment.

test/net/sourceforge/kolmafia/session/RelayAutomationTest.java
midgleyc: To be more specific, the `saveSettingsToFile` change is handled in that class (among other things I haven't checked).

Veracity0: Huh. I have not looked at that before. I see that it does the following:
```
   GenericRequest.passwordHash = "";
   KoLCharacter.reset("");
   Preferences.saveSettingsToFile = false;
 ```
I notice that the ```Preferences.saveSettingToFIle``` is NOT restored to true (the default) in the "After" extension.

I think that
```
    KoLCharacter.reset("relay automation");
    Preferences.reset("relay automation");
```
Need to be done by individual test files in BeforeAll and BeforeEach respectively.

Now that I have studied (a bit) the extension, I'll remove the 3 unnecessary lines and will pay attention in the future.
Although, there are plenty of test files that could be cleaned up. And also be djusted to use CleanUps. :)

Veracity0: It is done.

midgleyc (approved):

PR #1157, opened 2022-10-08
Update combats.txt

jaadams5:
See https://kolmafia.us/threads/invalid-adventure-area-the-barrel-full-of-barrels.28134/

My local tests now fail enumeratedTypesAreCaseInsensitive() apparently because the barrels are no longer valid as an enumerated type.  

I'm not sure this is the right fix and it certainly is not a data file integrity test for combats.txt but since AFAIK the same test fails locally for me but not in the GitHub workflow it undermines my trust in the workflow.

### DIFF
--- a/src/data/combats.txt
+++ b/src/data/combats.txt
@@ -238,7 +238,6 @@ The Ancient Hobo Burial Ground	-1	Spooky hobo	Zombo: 0
 The Archwizard's Tower	0	Archwizard: 0
 The Arid, Extra-Dry Desert	100	cactuary	giant giant giant centipede	plaque of locusts	rock scorpion	swarm of fire ants
 The Bandit Crossroads	100	fantasy bandit
-The Barrel Full of Barrels	-1	tiny barrel mimic	small barrel mimic	medium-sized barrel mimic	huge barrel mimic
 The Barrow Mounds	100	barrow wraith?
 The Bat Hole Entrance	100	albino bat	pine bat	regular old bat
 The Batrat and Ratbat Burrow	100	batrat	ratbat	screambat: 0	Batsnake: 0


jaadams5:
Checking case insensitivity for $location ==> expected: <Returned: true> but was: <Invalid adventure area: "The Barrel Full of Barrels"
Returned: true>

heeheehee-kolmafia:
I can't reproduce when running tests normally, although I can with
```shell
$ ./gradlew test --tests=CustomScriptTest.enumeratedTypesAreCaseInsensitive 
```
This points to your specific tendency of running the tests with `forkEvery = 1`.

The message you report is from a method of AdventureDatabase that's only invoked from a static block. Thus, in the "typical" scenario, this error message is most likely printed out (and ignored) in the first test case that causes AdventureDatabase.class to be loaded.

jaadams5:
> I can't reproduce when running tests normally, although I can with
> 
> ```shell
> $ ./gradlew test --tests=CustomScriptTest.enumeratedTypesAreCaseInsensitive 
> ```
> 
> This points to your specific tendency of running the tests with `forkEvery = 1`.
> 
> The message you report is from a method of AdventureDatabase that's only invoked from a static block. Thus, in the "typical" scenario, this error message is most likely printed out (and ignored) in the first test case that causes AdventureDatabase.class to be loaded.

I'm going to politely disagree if only because I am not running with fork=1.  I did so when there was test leakage that we were trying to plug, but since it increases the wall clock run time of testing on my system by a factor of 3 or more I no longer do it casually.

jaadams5:
Because I have been known to make mistakes...

Updated to 26827

Ran gradlew jacocotestreport

`Microsoft Windows [Version 10.0.19044.2006]
(c) Microsoft Corporation. All rights reserved.

C:\KoLmafia>gradlew jacocotestreport
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details

> Task :compileJava
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :compileTestJava
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.

> Task :test

CustomScriptTest > enumeratedTypesAreCaseInsensitive() FAILED
    org.opentest4j.AssertionFailedError at CustomScriptTest.java:105
Java HotSpot(TM) 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended

3675 tests completed, 1 failed, 31 skipped

> Task :test FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///C:/KoLmafia/build/reports/tests/test/index.html

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 3m 52s
6 actionable tasks: 3 executed, 3 up-to-date

C:\KoLmafia>`

jaadams5:
I did shadowjar, clean and then jacocotestreport and this time the test passed locally.  Trying to rerun immediately does nothing.

heeheehee-kolmafia:
My recollection is that there is a secondary effect at play where Gradle's JUnit test runner will run failing tests at the start of an execution, which will exacerbate this particular failure mode. (I can't find documentation of this, offhand.)

IMO the problem is that the test case is fragile, as it depends on the output matching exactly, up to any side effects such as informational log messages. One possible alternative would be to move this into textui/parsetree/DataTypesTest.java and instead check that

```java
assertEquals(
    DataTypes.parseValue(type, firstValue.toUpperCase(), true),
    DataTypes.parseValue(type, firstValue.toLowerCase(), true));
```

PR #1158, opened 2022-10-08
Profiler Test

jaadams5:
The class Profiler now has 100% coverage, up from 0%.

This is basically a CustomScriptTest with two changes.  One is that it uses profile to invoke the script.  The other is that the output cannot be predetermined because the profile timing depends upon the system running the test.

I have stuck the script file in relay but that is not an acceptable solution.

So it is Draft while I get feedback.  

I either need to find an expected output file that will be the same if items change or if the timing on the test system changes

OR

I need to find a place in ROOT where the script can live but will not be run as part of CustomScriptTest.

I'm also open to stronger assertions.


gausie:
Genuine question: how useful is this test? What does it show us? Does it show the profiler works as desired?

jaadams5:
Basically coverage++ 

Every other benefit comes as personal professional development :-)

I do have a better solution for script location but won't test and push today.

gausie:
I think it's probably technically a bad idea to increase coverage without adding any benefit - you could achieve the same by ignoring the file from coverage checking. In fact in that situation, we'd have somewhere explicitly reminding us that this class isn't actually covered!

On the other hand I can't think of a good way to actually test this, so perhaps just go for it

jaadams5:
Philosophically I'm OK with increasing coverage for its own sake.  But I also agree that if the resources used to do this were in some sense managed, this would not be how I'd allocate them "for the good of the community".  Perhaps there is an analogy to writing defensive code for file i/o when the file is under configuration control and there are tests for its integrity.

Anyway I did figure out how to put a script in scripts so it could be test executed but was not suitable for CustomScriptsTest.

jaadams5:
And I note the test success on Windows but not elsewhere.  Revert to draft until I figure that out.

Veracity0:
Philosophically, I am also opposed to coverage-for-the-sake-of-coverage.

"I called this method - and didn't bother to look at return values or side effects, so did it do the right thing? Who knows? But our tests now cover the lines that were executed, so, win!"

I want tests which actually test the functionality of the code being executed.

That said, I have written many a test for code which called hitherto un-executed-during-test methods in other modules, so coverage goes up more than expected - although diff coverage did not touch the new code being executed. I guess if the code I tested depended on correctness from the other modules it called (as it surely must), that's good indirect testing.

But simply calling methods for the sake of executing them without doing some legitimate testing is pointless.

jaadams5:
This remains philosophy until the OS issue that makes the test fail is resolved.

Philosophically, if I had said we have a low profile tool that adds output to the run of the script.  This test confirms that the script runs with the tool and output that is only generated by the tool is present, would we be having a different discussion?

The profile option runs a script, gathers some timing data (that is environment specific) and appends it to the script output.  A reason to use it is when there are two different scripts to so something and one might be faster than the other.  If I wanted to test that assertion I would create two scripts, run the profiler on both and observe which one was faster.  Ideally that would be the one I had written to be faster.  So the test would run both, confirm expected output and then make assertions about which times were higher (slower).  Is that a more useful, "does something" test?

jaadams5:
Added print to failing test

Unable to invoke /Excluded/CountItems.ash

Not sure why, but it is a data point.

jaadams5:
It seems like all of the angst about failing tests was programmer error.  /Exclude/CountItems.ash is an absolute file name on *nix but not Windows.  Thus it is not a valid script.   Exclude/CountItems.ash however is just fine.

jaadams5:
Didn't intend to close.

When KoLmafiaCLI.findScriptFile() gets tests some sensitivity to absolute files across OS might be of interest.

jaadams5:
Tests working.  Do we keep this or close it?

midgleyc (approved):

midgleyc:
Seems reasonable to me to have it in. It tests something meaningful.

PR #1159, opened 2022-10-09
Fix boris day

libraryaddict:
Fixes https://kolmafia.us/threads/possible-consumables-database-error-prevents-mafia-from-starting.28137/


Veracity0:
Good call, but I just submitted a PR with the identical fix that also cleaned up unnecessary Integer to int conversions.

PR #1162, opened 2022-10-09
Revert rest change to ChateauRequest.

phulin:
A change in #1137 ([here](https://github.com/kolmafia/kolmafia/commit/db6f43f0562286bc9eb2013c509663fc1052221b#diff-4e27d487224bb81a99bdc43788d64468183b9e44d789954232042a5c3062c9b3L116-R115)) caused Chateau rests to register incorrectly if users click on some parts of the bed image in the Chateau.

In particular, clicking on the image of the bed triggers a request with `action=chateau_restbox` no matter whether the rest is free or not, so clicking on the bed will incorrectly set free rests to the maximum. Clicking on the "Rest in your Bed" text uses `action=chateau_restlabelfree` if the rest is free.

@Veracity0 Flagging that this reverts a change from a few days ago - not sure why that change was made originally.


phulin:
Will write a test for this... stay tuned.

src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -1394,11 +1394,11 @@ public void externalExecute() {
   public static final boolean shouldIgnore(final GenericRequest request) {
     String requestURL = GenericRequest.decodeField(request.formURLString);
     return requestURL == null
-        ||
+        //        ||
```
Veracity0, line 1397: Absolutely not. Remove your private change. :)

test/net/sourceforge/kolmafia/request/ChateauRequestTest.java
```diff
@@ -0,0 +1,68 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class ChateauRequestTest {
+  @BeforeAll
+  public static void beforeAll() {
+    KoLCharacter.reset(true);
+    KoLCharacter.reset("ChateauRequestTest");
+    Preferences.reset("ChateauRequestTest");
+    Preferences.setBoolean("chateauAvailable", true);
```
Veracity0, line 22: Do these using withProperty() Cleanups

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/GenericRequest.java
gausie: Luckily this will cause a test I added to fail

Veracity0: Absolutely not. You may not submit these private changes.

test/net/sourceforge/kolmafia/request/ChateauRequestTest.java
Veracity0, line 22: Do these using withProperty() Cleanups.

Veracity0 (changes requested):
The test failures were looking right at your private changes. ;)

src/net/sourceforge/kolmafia/request/GenericRequest.java
phulin: usually I don't make this mistake... thank you for adding that test, Sam :)

phulin: Fixed.

test/net/sourceforge/kolmafia/request/ChateauRequestTest.java
phulin: Made an attempt, let me know if this is the pattern you were thinking of.

Veracity0: Here's what I would do.

ChateauRequestTest is for testing everything about the Chateau.
You are testing free rests in particular.

The pattern I would like is this:

```
public class ChateauRequestTest {

  @BeforeAll
  public static void beforeAll() {
    KoLCharacter.reset("ChateuRequestTest");
    Preferencesr.reset("ChateuRequestTest");
  }

  @BeforeEach
  public void beforeEach() {
    // stuff
  }

  @Nested
  class FreeRest {

    public Cleanups propertyCleanups() {
      // ... stuff that applies to free rests
      return cleanups;
    }

    @Test
    public void test1() {
      var cleanups = new Cleanups(propertyCleanups());
      try (cleanups) {
        // test code
      }
    }
  }
}
```

gausie (dismissed):

test/net/sourceforge/kolmafia/request/ChateauRequestTest.java
phulin: Restructured the test - I think this is the right way to do it. Let me know if you disagree.

phulin: (And thanks for the help - it's been at least a decade since I used JUnit...)

Veracity0: Very nice!

Veracity0 (approved):

PR #1163, opened 2022-10-09
Using the Autumn-Aton removes it from inventory

horrible-little-slime:


jaadams5 (approved):

gausie (commented):
This is quite easy to add a test for (just add to the existing test for sending him off) so would I'd appreciate if you could do ❤️

gausie (approved):

PR #1165, opened 2022-10-10
Allow taking one step at a time in the Rabbit Hole chess puzzle

Veracity0:
The goal is to have a "Step" button which will submit "chess step" which will redirect to taking the calculated "correct" next step for solving the chess puzzle so you can actually watch what happens, step by step.

1) Add Step button to the chess puzzle

### DIFF
--- a/src/net/sourceforge/kolmafia/session/RabbitHoleManager.java
+++ b/src/net/sourceforge/kolmafia/session/RabbitHoleManager.java
@@ -1123,7 +1123,7 @@ public abstract class RabbitHoleManager {
   }
 
   public static final void decorateChessPuzzle(final StringBuffer buffer) {
-    // Add a "Solve!" button to the Chess Board
+    // Add a "Step" and a "Solve!" button to the Chess Board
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/session/RabbitHoleManagerTest.java
```diff
@@ -0,0 +1,233 @@
...
+
+  @Nested
+  class Chess {
+    @AfterEach
+    public void afterEach() {
+      InventoryManager.resetInventory();
+      RelayRequest.specialCommandIsAdventure = false;
+      RelayRequest.specialCommandResponse = "";
+      RelayRequest.specialCommandStatus = "";
+      RelayRequest.redirectedCommandURL = "";
```
midgleyc, line 50: Might be worth making a `RelayRequest.reset()` to group these, given they'll be wanted after each test interacting with the relay request.

midgleyc (approved):

midgleyc (approved):

PR #1171, opened 2022-10-11
feat: time tests and test methods

midgleyc:
Not sure we actually want this to go in (perhaps behind a flag, if that's possible?) but I've noticed that sometimes the tests here on GitHub take a very long time, and I'm curious as to which test is taking up the time.

Coverage might be slowing it too, so perhaps we only want to cover on Ubuntu?


gausie:
We could add it as a separate workflow that runs on dispatch or whatever the event is that lets you run it on button press

Veracity0:
Considering that the tests for this have been stuck on Windows forever, this is timely.

You may be able to answer your question with this run. 😀

jaadams5:
I routinely work in Windows booting up a Debian VM if I really have to.

Tests take less than six minutes wall clock time.  I haven't seen anything take more than that since we cleaned up the TokenServer(?  fredg's work) tests.

I just ran this locally.  Still under 6 minutes.  I haven't found where the output went and I didn't have a large enough scroll buffer but the last FightRequest test (Commerce Ghost or the next one) clocks in at about 24 seconds.  There are tests in FileMonitortest that take 15s, 6s, and 22s.  Something in StateCheckWrappersTest taks about 8s.  Doesn't seem to be much else over 5s.

I will look at GitHub output and mine and see if the patterns are similar, probably tomorrow.

I will remind folks that the root cause could be one or more tests or it could be something GitHub is doing with Windows that gives a different environment.

midgleyc:
No major investigation but some odd stuff.

For example, `PreferencesTest` took 231590 ms. Totting up the individual methods gives 62 ms.

So a lot of time must be spent in before and after each (or just by JUnit getting ready to run things, I guess):

```java
  @BeforeEach
  public void initializeCharPrefs() {
    KoLCharacter.reset("fakePrefUser");
    KoLCharacter.reset(true);
    Preferences.saveSettingsToFile = false;
  }

  @AfterEach
  public void resetCharAndPrefs() {
    KoLCharacter.reset("");
    KoLCharacter.reset(true);
    KoLCharacter.setUserId(0);
    Preferences.saveSettingsToFile = false;
  }
```

Now this resets the character twice while it shouldn't need to, but at 231 seconds for 35 tests, this is about 6 seconds each, which still seems like rather a lot.

I noticed that in `FightRequestTest`, the nested Lovebugs test sets `Preferences.saveSettingsToFile` to true for the rest of the tests in that class. I filed https://github.com/kolmafia/kolmafia/pull/1172 thinking it might be slow file system interaction, but some filesystem-heavy tests seem completely fine, so I don't think it's that overall.

jaadams5:
All this was done without recent updates.

These tests took more than 5 seconds both locally and on GitHub.

[KoLAdventureValidationTest]
[ModifierExpressionTest]
[RestoreExpressionTest]
[MaximizerTest]
[PreferencesTest]
[FightRequestTest]
[FilesMonitorTest:filesMonitorTest]
[FilesMonitorTest:filesMonitorNewFileTest]
[FilesMonitorTest]
[StateCheckWrappersTest]

There are at least 100 tests that exceeded 5 seconds on GitHub but not locally.  There are also a boatload that were an order of magnitude faster locally than on GitHub.

Not sure what any of this means :-)

midgleyc:
I can't see a way to easily make this flaggable. I'm currently using a specific file to apply this to all tests, and I don't think it's possible to either tell JUnit to use a different file under some circumstances or selectively change the contents of the file.

If this always runs, the only effect is that the output fills with messages you're not necessarily interested in.

The outputs should ideally be simply the time and class / method, but I think JUnit wraps the log in a note of where it came from.

jaadams5:
The output I would tweak seems to be added by JUnit.  The extra clutter is most likely to matter when a test fails and some work has to be done to see which one.  Gradle puts out an elapsed time and this adds about 2 minutes to the overall test task on my system.

Those are all things I could live with if someone wanted to commit this.  But I could also see capturing this as a patch file and applying it locally or as a new PR if someone wanted to do something with the timing data.

I guess at the moment I can accept someone else's preference/decision.

gausie:
What's the status here?

jaadams5:
>I can't see a way to easily make this flaggable. 

I think we are waiting to see if anyone can make this flaggable.  I like the idea but would like to be able to turn it on or off because it does take some time locally.  

My perfect solution would be to have a flag that allowed me to turn it on or off, locally and on or off on GitHub.  I don't really care if it is always on on GitHub but I do want to toggle whether it runs locally or not.

As it stands not pushing this PR gives us that.  If we want to turn it on for GitHub we update to main and then run.  If I (we?) want it locally I just grab this branch.

midgleyc:
If it's not going to be merged in I could close the PR but keep the branch alive, then you can cherry-pick it into your branches, run to check, then delete it after. That's what I've been doing.

The issue with making it flaggable is that I'm using [automatic registration](https://junit.org/junit5/docs/current/user-guide/#extensions-registration-automatic) and that requires a particular file. I'm wondering if I knew gradle better if I could add something to some test run call that read an environment variable and modified the file to add / remove the line registering this one. But I also want it to run from inside my IDE.

midgleyc:
perhaps I could always register the class, but only have it do anything if a particular env var is set?

jaadams5:
I just ran locally.  8+ minutes without this and 9+ minutes without.  Close enough that I won't use timing as a reason to insist it be switchable.  It is tremendously verbose but I think that can be adjusted to.

midgleyc:
I think, given that switching it on or off is always going to require some change, the easiest thing to do is to delete the registration line in `test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension` if you don't want it to run.

Options like "create another gradle task for this" don't satisfy my desire to be able to run this on an individual test from my IDE.

jaadams5:
> I think, given that switching it on or off is always going to require some change, the easiest thing to do is to delete the registration line in `test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension` if you don't want it to run.

I think this means we are in agreement??

jaadams5 (approved):

PR #1173, opened 2022-10-12
feat: location level

midgleyc:
The Autumnaton returns seasonal items according to location "level", which on the kol side is a particular tag in "special": "highlevel", "midlevel" or "lowlevel". No tag is interpreted as "midlevel".

"Level" was already used for water level, so I went for "DiffLevel".

I didn't want to use "difficulty" to avoid confusion with the other difficulties (water level, turtle taming, barrel god) which all use the stat requirement.

While I used a string, I'm imagining it to only have the values "unknown", "low", "mid", or "high". Similarly, env is "none", "unknown", "indoor", "outdoor", "underground", "underwater". It might be worth swapping them out for enums at some point.


src/data/adventures.txt
```diff
@@ -7,696 +7,696 @@ Campground	adventure=543	Env: outdoor	Your Mushroom Garden
 
 # Clan areas
```
jaadams5, line 9: Does this change the version of the file?  Should it?  If so we should also look at tests since I know some tests have a hardwired version rather than just reading the latest/current version.

Veracity0 (changes requested):
This adds an "attribute" to the "attributes" field of adventures.txt.
This does not change the number of fields per line, so it does not warrant a version number bump, in my opinion, the comment listing which attributes there are at the top of the file should be updated.

In fact, it would be nice if the attributes were documented; until you mentioned that Level was water level, I did not know.

src/data/adventures.txt
Veracity0: version number should change if you add a new tab delimited field.
Programs that read the file presumably are coded to work with a specific version.
If tests are coded to work with a specific version, if the file format changes, they should/must be updated.

Veracity0 (approved):

PR #1176, opened 2022-10-12
use link for autumnaton not inline. more drops for Othello

Veracity0:
1) Autumn-aton redirects to a choice, so use link cannot be in-line
2) Othello also drops handkerchief, dagger, and military jacket


midgleyc (approved):

midgleyc:
Actually, I think there's a tab in Othello's drops where there should be a space. Does that break things?

Veracity0:
I see tabs separating each drop. That is correct.
I see spaces between each word and between the item and the (0).
When I look at Location Details, I see all 4 items.
And when I encounter Othello, all 4 items are listed.

midgleyc:
My mistake. I thought drops were supposed to be space separated, not tab separated.

PR #1179, opened 2022-10-14
refactor: use types internally for environment and difflevel

midgleyc:


jaadams5 (dismissed):
No new tests?  ;-)

midgleyc:
It looks like the only uncovered areas are:
* AreaCombatData.toString(), but I can't see where this is called (outside of RelayRequest's "safety") or what it should do
* Two error branches that shouldn't be called
* Gibberer adventure increment, for which I'd want a debug log from somebody with a Gibberer.

Veracity0:
I have a Gibberererer.

I assume you want the "time slows down message" which supposedly gives an extra rollover adventure. Supposedly "like a wild hare".

Do you want a wild hare message, too?

I'm out of turns today, but tomorrow is another day.

midgleyc:
Yes please: both would be good for tests.

gausie (dismissed):
Love it, thank you. I suppose I should have waited for the above comments to be resolved, but still.

midgleyc:
I'll do the tests once I get the HTML, but I agree this can go in as-is.

midgleyc:
made it in under the wire, V

I'll get the tests done now

Veracity0:
I just uploaded .txt files for both the hare and gibberer.
txt, not html, so you can see which requests were submitted.
Obviously, you will extract the html. :)

midgleyc:
you probably want to relogin, too: you've given enough info for somebody to impersonate you, going by https://kolmafia.us/threads/replace-pwd-hashes-in-debug-log-with-generic-string.28127/ (the session id + cors header + user agent)

Veracity0:
I logged out and came downstairs to have breakfast.

Veracity0:
And do feel free to delete the .txt once you have extracted the html and request. :)

PR #1183, opened 2022-10-15
feat: default sharePriceData to true

midgleyc:
Inspired by https://kolmafia.us/threads/autoscend-ash-demands-ancient-magi-wipes-cant-find-file-divides-by-zero.28158

I think this is a potential problem caused by #687, which removed the bundled mallprices.txt file: if the user is not sharing mall prices, then they won't get any price data, which can cause scripts to crash.

One other option would be to download the file if it doesn't exist (but not keep it up to date), which would ensure the user always has some price data.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -136,7 +136,7 @@ global	saveState
 global	saveStateActive
 global	scriptButtonPosition	0
 global	scriptList	restore hp | restore mp
-global	sharePriceData	false
+global	sharePriceData	true
 global	showAllRequests	false
 global	showExceptionalRequests	false
 global	statusDropdown	0
--- a/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
@@ -38,6 +38,8 @@ public class MallPriceDatabase {
   private static final HashSet<String> submitted = new HashSet<>();
   private static int modCount = 0;
 
+  public static final File PRICE_FILE = new File(KoLConstants.DATA_LOCATION, "mallprices.txt");
+
   private static final int CONNECT_TIMEOUT = 15 * 1000;
 
   static {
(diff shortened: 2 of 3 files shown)


midgleyc:
I'm not sure whether defaulting sharing price data to true or pulling down an instance of the file if it doesn't exist is the better solution.

Arguments for changing the default:
* the user presumably wants up-to-date price data, and probably won't know they need to set this preference for that

Arguments against:
* the user possibly doesn't want to upload files from their computer to a remote server by default, and it's kinder to make this opt-in

I think since this code was written, programs have tended to have telemetry and other server communication as opt-out, instead of opt-in. This is a bit different from telemetry, but in the same ballpark.

jaadams5:
The historical reasons for making it an option seem to have been overtaken by events.

I would say if the file is not there then mafia should download it.

If the user has opted in to price sharing then the file is always downloaded at the beginning of the session and uploaded at exit.

The file has to be there for mafia to work so downloading it if it is not should not be under user control.

jaadams5 (commented):
This won't automatically fix things for a user who is missing the file and already has the preference set to false, correct?

midgleyc:
Yeah. So maybe we need to always download it if it isn't there.

I'm not sure this should be a preference at all, now: having a preference which is essentially "improve my privacy, but break some things" isn't great :P

gausie:
> I'm not sure this should be a preference at all, now: having a preference which is essentially "improve my privacy, but break some things" isn't great :P

Agreed, remove imo

jaadams5:
Maybe I'm not thinking it through but if mafia is changed to guarantee that the file will always be present then the preference still has a place.  If the preference is false then mafia will DL the file under some programmer defined condition - either when it's not there or always at the start of a session.  If the preference is true then the file is DL'd at the start of a session and UL'd at the end.

gausie:
I just don't see the point of the preference, the only reasons I can think
of are not ones I'd care to entertain usually

jaadams5:
If you are paranoid or have limits on your uploaded bandwidth the preference will give you a felling of control even if, in 2022, there is no measurable effect.  Similarly, making it go away will give some people a feeling that something has been taken away from them.

But if you want to eliminate the preference completely and make the remaining code function the way it does now when prices are shared that is also OK.

midgleyc:
I think, having read through the comments:
* we should leave the preference in, just in case somebody wants to turn it off
* we should default it to share data, because having accurate historical data is generally beneficial, and many scripts assume this is the case
* we should download the file if it doesn't exist on login no matter the value of the preference

jaadams5 (approved):

PR #1184, opened 2022-10-15
Adjust cosmic bowling ball return if it gets out of sync based on the start-of-combat messages

gausie:
I'm getting this number out of sync at the moment. While I determine why that's the case, we can ensure that we handle the situation gracefully by reacting to the start of combat messages.


test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -715,6 +715,34 @@ public void canReabsorbThirdMonster() {
...
+      // Off in the distance, you hear your cosmic bowling ball rattling around in the ball
+      // return system.
+      "1, 7, 8",
+      "1, 10, 9",
+      // You hear your cosmic bowling ball rattling around in the ball return system.
+      "2, 9, 7",
+      "2, 2, 4",
+      // You hear your cosmic bowling ball rattling around in the ball return system nearby.
+      "3, 5, 3",
+      "3, 1 ,2",
```
midgleyc, line 729: minor, but your CSV is slightly off here
```suggestion
      "3, 1, 2",
```

midgleyc (approved):

PR #1189, opened 2022-10-15
Teach NPCPurchaseRequest to parse item with quantity

Veracity0:
I got tired of seeing pseudo-coinmaster.txt data every time you looked at the flowertradein shop.

You do have to look at it - and learn current prices - every time, but we are still not ready to handle it.

- I make PCPurchaseRequest.parseShopResponse able t parse "Chroner (15)", say.
- But I'm not prepared to make the Mercantile Flower Exchange a full-fledged mixed-curreny coinmaster, yet.


gausie:
I see a test fixture but no test code - is there something that was left out of the branch by mistake?

Veracity0:
No. It would be necessary and sufficient for coding up FlowerExchangeRequest, so I included it to get it into our (test) code base, should I (or any of us) want to do that later. Almost certainly after Twitch has gone away this time, unfortunately.

I'd want coinmasters.txt to be able to read the followibg syntax:

```
The Central Loathing Floral Mercantile Exchange	buy	rose (2)	Chroner (1)	ROW759
The Central Loathing Floral Mercantile Exchange	buy	red tulip	Chroner (16)	ROW760
```
In other words, put which currency is spent right into the data file.
You could omit it, if there is only once currency.
I do not have time, right now, unfortunately.
But there are a variety of others - Cosmic Ray's Bazaar, for example - which could use that.

Although - do we know (or expect) that the flower exchange needs Twitch to be open in order to trade for Chroners? The shops that USE the Chroners will be unavailable.

gausie:
> Although - do we know (or expect) that the flower exchange needs Twitch to be open in order to trade for Chroners? The shops that USE the Chroners will be unavailable.

I was thinking this too (especially wondering if I should keep https://tulipbot.ar.gy up after TTT is closed just for fun). My instinct is that it will stay open.

midgleyc (approved):
I think a test could be written that calls `parseShopResponse`, but it would have to be removed after `coinmasters.txt` was updated? Or is there a way to blank known shops?

If too hard I think this is fine as-is.

Veracity0:
I did some work this morning to see if I could do a FlowerExchangeRequest and I discovered a number of shortcomings in how we handle shop.php with a whichrow attribute:

- we assume that you will get a single object per purchase.
- we assume that each shop has only one row that sells a particular item

Those two assumptions are both violated by the flower exchange

- we do allow the same item to appear in multiple "shops" - the legend keys in the DA vending machine and in Cosmic Ray's Bazaar.
- we do support shops that sell both for items and Meat - Cosmic Ray's and the Black Market.
- we do support shops that have multiple currencies - Mr. Store, Cosmic ray's, ...
- we do not support "shops" that require multiple disparate items. Grandma Sea Monkey, Pixels, ...
Well, we do support them - as creation methods.
- we do support coinmasters that deal in "tokens" - Quartersmaster, Dimesmaster, Swagger Shop. Those are not shop.php
- some shops should be refreshed each time you visit - flower exchange prices change, Mr. Store adds and removes items, the Hermit sells out

I think the key attributes of shop.php:

- It sells "item (n)" in exchange for one or more other items.
- It can sell the same item in exchange for different currencies. Flower Exchange...
- Every item-for-currency is given a distinct row.
- ROWs are not shared accross shops; same item, different ROW
- multi-buy is the number of transactions you want:
2 roses -> 1 Chroner means spending 1000 roses is 500 transactions to get 500 chroner
1 red tulip for 15 Chroner means spending 10 red tulips is 10 transactions to get 150 Chroner

I envision a Project:

- a Row is rowId, shopid, AdventureResult purchas, AdventureResult[] costs
- - Array of costs if we want to include "creation methods" like Grandma, pixels, star charts - which are "shop.php".
- shop.txt could replace npcstores.txt and coinmasters.txt
- - Every npc store (Meat only) has rows and uses shop.php
- - Old coinmasters do not use shop.php and do not have rows
- - non-item tokens still exist (dimes, quarters, swagger, ...
- - We can parse shops that have mutable stocks.
- - We can parse shops whose inventory have to be unlocked (or learned by the community) - PirateRealm, Lathe, etc.

Perhaps shops.php will have, essentially, AventureResults both for items for sale but for currencies.
```
Star Chart	starchart	ROW141	Richard's star key	star chart	star (8)	line (7)
The Central Loathing Floral Mercantile Exchange	flowertradein	ROW759	Chroner	rose (2)
The Central Loathing Floral Mercantile Exchange	flowertradein	ROW760	Chroner (16)	red tulip
```

PixelRequest is a CreateItemRequest - but it is shop.php with rows.
Similar for star chart, Nash Crosby's Still, Grandma

That's another wrinkle; "acquire" will create things or buy them - but coinmaster purchases count as creations?
Should we move those creation methods out of concoctions?
Or make them both concoctions and shops?

Very preliminary thinking about this.

I don't want to do more on the flower exchange without carefully thinking about this, rather than throwing in a hack to support the broken assumption.

Veracity0:
I'm going to merge this without a test, for now.

Once I finish my "add 'step' buttons to puzzle animations" project, I'll start the "shop.php" project, and I expect that parseShopResponse will have major refactoring.

PR #1190, opened 2022-10-15
feat: autumnaton locations

midgleyc:
Add location list to autumnaton command. Also add ASH command for getting locations.

This is a useful increment, but there's more I want to do here:
* display upgrades, and what they do, if you don't have them
* cache the locations like `LocketManager` does

To update the cache, I need to know exactly what causes the location to appear (is a turn necessary, or is a combat enough?) and also where this should go in the codebase (somewhere in AdventureRequest?).

Checked "visited": initial copperhead NC is enough, as is free lianas. Didn't try canceling the copperhead NC. Next run.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -2747,9 +2747,7 @@ public abstract class RuntimeLibrary {
     params = new Type[] {};
     functions.add(
         new LibraryFunction(
-            "get_autumnaton_locations",
-            new AggregateType(DataTypes.LOCATION_TYPE, 0),
-            params));
+            "get_autumnaton_locations", new AggregateType(DataTypes.LOCATION_TYPE, 0), params));
   }
 
   public static Method findMethod(final String name, final Class<?>[] args)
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -9827,4 +9833,25 @@ public static Value zap(ScriptRuntime controller, Value item) {
 
     return DataTypes.makeItemValue(itemId, true);
   }
+
+  public static Value get_autumnaton_locations(ScriptRuntime controller) {
+    if (!InventoryManager.hasItem(ItemPool.AUTUMNATON)) {
+      return new ArrayValue(new AggregateType(DataTypes.LOCATION_TYPE, 0));
```
jaadams5, line 9839: Does KoL only supply the list if the AUTUMNATON is in inventory?  I can see a use case for planning where it might be possessed but elsewhere as well as a case where the locations are of interest even if the item is not owned.  Not a blocker.

jaadams5 (dismissed):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
midgleyc: at the moment, there's no caching, and we need to successfully use the autumn-aton to get the list of locations. I want to add caching, but I need to do some more investigation first.

PR #1191, opened 2022-10-15
Decrement bastille loaner voucher item count when used

libraryaddict:
Currently bastille loaner voucher item counts are not updated when they're used.

When used, even if the player chooses to walk away the wiki reports that the voucher is used.


midgleyc (approved):
Thanks!

PR #1192, opened 2022-10-16
fix: lowercase for environment + difficulty level on proxy record

midgleyc:
I mixed up `name` and `name()`, and assumed coverage meant something was tested instead of just that we had some test that ran that bit of code.

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -3322,7 +3322,7 @@ public class Modifiers {
 
     Modifiers.currentLocation = location.getAdventureName();
     Modifiers.currentZone = location.getZone();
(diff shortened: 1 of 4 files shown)


jaadams5 (approved):
Not sure what you mean by "coverage meant something was tested instead of just that we had some test that ran that bit of code." but that's irrelevant.

midgleyc:
The code coverage stats tell you whether a particular line was run in any test, rather than whether there were any assertions that related to the code in question. For example, it is the difference between

```java
myfunc();
```

and

```java
assertThat(myfunc(), is(1))
```

The former calls, but doesn't test anything except that the code doesn't crash; the latter tests the actual function.

In this case it's much less obvious: some test calls a complicated bit of code that includes the relevant lines, then tests something else, and not those lines in particular.

PR #1200, opened 2022-10-20
Correct Crystal Ball tracking

libraryaddict:
As said by cannon on ASS discord:

```
orb predictions are valid if:
* your turnsplayed is <= the turnsplayed when you got the prediction + 1
OR
* your [lastadv] flag is the zone the prediction is in. 
Toastergazing will only wipe the latter kind of prediction, when the first condition isn't met
```

Crystal Ball predictions has some inaccuracies.

## Pondering overrides existing predictions, setting every prediction turn count to the player's current turn as if it was just made.

Changed to use the previous prediction turn count if found, otherwise use the current turn

## A prediction can be invalidated using the last adventured location. However, mafia doesn't track it the same way as KoL does. Mafia actually ignores some of the zones, or adds its own.

Added a new field which only keeps track of what KoL tells us,.

Some of Mafia's locations are not in KoL (Cobb Throne Room) and others are explicitly ignored in mafia.

Such as Cake Arena or Mining.

Additionally, as it's set in a single location it should let us invalidate predictions reliably.

## Predictions are invalidated on the wrong turn

This is likely due to the previous area in which prediction invalidations were handled.

Now we use a combination of location changes, and when turn count changes.

## Some monsters have ambiguous names and cannot be parsed by their name. `Ninja Snowman (Chopsticks)` `Ninja Snowman (Hilt)` are both known as `Ninja Snowman`

We now find the first monster to have a Manual name matching what we're looking for. It isn't 100% accurate, but this is a limitation due to KoL not exposing the monster ID.

Additionally, when we're checking if the monster we've encountered is the predicted one, we're checking if they both share the same manual name. If they do, then we don't know better and must claim they are the predicted monster.

## Locations and Monsters with ` in ` as part of their names can be parsed incorrectly or not at all

Changed how we match those predictions from a simple regex solution to a looping seeking a match for the full prediction string.

## Adventuring somewhere that does not do a charpane request, such as an encounter that did not take a turn: Does not change the [Last Adventure]

Now checks for `Adventure Again (Santa's Parade)` using regex, if found it will set the last zone using that.

## Known issues
-  Predictions are only cleared when checking a location's combat and non-combat queue. Some items can bypass that, such as rusty hedge trimmers. Restoring a prediction that should've been invalidated. (Eg, predict in twin peaks, spend 10 turns elsewhere without ball, use rusty hedge trimmers, ponder)
- Ambiguous monsters due to limitations can predict the wrong monster with the same name, which is an issue if you were seeking a specific monster or were unaware of this.


gausie:
> ## A prediction can be invalidated using the last adventured location. However, mafia doesn't track it the same way as KoL does. Mafia actually ignores some of the zones, or adds its own.
> 
> Added a new field which only keeps track of what KoL tells us,.
> 
> Some of Mafia's locations are not in KoL (Cobb Throne Room) and others are explicitly ignored in mafia.
> 
> Such as Cake Arena or Mining.
> 
> Additionally, as it's set in a single location it should let us invalidate predictions reliably.

Is this not just an adventure.php thing?

libraryaddict:
> > ## A prediction can be invalidated using the last adventured location. However, mafia doesn't track it the same way as KoL does. Mafia actually ignores some of the zones, or adds its own.
> > Added a new field which only keeps track of what KoL tells us,.
> > Some of Mafia's locations are not in KoL (Cobb Throne Room) and others are explicitly ignored in mafia.
> > Such as Cake Arena or Mining.
> > Additionally, as it's set in a single location it should let us invalidate predictions reliably.
> 
> Is this not just an adventure.php thing?

Sorry, to elaborate.

Throne room is not a kol location for last adventured. But it is for mafia.

Cake arena is a kol location, but not for mafia.

gausie:
> Throne room is not a kol location for last adventured. But it is for mafia.

Is this always the case? Should we remove it from our consideration for last adventure location?

> Cake arena is a kol location, but not for mafia.

Should we make it a mafia location?

libraryaddict:
I'm pretty sure there's mentions of code that explicitly does not want those locations to be in mafia. Such as cellar, mining.

That said, occurs to me that we depend on locations existing for predictions. So really we can still handle missing lastadv.

The real trick is knowing when mafia is reporting a zone change when kol disagrees. Or if kol reports one but mafia disagrees. That's the only real issue AFAIK.

And you would want to know the actual kol last adv no matter how many times mafia changes it. 

Such as free fights shenanigans.

Or summed up, tracking kol actual reported last adventure is the easiest solution. Mafia is using different logic on what can be considered an adventure.

gausie:
I do think it is worth tracking when KoL disagrees with the value of last adventure. And ideally working out *why* that is. 

We parse last adventure from the charpane and update it so why does that ever not match? There is only one special case I can see, and that is when the last adventure was in "The Naughty Sorceress' Tower".

I'm still skeptical of the need for this "vanilla" pref, and would prefer adapting our code to match the state of KoL where possible.

libraryaddict:
It sounds like that to keep it in line, we can no longer report that last adventure was throne room. And doubtlessly other places, probably island war bosses for example.

I think that's a concern in that it would throw people off. Unless last adventure pref isn't related to the internal code?

libraryaddict:
Actually, it occurs to me that instead of setting it as a field. We could instead just set the [lastadv] as a pref.

Florist Friar apparently works on a similar mechanic (Untested), so perhaps we could set the field in a preference instead.
We have "lastAdventure" for mafia, lets leave that untouched.

So instead we can use "lastZone"? Doesn't promise to resolve to a valid location. Although, we might want to add those locations just so we can work some magic with them. Such as checking what type of zone it is. Indoors? Outdoors?

libraryaddict:
Hmm curious, had a bad prediction.

Started in Defiled Cranny, predicted gluttonous ghoul. Ended up sinking the evilness from 35 to 22. 
Fight ended, pondered, prediction still in there.

Switched to fantasy airship after some prep stuff (Parka, equipping) then encountered the introduction adventure which is free.

Pondered again, no predictions.

What I think happened here is that KoL invalidated my prediction in the cranny because it was no longer possible, it didn't eval it at the end of the fight in the cranny as one might expect.

So not an actual issue as such. This would match with twin peaks in funny behavior somewhat.

libraryaddict:
Marked as a draft again, had two more incorrect ponders I can't explain.

libraryaddict:
One issue somewhat, is that at least one location and possibly others do not tell Mafia when they've changed.

No charpane refresh, no api.php calls needed, not even an "Adventure Again" inside the page.
An example here is toastergazing, visiting the shore, choosing the vacation shop to get out of the choice.
It changes your last adventured, but we don't know it.

Of course, mafia could track this itself. But mafia is also wrong about places that count as "Last Adventure" which is why this PR tracks it.
A potential hardcoded list is an option to be sure. Or call api.php when we thought there should've been an adventure, but no api or chanepane requests was made.

libraryaddict:
@Veracity0 do you have any thoughts about how we're tracking zones for this pr?

Given the way mafia zone tracking is different from kol, we can't just reuse the mafia tracking since we need to know what kol actually thinks the last adventure is.

But although the current changes works, it doesn't work properly if kol never requests a charpane refresh and doesn't show the adventure again message. Meaning mafia has no idea that kol considers us to have adventured in a new location.

This is noticeable if you go to the shore, then visit the shop.
Charpane will not be refreshed, but if you refresh it then it shows that our last location was changed. This happens on other locations as well I think.

I don't have a good solution for this. It's more of a kol bug given it should definitely refresh our charpane. But not likely to be fixed.

src/net/sourceforge/kolmafia/session/CrystalBallManager.java
```diff
@@ -135,12 +162,26 @@ private static String parseCrystalBallMonster(final String responseText) {
...
+    // We use the vanilla last adventure here as sometimes mafia does not recognize it
+    // And we still want to invalidate the predictions, even if mafia doesn't recognize it
+    String lastAdventureName = KoLAdventure.lastZoneName;
+
+    if (lastAdventureName == null) {
       return;
     }
 
-    String lastAdventureName = KoLAdventure.lastLocationName;
+    if (CrystalBallManager.predictions.isEmpty()) {
```
midgleyc, line 182: minor: you could shift this to the very top of the function to avoid copying the `lastAdventureName` in the (hopefully common) case that there are no valid predictions because the user isn't using the crystal ball.

midgleyc (changes requested):
Thanks!

Sorry for the delay; the longer and more complicated something is the longer it takes anybody to get to it :P

src/net/sourceforge/kolmafia/session/CrystalBallManager.java
libraryaddict: Change made

midgleyc (approved):

PR #1201, opened 2022-10-21
fix typo in item name

Veracity0:


midgleyc (dismissed):
we might want to have a test that runs through these and confirms they're all valid item names, I think this is the second time this has happened

Veracity0:
This is the second time. There was a report on kolmafia.us.

jaadams5 (approved):

PR #1202, opened 2022-10-21
Update chat to match KoL behaviour

oxc:
This PR consists of two parts:

1. Both KoL mchat and lchat have switched to 3000/10000ms somewhen in the last 8 years. Adjust our values to do the same.

lchat:
image
mchat (away time comes from server, not from source code):
image
image

2. KoL mchat is adding a random aa=%RANDOMNUMBER% parameter to each newchatmessages.php request. Probably some kind of cache buster? I've added this to ChatRequest for the mchat relay requests.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ChatRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ChatRequest.java
@@ -2,6 +2,7 @@ package net.sourceforge.kolmafia.request;
 
 import java.nio.charset.StandardCharsets;
 import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.KoLConstants;
 import net.sourceforge.kolmafia.chat.ChatManager;
 
 public class ChatRequest extends GenericRequest {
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/request/ChatRequest.java
```diff
@@ -21,7 +21,9 @@ public ChatRequest(final long lastSeen, final boolean tabbedChat, final boolean
     StringBuilder newURLString = new StringBuilder("newchatmessages.php?");
 
     if (tabbedChat) {
-      newURLString.append("j=1&");
+      newURLString.append("aa=");
+      newURLString.append(Math.random());
```
jaadams5, line 25: Why?

oxc: It's both in the description and the commit message:

> KoL mchat is adding a random aa=%RANDOMNUMBER% parameter to each newchatmessages.php request. Probably some kind of cache buster? I've added this to ChatRequest for the mchat relay requests.

oxc:
I've change the code to use the RNG constant, so I can I seed it during tests.

gausie (approved):
Works for me, thanks!

src/net/sourceforge/kolmafia/request/ChatRequest.java
jaadams5: OK.  Not everyone who is reading code is also reading description and commit messages and this was not something I expected from the description.  

Resolved.

jaadams5 (approved):

PR #1203, opened 2022-10-21
Support for using saved volcano maze maps

Veracity0:
There are, supposedly, six sets of five volcano maze maps.
People have been whining for years that KoLmafia should use the published solutions, rather than deriving them each time.

I refuse to remove the path derivation logic, but I am willing to load in saved map cycles and determining which cycle/map we are on by looking at KoL's volcano map when you visit it.

1) I have 4 of the 6 map sequences. They are now in volcanomaze.txt
2) If you have "useCachedVolcanoMaps" set to true (the default; I only have the property for use in testing), when you visit the maze, we look up the platforms, and if we find a matching map, we set all five of the volcanoMapX properties to correspond to the map sequence you have.
You can "step" or "solve", and KoLmafia will not have to explore and possibly require you to jump in and restart.
Note that you can manually step and get off the solution path, in which case "step" and "solve" will require you to jump in.
3) If we don't recognize the map, we derive it, as before, and when you finish the discovery, we print out something like this to the gCLI and session log:
```
--------------------
5	1	0,3,17,19,21,24,26,28,31,33,37,48,54,63,66,68,70,74,82,85,89,93,109,111,116,118,123,126,127,134,143,149,155,157,163,166
5	2	1,4,14,16,20,22,25,38,43,47,52,57,59,67,71,75,77,80,86,87,88,103,107,119,122,131,138,141,146,148,151,154
5	3	2,8,12,18,23,29,36,39,40,46,55,56,58,65,73,76,83,90,94,96,99,101,105,106,113,117,121,124,129,132,140,147,152,158,168
5	4	6,9,11,27,32,35,42,44,50,53,61,64,69,79,91,95,98,100,104,115,120,125,133,136,142,144,153,159,161,164
5	5	5,7,10,13,15,30,34,41,45,49,51,60,62,72,78,81,92,97,102,108,110,112,114,128,130,135,137,139,145,150,156,160,162,165,167
--------------------
```
Which are exactly the five lines needed for volcanomaze.txt
4) I wrote tests for detecting the entire map sequence from the first (cached) map you see.

There are 2 map sequences I do not have. I asked the community for help.
No joy, yet, but if they respond before this is approved, I will include them.
If not, I will shrug and merge and fill in map sequences as I discover them myself in future ascensions.


src/data/defaults.txt
```diff
@@ -159,6 +159,7 @@ global	useLastUserAgent	false
 global	useSystemTrayIcon	false
 global	useTabbedChatFrame	true
 global	useToolbars	true
+global	useCachedVolcanoMaps	true
```
midgleyc, line 162: Is this worth letting be configurable? When would a user want to turn this off?

midgleyc (approved):

src/data/defaults.txt
midgleyc: Oh hmm auto-merge sure interferes with my approach of approving with optional comments!

Veracity0: My tests that demonstrate that we can derive new map sequences turn it off.
The user will not want to turn it off. It need not be configurable.

Sorry about automerge.

midgleyc: It's okay, I should have checked myself.

PR #1208, opened 2022-10-23
fix: no visible delay in call script

midgleyc:
Add items to the MRU List in parallel to avoid a pause before updating the GCLI after calling a script.

The `synchronized` keywords in this file can cause delays elsewhere, but this seemed the most obvious.


jaadams5 (approved):
Thanks.  I had poked at this but running in parallel was not something I thought of.

PR #1214, opened 2022-10-25
Bugbear Invasion, Diff Levels, Volcano Maze #5, Relay Warnings, Autumn-aton logging

Veracity0:
Started a Bugbear Invasion run in a Canadia sign.

The "bugbears" command is very handy. We have good tracking for which of the Mothership zones are open. canAdventure support will be easy.

Bugbear Invasion Fixes:

1) Two of the bugbears have capital letters in their names. That threw off fight tracking for them, which assumed the names were all lower case.
2) canAdventure for bugbear zones
3) place.php?whichplace=bugbearship&action=bb_bridge leads to the fight with the Bugbear Captain. Log it appropriately (like the Sorceress Tower)

Diff Levels:

4) Camp Logging Camp and the Outskirts of Camp Logging Camp
5) The 3 Verge of War (small island) zones we were missing. Unsurprisingly, they are all high.
6) Battlefield (Hippy Uniform) - high
7) I Sent Autumn-aton to two Junkyard zones Exactly like the post-war Junkyard, it returns with no items whatsoever.
Just like the post-war version, I submitted a bug report.
We assume all of those are "outdoors high", but we cannot confirm it.
8) The Themthar Hills - high
9) McMillicancuddy's Barn - high
10) McMillicancuddy's Pond - high
11) The Feeding Chamber, The Royal Guard Chamber, The Filthworm Queen's Chamber - high
12) Medbay, Waste Processing, Sonar - level 1 zones - all mid
13) Science Lab, Morgue, Special Ops - level 2 zones - all mid
14) Engineering, Navigation, Galley - level 3 zones - all mid
15) The Nemesis' Lair - mid
16) Le Marais D&egrave;gueulasse:
The Edge of the Swamp, The Ruined Wizard Tower, The Weird Swamp Village - all mid

Other Features:

17) When you send your Autumn-aton off, log it with adventure name.
Yes, it logs the URL, and you can see the adventure number there, but, it would be nicer...
Similarly, log upgrade installations
18) Getting a Drunkula Wineglass warning with no Wineglass in inventory is unhelpful. Soup up the warning.
19) RelayRequestWarningsTest for the most part did not use Cleanups. In fact, it left your path set to You, Robot - which screwed up my WIneglass test. I split the Mohawk wig test into multiple individual tests using cleanups, which fixed that issue. Something similar should be done for the Survival Knife and Machete tests,
20) I got the 5th Volcano Maze!

There are four more Swamp zones - which I expect are also mid.
And there is more cleanup needed for the RelayRequestWarningsTest.

Those all can wait. I want to get this in.


src/net/sourceforge/kolmafia/session/AutumnatonManager.java
```diff
@@ -87,6 +91,37 @@ private static void parseQuest(final String responseText, final int snarfblat) {
...
+        // Installing an upgrade
+        int choice = ChoiceUtilities.extractChoiceFromURL(urlString);
+        String desc = ChoiceManager.choiceDescription(choice, decision);
+        RequestLogger.printLine(desc);
+        RequestLogger.updateSessionLog(desc);
+        return true;
+      }
+      case 2 -> {
+        // Sending on a mission
+        int snarfblat = getSnarfblat(urlString);
```
gausie, line 114: I suppose this can be null

gausie (approved):

src/net/sourceforge/kolmafia/session/AutumnatonManager.java
Veracity0: Unexpected, but, I suppose.
getSnarfblat returns an int - but if not found, it returns null? And Java accepted that?
And looking up KoLAdventure by snarfblat can return null if unknown adventure.

Both fixed.

jaadams5 (approved):
I may have just had the null adventure.  Will see if it repeats?

PR #1218, opened 2022-10-28
11 more Diff Levels. Felonia's item drops.

Veracity0:


midgleyc (approved):
I think we're going to be seeing a lot of mids ;)

PR #1219, opened 2022-10-29
Mark RecordType.get(names) param as readonly

oxc:
This lets the compiler know that the input array is not modified by the function. This is helpful when you have a const array of item names that you want to lookup:

```ts

const GIFT_PACKAGES = [
  'plain brown wrapper',
  'less-than-three-shaped box',
  'exactly-three-shaped box',
  'chocolate box',
  'miniature coffin',
  'solid asbestos box',
  'solid linoleum box',
  'solid chrome box',
  'cryptic puzzle box',
  'refrigerated biohazard container',
  'magnetic field',
  'black velvet box'
] as const;

const items = Item.get(GIFT_PACKAGES); // does not work without this change
```


midgleyc (approved):
Thanks!

PR #1220, opened 2022-10-29
Make maximizer `effective` keyword ignore June Cleaver

Alium58:
https://kolmafia.us/threads/june-cleaver-maximizer-effective-keyword.28217/


midgleyc (changes requested):
The change itself seems good, but please add a test.

Alium58:
Created a few test for the effective keyword. Let me know what ya think of them

midgleyc (approved):
Looks good, thank you!

PR #1222, opened 2022-10-31
Some skill parsing things and a no longer failing test.

jaadams5:
I tried to write something that handled the fact that my charsheet has 17 as the ID for Summon Hilarious Objects instead of 7226.

In the process I noticed a previous test that passed with Torso Awaregness as a skill name (from KOL) even though we think the skill is Torso Awareness.

So I added a check id and name to that test and it now fails because Torso Awaregness  is not a skill.

Similarly my other test is expected to fail.

The skill is Torso Awareness on my charsheet in plain KoL as of 10/30/22.  

If you try and display the description of Summon Hilarious Objects on my charsheet it fails, although KoL does cast it.  I am not convinced, however, that KoLmafia is correctly casting the skill.

Not sure where to go with this yet.


jaadams5:
I did four things.

I added a "fix" for the KoL bug that uses 17 as the skillId for Summon Hilarious Objects that just changes the id to 7226.

I cleaned some lint in CharSheetRequest.

I added a test that parses a real charsheet with skillId of 17 and shows it was replaced with 7226.

I noticed an old test that had Torso Awaregness which is no longer the name of the skill.  I edited the data file and the test accordingly.

I am open to the suggestion that this PR be rejected.  If it is, I would hope we would improve how we handle skills with the same name.

gausie (approved):

PR #1223, opened 2022-10-31
perf: improve RequestLogger for "external" requests

oxc:
RequestLogger.registerRequest checks for quite a number of well-known requests, and marks them as wasLastRequestSimple.

This is also done for requests that are considered "external", i.e. basically every relay request and visit_url call, if I'm not mistaken.

For such requests, we can skip all the instanceof checks (which are surely measurable in that quantity), by flipping the OR conditions. Due to the lazy-evaluation of the Java logical OR, the right hand side is only evaluated if required.

I'm aware that this is probably premature optimization, especially since the single registerRequest calls will take much more time in comparison (this occurred to me during writing this description, so I'm posting it anyway), so please feel free to disregard this PR if you feel it unnecessary.

### DIFF
--- a/src/net/sourceforge/kolmafia/RequestLogger.java
+++ b/src/net/sourceforge/kolmafia/RequestLogger.java
@@ -383,7 +383,7 @@ public class RequestLogger extends NullStream {
             || request instanceof RelayRequest
             || request instanceof PlaceRequest;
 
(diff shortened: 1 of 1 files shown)


midgleyc (approved):
Thanks!

PR #1224, opened 2022-11-02
feat: initial items for cookbookbat

midgleyc:
Get those items started.

Also can we just sort `modifiers.txt` alphabetically? And maybe autosort it? And same for the other files? I want to just add stuff at the end and run a script or something.

### DIFF
--- a/src/data/familiars.txt
+++ b/src/data/familiars.txt
@@ -314,3 +314,4 @@
 285	Arachnelf	arachnelf.gif	none	festive egg sac	Site Alpha ID badge	3	1	3	1
 286	Synthetic Rock	synthrock.gif	none	synthetic rock		0	0	0	0
 287	Grey Goose	greygoose.gif	item0	grey gosling	grey down vest	1	2	3	3
+288	Cookbookbat	bbat_fam.gif	drop	mummified entombed cookbookbat		0	0	0	0
--- a/src/data/fullness.txt
+++ b/src/data/fullness.txt
@@ -655,6 +655,7 @@ pear tart	3	7	awesome	8-14	0	25-45	0
 pecan pie	3	5	good	7-9	20-25	0	0	30 Pecan Pied Piper (+25 candy drop)
 Peche a la Frog	3	1	decent	3-4	0	0	0
 penguin focaccia bread	7	1	good	18-24	21-38	21-38	21-38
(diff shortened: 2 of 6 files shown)


Veracity0:
By alphabetically, I assume you mean case-insensitive.
Exactly which files (in addition to modifiers) do you want to be alphabetical that are not currently alphabetical?

Regarding modifiers.txt

- I find it useful to be able to manually look at "hats that grant Adventures", say.
The maximizer will display ONE hat which does that, but if a multi is thinking about buying something affordable, having the Admiral's hat suggested is not helpful.
- Speaking of which:
```
# Admiral's hat: +50 Skeleton Anger
Item	Admiral's hat	Adventures: +5, Moxie Percent: +30, Familiar Effect: "atk"
```
We include non-modifier enchantments as comments. I've noticed that some people don't bother to put thos into modifiers.txt. I've have to go add them in later.
Or consider this:
```
# Item survival knife: Aids in desert exploration
Item	survival knife	Muscle: +5, Mysticality: +5, Moxie: +5, Lasts Until Rollover
```
I actually don't like the "Item" in front. I understand that registering a new item includes it. To make it easier to uncomment it if it becomes a modifier, I guess, but that's not really useful; we would make the modifier keyword/expression different.
The point being, any automated "sorting" would have to be smart.

Regarding other data files:

- things with ids - effects, items, familiars - typically already get "added at the end".
- coinmasters, npcstores - should be grouped within stores. I don't much care what order the shops are in.
- concoctions? Here I agree. They were originally grouped by hola using ... some ... logic.
```
# Burritos, pizza and statue lime pies.  These are a class all their own
# because of their popularity.  Other really popular items, like spectral
# pickles, are also included.
```
I could see things grouped my mixing method:

MIX
MIX_FANCY
ACOCK
SCOCK

for example.

Veracity0 (approved):

midgleyc:
Yeah, `modifiers.txt` is the main one. When I'm trying to put things in the right place I search "Item<tab><starting characters>" and sometimes it jumps off to a whole other section.

Any sorting thing would have to have some level of intelligence to be able to deal with the comments at the top. I would like it if, rather than everything starting "Item", it could start "Hat", "Pants", etc -- corresponding to the section. It would make finding lines easier.

The comments not starting with the item type would hurt automation slightly, but it'd be manageable I think.

PR #1225, opened 2022-11-03
Various cleanups, tests, and small features.

Veracity0:
1) I noticed that the new IOTM often appeared with an extra space in front of its item name. In Mr. Store, for example.
I looked at all the places when we extract items names from KoL pages where we check the name against the expected name given the itemId. I made them all trim() the extracted name.
2) There were many places where we looked up itemId from descId and, if not found, registered a new item from the descId.
I added ItemDatabase.lookUpItemIdFromDescription, which will lookup the item using the descid and will register a new item if not found. In either case, return the itemId.
3) There were several examples of old code which declared that they could not register new items given just a descId, "since they didn't have itemId". Old, I say, because we have long had the ability to register new items using just the description. I made those places use the new method to look up and register if required.
4) I added tracking the Grimstone hare game in "hareMillisecondsSaved" incremented after each fight. KoL logs the final time as something like "29:58.162" - which is to say, 30 minutes - accumulated milliseconds.
5) Pandamonium is open only when you visit it. I have saved HTML for doing that.
canAdventure() allows Panadmonium zones if Azazel quest is started, Friars is finished, or you have the ritual items.
prepareForAdventure will run the ritual, if needed, and start Azazel, if needed.
6) Unit tests for KoLAdventure.canAdventure() with the 5 grimstone zones.
7) Legendary recipes and foods, courtesy of reverkiller


src/data/fullness.txt
```diff
@@ -248,6 +249,7 @@ D roll	2	4	awesome	6-8	0	20-30	0	20 On a Roll (+25 Food Drop)
 dead lights pie	2	2	awesome	7-9	5-10	20-40	5-10
 dead meat bun	3	4	decent	5	10	0	0
 decent brain	1	1	decent	2	3-5	3-5	3-5	Zombie Slayer
+Deep Dish of Legend	2	1	EPIC	0	0	0	0	Unspaded
```
midgleyc, line 252: effect should be noted

src/data/fullness.txt
```diff
@@ -129,6 +129,7 @@ bucket of honey	1	4	good	3	0	0	0	50 Sugar Rush
 bunch of square grapes	2	7	awesome	6-7	1-5	5-20	1-5
 C.H.U.M. chum	3	1	crappy	1	0	0	0	20 Majorly Poisoned
 cactus fruit	2	7	awesome	6-7	6-20	1-5	1-5	lose 8-10 HP
+Calzone of Legend	2	5	EPIC	0	0	0	0	Unspaded
```
midgleyc, line 132: effect should be noted

midgleyc (changes requested):

src/data/fullness.txt
Veracity0: Done.

src/data/fullness.txt
Veracity0: Done.

midgleyc (approved):

PR #1226, opened 2022-11-03
feat: more cookbookbat items

midgleyc:
Just the items / effects. Preferences / concoctions yet to come.


Veracity0:
This is exciting.

Do you have the text for reading a Recipe and unlocking the concoction?

Because given that, we should be able to add the concoction for "Jarlsburg's vegetable soup", add the appropriate property:
```
user	unknownRecipe10971	true
```
and set that to false when you read "Recipe of Before Yore: Jarlsberg's vegetable soup".

And so on for all the other Recipes for which we know the item of the result.

Veracity0 (approved):

PR #1227, opened 2022-11-04
Only change ascension class if different.

phulin:
The lack of this check was causing a concoction database refresh on every api.php request.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -1214,20 +1214,22 @@ public abstract class KoLCharacter {
    * @param ascensionClass The name of the character's class
    */
   public static final void setAscensionClass(final AscensionClass ascensionClass) {
-    if (KoLCharacter.ascensionClass != ascensionClass) {
-      KoLCharacter.ascensionClass = ascensionClass;
-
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -1214,18 +1214,20 @@ public static final Correspondent getEudora() {
    * @param ascensionClass The name of the character's class
    */
   public static final void setAscensionClass(final AscensionClass ascensionClass) {
-    KoLCharacter.ascensionClass = ascensionClass;
+    if (KoLCharacter.ascensionClass != ascensionClass) {
```
Veracity0, line 1217: Can I suggest you simply do this at the top of the method:
```
   if (KoLCharacter.ascensionClass == ascensionClass) {
      return;
   }
```
That way, you do not have to re-indent the rest of the method because it is inside a conditional.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: done :)

Veracity0 (approved):

PR #1228, opened 2022-11-04
fix(JavaScript): allow ConsString to be passed for StringBuffer

oxc:
Library functions that currently take a StringBuffer as first argument have a special handling in LibraryFunctionStub, so that they can take string values instead when called from Javascript.
However, Rhino sometimes creates a ConsString instances for concatenated strings. They are handled as string Values by our code, but they are not caught by the LibraryFunctionStub StringBuffer handling, so it looks for a function that takes a String Value instead of a StringBuffer one.

This leads to error messages like:
```
> js const foo = "foo"; bufferToFile(foo + "bar", "data/foobar.txt")

Script exception: Function 'buffer_to_file( string, string )' undefined. This script may require a more recent version of KoLmafia and/or its supporting scripts. (command line#1)
````

This change adds support for ConsString in the StringBuffer handling.

It might have been easier to just accept CharSequence, but I'm not sure if there can be StringBuffers passed, which would then be re-wrapped.

Ultimately, it might be better to modify the detection of StringBuffer methods when strings are passed, or to allow any CharSequence to be passed in the first place.


gausie (approved):
Thank you for this, hopefully this is the last place ConsString causes us issues!

PR #1230, opened 2022-11-05
Add deferUpdate to GearChangeFrame.

phulin:
This should make a substantial performance improvement. We will now update the equipment lists only once when doing an equipment update, which happens every adventure.

I've also changed `ensureValidSelections` to `updateAllModels` as I believe that's a more consistent and descriptive name.


phulin:
Ah... seems like I'm not supposed to be initializing a frame on a test. Not sure how to get around that with the way the code is currently structured.

jaadams5:
>This should make a substantial performance improvement.

Any chance you could get before and after profile data that are comparable enough to support this assertion?

phulin:
Yep, will do.

phulin:
Ran the following test a few times and saw between a 20 and 40% decrease in execution time with the deferred updates turned on. Still need to figure out the test issue.

```java
        KoLConstants.inventory.add(ItemPool.get(entry.getKey(), 1));
      }
      var outfitOne =
          List.of(...)
              .stream()
              .map(id -> id < 0 ? EquipmentRequest.UNEQUIP : ItemPool.get(id, 1))
              .toList();
      var outfitTwo =
          List.of(...)
              .stream()
              .map(id -> id < 0 ? EquipmentRequest.UNEQUIP : ItemPool.get(id, 1))
              .toList();

      EquipmentManager.setEquipment(outfitOne.toArray(new AdventureResult[outfitOne.size()]));
      EquipmentManager.setEquipment(outfitTwo.toArray(new AdventureResult[outfitTwo.size()]));
      var start = Instant.now();
      for (var i = 0; i < 10; i++) {
        EquipmentRequest.switchEquipment(
            outfitTwo.toArray(new AdventureResult[outfitTwo.size()]),
            outfitOne.toArray(new AdventureResult[outfitOne.size()]));
        EquipmentRequest.switchEquipment(
            outfitOne.toArray(new AdventureResult[outfitOne.size()]),
            outfitTwo.toArray(new AdventureResult[outfitTwo.size()]));
      }
      var end = Instant.now();
      System.out.println("Elapsed: " + Duration.between(start, end).toMillis());
```

midgleyc:
Looks like it might be https://arbitrary-but-fixed.net/2022/01/21/headless-gui-github-actions.html?

Veracity0:
You cannot make a JFrame unless you have a real graphics environment.
However, you can make JPanels and test them. We do that in UseItemEnqueuePanelTest, for example.

What if GearChangeFrame was essentially just a container for GearChangePanel.
I did that for SynthesizeFrame.
```
public class SynthesizeFrame extends GenericFrame {
  public SynthesizeFrame() {
    super("Sweet Synthesis");

    JPanel centerPanel = new JPanel(new BorderLayout());

    centerPanel.add(new SynthesizePanel(), BorderLayout.CENTER);
    centerPanel.add(new StatusPanel(), BorderLayout.SOUTH);

    this.setCenterComponent(centerPanel);
  }
}
```
If all the "guts" of GearChangeFrame is in GearChangePanel, you could instantiate the latter and test it.

There would be other refactoring needed. GearChangeFrame.updateEquipmentModels calls a couple of SwingUtilities methods. It wants to call updateEquipmentModelsInternal() in the SwingEventThread. But, seems to me, it could invoke it directly if GraphicsEnvironment.isHeadless().

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
```diff
@@ -0,0 +1,1497 @@
+package net.sourceforge.kolmafia.swingui.panel;
+
+import java.awt.*;
```
midgleyc, line 3: we expand the *s on src/ files (or we try to, at least)

midgleyc (dismissed):
Also, the tabs have gone from the top to the bottom. Was there a reason for this change?

Veracity0:
Nice. checked out the PR so I could easily look at what you did with GearChangePanel.

1) GenericFrame had support for tabs. So, obviously you needed to handle them yourself in your panel.
- GenericFrame had support for tabbing forward and back.
I am fine with not having that for this 2-tab panel.
- GenericFrame had support for remembering which tab you were on when you closed it.
Again, I am fine with not doing that; let it always start with the Equipment tab, not the Customizable tab.
- Please put the tabs at SwingConstants.TOP
2) I like the deferredUpdateLevel concept. Reminds me of ConcoctionDatabase refreshes. ;)

And this is outside the scope of this PR, but I wish we had implemented some things as Listeners.
For example, when you add a familiar - or free the king or break ronin in a path that restricts familiars - I wish we could just signal "(familiars)" and both the GearChangePanel and any other GUI element that cares about that could update, rather than explicitly calling GearChangeFrema.updateFamiliars().

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
```diff
@@ -0,0 +1,1496 @@
...
+    this.outfitSelect =
+        new OutfitComboBox((LockableListModel<SpecialOutfit>) EquipmentManager.getOutfits());
+    this.customSelect =
+        new OutfitComboBox((LockableListModel<SpecialOutfit>) EquipmentManager.getCustomOutfits());
+
+    this.equipmentPanel = new EquipmentPanel();
+    this.customizablePanel = new CustomizablePanel();
+
+    this.tabs = KoLmafiaGUI.getTabbedPane();
+    this.tabs.setTabPlacement(SwingConstants.BOTTOM);
```
Veracity0, line 112: Please use SwingConstants.TOP, as it was before.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
```diff
@@ -0,0 +1,1496 @@
...
+      }
+    }
+  }
+
+  // For performance reasons, allow deferral of updateAllModels, which recomputes the lists that
+  // show in
+  // GearChangeFrame. This way we can perform multiple equipment updates at once without triggering
+  // several runs
+  // through the entire list of equipment in a player's inventory to see if it is equippable.
+  public static void deferUpdate() {
```
Veracity0, line 1269: Could you fix the formatting of that comment, please?
Looks like spotlessapply broke long lines into two.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
phulin: Huh. I compared to my jar version of mafia to ensure it looked the same. The jar that I just downloaded has the tabs on the bottom. Maybe it's OS-dependent somehow? I don't see a preference that would make the positioning user-specific.

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
phulin: thanks, this was my naughty IDE.

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
phulin: Will do.

src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
Veracity0: Java doc for JtabbedPane.setTabPlacement()

```public void setTabPlacement(int tabPlacement)```
Sets the tab placement for this tabbedpane. Possible values are:
- JTabbedPane.TOP
- JTabbedPane.BOTTOM
- JTabbedPane.LEFT
- JTabbedPane.RIGHT

The default value, if not set, is SwingConstants.TOP.

I have no idea why it is BOTTOM for you. Perhaps it is affected by look and feel?

GenericFrame does not call that - and GearChangeFrame has tabs at the top.
In your PR, I commented out the line - and the tabs are at the top.

I think you should remove that call and just let os/look-and-feel choose it.

phulin: Done.

phulin:
I think I've addressed all comments.

Veracity0 (approved):

PR #1231, opened 2022-11-05
Use HashMap over TreeMap in ConsumablesDatabase.

phulin:
This commit adds a CaseInsensitiveHashMap class and test, and then uses them to replace the TreeMaps in ConsumablesDatabase. These are performance-critical maps as they are read from a few thousand times during every concoction database refresh. They are modified rarely and don't grow in size essentially ever.

In the following speed test, this change caused a roughly 40% decrease in execution time for `getAdventureRange`, a function which takes up high single digits of execution time in my profiling tests.

```java
    for (int i = 0; i < 500; i++) {
        ConsumablesDatabase.getAdventureRange(entry.getValue());
      }
    }
```


midgleyc (approved):
Thanks!

PR #1232, opened 2022-11-05
Tests for concoction database refreshing

Veracity0:
This is a learning experience.

I have two fights saved:

- A Fruit Golem dropping three consumables
- A BASIC Elemental dropping a GOTO

I intend to use the response txts - two per fight: the start and and the finish - to check how many refreshes are done.

Three ways:

- GenericRequest - raw request/response
- KoLmafia.makeRequest with a single iteration - AdventureRequest
- KoLmafia.makeRequest with multiple iterations

So far, I have only the first (for the two monsters).

Discoveries:

- My response text had a large number of combat items. KoLmafia compares the "Use Item" dropdowns on a fight page with inventory - and resynchs inventory, which can cause a refresh, if the items are concoction ingredients, say.
My fix was to empty out the combat item dropdowns in my saved HTML.
- After we finish a fight, we call api.php. I'm using an empty response, but a real one would show a turn being used - and that would expect a refresh, since an "Adventure" is a virtual ingredient for mixing methods that take turns.

I'm going to have to think about how to simulate that when I start testing automation.


Veracity0:
I'm not feeling it any more. I don't see the point.

gausie:
> I'm not feeling it any more. I don't see the point.

Fair enough!

PR #1233, opened 2022-11-06
Fix appearance_rates for jungles of ancient loathing & seaside megapolis

Rinn:
I don't have the page html to write tests for the quest progress but I wrote tests for the encounter rates.


midgleyc (approved):
Could `Cleanups`ify the tests (and split them out), but the `AreaCombatTestData` is already resetting the character between tests so no harm as is I think.

Thanks!

PR #1234, opened 2022-11-06
feat: add "concoction_price" ash function for VYKEA and items

midgleyc:
When meat farming, you want to know how much it would cost to acquire a particular thing compared to the benefit it brings. You don't want to use `retrieve_price`, because you want to value the things in inventory at mall value instead of at zero (and that doesn't work for VYKEA anyway).

Add a function that tells you how much meat it would take to acquire a particular VYKEA companion (consumed items only, not counting the hex key or extra planks / rails / brackets).

For items, it is kind of like `retrieve_price`, but assumes you can't access any of the items. For implementation simplicity, doesn't recurse into ingredients (it can't re-use `InventoryManager.priceToAcquire` because that one prices accessible items at 0). Also doesn't seem to check Five Finger Discount / trousers, but I might have just written the test wrong.

It doesn't investigate recursive prices to satisfy a different potential use case: if `concoction_price(x) < mall_price(x)`, you can make some meat by creating and selling the item. If it did investigate recursive prices you'd want to know what the minimal collection was, and things change as you recurse (e.g. free crafts). Easier to just not do that.


gausie (approved):
Nice!

PR #1235, opened 2022-11-06
Refactor of ConsumablesDatabase

phulin:
Refactors ConsumablesDatabase to get rid of the various name-to-X maps and replace them with a single name-to-object map. This should be a performance and, I would argue, readability improvement.


phulin:
This is ready for review, although note that it's based on the GearChangeFrame PR and so has a couple extra changes in it.

midgleyc:
Could you please split out the non-refactor into a separate PR? I think this is just `isSaberForceZone`, but could also be the substring match -- but that's the same file as the refactor.

phulin:
> Could you please split out the non-refactor into a separate PR? I think this is just `isSaberForceZone`, but could also be the substring match -- but that's the same file as the refactor.

Done. Also fixed the tests.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -133,154 +124,94 @@ public record DustyBottle(int id, String name, String alias) {}
...
-        "inebriety.txt", KoLConstants.INEBRIETY_VERSION, ConsumablesDatabase.inebrietyByName);
+        "inebriety.txt", KoLConstants.INEBRIETY_VERSION, KoLConstants.CONSUME_DRINK);
     ConsumablesDatabase.readConsumptionData(
-        "spleenhit.txt", KoLConstants.SPLEENHIT_VERSION, ConsumablesDatabase.spleenHitByName);
+        "spleenhit.txt", KoLConstants.SPLEENHIT_VERSION, KoLConstants.CONSUME_SPLEEN);
     ConsumablesDatabase.readNonfillingData();
   }
 
   static {
+    for (int i = 0; i < 1 << 5; ++i) {
```
gausie, line 140: 1 << 5 is basicly a magic number here - split into a named const or calculate it staticly too

gausie: Just noticed it was always there. Still.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -329,73 +265,50 @@ private static void saveConsumptionValues(String[] data, Map<String, Integer> ma
...
+          adventures = "5-9";
         }
         case "vodka and cranberry" -> {
-          ConsumablesDatabase.qualityByName.put(name, ConsumableQuality.GOOD);
-          ConsumablesDatabase.saveAdventureRange(name, unitCost, "6-9");
-        }
-        default -> {
-          ConsumablesDatabase.qualityByName.put(name, ConsumableQuality.find(quality));
-          ConsumablesDatabase.saveAdventureRange(name, unitCost, adventures);
+          quality = ConsumableQuality.GOOD;
```
gausie, line 280: This doesn't look right

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -924,120 +833,90 @@ public static double getAdventureRange(final String name) {
...
-    muscleFactor *= ConsumablesDatabase.conditionalStatMultiplier(name);
-    int statUnit = ConsumablesDatabase.getStatUnit(name);
-    int statBonus = ConsumablesDatabase.conditionalExtraStats(name);
-    String range = ConsumablesDatabase.extractStatRange(muscle, muscleFactor, statUnit, statBonus);
+    String statRange = consumable.statRangeStrings[stat];
+    int modifier =
+        switch (stat) {
+          case Consumable.MUSCLE -> Modifiers.MUS_EXPERIENCE_PCT;
+          case Consumable.MYSTICALITY -> Modifiers.MYS_EXPERIENCE_PCT;
+          default -> Modifiers.MOX_EXPERIENCE_PCT;
```
gausie, line 883: Would rather this failed(? I know it can't) a little safer

gausie (commented):
First pass comments

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
phulin: why not? The lines are the same and in the same sequence, I just removed the default case and have it fall through below to reduce code duplication.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
phulin: Done.

phulin:
@gausie I think I increased both files to 100%, but I'm not sure how to run the action locally.

src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
```diff
@@ -1255,7 +1256,15 @@ public static final synchronized void refreshConcoctionsNow() {
     Preferences.increment("_concoctionDatabaseRefreshes");
     ConcoctionDatabase.refreshNeeded = false;
 
-    List<AdventureResult> availableIngredients = ConcoctionDatabase.getAvailableIngredients();
+    List<AdventureResult> availableIngredientsList = ConcoctionDatabase.getAvailableIngredients();
+
+    // In addition to the list, we create a second data structure here for better performance.
+    // Because we do many lookups to the available ingredients to see how many there are,
+    // having an O(1) lookup helps a lot.
+    Map<Integer, AdventureResult> availableIngredients = new ArrayListMap<>();
```
midgleyc, line 1264: I think we could specify the initial size of this map -- it's the size of the populating list

src/net/sourceforge/kolmafia/utilities/ArrayListMap.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Map from Integer to value backed by an ArrayList. Useful when keys are nonnegative integers and
+ * are dense, with very fast reads.
+ *
+ * <p>Silently drops elements with index < 0 to be consistent with existing behavior. Does not allow
+ * null values; null indicates the absence of a value.
+ *
+ * @param <V> Type of map value.
+ */
+public class ArrayListMap<V> extends AbstractMap<Integer, V> {
```
midgleyc, line 20: is this substantially faster than a `HashMap<Integer, V>` in practice?

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
gausie: Am I reading this wrong or is the default for items to be good here? I was reviewing from mobile

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
gausie: 🎉

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
midgleyc: This is under "vodka and cranberry".

But it would be good to have a test that non-Boris items have the same quality / adventures as normal on Boris.

src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
phulin: Yep, good idea.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
phulin: Added a couple lines to the test for this.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
phulin: Fixed.

midgleyc (approved):
Thanks!

PR #1237, opened 2022-11-06
Work around compiler error

Veracity0:
See discussion at this bug report.

https://kolmafia.us/threads/one-char-doesnt-run-after-upgrade-to-kolmafia-26897-m-jar.28240/


midgleyc (approved):
if it works, all good :)

PR #1238, opened 2022-11-07
Fix queue being added to without reason when pondering

libraryaddict:


gausie:
Please add a comment explaining the difference between the new functions. Or refer rename one to addAndEnqueue

gausie:
Thanks! I would say "ToMap" is overly leaky of the exact implementation but happy to approve. If you want to change it, go ahead.

gausie (approved):

PR #1246, opened 2022-11-08
Update location only if changed.

phulin:
Another performance hotspot. I don't believe there is any reason to recalculateAdjustments, setLocation, updateStatus, or fire the preference change if location hasn't changed.


gausie:
Check out the coverage report, should be very easy to hit 100% on your two new classes

jaadams5 (dismissed):

phulin:
> Check out the coverage report, should be very easy to hit 100% on your two new classes

Which ones? This PR doesn't have any new classes.

jaadams5 (approved):
But I would like @gausie to do the push when the coverage/test question is resolved to their satisfaction.

gausie:
> > Check out the coverage report, should be very easy to hit 100% on your two new classes
> 
> Which ones? This PR doesn't have any new classes.

I must have had two tabs, meant to comment this on the Consumables refactor.

PR #1248, opened 2022-11-08
More test cleanup

jaadams5:
Some tests were leaving .ser files behind in data.  This deletes them.

Also some lint, since I was already there.


jaadams5:
> It feels like it would be better to add a static flag or something that overrides `serializeAfterwards` to always be false, add and then add a cleanup that sets and unsets that flag.

OH.  I like that much better.  Let me investigate.  Thank you.

jaadams5:
#1256 has been submitted as (better) alternative to getting the file out of data.

PR #1251, opened 2022-11-08
feat(relay): allow to specify the relay port

oxc:
Allow to specify a specific relay port for a user.

I want my characters to always be reachable under the same port, regardless of which order I start them in. Also for this reason, when a relayPort is explicitly specified in preferences, there will be no checking for other free ports, only the one will be tried.

### DIFF
--- a/src/net/sourceforge/kolmafia/webui/RelayServer.java
+++ b/src/net/sourceforge/kolmafia/webui/RelayServer.java
@@ -96,18 +96,34 @@ public class RelayServer implements Runnable {
   @Override
   public void run() {
     boolean startedSuccessfully = true;
-    Integer minPort = 60080;
-    Integer maxPort = minPort + 10;
-    RelayServer.port = minPort;
-    while (!this.openServerSocket()) {
-      if (RelayServer.port < maxPort) {
-        ++RelayServer.port;
-      } else {
+
+    int relayPort = Preferences.getInteger("relayPort");
+    if (relayPort != 0) {
+      RelayServer.port = relayPort;
+      if (!this.openServerSocket()) {
         KoLmafia.updateDisplay(
             KoLConstants.MafiaState.ERROR,
-            "Failed to find free port in range " + minPort + " to " + maxPort + ".");
+            "Port " + relayPort + " which is set as relayPort is already in use.");
         startedSuccessfully = false;
-        break;
+      }
+    } else {
+      int minPort = 60080;
+      int maxPort = minPort + 10;
+      RelayServer.port = minPort;
+      while (!this.openServerSocket()) {
+        if (RelayServer.port < maxPort) {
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/webui/RelayServer.java
```diff
@@ -96,16 +96,28 @@ public static final void stop() {
   @Override
   public void run() {
     boolean startedSuccessfully = true;
-    Integer minPort = 60080;
-    Integer maxPort = minPort + 10;
+
+    int minPort, maxPort;
+    int relayPort = Preferences.getInteger("relayPort");
```
midgleyc, line 101: could you add it to defaults, please?

oxc: Ooh, I thought there must be something like this :) With which value would you add it? 0 or 60080?
If I add it with 60080, it will not work together with the "try 10 ports by default, but not if one has been set explicitly".

oxc: I've added it as 0 for now.

jaadams5: I don't think this will be successful if one character has no relayPort set and another has relayPort set to 60080 and that character logs in second.   I'd consider renaming the preference to something like characterSpecificRelayPort and having some error checking that flagged a potential error if a specific port was in the range of ports that would be checked in the absence of a specific port.  Some kind of test might be appropriate if anyone is clever enough to figure it out.

gausie: I think this is fine. Change that error message to tell the user that their explicitly chosen port is already in use if that is the case.

midgleyc: I agree. I think the cases "relayPort is set" and "relayPort is not set" are distinct. The efforts to reuse the existing `minPort` / `maxPort` stuff I think is more confusing.

I would like to see:
* is relayPort nonzero? If so, start the server there, else print error message about relayPort being inaccessible.
* otherwise, do what we did originally with a port range, printing the error message about minPort and maxPort and informing about `relayPort`.

oxc: Yes, this is also how I would have implemented your suggestions. I will do it like this tonight and get back to you.

oxc: I've implemented the changes as suggested.

midgleyc (approved):
Thanks!

oxc:
Sorry, I forgot to fetch before rebasing. Now I've dismissed your review 🤦🏻

midgleyc (approved):

PR #1252, opened 2022-11-09
Cookbookbat recipe drops tracking

Zauschneria:
Adding a property to track cookbookbat recipe drops and setting it up as a drops familiar


Zauschneria:
Yeah, it is set up for there to be 1 recipe drop per day, I could switch the property to a boolean, but with a counter it would make any accidental overcounting really obvious. I added the check for active familiar when updating the property which should prevent that first recipe from counting. I would be interested to see if buying a recipe from the mall with a cookbookbat active causes issues, but it seems like a pretty niche edge case.

Veracity0:
Actually, since you have that inside "if (adventureResults)", that should be enough.
That means the drop came as the result of adventuring - usually combat, but I think NCs count there too.
In particular, it does not include buying from the mall or, resumably, growing the familiar larva.

src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -3324,6 +3324,26 @@ private static void gainItem(boolean adventureResults, AdventureResult result) {
...
+      case ItemPool.ROBY_BORIS_BREAD:
+      case ItemPool.ROBY_ROASTED_VEGETABLE_OF_J:
+      case ItemPool.ROBY_PETES_RICH_RICOTTA:
+      case ItemPool.ROBY_ROASTED_VEGETABLE_FOCACCIA:
+      case ItemPool.ROBY_PLAIN_CALZONE:
+      case ItemPool.ROBY_BAKED_VEGGIE_RICOTTA:
+      case ItemPool.ROBY_DEEP_DISH_OF_LEGEND:
+      case ItemPool.ROBY_CALZONE_OF_LEGEND:
+      case ItemPool.ROBY_PIZZA_OF_LEGEND:
+        if (adventureResults && KoLCharacter.currentFamiliar.getId() == FamiliarPool.COOKBOOKBAT) {
```
Veracity0, line 3343: As mentioned in another combat, adventureResults precludes any way of obtaining the item that is not adventuring - combats or NC. No mall, no initial drop from growing the familiar.

You don't need to check that it is the equipped familiar.

test/net/sourceforge/kolmafia/session/ResultProcessorTest.java
```diff
@@ -252,4 +257,54 @@ public void gettingSecondMaelstromStone(int stone) {
...
+
+      try (cleanups) {
+        ResultProcessor.processResult(true, recipe);
+
+        assertThat("_cookbookbatRecipeDrops", isSetTo("1"));
+      }
+    }
+
+    @Test
+    public void cookbookbatPropertyNoUpdateIfNotActiveFamiliar() {
```
Veracity0, line 297: Change this test to not check active familiar - since that's irrelevant - to checking with combatResults false.
```
ResultProcessor.processResult(false, ItemPool.get(ItemPool.ROBY_BAKED_VEGGIE_RICOTTA))
```

Veracity0 (changes requested):

gausie:
Make it a boolean pref. If you're worried about overcounting add test cases to confirm that it doesn't happen!

Zauschneria:
So the drops familiar logic only appears to work with numbers https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanel.java#L2818-L2836 so it has to be an int without some more significant refactoring, you can see that there are already other drop familiars with only one drop per day: https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/FamiliarData.java#L1092-L1098

src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -1146,6 +1146,9 @@ private static void loadDropFamiliars() {
             "snowglobe",
             "_snowglobeDrops",
             1));
+    DROP_FAMILIARS.add(
+        new DropInfo(
+            FamiliarPool.COOKBOOKBAT, -1, "cookbookbat recipe", "_cookbookbatRecipeDrops", 1));
```
Rinn, line 1151: _cookbookbatRecipeDrops needs to be added to defaults.txt

Rinn (changes requested):

src/data/defaults.txt
```diff
@@ -1665,6 +1665,7 @@ user	_communismUsed	false
 user	_companionshipCasts	0
 user	_confusingLEDClockUsed	false
 user	_controlPanelUsed	false
+user    _cookbookbatRecipeDrops 0
```
Rinn, line 1668: need tabs not spaces

Rinn (changes requested):

libraryaddict:
Would this work as a solution for https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/FamiliarData.java#L1014 or would it be too messy/hacky?

```
    public int dropsToday() {
      if (Preferences.getDefault(this.dropTracker).equals("false")) {
        return Preferences.getBoolean(this.dropTracker) ? 1 : 0;
      }

      return Preferences.getInteger(this.dropTracker);
    }
```

gausie:
> Would this work as a solution for https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/FamiliarData.java#L1014 or would it be too messy/hacky?
> 
> ```
>     public int dropsToday() {
>       if (Preferences.getDefault(this.dropTracker).equals("false")) {
>         return Preferences.getBoolean(this.dropTracker) ? 1 : 0;
>       }
> 
>       return Preferences.getInteger(this.dropTracker);
>     }
> ```

I think that's exactly the right solution

Zauschneria:
I will convert all the other 1 drop familiars to booleans in another PR

Veracity0 (approved):

Rinn (approved):

PR #1256, opened 2022-11-09
Keep data in test clean

jaadams5:
This is an alternative to  #1244.

Rather than rewrite the custom script test as something else, before/after all were added to copy a file from request to data and delete at the end of testing.

This is probably a better solution because it can be easily extended to other custom scripts that require files in data and preserves all of the test benefits of the original.

The file was not renamed to preserve initial traceability to the original.


gausie (approved):
This is probably fine! Fwiw you can rename the files and git will be smart about traceability

PR #1258, opened 2022-11-10
Add pseudo LimitModes ASTRAL, BIRD, MOLE, ROACH

Veracity0:
Using an Astral Mushroom gives you five turns of Half-Astral.
The first adventure you take after that sends you to a choice where you choose Bad Trip, Mediocre Trip, or Great Trip.
That did not cost a turn, but the next 5 turns must be spent at a particular location.
When the effect wears off, you are no longer limited.
There is no "exit" adventure.

Using a llama lama gong sends you right into a choice adventure where you can choose among
- 3 turns of Form of...Cockroach
- 12 turns of Shape of...Mole!
- 15 turns of Form of...Bird!

The Cockroach sends you straight into an uninterruptible chain of three choices.
And the end, the effect wears off and you talk to your Llama and get a reward.

The Mole opens up Mt. Molehill where you must adventure 12 times.
When the effect wears off, your next adventure lets you talk to your Llama.

The Bird opens up no new area, but all your Combat skills are replaced with bird skills, you may not use items in combat, and you can Pickpocket. Consuming special "bug" drops open up more combat skills.
When the effect wears off, your next adventure lets you talk to your Llama and get a reward.

If you are under the effect of either an astral mushroom (Half-Astral is in effect) or a llama lama gong (one of the effects is active or it has worn off but you have not gotten the Welcome Back NC), you may not use another astral mushroom or llama lava gong.

I can model all of the above as pseudo LimitModes.
KoL returns limitmode in api.php as 0 for all of them.
1) When setLimitMode(LimitMode.NONE), if you are Half-Astral, set it to LimitMode.ASTRAL
When you you take the choice to choose a trip, set "currentAstralTrip" to Bad Trip, Mediocre Trip, Good Trip
When Half-Astral wears off, set LimitMode.NONE, which will also clear currentAstralTrip
2) While you are in LimitMode.ASTRAL, you can adventure only in your chosen trip.
If you have not chosen it, prepareForAdventure will select it, based on which location you selected.
astral mushroom and llama lama gong are limited as described above.
3) If you try to adventure in an Astral zone, canAdventure will allow it if you are already in LimitMode.ASTRAL - and have either selected that zone or have not yet selected it - and prepareForAdventure will use an astral mushroom and select the trip, if needed.
4) When you use a gong and select a form, set currentLlamaForm to Mole, Roach, or Bird
setLimitMode(LimitMode.NONE) will look at that setting and give you LimitMode.MOLE, LimitMode.ROACH, or LimitMode.BIRD.
5) When you are in LimitMode.MOLE, you can adventure only in Mt. Molehill
astral mushrooms and llama lama gongs are limited as above.
When you try to adventure again, Welcome Back will clear currentLlamaForm.
6) If you try to adventure in Mt. Molehill, canAdventure will allow it if you are not in a LimitMode and have a gong. prepareForAdventure will use a gong and select mole form, if necessary.
7) When you are in LimitMode.BIRD,
astral mushrooms and llama lama gongs are limited as above.
When the effect wears off, Welcome Back will clear currentLlamaForm.
8) When you are in LimitMode.ROACH,
astral mushrooms and llama lama gongs are limited as above. Not that you can use them in a choice chain, anyway.
When the effect wears off, Welcome Back is part of the chain and clears your currentLlamaForm

I have LimitMode tests for entering and leaving LimitMode.ASTRAL, LimitMode.MOLE, LimitMode.BIRD, and LimitMode.ROACH

------

Future considerations (perhaps not in the PR, which was to enable canAdventure/prepareForAdventure fixes):

We have hardcoded "is BIRDFORM active?" checks in various places. 
More efficient to just check if currentLimitMode is LimitMode.BIRD, rather than searching the activeEffectsList.
And maybe we could use it to check combat items and combat skills? Perhaps not.


src/net/sourceforge/kolmafia/session/ChoiceControl.java
```diff
@@ -4742,6 +4757,27 @@ public static void postChoice2(final String urlString, final GenericRequest requ
...
+            switch (ChoiceManager.lastDecision) {
+              case 1 -> "Roach";
+              case 2 -> "Mole";
+              case 3 -> "Bird";
+              default -> "";
+            };
+
+        // We are now in a pseudo LimitMode
+        Preferences.setString("currentLlamaForm", form);
+        KoLCharacter.setLimitMode(LimitMode.NONE);
```
midgleyc, line 4772: a comment saying that this has the side effect of setting the actual limit mode would be appreciated

src/net/sourceforge/kolmafia/session/LimitMode.java
```diff
@@ -46,11 +57,28 @@ public void reset() {
     resetFunction().ifPresent(Runnable::run);
   }
 
+  public void finish() {
+    switch (this) {
```
midgleyc, line 61: comment: llama forms are handled in `ChoiceControl`.

test/net/sourceforge/kolmafia/session/LimitModeTest.java
```diff
@@ -394,4 +433,175 @@ void someLimitMCD(final LimitMode lm) {
...
+        assertThat(Preferences.getString("currentAstralTrip"), is(""));
+      }
+    }
+  }
+
+  @Nested
+  class Mole {
+    @AfterEach
+    public void afterEach() {
+      KoLConstants.activeEffects.clear();
```
midgleyc, line 521: this should be unnecessary due to the cleanups that add and clear effects?

test/net/sourceforge/kolmafia/session/LimitModeTest.java
```diff
@@ -394,4 +433,175 @@ void someLimitMCD(final LimitMode lm) {
...
+    }
+  }
+
+  // *** Pseudo LimitModes
+
+  @Nested
+  class Astral {
+    @AfterEach
+    public void afterEach() {
+      KoLConstants.activeEffects.clear();
```
midgleyc, line 462: this should be unnecessary due to the cleanups that add and clear effects?

src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -669,9 +669,18 @@ public static String parseChoiceEncounter(
     }
 
     if (urlString.equals("choice.php")) {
-      // Since we cannot walk from this choice, this must have been KoL
-      // redirecting back to the choice after a page refresh
-      return null;
+      // Although KoL usually redirects to choice.php?forceoption=0, older
```
midgleyc, line 672: this is unfortunate :(

I think the only solution would be to note down where it happens and do some special exceptions, or try to get CDM involved?

midgleyc (dismissed):

src/net/sourceforge/kolmafia/request/AdventureRequest.java
Veracity0: I think I figured out a workaround

 If you get exactly "choice.php" and your last encounter is the same as this one, it is a refresh.

I don't think there will be:

choice.php (encounter name "Hello")
choice.php?whichchoice=X&option=Y
choice.php (encounter name "Hello")

I'll write some tests for this and adjust it.

Veracity0: Yes! When in Mole or Bird form, you adventure for 12 or 15 turns and the next adventure you take redirects to "choice.php" to the "Welcome Back!" encounter.

I used the HTML for leaving reincarnation as a mole and verified that "Welcome Back!" got added as an encounter.
(In the relay browser I refreshed that page multiple times and it did not count the refreshes as new encounters)
And I restored the check in the ChoiceManagerTest for detecting that situation.

test/net/sourceforge/kolmafia/session/LimitModeTest.java
Veracity0: Yeah. That was before I invented Player.withNoEffects

test/net/sourceforge/kolmafia/session/LimitModeTest.java
Veracity0: Ditto.

src/net/sourceforge/kolmafia/session/ChoiceControl.java
Veracity0: Done

src/net/sourceforge/kolmafia/session/LimitMode.java
Veracity0: Done

midgleyc (approved):

Veracity0:
I added tests for entering and leaving LimitMode.BIRD and LimitMode.ROACH.
I look forward to seeing what the final coverage numbers are.

Assuming all goes well, I will merge without awaiting further review, since all that is new is tests.

PR #1259, opened 2022-11-10
Don't add dwarf gear text if the items are not equipped.

phulin:
Evaluating the defense regex in particular is quite slow, but we shouldn't evaluate them at all if you don't have the items on.


gausie (approved):
I tend to prefer fixtures made from actual game HTML but I think this is fine. Where did you get the native text from?

phulin:
Got it from the game, although I tweaked the numbers. I can replace with a full HTML file if you like.

gausie:
Up to you really, I'm happy to merge now

PR #1261, opened 2022-11-10
More test clean up of residue in data/

jaadams5:
Added an option not to write the adventure queue and turns files and then disabled the write in tests that created them but were not cleaning them up.

Changed the username for several tests to make it easier to find which test was leaving files based on the username behind.

Lint and spotless.

Since the tests that got changed were not using the Cleanups framework I elected not to change them to do so.  baby steps.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
@@ -8,6 +8,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
+import java.io.Serial;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashSet;
@@ -32,12 +33,10 @@ import net.sourceforge.kolmafia.utilities.RollingLinkedList;
  */
 
 public class AdventureQueueDatabase implements Serializable {
-  private static final long serialVersionUID = -180241952508113931L;
+  @Serial private static final long serialVersionUID = -180241952508113931L;
 
-  private static TreeMap<String, RollingLinkedList<String>> COMBAT_QUEUE =
-      new TreeMap<String, RollingLinkedList<String>>();
-  private static TreeMap<String, RollingLinkedList<String>> NONCOMBAT_QUEUE =
-      new TreeMap<String, RollingLinkedList<String>>();
+  private static TreeMap<String, RollingLinkedList<String>> COMBAT_QUEUE = new TreeMap<>();
+  private static TreeMap<String, RollingLinkedList<String>> NONCOMBAT_QUEUE = new TreeMap<>();
 
   // for testing only, otherwise leave at true;
   public static boolean allowSerializationWrite = true;
@@ -85,16 +84,15 @@ public class AdventureQueueDatabase implements Serializable {
   }
 
   private static void resetQueue(boolean serializeAfterwards) {
-    AdventureQueueDatabase.COMBAT_QUEUE = new TreeMap<String, RollingLinkedList<String>>();
-    AdventureQueueDatabase.NONCOMBAT_QUEUE = new TreeMap<String, RollingLinkedList<String>>();
+    AdventureQueueDatabase.COMBAT_QUEUE = new TreeMap<>();
+    AdventureQueueDatabase.NONCOMBAT_QUEUE = new TreeMap<>();
 
     List<KoLAdventure> list = AdventureDatabase.getAsLockableListModel();
 
     for (KoLAdventure adv : list) {
-      AdventureQueueDatabase.COMBAT_QUEUE.put(
-          adv.getAdventureName(), new RollingLinkedList<String>(5));
+      AdventureQueueDatabase.COMBAT_QUEUE.put(adv.getAdventureName(), new RollingLinkedList<>(5));
       AdventureQueueDatabase.NONCOMBAT_QUEUE.put(
-          adv.getAdventureName(), new RollingLinkedList<String>(5));
+          adv.getAdventureName(), new RollingLinkedList<>(5));
     }
 
     if (serializeAfterwards) {
@@ -199,8 +197,7 @@ public class AdventureQueueDatabase implements Serializable {
         ObjectOutputStream out = new ObjectOutputStream(fileOut);
 
         // make a collection with combat queue first
-        List<TreeMap<String, RollingLinkedList<String>>> queues =
-            new ArrayList<TreeMap<String, RollingLinkedList<String>>>();
+        List<TreeMap<String, RollingLinkedList<String>>> queues = new ArrayList<>();
         queues.add(COMBAT_QUEUE);
         queues.add(NONCOMBAT_QUEUE);
         out.writeObject(queues);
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
```diff
@@ -187,22 +188,23 @@ public static RollingLinkedList<String> getZoneNoncombatQueue(String adv) {
...
-      // make a collection with combat queue first
-      List<TreeMap<String, RollingLinkedList<String>>> queues =
-          new ArrayList<TreeMap<String, RollingLinkedList<String>>>();
-      queues.add(COMBAT_QUEUE);
-      queues.add(NONCOMBAT_QUEUE);
-      out.writeObject(queues);
-      out.close();
-    } catch (IOException e) {
-      e.printStackTrace();
+    if (allowSerializationWrite) {
```
midgleyc, line 191: would probably prefer to see this as:
```
if (!allowSerializationWrite) return;
```

to make it obvious the method does nothing else

src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
```diff
@@ -236,24 +238,14 @@ public static void deserialize() {
...
-      file.delete();
-      AdventureQueueDatabase.resetQueue();
-      return;
-    } catch (EOFException e) {
-      // Malformed data. Wipe the bogus file.
-      file.delete();
-      AdventureQueueDatabase.resetQueue();
-      return;
-    } catch (IOException e) {
+    } // Old version of the combat queue handling.  Sorry, have to delete your queue.
```
midgleyc, line 246: this combination has put the comments in the wrong place.

src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
```diff
@@ -132,17 +136,19 @@ public static int getTurns(final String loc) {
...
-
-    try {
-      FileOutputStream fileOut = new FileOutputStream(file);
-      ObjectOutputStream out = new ObjectOutputStream(fileOut);
-
-      out.writeObject(AdventureSpentDatabase.TURNS);
-      out.close();
-    } catch (IOException e) {
-      e.printStackTrace();
+    if (allowSerializationWrite) {
```
midgleyc, line 139: would probably prefer to see this as:
```
if (!allowSerializationWrite) return;
```

to make it obvious the method does nothing else

src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
```diff
@@ -170,36 +176,30 @@ public static void deserialize() {
...
       file.delete();
       AdventureSpentDatabase.resetTurns();
-      return;
-    } catch (ClassCastException e) {
-      // Old version of the combat queue handling.  Sorry, have to delete your queue.
-      file.delete();
-      AdventureSpentDatabase.resetTurns();
-      return;
-    } catch (IOException e) {
+    } // Old version of the combat queue handling.  Sorry, have to delete your queue.
```
midgleyc, line 184: this combination has put the comments in the wrong place.

test/net/sourceforge/kolmafia/request/AdventureRequestTest.java
```diff
@@ -30,6 +35,12 @@ public class AdventureRequestTest {
   public void init() {
     KoLCharacter.reset("AdventureRequestTest");
     Preferences.reset("AdventureRequestTest");
+    AdventureQueueDatabase.allowSerializationWrite = false;
+  }
+
+  @AfterEach
+  public void restore() {
```
midgleyc, line 42: could be `AfterAll`. No test wants this to be true.

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -18,16 +18,27 @@
...
 public class AscensionRequestTest {
   @BeforeAll
   public static void initializeCharPrefs() {
-    KoLCharacter.reset("fakePrefUser");
+    KoLCharacter.reset("AscensionRequestFakeUser");
     KoLCharacter.reset(true);
     // Fix another test not cleaning up
     FightRequest.currentRound = 0;
+    AdventureQueueDatabase.allowSerializationWrite = false;
+    AdventureSpentDatabase.allowSerializationWrite = false;
```
midgleyc, line 35: why set it twice?

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
```diff
@@ -18,16 +18,27 @@
...
+  }
+
+  @AfterAll
+  public static void restoreWritePermissions() {
+    AdventureQueueDatabase.allowSerializationWrite = true;
```
midgleyc, line 40: why unset it twice?

midgleyc (changes requested):
Many comments.

Additionally, on the tests: we may want to set it to false in `ClearSharedStateBefore`, same as we do for `sessionId`, because while the tests are running we never want it true.

src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
jaadams5: Ok

src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
jaadams5: Not my natural idiom but OK.  Long long ago in a structured language no one remembers the coding standard was one return per function.  That habit persists even when it approaches being a Bad Habit :-)

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
jaadams5: Because I used the same name and AdventureQueueDatabase is not AdventureSpentDatabase ?

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
jaadams5: Because I used the same name and AdventureQueueDatabase is not AdventureSpentDatabase ?

Perhaps I should rename the variable?

test/net/sourceforge/kolmafia/request/AdventureRequestTest.java
jaadams5: I was mirroring the BeforeEach.  If I set it BeforeEach then I should clear it AfterEach but I don't think it matters until a test wants to write the file.

src/net/sourceforge/kolmafia/persistence/AdventureQueueDatabase.java
jaadams5: ok

src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
jaadams5: ok

test/net/sourceforge/kolmafia/request/AscensionRequestTest.java
midgleyc: oh, sorry. They have the same number of characters, so I didn't notice.

It's fine to leave it.

test/net/sourceforge/kolmafia/request/AdventureRequestTest.java
midgleyc: I suppose I think of it as a performance thing: with before each / after each it sets and unsets it every test. With before each / after all it sets it every test (perhaps unnecessarily), but only unsets it at the end: fewer actions.

src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
midgleyc: Delphi?

I think this pattern is called a "guard clause" nowadays. It tends to be preferred from what I've seen.

jaadams5:
I think I addressed all of @midgleyc 's comments.  That said, I think the AfterEach to AfterAll was not the right move.  I need to run the tests locally which has gotten harder since something has doubled the wall clock time to do so...

midgleyc (approved):

jaadams5:
Local test worked.  Thanks.

PR #1262, opened 2022-11-10
Cleanup another loose file in data

jaadams5:
A custom script creates a file but does not delete it.  Rewriting the script didn't work for me so the test fixture deletes it.

I chose to do this in CleareSharedState because it was easy and made sense.  A case could be made that ClearSharedStateAfter is an artifact of a time when we were still discovering how we wanted to do testing and that it should be deprecated in favor of making the tests do their own cleanup.  But I'm not making that case now.


jaadams5:
Tests are passing locally on Windows :-(  I will dig into this but not immediately.

jaadams5:
Apparently content-types.properties needs to be in data on ubuntu.  Will defer deleting it for the moment.

Veracity0:
Regarding "tests cleaning up after themselves".

Suppose I have HTML from KoL of a FightRequest. Straight from my DEBUG log.

The "use items" dropdown on the Fight page shows all 60 combat usable items I have in inventory.

KoLmafia parses that as part of handling the Fight and concludes that inventory must be out of sync. Every item it didn't know about is added to inventory.

So, after parsing the Fight page, there are now an extra 60 items in inventory.

What is your preferred technique for handling this?

midgleyc:
I think "ClearSharedStateAfter" is a good thing. Many tests can modify large amounts of state, and it's easier to just have a single file that takes down as much as possible between tests.

As a good example, take `GenericRequest.sessionId = null`. Anything faking requests wants to set this and have it unset at the end of the test. It's easier to do it in one place to be sure it gets done.

gausie:
> I think "ClearSharedStateAfter" is a good thing. Many tests can modify large amounts of state, and it's easier to just have a single file that takes down as much as possible between tests.
> 
> As a good example, take `GenericRequest.sessionId = null`. Anything faking requests wants to set this and have it unset at the end of the test. It's easier to do it in one place to be sure it gets done.

To expand on this: I think that we should continue to be more and more explicit with cleanups with a perhaps unattainable goal of removing ClearSharedStateAfter, but should embrace it until then

jaadams5:
I think we are in agreement that some after everything cleanup is still appropriate.  So is this use of that capability Approved?  :-)

gausie (approved):

PR #1264, opened 2022-11-10
Simplify new equipment list.

phulin:
This code is a bottleneck. It's possible there is a reason the current code is the way it is, but I can't see one - unless there is some object-identity issues that are being worked around. newItems is mutated in this method but then the new value is never referenced afterwards.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
+++ b/src/net/sourceforge/kolmafia/swingui/panel/GearChangePanel.java
@@ -1236,8 +1236,7 @@ public class GearChangePanel extends JPanel {
       final LockableListModel<AdventureResult> currentItems,
       final List<AdventureResult> newItems,
       final AdventureResult equippedItem) {
-    currentItems.retainAll(newItems);
-    newItems.removeAll(currentItems);
+    currentItems.clear();
     currentItems.addAll(newItems);
     currentItems.setSelectedItem(equippedItem);
   }


jaadams5:
> This code is a bottleneck.

Any basis for this assertion besides intuition and static analysis?

phulin:
> Any basis for this assertion besides intuition and static analysis?

Profiling while running a typical adventuring script indicates this function is taking something like 7-8% of execution time, because retainAll iterates through every element of the relevant list and checks whether the other list contains it.

jaadams5:
> > Any basis for this assertion besides intuition and static analysis?
> 
> Profiling while running a typical adventuring script indicates this function is taking something like 7-8% of execution time, because retainAll iterates through every element of the relevant list and checks whether the other list contains it.

OK.  I presume you have before and after timings? :-)

I too wish I knew why it was done that way but I cannot construct a test data where the results are different.  And the existing tests covered before and after so at least things were exercised.

IntelliJ reports several Lint type things that could be addressed.  If you are running it, you might consider doing Analyze and addressing most of the issues.  If not I'll try and do that, probably for another PR.

I'll Approve but wait a while before I push just in case someone does know why things were done the way they were.

jaadams5 (approved):

midgleyc:
I don't see an issue with this change.

`currentItems` is a `SortedListModel`. If it weren't, after this change the list could be in a different order to before, as it would contain all the items it contained before, followed by all the ones it contained after. But because it's sorted, this isn't a problem.

The other change is to remove `newItems.removeAll(currentItems);`. `newItems` comes from `equipmentLists`, which is generated from scratch in `updateAllModels` using `List<List<AdventureResult>> equipmentLists = this.populateEquipmentLists();` and never referenced after this method, so this removal seems fine too.

PR #1265, opened 2022-11-10
Reduce frequency of name-to-id-to-name conversions.

phulin:
`getEffectModifiers` currently starts with an effect id, which then gets turned into a lookup string, which then gets parsed back into an id in `getLookupName`. Instead we should just use the id.


phulin:
Will fix tests shortly.

midgleyc (approved):
Thanks!

PR #1266, opened 2022-11-11
Delete TCRS files when created as side effect of test

jaadams5:
Another PR to clean out data/ after tests.

Could not figure out how to put the deletes in a Cleanup unless there was a new one for TCRS,


gausie:
My only comment is that this could probably be a nice Java 8 stream expression

jaadams5:
> My only comment is that this could probably be a nice Java 8 stream expression

I'll look at that and try harder :-)_

jaadams5:
> My only comment is that this could probably be a nice Java 8 stream expression

This feels awkward to me but...

gausie:
> > My only comment is that this could probably be a nice Java 8 stream expression
> 
> This feels awkward to me but...

It looks great to me! What feels awkward?

gausie (approved):

gausie:
One small change is that I would map the stream `toFile()`

test/net/sourceforge/kolmafia/persistence/ConsumablesDatabaseTest.java
```diff
@@ -297,6 +302,16 @@ static void afterAll() {
       DebugDatabase.cacheItemDescriptionText(
           ItemPool.RING, html("request/test_normal_desc_item_ring.html"));
       TCRSDatabase.resetModifiers();
+      File root = KoLConstants.DATA_LOCATION;
+      String[] contents = root.list();
+      if (contents != null) {
```
Aenimus: If you are going to keep this iteration, I would recommend a compare to null and return early rather than nest everything.

gausie: Agreed, though the stream is better

jaadams5:
> > > My only comment is that this could probably be a nice Java 8 stream expression
> > 
> > 
> > This feels awkward to me but...
> 
> It looks great to me! What feels awkward?

I'm not convinced that Files.walk is the place to start although I didn't really find an alternative.  I had a hard time getting it to work because I wasn't always sure when I had a File and when I had a Path.  And I still am inclined to write in Java < 8 or so and letting the IDE teach me there are language changes that make something better :-)

test/net/sourceforge/kolmafia/persistence/ConsumablesDatabaseTest.java
jaadams5: Right.  That interaction is gone but as discussed elsewhere I have an old, bad? habit from a time when one return per function was expected.  I don't think it was ALGOL W so it was probably FORTRAN with preprocessing so the coder could pretend there was some structure.  The preprocessor probably has not existed for several decades.

PR #1269, opened 2022-11-12
feat: cookbookbat cooks

midgleyc:
Cookbookbat offers 5 free cooks per day.

Not too sure as to whether the ConcoctionDatabase changes are right: I don't understand what `adventureSmithingLimit` and the like do, and whether we need a corresponding one for cooking. Also everything that refers to Jewelrycrafting taking adventures can presumably be removed?


gausie:
This probably wants tests for crafting order

midgleyc:
I think this probably wants tests all over the place ;). It's just a lot of work to set up for the first time.

Plus ConcoctionDatabase wants a bit of cleanup after tests are in place.

gausie (approved):

midgleyc:
Not having any luck making these tests. In `ConcoctionTest`:

```java
@Test
public void smithingTakesAdventures() {
  var cleanups = new Cleanups(withItem(ItemPool.TENDER_HAMMER), withMeat(1000), withAdventuresLeft(5));

  ConcoctionDatabase.refreshConcoctions();

  var con = ConcoctionPool.get(ItemPool.MACE_OF_THE_TORTOISE);
  assertThat(con.getAdventuresNeeded(1), is(1));
}
```
This fails, and debugging says the character can't smith. I don't understand how or why.

Veracity0:
> Not having any luck making these tests. In `ConcoctionTest`:
> 
> ```
> var con = ConcoctionPool.get(ItemPool.MACE_OF_THE_TORTOISE);
> ```
> 
> This fails, and debugging says the character can't smith. I don't understand how or why.

Mace of the Tortoise is not creatable; it is given to you by the ghost of The Unknown Turtle Tamer.

PR #1271, opened 2022-11-12
feat: Oliver's Place

midgleyc:
Dump of items, effects, zone etc.

Yeah it's mid outdoor in-game, hope that'll get changed, can recheck later.


Veracity0 (approved):
This is a good start. I see all sorts of things that will be needed in code, rather than data (a setting to track whether you've used it. prepareForAdventure, canAdventure support, and so on.

The new zone depends on an item - either the deed or, if there ends up being one, a day-pass.
I assume some of the new items are monster drops.
Is the government per-diem reusable? Once per day, I assume. Breakfast?
Presumably, the deed can only be used once per character and there is a message.
Do we need to record the Steam code? I assume not; presumably, like the horsery, it is on the plaque on the wall.

This is fine for now.

midgleyc:
Yes, there is a lot to add. The pool table gives meat once per day with enough pool skill.

The code is present, but it's on the graffitied table, not the plaque. The plaque lets you rename the place.

PR #1273, opened 2022-11-12
tests: concoction tests

midgleyc:
Follow-up to #1269. Started rewriting the world, so split out.

### DIFF
--- a/test/internal/helpers/Player.java
+++ b/test/internal/helpers/Player.java
@@ -2,6 +2,7 @@ package internal.helpers;
 
 import static org.mockito.Mockito.mockStatic;
 
+import internal.helpers.Cleanups.OrderedRunnable;
 import internal.network.FakeHttpClientBuilder;
 import internal.network.FakeHttpResponse;
 import java.net.http.HttpClient;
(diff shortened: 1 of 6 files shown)


test/internal/helpers/Player.java
```diff
@@ -1789,4 +1789,9 @@ public static Cleanups withValueOfAdventure(final int value) {
     cleanups.add(ConcoctionDatabase::refreshConcoctions);
     return cleanups;
   }
+
+  public static Cleanups withConcoctionRefresh() {
+    ConcoctionDatabase.refreshConcoctions();
+    return new Cleanups(new OrderedRunnable(ConcoctionDatabase::refreshConcoctions, 10));
```
jaadams5, line 1795: Magic number?  Not having finished everything I presume there is a desire to have something clean up last and 10 is an arbitrary value that will do that until there are > 10 cleanups?

jaadams5 (approved):
Consider addressing the magic number 10 but that's not a blocker.

test/internal/helpers/Player.java
midgleyc: yeah, any number greater than one is fine.

PR #1276, opened 2022-11-12
feat: remove turn-costing jewelcrafting, add turn-free cooking

midgleyc:
Jewelrycrafting doesn't take turns, so remove code that checks free crafts related to it.

Cooking now has a separate source of free crafts, so add the "cookingLimit" and "turnFreeCookingLimit" concoctions. Still not really sure what this area of the code does.


jaadams5 (approved):
Is there an opportunity for a test?

midgleyc:
Not more than I've already done easily.

Most of the functions are hard to call, or hard to test -- e.g. "what does calculate3 do?" is a reasonable question.

PR #1279, opened 2022-11-13
The gong command finishes your journey of reincarnation

Veracity0:
We had this funky code in the responsetext parsing of UseItemRequest to detect that you hadn't finished you journey of reincarnation when you attempted to use a llama lama gong. If your effect had run out, it would visit the Noob Cave and accept your reward and then retry using the item.

Turns out this was to allow the "gong" command to immediately use a gong even if the user hadn't seen the Welcome Back adventure.

1) I moved the equivalent to the gong command itself: before it tries to use a gong, it sees if it needs to visit the Welcome Back NC, and if so, does it by visiting the Noob Cave.

2) I also learned that UseItemRequest now tells GenericRequest to follow redirects, which will leave it in a choice when the initial choice was so-followed; it did not automate the choice. Which is to say, this command has been broken ever since we made UseItemRequest automatically follow redirects.

   I added a fluid method to UseItemRequest to allow this instance to direct GenericRequest to NOT follow the redirect, but to automate the choice chain.

Rant:

It is really irritating when I make tests that work very hard to not leak state, but when incorporated into the whole set of tests, completely unrelated tests start failing. In this case, when I stuck a debug print into the failing tests, they stopped failing on my system when running the whole suite.

I managed to get one to print something useful: "You are currently in a fight or a choice" - which is why retrieving the stash failed.

The was not me. I called choices, but verified with assertions that I was not leaking ChoiceManager.handlingChoice.

But, due to having an extra file of tests, apparently a test which DID leak moved in front of other tests and broke them.

I added more stuff to ClearSharedStateBefore and the tests no longer fail.

Global static variables are making me more and more irritated.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/UseItemRequest.java
+++ b/src/net/sourceforge/kolmafia/request/UseItemRequest.java
@@ -20,7 +20,6 @@ import net.sourceforge.kolmafia.SpecialOutfit.Checkpoint;
 import net.sourceforge.kolmafia.ZodiacSign;
 import net.sourceforge.kolmafia.moods.ManaBurnManager;
 import net.sourceforge.kolmafia.moods.RecoveryManager;
-import net.sourceforge.kolmafia.objectpool.AdventurePool;
 import net.sourceforge.kolmafia.objectpool.EffectPool;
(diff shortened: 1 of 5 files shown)


gausie (approved):
Super support!

PR #1280, opened 2022-11-13
Keep consumable attributes in TCRS

gausie:
Attributes like SAUCY and CANNED are true in TCRS if they are true in the game in general. They're useful for diet planning so I'm pulling them through.

A few observations:
* I made a few small improvements to tests
* We should collect the actual substat awards from consumables - they don't get rerolled like potions do so it shouldn't be an effort that has to be repeated.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
@@ -12,6 +12,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.TreeMap;
 import java.util.stream.IntStream;
 import net.java.dev.spellcast.utilities.DataUtilities;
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -674,22 +675,26 @@ private static void addEffectSource(
 
   private static void applyConsumableModifiers(
       final int usage, final String itemName, final TCRS tcrs) {
-    Integer lint = ConsumablesDatabase.getLevelReqByName(itemName);
-    int level = lint == null ? 0 : lint.intValue();
+    var consumable = ConsumablesDatabase.getConsumableByName(itemName);
+    Integer lint = ConsumablesDatabase.getLevelReq(consumable);
+    int level = lint == null ? 0 : lint;
```
Veracity0, line 680: I haven't studied all usages of getLevelReqByName (which wraps getLevelReq), but I have to wonder why returning null, if it is not present, is better than returning 0.

Investigating/fixing that is not required for this PR.

Veracity0 (approved):

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
gausie: Good point. I'm not sure why it does that either

PR #1281, opened 2022-11-13
Adventure Validation for Oliver's Place

Veracity0:
1) Added property: ownsSpeakeasy
- If you use the deed, set to true
- we see it in town_wrong, set to true
2) preValidateAdventure will look for it in town_wrong, if ownership unknown
It will also detect ownership if you have a milk cap (a quest item) in inventory

I don't know if there will be a day pass.
The Horsery does not have one. That has no factoids.
Boxing Daycare does have one. That also has no factoids.

I'm guessing not, since this is associated with a steam game, like the Horsery.
But if so, we can adjust.


gausie:
I'm also not sure that "Oliver's Place" is the best name, as it can be renamed. "Speakeasy" might be better, since that is the whichplace for it.

Veracity0:
I agree that Speakeasy is a better zone name than Oliver's place and "ownsSpeakeasy" is a better property.
Will fix.

An Easter Egg in the HTML fixture I included - an update to the map for the Wrong Side of the Tracks - is that the Speakeasy is named "Veracity's Place". KoL builds the image label out of individual letters, rather than having it be part of a .gif, as in other map images.

gausie:
> I agree that Speakeasy is a better zone name than Oliver's place and "ownsSpeakeasy" is a better property. Will fix.
> 
> An Easter Egg in the HTML fixture I included - an update to the map for the Wrong Side of the Tracks - is that the Speakeasy is named "Veracity's Place". KoL builds the image label out of individual letters, rather than having it be part of a .gif, as in other map images.

We should probably store the name in a pref, just for fun. No need to be in this PR

gausie (approved):

PR #1282, opened 2022-11-13
Yet another test file in data/ to be cleaned up.

jaadams5:
This should fail on Ubuntu.


jaadams5:
The Ubuntu test fails with

2022-11-14T00:13:13.9291441Z build/test/net/sourceforge/kolmafia/CustomScriptTest.class:83 | org.opentest4j.AssertionFailedError: test_finally_return.ash output does not match:  ==> expected: <You gave me 50 Good enough! I received 50 You gave me 100 I prefer 101 I received 101 You gave me 200 No way!> but was: <You gave me 50 Good enough! I received 50 You gave me 100 I prefer 101 I received 101 You gave me 200 No way! I received 0>

Running on Windows the final "I received 0" is not there.  I'll leave this out there for a bit in case anyone who runs js on Ubuntu wants to look at it.

jaadams5:
Test is no longer failing so...

gausie (approved):

PR #1283, opened 2022-11-14
Improve banish manager tests + ice house handling

gausie:
- Improve ice house support and modernise BanishManager tests
- Add some missing coverage


midgleyc (approved):
Thanks!

PR #1284, opened 2022-11-14
Store speakeasy name

gausie:


Veracity0 (approved):
I love it. I have similar code in my PR which makes a CoinMaster, but this is more thorough and has tests.

PR #1285, opened 2022-11-14
Fancy Dan coinmaster and other Oliver's Place changes

Veracity0:
- Fancy Dan is a coinmaster that takes milk caps and drink chits
- One government per-diem can be used per day (_governmentPerDiemUsed) and it not consumed if you try to use a second one.
- Add "spend" use link when you get a milk cap or a drink chit.
- Log visits in selected areas of the Speakeasy


midgleyc (approved):
Thanks!

PR #1286, opened 2022-11-14
Reinitialize encoded values on reset.

phulin:
This was causing a bug when `saveSettingsOnSet` was off, so when a user closed mafia it was accidentally saving the old values.

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -548,6 +548,8 @@ public class Preferences {
 
   /** Resets all settings so that the given user is represented whenever settings are modified. */
   public static synchronized void reset(String username) {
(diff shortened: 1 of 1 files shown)


Veracity0 (approved):

Veracity0 (approved):
Well, one comment.

If Preferences.saveSettingsToFile is false, we won't actually write the file.
That is ONLY for use in tests.

So, re-encoding them here is wasted work if that variable is false - which will marginally slow down tests.

You may want to think about that, but it's not going to block this, which is an important user-affecting fix.

phulin:
> Well, one comment.
> 
> If Preferences.saveSettingsToFile is false, we won't actually write the file. That is ONLY for use in tests.
> 
> So, re-encoding them here is wasted work if that variable is false - which will marginally slow down tests.
> 
> You may want to think about that, but it's not going to block this, which is an important user-affecting fix.

Will do. Thanks for the quick mege.

PR #1287, opened 2022-11-15
Fluid construction for CoinmasterData

Veracity0:
If you have ever added a new Coinmaster, you've noticed that you need to create a Coinmaster Data for it.
This object has a single BOA constructor with 25 arguments.
Almost every Coinmaster uses only a fraction of those arguments.
Your best hope is to find a Coinmaster that behaves fairly closely to the one you want to create and copy and modify.

This PR adds "fluid" construction of CoinMasterData.

For example:
```
  public static final CoinmasterData DINOSTAUR =
      new CoinmasterData(master, "Dinostaur", DinostaurRequest.class)
          .withToken("Dinodollar")
          .withTokenPattern(TOKEN_PATTERN)
          .withItem(COIN)
          .withShopRowFields(master, "dino");
```
The fields can come in any order, of course.
withShopRowFields adds 9 fields that are used by any shop.php requests that use "rows"
Fields that do not apply to the Coinmaster simply default and are not specified or used.

Here's another one:
```
  public static final CoinmasterData FANCY_DAN =
      new CoinmasterData(master, "olivers", FancyDanRequest.class)
          .withShopRowFields(master, "olivers")
          .withBuyPrices()
          .withItemBuyPrice(FancyDanRequest::itemBuyPrice);

  private static AdventureResult itemBuyPrice(final Integer itemId) {
    return buyCosts.get(itemId);
  }
```
Notice that it has no "item" (since it trades in multiple items) or "token" (ditto) or "tokenPattern" (ditto).
And since it has multiple currencies, it needs a custom method to get the buy cost (an AdventureResult of the currency needed - itemId and count).

I added javadoc comments to all of the fluid "withFIELD" methods.
Tell me if you think they are suitable.

Future PR:

- Use LockableListFactory to generate lists for buyItems and sellItems
- More support for multiple currencies
- More support for shops with variable inventory.


gausie:
> If you have ever added a new Coinmaster, you've noticed that you need to create a Coinmaster Data for it.

I did it once and have been scared to do so ever since!

So thank you!! I love this. It's probably worth doing good javadoc strings here too so they're super fluid to author in a good IDE. Thank you thank you!

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
-  private final boolean canPurchase;
+  private String itemField = null;
+  private Pattern itemPattern = null;
+  private String countField = null;
+  private Pattern countPattern = null;
+  private String storageAction = null;
+  private String tradeAllAction = null;
+  private boolean needsPasswordHash = false;
+
+  // False if the coinmster doesn't sell anything that goes into
```
midgleyc, line 76: coinmaster

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
+  /**
+   * Defines text that indicates you have at least one token available.
+   *
+   * <p>Occasionally (we have one example), a coinmaster has a single token, but instead of a simple
+   * string indicating no tokens are available, there is a string indicating you have at least one.
+   *
+   * <p>This is mutually exclusive with <code>tokenTest</code>. Provide one or the other (or
+   * neither).
+   *
+   * @param positiveTest - String indicating at least one token is available
```
midgleyc, line 165: it's not a string, but a boolean indicating whether "tokenTest" is a positive or negative test

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
+   * Defines text that indicates you have no tokens available to spend.
+   *
+   * <p>When examining the responseText that shows the coinmaster's inventory, usually the available
+   * amount of each currency is displayed.
+   *
+   * <p>If there is only one currency - the token - and there is a simple string indicating that
+   * none are available, provide it here. If it is present, we need not do a pattern search to
+   * calculate the number of available tokens.
+   *
+   * <p>This is mutually exclusive with <code>positive</code>. Provide one or the other (or
```
midgleyc, line 145: This seems wrong -- it's actually the case that this is a negative test by default, but if `positiveTest` is true it's a positive test.

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
+  public CoinmasterData withProperty(String property) {
     this.property = property;
+    return this;
+  }
+
+  /**
+   * Specifies that the "rows" of this <code>shop.php</code> coinmaster are all listed in <code>
+   * coinmasters.txt</code>.
+   *
+   * <p>If the coinmaster uses modern <code>shop.php</code>, every purchasble item has a "row"
```
midgleyc, line 229: purchasable

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
+   */
+  public CoinmasterData withItemRows(String master) {
+    return withItemRows(CoinmastersDatabase.getRows(master));
+  }
+
+  /**
+   * Provides the "rows" of this <code>shop.php</code> coinmaster when they are not (all) specified
+   * in <code>coinmasters.txt</code>.
+   *
+   * <p>If the coinmaster uses modern <code>shop.php</code>, every purchasble item has a "row"
```
midgleyc, line 246: purchasable

src/net/sourceforge/kolmafia/CoinmasterData.java
```diff
@@ -6,119 +6,662 @@
...
+  /**
+   * Provides a <code>List</code> of <code>AdventureResult</code>s of the items that you can buy
+   * from this coinmaster.
+   *
+   * <p>If not all of the items are in <code>coinmasters.txt</code> - or a <code>List</code> has
+   * been created ahead of time elsewhere - pass it in here.
+   *
+   * <p>*** It is important that the <code>List</code> be a <code>LockableListModel</code>, since it
+   * will be used as the model for this coinmaster in <code>CoinmastersFrame</code> (I need to
+   * figure out how to enforce this, since that will not be the case if we are running headless)
```
midgleyc, line 321: if we're running headless there is no frame so it doesn't matter, right?

midgleyc (changes requested):
I didn't go over the requests in detail, but here's a review of the documentation.

Thanks for doing this, should make coinmasters simpler.

src/net/sourceforge/kolmafia/CoinmasterData.java
Veracity0: Yes. Except our convention seems to be this:
```
  List<AdventureResult> activeEffects = LockableListFactory.getInstance(AdventureResult.class);
```
rather than this, from CoinmastersDatabase:
```
  public static final LockableListModel<AdventureResult> getNewList() {
    return new LockableListModel<AdventureResult>();
  }
```
I want to study this more and, perhaps, deal with it in a subsequent PR.

src/net/sourceforge/kolmafia/CoinmasterData.java
Veracity0: Done. Also the typo in the next line.

src/net/sourceforge/kolmafia/CoinmasterData.java
Veracity0: Oh, right.

src/net/sourceforge/kolmafia/CoinmasterData.java
Veracity0: Right. fixed.

src/net/sourceforge/kolmafia/CoinmasterData.java
Veracity0: Fixed

midgleyc (approved):

PR #1292, opened 2022-11-16
Allow to_item to accept floats

libraryaddict:
In some incidents when running javascript your numbers will be converted into floats, despite no obvious operations done. That is to say, 1 will turn into 1.0 and throw your code into disarray. And yes, you could be using Item.get() instead. But we expose toItem and should support it.

Realistically, we don't want to be accepting floats. We could reluctantly accept 1.0 but we don't want to accept 1.1

In practical experience, it doesn't really matter if we accept decimal places.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -3431,7 +3431,8 @@ public abstract class RuntimeLibrary {
   }
 
   public static Value to_item(ScriptRuntime controller, final Value value) {
-    if (value.getType().equals(DataTypes.TYPE_INT)) {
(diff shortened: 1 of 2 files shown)


gausie:
Firstly, add tests to confirm that's 4.1 gets cast to 4. I would rather this failed though. Secondly, I think this fix should happen in JavaScript land, not in RuntimeLibrary.

In fact, I'll put a fix in today if you can't.

libraryaddict:
> Firstly, add tests to confirm that's 4.1 gets cast to 4. I would rather this failed though. Secondly, I think this fix should happen in JavaScript land, not in RuntimeLibrary.
> 
> In fact, I'll put a fix in today if you can't.


Yeah, that would be best. I don't understand that area very well

PR #1294, opened 2022-11-16
Track oliver's place free fights

horrible-little-slime:


gausie:
Thanks Neil. Can you add tests?

horrible-little-slime:
> Thanks Neil. Can you add tests?
> 
> We should also set to 3 if we see a (1) next to the location in the Speakeasy.
> 
> I also note that "Looks like the fight is brawl is heating up, further encounters here will cost an adventure." appears in combat when we hit 3, so that is worth using to update things.

Done

gausie:
Amazing! The only thing missing is the checking for a `(1)` against the location I think

gausie (approved):
Perfect thank you

PR #1297, opened 2022-11-17
Avoid aggressive dev gifting

gausie:
Exclude the stash because the user may hit clan karma limits, which will cause mafia to attempt to acquire the difference from the mall.

I've just sent someone 🥩10,000,000 who fell foul to this and would rather not keep refunding people in that way :)


Veracity0 (approved):
Ouch.

PR #1298, opened 2022-11-17
Rename Oliver's Place to Speakeasy

gausie:
We call it Speakeasy elsewhere


horrible-little-slime:
defaults.txt is no longer alphabetical

gausie:
Oh no! I'll fix that later

PR #1301, opened 2022-11-18
Add missing "drops items" override for the Puck Men

gausie:
- Add Puck Man and Ms Puck Man to have a variable DROPS_ITEMS modifier
- Modernize some switches

I want to make a familiardrops.txt that captures this info and the Drop Familiars info in FamiliarData too, but that's for another time

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -2257,136 +2257,131 @@ public class Modifiers {
 
   private boolean overrideItem(final int itemId) {
     switch (itemId) {
-      case ItemPool.TUESDAYS_RUBY:
-        {
-          // Set modifiers depending on what KoL day of the week it is
-          var dotw = DateTimeManager.getArizonaDateTime().getDayOfWeek();
+      case ItemPool.TUESDAYS_RUBY -> {
+        // Set modifiers depending on what KoL day of the week it is
+        var dotw = DateTimeManager.getArizonaDateTime().getDayOfWeek();
 
-          this.set(Modifiers.MEATDROP, dotw == DayOfWeek.SUNDAY ? 5.0 : 0.0);
-          this.set(Modifiers.MUS_PCT, dotw == DayOfWeek.MONDAY ? 5.0 : 0.0);
-          this.set(Modifiers.MP_REGEN_MIN, dotw == DayOfWeek.TUESDAY ? 3.0 : 0.0);
-          this.set(Modifiers.MP_REGEN_MAX, dotw == DayOfWeek.TUESDAY ? 7.0 : 0.0);
-          this.set(Modifiers.MYS_PCT, dotw == DayOfWeek.WEDNESDAY ? 5.0 : 0.0);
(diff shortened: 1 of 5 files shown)


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2257,136 +2257,131 @@ public static final String trimModifiers(final String modifiers, final String re
...
   private boolean overrideSkill(final String name) {
-    switch (name) {
-      case "Ferocity":
-        if (KoLCharacter.isVampyre()) {
-          this.set(Modifiers.HP, -10.0);
-        } else if (KoLCharacter.isAvatarOfBoris()) {
-          this.set(Modifiers.CRITICAL_PCT, 25.0);
-        }
-        return true;
+    if ("Ferocity".equals(name)) {
```
Veracity0, line 2336: Why remove the switch?

It is true that there is only one case, but if another skill comes along, we'll just have to turn it back into a switch in order to add a second case.

gausie: I let IntelliJ take the wheel here.

Though this reminds, me that it feels off. One of these is `[11002]Ferocity` and the other is `[24017]Ferocity`. Is there a way we can just maintain that disambiguation?

Veracity0: Use those as the names? SkillDatabase.getSkillId() will accept a bracketed number as the skill id.

gausie: I *think* it's a bit more complicated than that

gausie: Yes I see - modifier lookups for skills don't deal in ids at all. I have a bit of code to write.

Veracity0: What am I missing?
```
# Ferocity (Boris): Critical Hit Percent: +25
Skill	[11002]Ferocity	Critical Hit Percent: +25
# Ferocity (Vampyre): Your combat skills do 50% more damage
Skill	[24017]Ferocity	Spell Damage Percent: +50, Maximum HP: -10
```
yields:
```
> ash $skill[Ferocity].name

Multiple matches for "Ferocity"; using "[24017]Ferocity". (char 8 to char 16) Clarify by using one of:
$skill[[11002]Ferocity]
$skill[[24017]Ferocity]
Returned: Ferocity

> ash string_modifier($skill[[11002]Ferocity], "Modifiers")

Returned: Critical Hit Percent: +25

> ash string_modifier($skill[[24017]Ferocity], "Modifiers")

Returned: Spell Damage Percent: +50, Maximum HP: -10
```

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/Modifiers.java
gausie: Actually having the skill will not match as a result

gausie: My following commits fix the issue :)

Veracity0 (approved):

PR #1303, opened 2022-11-18
Use LockableListFactory

Veracity0:
1) "Recipe of Before Yore: baked veggie ricotta" has recipe for "baked veggie ricotta casserole"
2) CoinmasterDatabase uses LockableListFactory to create LockableListModel objects
3) stuffed dinosaur has power 10


midgleyc (approved):
Thanks!

PR #1305, opened 2022-11-19
Fix context menu bug

gausie:
Accidentally introduced a bug with the watchable prefs feature, fixed it here.

PR #1306, opened 2022-11-19
fix context menus always showing first element of list or table

Veracity0:
ShowDesciptionList and showDescriptionTable each had a "lastSelectIndex" field which was set by the context menu popup. When an element in the context menu was selected, the action was invoked on the element at that index, unless the index was -1, in which case it would fetch the current selected index.

Those variables were not initialized. Hence, as ints, they defaulted to 0.

When PopupListener was refactored into an independent widget, it no longer had access to set "lastSelectedIndex".

Therefore, ShowDescriptionList/ShowDescriptionTable looked at the uninitialized index, got a zero, and always executed the action on element 0.

This PR removes lastSelectedIndex from both ShowDescriptionList and ShowDescriptionTable and instead gets the currently selected index - which PopupLIstener ensured was valid; if not, the popup was not displayed.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/listener/PopupListener.java
+++ b/src/net/sourceforge/kolmafia/swingui/listener/PopupListener.java
@@ -62,7 +62,7 @@ public class PopupListener extends MouseAdapter {
 
     if (!table.isRowSelected(row)) {
       table.clearSelection();
-      table.addRowSelectionInterval(row, row);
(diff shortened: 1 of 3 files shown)


Veracity0:
gausie beat me to it by five minutes. :)

https://github.com/kolmafia/kolmafia/pull/1305

gausie:
Sorry!

PR #1307, opened 2022-11-19
Go back to lazy parsing of modifier strings.

phulin:
Switching to eager parsing of modifier strings had the unfortunate effect of making `resetModifiers` and thus `KoLCharacter.reset` much slower, causing the test suite to time out. This PR returns to lazy parsing, but this time keeping strings and modifiers in two separate maps for better typing.


phulin:
(Current build has checks run in 15m 31s on GitHub, this PR has checks run in 4m 11s)

gausie (approved):

PR #1312, opened 2022-11-23
If in gnoob, check charpane on login for absorbs done

libraryaddict:
Interestingly enough, it does check charpane in the very first call of a session request.

But it does that for valhalla and has not loaded the api at this point so is not aware of the character's class.


Veracity0:
It's been a long time since I've done this path, but I am confused by this.
If I recall, you absorb items and either get enchantments or special passive skills.
I assume the passive skills are on the charsheet.
What does the charpane show you?

Veracity0:
What does "check charpane for absorbs done" mean?

Haven't done this path for a long time. I see we go to charpane rather than api.

Why? What is this doing?

libraryaddict:
> What does "check charpane for absorbs done" mean?
> 
> Haven't done this path for a long time. I see we go to charpane rather than api.
> 
> Why? What is this doing?

We can absorb up to 15 items, the API doesn't keep track of how many we have absorbed. We must fetch how many absorbs we've completed as far from the charpane itself.

gausie:
Maybe update your code comment to explain a little more what is being done. Also in general, write PR descriptions as if we don't know anything about the change - your description here is an addendum missing the actual text!

gausie:
Now do it again but much shorter :)

midgleyc:
"Get current absorbs" is likely enough, in the same vein as the You Robot energy comment.

Why don't we know the class when parsing the char sheet for the first time? Where do we learn it?

libraryaddict:
> "Get current absorbs" is likely enough, in the same vein as the You Robot energy comment.
> 
> Why don't we know the class when parsing the char sheet for the first time? Where do we learn it?

We don't know it because we're parsing it to test for valhalla. Refreshing works because by this point it's already learned our class.

gausie:
I genuinely don't like the tone of the commit messages so I'm closing this PR

libraryaddict:
> I genuinely don't like the tone of the commit messages so I'm closing this PR

Ah I apologize, when I said "dumb it down" it was meant in the context of "simplify it down" not "dumb it down to his level". The "for gausie" was meant as a ping and a "for him" given you'd asked for the changes.

But yeah, fair.

It wasn't meant in an antagonistic way, I apologize.

gausie:
Ok. I appreciate the explanation

gausie (approved):

PR #1315, opened 2022-11-25
Track consumption of T4 Cookbookbat foods

horrible-little-slime:


test/net/sourceforge/kolmafia/request/EatItemRequestTest.java
```diff
@@ -61,4 +64,44 @@ void milkResponseSetsPreference() {
...
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    "10991, pizzaOfLegendEaten",
+    "10992, calzoneOfLegendEaten",
+    "11000, deepDishOfLegendEaten"
+  })
+  public void canTrackCookbookbatFoodsSuccess(Integer itemId, String prefname) {
+    assertFalse(Preferences.getBoolean(prefname));
```
Veracity0, line 75: I would put all of these tests inside something like:
```
    try (withProperty(prefname, false) {
    }
```
so that the pref is restored to its original state after each test finishes.

As is, each of the three tests leaves the property set to true.

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/request/EatItemRequestTest.java
horrible-little-slime: I assumed the BeforeEach would handle that, but it's good practice to use a cleanups regardless, so I'll do just that.

Veracity0 (approved):

PR #1317, opened 2022-11-26
fix: allow eating a glitch monster after using it

midgleyc:
Allow eating a glitch monster after using it. This is the only item which is both usable and edible, which some parts of the code assume is impossible. The drunki-bears have both fullness and drunkenness, and so should probably also be treated differently by maximumUses, but for now aren't.

Some things I'm not too sure about:
* `EatItemRequest` looks at the main consumption type of an item. I hope it just deals with food and food helpers like the salad fork.
* I don't know why it didn't pass the consumption type initially, instead doing the fullness check. Code reuse?
* In general this part of the codebase doesn't have enough tests for me to be confident doing large changes.


midgleyc:
okay, so from reading the code fortune cookie is /also/ usable and edible, and this is handled in at least one place. Might try some refactoring / tests.

jaadams5 (approved):

PR #1318, opened 2022-11-26
refactor: change KoLConstant's consume types to an enum

midgleyc:
Change consumption types to an enum, and rename for clarity. Container stays container for now, but I am tempted to make it "BACK" as containers haven't been relevant for a long time!

Commit by commit, the first has the actual refactor, and the others were automated reformatting and renaming.

Add "Unknown" for the purpose that "-1" was serving previously.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
@@ -493,7 +493,7 @@ public class ConcoctionDatabase {
     ConsumptionType consumpt = ItemDatabase.getConsumptionType(id);
 
     if (c.getFullness() > 0
-        || consumpt == ConsumptionType.CONSUME_FOOD_HELPER
+        || consumpt == ConsumptionType.FOOD_HELPER
         || ConcoctionDatabase.canQueueFood(id)) {
       queue = ConcoctionDatabase.queuedFood;
       queuedIngredients = ConcoctionDatabase.queuedFoodIngredients;
@@ -502,7 +502,7 @@ public class ConcoctionDatabase {
         ConcoctionDatabase.queuedInebriety++;
       }
     } else if (c.getInebriety() > 0
-        || consumpt == ConsumptionType.CONSUME_DRINK_HELPER
+        || consumpt == ConsumptionType.DRINK_HELPER
         || ConcoctionDatabase.canQueueBooze(id)) {
       queue = ConcoctionDatabase.queuedBooze;
       queuedIngredients = ConcoctionDatabase.queuedBoozeIngredients;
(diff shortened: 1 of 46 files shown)


src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
```diff
@@ -838,18 +839,18 @@ private static int lastUnconsumed(int quantity, ConcoctionType type) {
...
-    // KoLConstants.NO_CONSUME - create or retrieve items
-    // KoLConstants.CONSUME_EAT - eat food items
-    // KoLConstants.CONSUME_DRINK - drink booze items
-    // KoLConstants.CONSUME_SPLEEN - use spleen items
-    // KoLConstants.CONSUME_GHOST - binge ghost with food
-    // KoLConstants.CONSUME_HOBO - binge hobo with booze
-    // KoLConstants.CONSUME_USE - use potions
-    // KoLConstants.CONSUME_MULTIPLE - use potions
-    // KoLConstants.CONSUME_AVATAR - use potions
+    // ConsumptionType.NO_CONSUME - create or retrieve items
```
Veracity0, line 845: These are not the new enum tags.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -619,17 +620,17 @@ public static void setEffectName(final Integer itemId, String name) {
   }
 
   private static void addEffectSource(
-      final String itemName, final int usage, final String effectName) {
+      final String itemName, final ConsumptionType usage, final String effectName) {
     int effectId = EffectDatabase.getEffectId(effectName);
     if (effectId == -1) {
       return;
     }
     String verb =
```
Veracity0, line 628: I've not mentioned converting to modern switch statements, but this would be a perfect use for a switch expression.

Veracity0 (changes requested):
Regarding "container": KoL itself uses "type=container" in URLs.
EquipmentRequest.slotNames has "back"
EquipmentRequest.phpSlotNames has "container"
EquipmentRequest.CONTAINER_PATTERN has "container"

A future PR could certainly replace almost all usage of "container" to "back", with comments explaining the handful of remaining "container".

I also did not suggest modernizing many switch statements with "->" (although you could ) because, as you say, test coverage is sparse.

src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
midgleyc: updated, though I'm not sure whether the comment is still correct.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
midgleyc: I avoided switching to expressions to keep the difference lower, but I've converted this and another ternary if.

Veracity0 (approved):

PR #1321, opened 2022-11-27
feat: convert attributes to enumset

midgleyc:
Further refactoring to add types.

Could do with more (or, for the bulk of it, any) tests. Will run through on my turns tomorrow to check for major problems.


gausie:
I think this is vital for modernising out codebase so thank you. We perhaps need to sort out all the different types of Attribute, as the term is used everywhere.

midgleyc:
For these item attributes:
* the first 4 are acccess modifiers: tradeable, discardable, gift, quest. These are duplicated in, and initially read from, `accessById`.
* the remaining the "secondary uses" as in items.txt. These act as flags. Some refer to a KoL property (e.g. chocolate) and some to a KoLMafia-only thing (e.g. matchable).

jaadams5 (approved):
Lots to look at and I fear I missed something but....

PR #1322, opened 2022-11-28
Cookbookbat quest tracking

horrible-little-slime:
I may or may not have time to work on this more later in the week; if somebody has the time and inclination before then, I encourage them to take the reins.

I ran into a snaggle when I was trying to account for all of the cookbookbat quest messages--some of them don't contain all the information of the quest, which both makes the quest-resetting logic annoying and makes a lot of the parsing logic annoying--trying to check a matcher for a capture group that doesn't exist on it will throw an exception. Luckily, it always gives all of the necessary information when you _start_ the quest, so I think we can get away with ignoring the other messages.

Here, "cookbookbatQuestTurns" is being used to track two separate things, which may be problematic. We're using it to determine how long the current quest has been active _or_ how long we haven't had a quest. Both are worth tracking!


gausie:
True story: the first thing I did was scroll down to look at test assertions just so I could see what the expected behavior was and comment on the `cookbookbatQuestTurns` usage. This is not a requirement but I strongly suggest - write tests before writing implementation where possible!

We could do something smart with monster / zone: if we see a zone or monster that doesn't match our current understanding (if we have one) wipe both prefs and update this one. Then when we see the other pref we can just update that too.

jaadams5:
>write tests before writing implementation where possible!

Another true story - there is a software development process - Test Driven Development - that (simplistically) writes the tests before the code.  In my experience using it didn't always make sense but it worked very well when you were starting from almost a clean slate or the new thing was a single class or object.  Since this is not quite straightforward I also wonder of writing the tests first would be helpful.

I also note the assertion that the complete information is only guaranteed when the quest is started.  That suggests there needs to be some provision in case the quest is started outside of KoLmafia.

gausie:
@horrible-little-slime how you doin bb

horrible-little-slime:
> @horrible-little-slime how you doin bb

Poorly, in terms of the energy I have to devote to this!

PR #1323, opened 2022-11-30
More fights that are only sometimes free

gausie:
- More monsters that are conditionally free
- Tidy up file
- Add tests for conditional free fights

### DIFF
--- a/src/net/sourceforge/kolmafia/MonsterData.java
+++ b/src/net/sourceforge/kolmafia/MonsterData.java
@@ -87,10 +87,6 @@ public class MonsterData extends AdventureResult {
     }
   }
 
-  static {
-    Attribute[] attributes = Attribute.values();
-  }
-
   // ***********************************************************
   // Parsing an attribute string - from monsters.txt or manually
   // crafted by MonsterManuelManager - into a map from Attribute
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -87,10 +87,6 @@ public String getOption() {
     }
   }
 
-  static {
-    Attribute[] attributes = Attribute.values();
```
midgleyc, line 91: I believe this was here to call the constructor for every attribute, to populate `optionToAttribute`.

Is it unnecessary?

gausie: Yeah, the constructors are called when the enum is loaded by the ClassLoader

src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -1838,51 +1788,33 @@ public boolean willUsuallyMiss(final int defenseModifier) {
   }
 
   public void appendItemDrops(StringBuilder buffer) {
```
midgleyc, line 1790: please add some tests for this refactor (confirming the same behaviour). I think it's probably fine, but it's hard to be sure purely by reading.

src/net/sourceforge/kolmafia/MonsterData.java
Veracity0: That is not correct.

- I commented out that line and recompiled.
- logged in a character.
- gCLI reports this:
```
Monster: "1335 HaXx0r": unknown option: Atk:
Monster: "1335 HaXx0r": unknown option: 81
Monster: "1335 HaXx0r": unknown option: Def:
Monster: "1335 HaXx0r": unknown option: 72
Monster: "1335 HaXx0r": unknown option: HP:
Monster: "1335 HaXx0r": unknown option: 71
Monster: "1335 HaXx0r": unknown option: Init:
Monster: "1335 HaXx0r": unknown option: 50
Monster: "1335 HaXx0r": unknown option: Meat:
Monster: "1335 HaXx0r": unknown option: 30
Monster: "1335 HaXx0r": unknown option: P:
Monster: "1335 HaXx0r": unknown option: dude
Monster: "1335 HaXx0r": unknown option: EA:
Monster: "1335 HaXx0r": unknown option: "bad
Monster: "1335 HaXx0r": unknown option: spelling"
Monster: "1335 HaXx0r": unknown option: Article:
Monster: "1335 HaXx0r": unknown option: a
```

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/MonsterData.java
gausie: @midgleyc @Veracity0 hmm. ok, fixed it I think

src/net/sourceforge/kolmafia/MonsterData.java
gausie: I knew I should do this and to my shame I chose not to

gausie: Done!

Veracity0 (approved):

src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -1676,13 +1629,22 @@ public boolean isDrippy() {
...
-        && Preferences.getInteger("_snojoFreeFights") < 10) {
-      EnumSet<EncounterType> temp = this.type.clone();
+    // Some monsters are conditionally free
+    var forceFree =
+        switch (this.name) {
+          case "X-32-F Combat Training Snowman" -> Preferences.getInteger("_snojoFreeFights") < 10;
+          case "biker", "\"plain\" girl", "jock", "party girl", "burnout" -> Preferences.getInteger(
+                  "_neverendingPartyFreeTurns")
+              < 10;
+          default -> false;
```
Veracity0, line 1639: How about _speakeasyFreeFights?

gausie: The monsters themselves aren't inherently free - you could fax or wish them and they'd cost adventures. NEP monsters are free by definition - in a standard run right now you can even reminisce one for a free scaling fight!

The snowman is actually more questionable but is essentially true since it can't be copied.

src/net/sourceforge/kolmafia/MonsterData.java
midgleyc: another good test would be velvet underground dweller, who only has bounty items.

PR #1324, opened 2022-11-30
Not allowed to use hatter command in TCRS

gausie:
The potion won't ever give you the effect because it is nohookah, so this command will just waste a DRINK ME


midgleyc (approved):
Probably don't need a fake client because the test never checks the requests, but minimal harm to have it.

Thanks!

gausie:
> Probably don't need a fake client because the test never checks the requests, but minimal harm to have it.
> 
> Thanks!

Yeah this is for someone who comes along to test the rest of the command, which I decided not to do

PR #1325, opened 2022-12-01
fix: fix maximum uses for non-ASCII named items

midgleyc:
As reported in https://kolmafia.us/threads/divide-by-zero-error-when-consuming-spleen-item.28353/. Fix from Veracity, tests from me.


Veracity0 (dismissed):
Thank you.

PR #1327, opened 2022-12-02
Tests and fixes for PHP Random functions

gausie:
@phulin and I are working on reverse-engineering the code for TCRS item modification. On gausie/kolmafia we're close to correctly generating the mods (including the precise item name change) for every item in every seed. Expect a PR on that soon, negating the need for the TCRS datafiles!

But in doing so, it seemed pertinent to test my implementations of PHP's rand functions. @phulin found that PHPMTRandom didn't deal well with negative seeds (not used in any of the instances we've handled so far so we didn't notice) so I have fixed that issue too.

- Add tests for PHPRandom
- Add tests for PHPMTRandom
- Add tests for negative seeds
- Fix bugs with negative seeds in PHPMTRandom


jaadams5 (approved):
tests are always good :-)

PR #1328, opened 2022-12-02
The mechanic with dusty bottles having aliased names was retired many years ago, I think this code can too

gausie:


Veracity0:
It is true that there is no longer a game mechanic that makes the bottles vary, but you are removing visible information which does no harm by being present.

Is the removed information in the Notes field in inebriety.text, at least?

gausie:
> It is true that there is no longer a game mechanic that makes the bottles vary, but you are removing visible information which does no harm by being present.
> 
> Is the removed information in the Notes field in inebriety.text, at least?

Do the bottles still get those names at all? I was unsure.

Veracity0:
> Do the bottles still get those names at all? I was unsure.
What names?

Yes, they are still "dusty bottle of {Marsala, Merlot, Muscat, Pinot Noir, Port, Zinfandel}.
Same as always, from when they were first implemented.
And, same as always, if you look at the descriptions while you have Lord Spookyraven's spectacles, they display an arcane rune. The rune identifies the quality - {glassy, average, bad, spooky, vinegar, great}.

What changed back in 2014 is that the displayed rune does not change per player per ascension; each dusty bottle of wine is now always associated with the same rune/effect.

Right now (before this PR), a bottle of wine displays the adjective as part of the name.
That includes in the item manager.
The "Notes" field tells you what that means:
"bad" -> lose 60-70% of hit points
"vinegar" -> 10 Full of Vinegar

But, as I mentioned:
```
> inv dusty

dusty bottle of glassy Marsala
dusty bottle of average Merlot
dusty bottle of great Zinfandel
```
you can see what you have without even looking at the notes.

This PR removes the adjective everywhere (inventory, the mall, ...), which, as I commented, removes information.

The fact that it was originally there for historical reasons and the bottles don't vary any more is not a particularly compelling reason to remove information, to my eye.

Are you intending to get rid of all aliases that don't vary across ascensions? Like "El Vibrato Punchcard (ATTACK)"? After all, you could just use "El Vibrato Punchcard (115 holes)".

gausie:
> > Do the bottles still get those names at all? I was unsure.
> > What names?
> 
> Yes, they are still "dusty bottle of {Marsala, Merlot, Muscat, Pinot Noir, Port, Zinfandel}. Same as always, from when they were first implemented. And, same as always, if you look at the descriptions while you have Lord Spookyraven's spectacles, they display an arcane rune. The rune identifies the quality - {glassy, average, bad, spooky, vinegar, great}. Note that KoL itself did not remove that legacy behavior - associating a rune with a dusty bottle type - even though it no longer has any in-game effect.
> 
> What changed back in 2014 is that the displayed rune does not change per player per ascension; each dusty bottle of wine is now always associated with the same rune/effect.
> 
> Right now (before this PR), a bottle of wine displays the adjective as part of the name. That includes in the item manager. The "Notes" field tells you what that means: "bad" -> lose 60-70% of hit points "vinegar" -> 10 Full of Vinegar
> 
> But, as I mentioned:
> 
> ```
> > inv dusty
> 
> dusty bottle of glassy Marsala
> dusty bottle of average Merlot
> dusty bottle of great Zinfandel
> ```
> 
> you can see what you have without even looking at the notes.
> 
> This PR removes the adjective everywhere (inventory, the mall, ...), which, as I commented, removes information.
> 
> The fact that it was originally there for historical reasons and the bottles don't vary any more is not a particularly compelling reason to remove information, to my eye.
> 
> Are you intending to get rid of all aliases that don't vary across ascensions? Like "El Vibrato Punchcard (ATTACK)"? After all, you could just use "El Vibrato Punchcard (115 holes)".

No, I think I just misunderstood totally what was going on here. I'll close this PR

PR #1330, opened 2022-12-05
Track the timezone set in Kingdom of Loathing

libraryaddict:
Although the default timezone is `Africa/Abidjan` from my personal experience, I've left the default of this preference as empty to indicate that mafia does not know what timezone has been set.

The use of this change is to help the user identity what timezone is being used when the game presents a time/date using the timezone set in the player's settings.

This allows future changes to mafia where the time can be converted back into a unix time for display purposes and for internal calculation. Such as store logs shown using local time, instead of the provided time by kol.

My own use is to filter store purchases to retrieve the purchases made between two date points, such as the time during which a script was run, or displaying purchases separated by kol day.

This PR could be extended further with the addition of either adding a record for TimeZone, or adding functions to retrieving the offset of a timezone, and daylight savings data.

However given the complicated nature of daylight savings, I've left that alone in interests of scripts using a library instead to handle the timezone, such as [moment](https://momentjs.com/).


src/net/sourceforge/kolmafia/request/AccountRequest.java
```diff
@@ -205,6 +205,11 @@ private static void parseCombatOptions(final String responseText) {
     KoLCharacter.setAutoAttackAction(autoAttackAction);
   }
 
+  private static final Pattern TIMEZONE_PATTERN =
+      Pattern.compile("<select name=\"timezone\">.*?</select>", Pattern.DOTALL);
+  private static final Pattern SELECTED_PATTERN_2 =
+      Pattern.compile("<option selected='selected'>(.*?)</option>");
```
midgleyc, line 211:
```suggestion
      Pattern.compile("<option selected='selected'>([^<]+)</option>");
```
might be slightly faster. Minor, though.

gausie:
>That suggests to me that the only benefit to capturing the setting in KoLmafia is for a user who can't remember how they set it and doesn't want to use a browser or a user who changes the account setting frequently.

Or for a script (probably a relay script) that want to run successfully on every user's mall log? Though it sounds like it's perfectly reasonable for such a script to just visit the relevant account page itself.

jaadams5:
> Or for a script (probably a relay script) that want to run successfully on every user's mall log? Though it sounds like it's perfectly reasonable for such a script to just visit the relevant account page itself.

To compare data from different users that have different account time zones you either need to capture what the time zone was when the data was dumped or the dumping process needs to normalize to a known timezone.

If I'm logged in and dumping data I can capture the timezone if I need it without making KoLmafia manage a preference.

I'm thinking this is a case where if the only user of the preference is a script, it is simple enough for the script to find out what it is.  If KoLmafia needed such a preference it would be in connection with a new Feature that injected the value to any output stream that had dates or used the preference to convert all the dates in the stream to a known time zone.

To the extent that mafia is doing the writing the simplest thing to do might just be to extend timestamps so that they always included the timezone.

I'm not going to have a hissy fit if it is committed but it does seem like something that is very niche.

libraryaddict:
Yeah, I'm going to close this as I think it'd be nicer for the script to manually check the first time it's run each day for example.

So that when the user does change their timezone without mafia knowing, the script still catches it.

gausie:
I think that since this is never needed in a limited mode (like the fight
or choice lock) there's no reason for the script not to pull the data
itself. I agree. We can close the PR and reject this feature

PR #1331, opened 2022-12-05
Let user configure Violet Fog under all three Astral zones.

Veracity0:
ChoiceAdventures defines configurable choices specifying the "zone" where the configuration gets sorted.
Even though a given choice only appears in a single adventure.php area (with two exceptions), we've chosen to collect all the choices from areas in a zone into the same panel.

That makes sense, I guess, for all the areas under "Woods" - Spooky Forest, Hidden Temple, Whitey's Grove.
Or HiddenCity. Or Island (Hippies, Fratboys, Pirates). Or McLarge (mine, extreme slope).

But it's trickier if you want to collect choices from areas that are not in the same Zone but are in different subzones.

Consider Spookyraven Manor. Manor0 - Manor3 are Basement, 1st, 2nd, and 3rd floors. Each contains three or four adventure.php areas. And several in Manor1 and several in Manor2 contain chioce adventures. We put those choices in zone "Manor1" and zone "Manor2". But what about "Light's Out"? That is a rare choice which appears in adventure.php locations on all four floors. We have chosen to put it in Manor1. We COULD put it in "Manor" - and all the other choices in the Manor there - but there is no mechanism for a parent zone like that to hold choice adventure configuration.

As we saw with the other case: Violet Fog appears in the {Bad, Mediocre, Great} Trip adventure.php areas - each of which is its own zone (since it was easier to support adventure validation). All are subzones of Astral.

One cannot share the same JComponent in multiple Swing containers. So, I need a single configuration panel for all three Astral subzones - and the ChoiceOptionPanel needs to associate it with each KoLAdventure's whose parentZone - not zone - is Astral.

That's what I did here.

If we wanted to combine the Manor1 and Manor2 choices into Manor, we could do the same thing and Lights Out would appear for every KoLAdventure in Spookyraven.

I did not do that.

I confess to being unpleased with ChoiceOptionsPanel. You change one dropdown? EVERY choiceAdventure setting is updated. Update a single preference? EVERY dropdown is updated. We have PreferenceCheckboxes elsewhere. Seems like each GUI element could have a single PreferenceListener associated.

That would be a big change.

I'm also not sure how to test almost anything in this package.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/panel/ChoiceOptionsPanel.java
+++ b/src/net/sourceforge/kolmafia/swingui/panel/ChoiceOptionsPanel.java
@@ -7,6 +7,7 @@ import java.awt.event.ActionListener;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.TreeMap;
 import javax.swing.JButton;
 import javax.swing.JComboBox;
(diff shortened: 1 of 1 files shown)


Veracity0:
I want to think about this some more.
Maybe going back to all three trips being directly in Astral again.

Veracity0:
> Maybe going back to all three trips being directly in Astral again.

I now recall the issues. The problem is that the three kinds of trip are mutually exclusive and if they are each their own zone, it simplifies the code.

Having them in the same zone would require LimitMode.limitAdventure to NOT simply call limitZone.
I could do that, but that would be a more complicated project.
Doing it would not invalidate THIS PR, which, as mentioned, is a potential solution for Lights Out.

I'm going with this, for now.

Veracity0:
Huh. Maybe that LimitMode change would be the hardest part. I DO have ample tests for that already.
I'll take another look.

midgleyc (approved):

PR #1332, opened 2022-12-07
Coinmaster Tokens

jaadams5:
I am prepared to be told this is the wrong way to do this or that it is not a good idea at all.

There are a couple of Coinmasters that use multiple tokens and there is no way for a script to know what the possible tokens might be.  I did not want to break anything so I added a new Coinmaster proxy for multiple tokens.  A Coinmaster either has exactly one token in which the proxy token is used and the proxy multiple_tokens is "" OR it has multiple tokens and the proxy token is "" and the proxy multiple_tokens is a comma delimited list of tokens.

I then decided it would be nice if any token that was really an item would be recognized as such in a script.  So I changed some token values to match the corresponding item name.

I decided LUCRE needed to be FILTHY_LUCRE for no especially good reason.

I ran lint on ProxyrecordValue since I was already there.

I wrote two custom test scripts.  One checks that every Coinmaster has a proxy value for token or multiple_tokens but not both.  The other makes a list of Coinmasters that have non-item tokens.

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
@@ -610,7 +610,7 @@ public class ItemPool {
   public static final int ANTIQUE_HAND_MIRROR = 2092;
   public static final int TOWEL = 2095;
   public static final int GMOB_POLLEN = 2096;
-  public static final int LUCRE = 2098;
+  public static final int FILTHY_LUCRE = 2098;
   public static final int ASCII_SHIRT = 2121;
   public static final int TOY_MERCENARY = 2139;
   public static final int EVIL_TEDDY_SEWING_KIT = 2147;
--- a/src/net/sourceforge/kolmafia/request/AWOLQuartermasterRequest.java
+++ b/src/net/sourceforge/kolmafia/request/AWOLQuartermasterRequest.java
@@ -20,7 +20,7 @@ public class AWOLQuartermasterRequest extends CoinMasterRequest {
 
   public static final CoinmasterData AWOL =
       new CoinmasterData(master, "awol", AWOLQuartermasterRequest.class)
-          .withToken("commendation")
+          .withToken("A. W. O. L. commendation")
           .withTokenPattern(TOKEN_PATTERN)
           .withItem(COMMENDATION)
           .withBuyURL("inv_use.php?whichitem=5116&ajax=1")
--- a/src/net/sourceforge/kolmafia/request/BountyHunterHunterRequest.java
+++ b/src/net/sourceforge/kolmafia/request/BountyHunterHunterRequest.java
@@ -22,11 +22,11 @@ public class BountyHunterHunterRequest extends CoinMasterRequest {
   private static final Pattern TOKEN_PATTERN =
       Pattern.compile("You have.*?<b>([\\d,]+)</b> filthy lucre");
 
-  public static final AdventureResult LUCRE = ItemPool.get(ItemPool.LUCRE, 1);
+  public static final AdventureResult LUCRE = ItemPool.get(ItemPool.FILTHY_LUCRE, 1);
 
   public static final CoinmasterData BHH =
       new CoinmasterData(master, "hunter", BountyHunterHunterRequest.class)
-          .withToken("lucre")
+          .withToken("filthy lucre")
           .withTokenTest("You don't have any filthy lucre")
           .withTokenPattern(TOKEN_PATTERN)
           .withItem(LUCRE)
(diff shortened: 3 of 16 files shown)


Veracity0:
I have to look a lot more closely at this. A few comments based on superficial scanning.

- You have a shit load of changes which have nothing to do with this. IDE suggestions? Whatever; made it a lot harder to discern the actual PR changes.
- Mr. Store has Mr. A and Uncle B.
- I believe token names appear in the title of the appropriate panel of the CoinmastersFrame. They were intentionally shortened to fit in that context. You have lengthened them. Do they still fit?
- Armory and Leggery tokens change every year. That is why it looks them up when you visit. Your intention is that we make a code change every year?

I'll comment more intently ehen zi am on my computer, rather than my phone. :)

jaadams5:
Thank you.

"ran lint" was supposed to be understood as short hand for "ran a static code analyzer and accepted the suggestions that seemed reasonable".  If this continues as a viable PR then I can certainly back those out and put them in a different PR.

I missed Uncle Bucks.  Will add.

I will confirm that the token name is used the GUI.  If this means I have to resize GUI elements than it is too intrusive compared to what I am trying to do.

I don't like the A&L implementation either but it doesn't bother me because every year we have to add new rewards as items and a new trophy for Standard so adding two new currencies seemed like a reasonable yearly chore.

I'm ultimately interested in a scriptable way to generate a list of Coinmaster tokens.  This change lets me iterate over all Coinmasters to do so.  An alternative solution might be a RTL function is_Coinmaster_Token(item) and iterate over items.

Thank you again.

Veracity0:
First of all, look at CoinmasterData.
```
A coinmaster deals in currencies other than Meat. If there is only one such currency, we call it the "token"
```
Note also that there is a method to derive all the currencies by looking at the "itemBuyPrice" of all items for sale:
```
  public Set<AdventureResult> currencies() {
    if (this.currencies == null) {
      this.currencies = new TreeSet<>();
      for (AdventureResult item : this.buyItems) {
        this.currencies.add(this.itemBuyPrice(item.getItemId()));
      }
    }
    return this.currencies;
  }
```
For Coinmasters with a static inventory with multiple currencies - the SpinMasterLathe, for example - we set them up at class initializatin time.
For Coinmasters with variable inventory and multiple currencies - MrStore, for example - we set them up when you visit the store and see what is on offer for what price.

Do not use "multiple tokens" with hard coded lists. Use "currencies" and let the CoinmasterData TELL you what they are.

I suggest you figure out which Coinmasters need to be visited to initialize their currencies.
Note that if a Coinmaster's inventory depends on player state - The Armory and Leggery, in particular - visiting will only detect items that are available for currencies you actually have.

Personally, I don't see the value of listing all the "pulverized standard rewards items". But, perhaps the Armory and Leggery could programmatically derive what its potential currencies are and what each is tradeable for. If you pulverize a Hardcore Standard reward, you can trade it for (any of) previous year's Hardcore rewards, right?

Perhaps we could have a data file: standard_rewards.txt:
```
2015	Hardcore	porcelain powder	Ceramic Suit	
2015	Normal	polyesterene powder	Synthetic Suit
2016	Hardcore	bakelite bits	Bakelite Brigandine
2016	Normal	wickerbits	Wicker Wear
2017	Hardcore	wrought-iron flakes	Wrought Wrappings
2017	Normal	aerosolized aerogel	Aeroutfit
2018	Hardcore	fiberglass fibers	Fiberglass Finery
2018	Normal	gabardine smithereens	Gabardine Guise
2019	Hardcore	marble molecules	Marble Materials
2019	Normal	chalk chunks	Chalk Chostume
2020	Hardcore	terra cotta tidbits	Terra Cotta Tackle
2020	Normal	paraffin pieces	Paraffinalia
2021	Hardcore	stained glass shards	Stained Glass Suit
2021	Normal	velour veneer	Velour Vestments
2022	Hardcore	flagstone flagments	Flagstone Finery
2022	Normal	loofah lumps	Loofah Loungewear
```
If you can trade the pulverized item for any of the (6) previous year's rewards, the outfit defines those.

jaadams5:
Thank you.  I had already decided to close this because Mr. Store actually has two tokens and using my new convention for multiple_tokens meant that I would break existing scripts because it would give them an empty string as the token.  The wisdom of that is supported by the fact that the token is used in the Coinmaster GUI and there is already code there that is handling multiple currencies (at least for Mr. Store).  This is not worth the overhead of doing the same thing differently in two places.

I agree that Set<AdventureResult> currencies() is a much better alternative and I will start there for my next exercise.

In general my application is

I have or can obtain X.
I can "transform" X into a member of a set Y.
Which member of Y should I choose if the goal is to sell it for meat?

I have, so far, not been able to get everything I want from coinmasters.txt.  The dynamic aspect of the data is problematic for speculative purposes - is it beneficial to obtain X? - because, for example, my wand of fudge control is usually in my DC.

I have a fantasy that there is a Unified Theory of Transformation and the right code could handle Coinmasters, the various creation methods, folding, zapping, using and anything else I have forgotten.  I think maybe I should give up on that Fantasy.

Probably should give up on speculation, too.

Thank you.

PR #1333, opened 2022-12-08
start of support for model train set

Veracity0:
adds the unbound & bound IOTMs
adds to campground items and workshed items.


midgleyc (approved):
Thanks!

PR #1337, opened 2022-12-10
Count turns spent on A Deserted Stretch of I-911

Veracity0:
Do not need Hare-Brained effect to adventure there.
You are given that effect when you attempt to adventure there.
The duration is remaining number of turns available in that zone.


gausie:
Don't forget to reset on ascension!

jaadams5 (approved):

PR #1338, opened 2022-12-11
Don't add html to the toString() of PurchaseRequest

libraryaddict:
As far as I can tell, it's meaningless as ListCellRenderer handles the rendering.

Always found it a little irritating when copying a mall search result came up with `<html><nobr><font color="gray">Manual of Numberology (1 @ 85,000,000): Cheap Stuff Here --&gt;</font></nobr></html>`

Especially when other list cell renderers didn't have this behavior.

So I looked into it, and it appears quite meaningless.

https://github.com/kolmafia/kolmafia/blob/main/src/net/sourceforge/kolmafia/swingui/widget/ListCellRendererFactory.java#L121

### DIFF
--- a/src/net/sourceforge/kolmafia/request/PurchaseRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PurchaseRequest.java
@@ -120,14 +120,6 @@ public abstract class PurchaseRequest extends GenericRequest
   @Override
   public String toString() {
     StringBuilder buffer = new StringBuilder();
-    String color = this.color();
-
-    buffer.append("<html><nobr>");
-    if (color != null) {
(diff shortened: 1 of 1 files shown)


jaadams5:
FWIW the code with the HTML does go back to 2013.  

How confident are you that CellRenderers are the only place that this toString() is used?  I'm having difficulty answering that and there are definitely cases where HTML in session logs is considered A Good Thing.

I'll approve but someone with more confidence than I about where it is actually used can push.

jaadams5 (approved):

midgleyc:
I can't see anything obvious and I think with our audience we'll know quite quickly if this is having negative effects for people.

Thanks!

PR #1339, opened 2022-12-11
feat: add ability to stop relay server

midgleyc:
Occasionally the relay server slows to a crawl for me. Historically I've restarted Mafia. I thought it would be nice to have the option to stop the relay server (and perhaps restart it later) if that happens.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/RelayBrowserCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/RelayBrowserCommand.java
@@ -1,16 +1,19 @@
 package net.sourceforge.kolmafia.textui.command;
 
 import net.sourceforge.kolmafia.webui.RelayLoader;
+import net.sourceforge.kolmafia.webui.RelayServer;
 
 public class RelayBrowserCommand extends AbstractCommand {
   public RelayBrowserCommand() {
-    this.usage = " [nobrowser] - start the relay server and/or open the relay browser.";
+    this.usage = " [nobrowser|stop] - start/stop the relay server and/or open the relay browser.";
   }
 
   @Override
   public void run(final String cmd, final String parameters) {
     if (parameters.equals("nobrowser")) {
       RelayLoader.startRelayServer();
+    } else if (parameters.equals("stop")) {
+      RelayServer.stop();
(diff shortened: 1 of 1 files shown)


jaadams5:
Why do we need this?  If the relay browser is too slow I shutdown the browser or stop it in the browser and refresh the page.    You must be doing something I have not considered :-)

jaadams5 (approved):

jaadams5:
> What do you mean when you say you "shutdown the browser"? There might already be a way to do what I'm trying to do.
> 
> Closing and re-opening Firefox doesn't fix the speed issue. I'm hoping that stopping the relay thread (and restarting it) might do something.

Ah,

I press the X to shutdown Brave.  On rare occasions I'll use Task Manager (Windows) to terminate Brave and related processes but that is infrequent.  But when it is gone from any task list I then restart using the Relay Browser item in Mafia.  

If you think the issue is the relay thread and not something else then this can be useful.

But I think a more interesting question is why the relay thread slows down if indeed that is the cause of what you are observing.  Are you just playing the game or running relay scripts that do more work than just build a page?  Ash or JavaScript?

I just approved this because it does address your pain point immediately but if you have the inclination I'd really be interested in some profiling.

midgleyc:
Yeah, so it's weird and has been going on for years, but is very inconsistent. After Mafia gets into a "slow state", each request will take a while to complete (consistently), and a restart will fix it.

I do run a fair few relay scripts (all ASH I think), and I play in the browser a lot. It occasionally will happen over the course of a session that Mafia gets slower, and a close / re-open will fix the issue.

Unusual things I do: I run a lot of chat commands, including from the GCLI.

Given what you've posted here, perhaps it's something to do with requests in general, and not just the relay thread? I don't understand how it can happen some of the time but not most of the time. Now that this is in I will be able to investigate if this helps ;)

jaadams5:
When restart fixes a speed thing the habits that date back to C and memory leaks kick in for me.  Java suggests holding references longer than needed.  Does the Garbage Collection monitor suggest anything?  I have a couple of characters with a lot of different items in the inventory so that is probably not it.  I avoid Chat so that could be something to look at.  You have a tool now so you don't need to respond here.  We can remember these points when you start to dig :-)

PR #1340, opened 2022-12-11
Fix array_rand extra rolls.

phulin:
The array_rand implementation in PHPRandom fails to short circuit after reaching the desired number of elements, which causes extra rolls to be drawn from the RNG. While the first result matches, future array_rand calls will be out of sync.

### DIFF
--- a/src/net/sourceforge/kolmafia/utilities/PHPRandom.java
+++ b/src/net/sourceforge/kolmafia/utilities/PHPRandom.java
@@ -68,7 +68,7 @@ public class PHPRandom extends Random {
     int[] result = new int[required];
     int j = 0;
 
-    for (int i = 0; i < count; i++) {
+    for (int i = 0; i < count && j < required; i++) {
       double chance = ((required - j) / (double) (count - i));
       if (nextDouble() < chance) {
         result[j++] = i;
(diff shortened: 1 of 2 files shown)


gausie:
> Context as to why this is needed? Any chance of a test that fails without this?

I mean if there is a bug in existing code I don't think context is necessarily needed.

jaadams5:
> I mean if there is a bug in existing code I don't think context is necessarily needed.

Given the absence of code documentation, what's the bug?  Is it an array indexing issue that is reported elsewhere that I have not seen or is it something downstream isn't working because the random number generator is called too many times or what?  What bad thing happens if this is not fixed?

gausie:
> > I mean if there is a bug in existing code I don't think context is necessarily needed.
> 
> Given the absence of code documentation, what's the bug? Is it an array indexing issue that is reported elsewhere that I have not seen or is it something downstream isn't working because the random number generator is called too many times or what? What bad thing happens if this is not fixed?

I agree, I want to see a test to confirm that it is a bug. But I don't need context on why a bug needs fixing once it's confirmed to be such.

jaadams5:
Maybe "context" wasn't the best way to ask "what is the bug?"  If that is answered then how it was discovered is just satisfying curiosity.

phulin:
Sorry, I pushed this this morning and then got distracted. I updated the description and will write a test.

midgleyc (approved):
Thanks!

PR #1341, opened 2022-12-12
Add Canticle of Carboloading to Breakfast

libraryaddict:
Feels a bit odd that this isn't already in breakfast.

Granted, it's not a skill that every class can use, but I believe it still belongs here.


jaadams5:
Consider

https://kolmafia.us/threads/canticle-of-carboloading-warning.13351/#post-98507

https://kolmafia.us/threads/canticle-of-carbloading-in-daily-deeds.7300/#post-55267

and especially 

https://kolmafia.us/threads/canticle-of-carboloading.4784/#post-34194

I'm thinking that a decade ago people did not want it automated and in 2010 had the ability to add it to breakfast themselves if they disagreed with the dev's decision not to add it to breakfast.

libraryaddict:
The first two posts read more like asking for carboloading support to mafia, and some was added.

The last one I think is not relevant, Carboloading is a skill you need to turn on in the preferences, much like the other automated skills before mafia will cast it for you.

It shouldn't cast the skill unless you turn said option on, I don't see a real reason not to add this. Other than the whole "Only one class can use it"

gausie:
I'm inclined to vote for adding this behaviour but...

It irks me that your response indicates poor comprehension. The last link
is literally a request for this behaviour that was denied because this can
be covered by a login script. It is relevant and your justification fails
to address or argue that.

libraryaddict:
> I'm inclined to vote for adding this behaviour but... It irks me that your response indicates poor comprehension. The last link is literally a request for this behaviour that was denied because this can be covered by a login script. It is relevant and your justification fails to address or argue that.

Right, but pretty much everything breakfast does can be covered in a login script. If I was going to go down their argument, I'd remove breakfast altogether.

gausie:
No you wouldn't. But you could create a fork that did that if you like.

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
```diff
@@ -76,6 +76,7 @@ public class UseSkillRequest extends GenericRequest implements Comparable<UseSki
     "Bowl Full of Jelly",
     "Eye and a Twist",
     "Chubby and Plump",
+    "Canticle of Carboloading"
```
gausie, line 79: Probably put this in adifferent position to the end of the list, because it's actually very old. If possible.

libraryaddict:
> No you wouldn't. But you could create a fork that did that if you like.

It was hyperbole, the point being that their argument can easily be constructed as being against 'breakfast' existing at all. Which is why I didn't mention the login script originally, as it didn't really compute.

That said, I admit my comment was a bit nonsensical, I'd just woken up and confused several comments and thoughts into a "We don't want to carboload automatically because it could upset our diet".

Regardless.

The first comment doesn't seem relevant to this PR. Its just asking for carboloading tracking.

The second asks for it to be added to daily deeds not breakfast, and isn't added to daily deeds as custom daily deeds will be a thing.

The third does ask for it, but is rejected in favor of recommending players use a login script. For no real determinable reason that I can see, as again, everything breakfast does could be moved into a login script. 

Ideally, breakfast would be a script to begin with. But as it has been a part of mafia for so long and integrated into so many workflows, it isn't a feasible change. There's no advantage to be gained.

Correct me if I'm wrong, but the difference between daily deeds and breakfast is that daily deeds is used for things that provide effects that would expire, use consumable items and consume adventures.

Breakfast is for things that are reusable and the only real cost is the MP consumed.

Although carboloading is mildly special in that you can store it across several days, perhaps you want to cast it after your diet. Still feels like it should be in breakfast though, since it can be toggled.

Veracity0:
FWIW, VMF calls a breakfast script (which should call breakfast) at the end of the day, not the beginning, as various breakfast actions can improve after adventuring.

I never have "get breakfast" checked on the login window.

gausie:
@libraryaddict I've asked you this before but I want to reiterate - when you write commentary online: write your comment, read it over critically, consider what makes sense and whether each part is necessary or likely to further your actual aims and then post.

jaadams5:
My fundamental issue is that this idea was rejected a decade ago.  No one has articulated what had changed to now  make it a good idea.  Nor has the KoLmafia.us community, who was part of the rejection decision, been involved in this conversation.  Maybe 10 years later no one cares but it would still be nice to know.

One criteria that devs have used over time to decide to accept or reject a feature is whether it can be easily scripted.  If it can then it is usually rejected.  This comes from some kind of conservative minimalism.  If it is not in KoLmafia then it is less likely to have unintended side effects and it does not have to be maintained by KoLmafia devs.  This is not, and never has been, universally applied but it is often the place to start.

In particular, many Daily Deeds were rejected because making a custom Daily Deed for that action was trivial.  Sometimes it required a usage preference to be implemented but even then the last step of creating the deed was left for those who wanted the feature.  Again, this was not, and never has been, applied applied by all devs at all times.

Recollections of past events are always suspect, especially when they involve cause and effect or claim why something happened.  In that spirit of uncertainty, my recollection was that Breakfast was originally a collection of things that could be done once a day, had no cost, needed no configuration and it did not matter when in they day they were done.  Daily Deeds were things that would have been in Breakfast except they violated one or more of those conditions.  The Breakfast cost condition was relaxed to allow some casting and that triggered the unexpected side effect of having to restore mana during Breakfast.  That meant Ronin and in run or aftercore became important to Breakfast because Breakfast now indirectly consumed resources that might have better uses.  So eventually some kind of fine tuned control over Breakfast needed to be implemented.  So Breakfast and Daily Deeds sort of evolved together and it was not always true that distinctions between them or criteria for inclusion or rejection were consistent.

A reason I am willing to make a discussion of this (besides a previous rejection) is the possibility the other devs have shifted away from using scripting difficulty as an inclusion criteria.  If that is the case then I need to make adjustments :-)

P.S. - Given that the preference _carboLoaded exists making this Breakfast instead of a custom Daily Deed just adds one mouse click for the user who wants it.

gausie:
Our line of support where a script would do the job is not well defined. No reason was given for the decision 10 years ago so I feel comfortable making the decision now!

I feel like having the option is user friendly without costing us anything at all.

For most users, simply casting this at the beginning of every day is the right thing to do. For the rest, scripting it into some decision making is a good option. That sounds like grounds for a breakfast casting option to me!

midgleyc (approved):
I also think it's helpful for Mafia to support stuff that the user might want to do at the start of the day (without having to e.g. compare prices, buy things etc.) so a script doesn't have to be written / shared.

Thanks!

PR #1343, opened 2022-12-13
Lint and a test

jaadams5:
I had some lint cleanup on ProxyRecordValue that got discarded with a closed PR.

I decided to restore it and "sweeten the pot" with a script that increased the test coverage.

If increasing the coverage of a class by writing custom scripts is reasonable then I can probably test a few more proxies.


src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -78,27 +78,27 @@ public Value aref(final int index, final AshRuntime interpreter) {
     }
 
     if (rv instanceof Integer) {
-      return DataTypes.makeIntValue(((Integer) rv).intValue());
+      return DataTypes.makeIntValue((Integer) rv);
```
Veracity0, line 81: You can modernize this even more by doing this:
```
    if (rv instanceof Integer iv) {
      return DataTypes.makeIntValue(iv);
    }
```
and so on for all the similar casts in this method.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
jaadams5: I like that idea.  Stand by.

jaadams5: @Veracity0 - Would you please resolve if I got it right?  Thank you.

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -77,32 +77,32 @@ public Value aref(final int index, final AshRuntime interpreter) {
       return (Value) rv;
```
Veracity0, line 77: Do it here too.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
Veracity0: That was all good.
There is one more instance in that same method just before your first change that could be fixed as well to eliminate a a cast.

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
jaadams5: Whoops.  Thank you.

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -1754,39 +1753,29 @@ public String get_nickname() {
...
-          return "stench.gif";
+      return switch ((Element) this.content) {
           // No image for Slime or Supercold in Manuel
-        case SLIME:
-          return "circle.gif";
-        case SUPERCOLD:
-          return "circle.gif";
-      }
-      return "";
+        case NONE, SLIME, SUPERCOLD -> "circle.gif";
```
gausie, line 1766: It seems like this could be replaced with `Element.getColor()`, which would be doubly nice because it would remove duplicated information in the codebase. Depending on whether we care about an extremely minor breaking change, you might want to add `Element.getLegacyColor()` that forces SLIME and SUPERCOLD to return circle.gif instead of the images in the Element enum

test/root/scripts/TestFamiliarProxy.ash
```diff
@@ -0,0 +1,40 @@
+familiar f = $familiar[blood-faced volleyball];
```
gausie, line 1: I think this test is fine, though it would definitely be better as a proper Java test since you'd get more context on failures if they happened. To my mind, ash tests are better conceptualised as "end to end" tests rather than unit tests.

gausie (commented):
All looks good, though I have some comments

test/root/scripts/TestFamiliarProxy.ash
jaadams5: Test was made because it was trivial.  Pick a proxy, copy the "add" lines from code, find and replace to turn them into ash print statements and the code at least is exercised.  I don't like testing a class implicitly by scripts but I will revisit and see if I can do something more class specific.  Thanks.

midgleyc: If you check `test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java`, we have some tests (e.g. `myPathCoercesToString`) which essentially evaluate a single ASH expression. I think it might be nice to make a generic reusable `attributeEvaluatesTo` parameterized test (as checking an attribute evaluation is something we do a lot), and have it take a stream of loads of possible arguments, including all these proxy values.

Veracity0 (approved):

jaadams5:
I'll continue to explore testing.

PR #1344, opened 2022-12-14
Initial support for trainset

libraryaddict:
This is initial support for trainset, using configuration settings as suggested by Gausie in Discord.

`lastFoodDropped` is missing tracking of actual food dropped due to a lack of knowledge around criteria of acquisition, and the codebase.

* What's the criteria for food dropping? End of fight? Tradeable & Discardable? Non-Conditional? Copyable monster? (Requires spading)
* Where should the code for tracking of said food drops be?
* How would we avoid the property being set when undesired, such as multiple food dropping
* Should this property be set when trainset is unavailable?
* What should the default property be? -1? Empty?

There's a possibility the property may need to be added in a future PR.

There's also other concerns with this PR, such as

* Are the trainset piece names acceptable?
* Should we replace `lastTrainsetConfiguration` with `trainsetConfigurableAt` which predicts a future turn? `trainsetPosition + 40`
* Should the FightRequest look at what trainset piece we've hit to determine if something has gone wrong with tracking?
* Should we attempt to parse trainset position while the configuration is in cooldown? This requires pixel parsing I believe, not sure how it would behave in mobile mode.
* Should we include trainset pieces that need url manipulation to reach? (Only the 10th piece qualifies as far as I'm aware)
* Are we missing tests?
* Should the `lastTrainsetConfiguration` when inaccuracy is detected, always be a higher number? Such as jumping `trainsetPosition` from 20 to 60 and `lastTrainsetConfiguration` from 20 to 22. Currently it may change `lastTrainsetConfiguration` from 20 to -10 if need be while leaving `trainsetPosition` untouched.

Note: 
There is a bug in trainset in which during configuration cooldown, accessible mode is not accessible and instead shows a blank screen. Which means we cannot use it to determine what station the train is at, and can't parse the dropdowns to view our current station.


gausie:
>     * Are the trainset piece names acceptable?
Sure 
>     * Should we replace `lastTrainsetConfiguration` with `trainsetConfigurableAt` which predicts a future turn? `trainsetPosition + 40`
No
>     * Should the FightRequest look at what trainset piece we've hit to determine if something has gone wrong with tracking?
Yes
>     * Should we attempt to parse trainset position while the configuration is in cooldown? This requires pixel parsing I believe, not sure how it would behave in mobile mode.
It's not pixel parsing really - on the kol end it will be coded with a certai coordinate per current position. So we need to reverse engineer and encode that map and then use it.
>     * Should we include trainset pieces that need url manipulation to reach? (Only the 10th piece qualifies as far as I'm aware)
Does it do anything? Can you include it in your track?
>     * Are we missing tests?
Is there any coverage missing for your changes?
>     * Should the `lastTrainsetConfiguration` when inaccuracy is detected, always be a higher number? Such as jumping `trainsetPosition` from 20 to 60 and `lastTrainsetConfiguration` from 20 to 22. Currently it may change `lastTrainsetConfiguration` from 20 to -10 if need be while leaving `trainsetPosition` untouched.

I think anything that gets it into a working state is fine when we are trying to recover from bad state info.

> Note: There is a bug in trainset in which during configuration cooldown, accessible mode is not accessible and instead shows a blank screen. Which means we cannot use it to determine what station the train is at, and can't parse the dropdowns to view our current station.

Why do we care about accessibility mode? Just parse the main page

src/net/sourceforge/kolmafia/session/TrainsetManager.java
```diff
@@ -0,0 +1,151 @@
...
+
+    return pieces;
+  }
+
+  public static void visitChoice(String html) {
+    int lastPosition = Preferences.getInteger("trainsetPosition");
+
+    Matcher matcher = CURRENT_STATION.matcher(html);
+
+    // The station cannot be read while the configuration is locked
```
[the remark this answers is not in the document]
libraryaddict: Posted a video in #iotm on discord

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -8129,6 +8133,21 @@ private static boolean handleCosmicBowlingBall(String text) {
     return false;
   }
 
+  private static boolean handleTrainset(String text) {
+    if (!text.contains("Your toy train moves ahead to the")) {
```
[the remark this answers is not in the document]
libraryaddict: The messages for some reason seem to have their html stripped out, the station names are in bold.

I didn't really care to look into it as the message itself should be unique. Although, it could be more unique.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -8129,6 +8133,21 @@ private static boolean handleCosmicBowlingBall(String text) {
...
+      return false;
+    }
+
+    Preferences.increment("trainsetPosition");
+
+    if (text.contains("What a coincidence, the diner is serving more of that food you just found!")
```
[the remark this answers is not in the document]
libraryaddict: I added that check here to keep it in the same group. `Your toy train moves ahead to the    What a coincidence, the diner is serving more of that food you just found!`. The text as provided by the FightRequest.

src/net/sourceforge/kolmafia/request/FightRequest.java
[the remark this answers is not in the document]
gausie: You can pass in the full html to handleTrainset if you want

src/net/sourceforge/kolmafia/session/TrainsetManager.java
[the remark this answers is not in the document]
gausie: Ok but I'm not sure what you mean by parse the pixels. I addressed this in my longer-form comment

src/net/sourceforge/kolmafia/request/FightRequest.java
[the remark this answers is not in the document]
gausie: I don't understand your response

src/net/sourceforge/kolmafia/session/TrainsetManager.java
```diff
@@ -0,0 +1,198 @@
...
+    int newPosition = Preferences.increment("trainsetPosition");
+
+    TrainsetPiece[] pieces = getTrainsetPieces();
+
+    // If trainset configuration is unknown
+    if (pieces.length != 8) {
+      return;
+    }
+
+    TrainsetPiece piece = TrainsetPiece.getByName(pieceName);
```
gausie, line 100: This branch onwards seems not to be tested, which would be useful

src/net/sourceforge/kolmafia/session/TrainsetManager.java
```diff
@@ -0,0 +1,198 @@
+package net.sourceforge.kolmafia.session;
+
+import java.util.Arrays;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import net.sourceforge.kolmafia.preferences.Preferences;
+
+public class TrainsetManager {
```
gausie, line 9: You can add a private constructor which will stop this class being considered uncovered

gausie:
Can you also add the configuration choice to walkaway-able choices?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -8143,6 +8148,23 @@ private static boolean handleCosmicBowlingBall(String text) {
     return false;
   }
 
+  private static final Pattern TRAINSET_MOVE =
+      Pattern.compile("^Your toy train moves ahead to the (.+?)\\.");
+
+  private static boolean handleTrainset(TagNode node) {
+    String text = node.getText().toString();
+
+    Matcher matcher = TRAINSET_MOVE.matcher(node.getText().toString());
```
midgleyc, line 8157: should probably reuse text here instead of getting the text again?

Also can you not pass in `str` from the caller instead of passing in the node again?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -6338,6 +6339,10 @@ private static void processNode(final TagNode node, final TagStatus status) {
 
       if (handleCosmicBowlingBall(str)) return;
 
+      if (FightRequest.handleTrainset(node)) {
```
midgleyc, line 6342: I think you could condition this on `status.toyTrain`.

Also the way this is written we have different places handling the logging and the preferences update -- could they be combined?

src/net/sourceforge/kolmafia/request/FightRequest.java
libraryaddict: Good catch, and no. The `str` has tags stripped out as far as I can see. Only the actual text in the html node is given.

src/net/sourceforge/kolmafia/request/FightRequest.java
libraryaddict: Huh, I didn't notice that. Fixed.

midgleyc: you marked this as resolved, but don't seem to have actually pushed the changes?

libraryaddict: Whoops

midgleyc (approved):

PR #1346, opened 2022-12-16
Ping-Pong Skill

Veracity0:
FWIW, here is the "You gain X Meat" message from defeating a Ping-Pong-Playing Trainbot, as FightRequest sees it in the HRML parsetree:
```
        <p>
          Just before the Trainbot shuts down for good, a tinny voice from somewhere inside it says &quot;You win!  Your excellent play deserves a reward.&quot;A slot opens and some Meat flops out.
          <center>
            <table>
              <tbody>
                <tr>
                  <td>
                    <img src="https://d2uyhvukfffg5a.cloudfront.net/itemimages/meat.gif" height="30" width="30" alt="Meat">
                  <td valign="center">
                    You gain 220 Meat.
```
pingpongSkill == 22

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1106,6 +1106,7 @@ user	peteMotorbikeMuffler
 user	peteMotorbikeSeat
 user	photocopyMonster
 user	pieStuffing
+user	pingpongSkill	0
 user	pirateSwagger	0
 user	pizzaOfLegendEaten	false
 user	plantingDate
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -3033,6 +3033,7 @@ user	choiceAdventure1472	0
 user	choiceAdventure1473	0
 user	choiceAdventure1474	0
 user	choiceAdventure1475	0
+user	choiceAdventure1486	0
 user	lockedItem4637	true
 user	lockedItem4638	true
 user	lockedItem4639	true
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -1649,6 +1649,7 @@ Item	pestoblade	Muscle: +3, Stench Damage: +3
 Item	pewter claymore	HP Regen Min: 6, HP Regen Max: 12, MP Regen Min: 6, MP Regen Max: 12
 # Pigsticker of Violence: Lets you stab opponents as they fail to attack you
 Item	Pigsticker of Violence	HP Regen Min: 20, HP Regen Max: 30, Critical Hit Percent: +10, Weapon Damage Percent: +60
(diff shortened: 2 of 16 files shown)


src/data/modifiers.txt
```diff
@@ -4100,6 +4100,8 @@ Item	Time Cape	Item Drop: [25*zone(Twitch)]
 # Time Cloak: Sometimes Find an Extra Chroner in the Time-Twitching Tower
 Item	Trainbot harness	Initiative: -200
 # toy jet pack: Jane!  Get Me Off This Crazy Thing!
+# Trainbot harness: Lets you rescue elves from passenger-occupied Crimbo train cars
+Item	Trainbot harness	Initiative: -200
```
gausie, line 4104: Look at line 4101!

gausie:
We should correct the ping pong skill from the quest log

src/data/defaults.txt
```diff
@@ -1106,6 +1106,7 @@ user	peteMotorbikeMuffler
 user	peteMotorbikeSeat
 user	photocopyMonster
 user	pieStuffing
+user	pingpongSkill	0
```
jaadams5, line 1109: wiki says "Is reset on ascension to 1.".  Might not matter if we can always look it up.  Is 1 a better default?

Veracity0:
> We should correct the ping pong skill from the quest log

That did not exist when I submitted this.

Now that it does, yes.

src/data/defaults.txt
Veracity0: Wiki had nada about this when I submitted.

If 1 is the default, yes.
And if it resets, that needs to be done as well.

src/data/modifiers.txt
Veracity0: Thanks! I will remove it, since it is not in the correct (alphabetic) place.

src/data/defaults.txt
Veracity0: I took a character with no ping-pong skill.
Nothing was in the quest log.
He got a ping-pong paddle.
Nothing was in the quest log.
He used on ping-pong ball.
```
You have achieved a skill level of 1 at playing ping-pong.
```
I would be surprised if your 1st skill level carried across ascension.
Should I believe the Wiki?

Veracity0: I'm resetting to 0 on ascension.
I made it read ping-pong skill from the quest log.
If it really is one on ascension, that would be a first: reset on ascension to something which is NOT your initial state.
Whatever. We will correct when you look at quest log.

jaadams5 (approved):

PR #1347, opened 2022-12-16
feat: new crimbo skills

midgleyc:
Yes, "Waiter" has two spaces after the colon. Maybe that'll get fixed. Probably not.

### DIFF
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -11071,6 +11071,9 @@
 11043
 11044	packaged model train set	494254899	modeltrain_box.gif	usable	t	0
 11045	model train set	959599372	modeltrain.gif	usable		0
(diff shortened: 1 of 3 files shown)


src/data/items.txt
```diff
@@ -11071,6 +11071,9 @@
 11043
 11044	packaged model train set	494254899	modeltrain_box.gif	usable	t	0
 11045	model train set	959599372	modeltrain.gif	usable		0
+11046	Crimbo training manual	990145553	trainmanual.gif	usable		0
```
gausie, line 11074: It's reusable right?

gausie (approved):

src/data/items.txt
midgleyc: yeah. And it might be curse-like, but I don't know what the page is.

PR #1353, opened 2022-12-18
track Elf Gratitude

Veracity0:


gausie (approved):

gausie:
This doesn't strip commas from the number when extracting from the quest log, though I'm not sure if we need to

Veracity0:
It doesn't need to. I did try this out.

All preferences are strings.

Preferences.getInteger("pref") will do string -> int conversion that strips commans.
ASH get_property("pref").to_int() does the same.

PR #1359, opened 2022-12-20
log white arm towel msg. use lost luggage count as elfGratitude gained

Veracity0:
1) When your familiar grabs something from the dining car, log the message.
2) When multiusing lost elf luggage, use count as amount of elfGratitude gained.


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -6302,6 +6304,10 @@ private static void processNode(final TagNode node, final TagStatus status) {
         FightRequest.handleTrainbotLuggageHook(str, status);
       }
 
+      if (won && status.armtowel) {
```
midgleyc, line 6307: I think there's now enough of these `won` blocks that I'd prefer to see:
```
if (won) {
  if (status.X) {
  [...]
}
```

Veracity0: I had thought about that and I agree.
I could also put the status.xxx checks into the individual methods, but I am intentionally not doong method calls unless necessary.
Similarly, I could even put all three into a single method, but I prefer having each method be specific to one thing.

midgleyc (approved):
Thanks!

PR #1360, opened 2022-12-22
Properly handle quantity limited NPC purchases and reset purchase limits properly between sessions

libraryaddict:
If you attempt to use `buy zeppelin ticket` you will notice that it buys a ticket, and reports a failure. No ticket purchased.

```
You spent 4,750 Meat
You acquire an item: Red Zeppelin ticket
Desired purchase quantity not reached (wanted 1, got 0)
```

The second commit is a working test.

The third commit is an improved coverage test that fails due to another issue where npc limits are not reset on character reset.

This is the cause for players being unable to buy from the Fireworks NPC Store if they leave mafia running over rollover.

The fourth commit is a fix for test, without the code, to prove the test is not the reason for failure of the next commit.

The fifth commit is the fix for that new issue.


test/net/sourceforge/kolmafia/request/NPCPurchaseRequestTest.java
```diff
@@ -2,25 +2,49 @@
...
   public void beforeEach() {
     HttpClientWrapper.setupFakeClient();
+
+    KoLCharacter.reset("NPCPurchaseTest");
+    Preferences.reset("NPCPurchaseTest");
+  }
+
+  @AfterEach
+  public void afterEach() {
+    KoLCharacter.reset(true);
```
midgleyc, line 47: we've been trying to move stuff like this into `Cleanups` to avoid expensive resets between tests.

Can you make new cleanups for:
```
withNPCStoreReset();
withCanInteract(boolean interact);
```

and shift the tests to use them?

gausie:
> If you attempt to use `buy zeppelin ticket` you will notice that it buys a ticket, and reports a failure. No ticket purchased.

No comment on this PR but as I've said in the past, you've got to get better at explaining your changes.

midgleyc (approved):
Thanks!

PR #1361, opened 2022-12-22
Coal car stuff

Veracity0:
More to come in other PRs, but I'd like to get this in quickly.

1) Coal-Shoveling Trainbot rare drop is the "really nice lump of goal" - a gift item which I have not seen yet
2) portable steam unit is once-per day
3) portable steam unit gives you an effect
4) If you have no crystal shards, Slagging Off offers only choice 3.
5) Maybe we'd like a way to alternate the options in Slagging Off? And automatically take choice 3 if you have no shards.
6) Arena parameters for Mini-Trainbot. That can certainly wait until we/I are done farming Crimbo stuff.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
+++ b/src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
@@ -6221,6 +6221,15 @@ public abstract class ChoiceAdventures {
         "Crimbo Train (Dining Car)",
         new Option("acquire 3 lost elf trunks", 1),
         new Option("decrease Trainbot strength", 2));
+
+    // Slagging Off
+    new ChoiceAdventure(
+        1489,
+        "Crimbo22",
+        "Crimbo Train (Coal Car)",
+        new Option("acquire crystal Crimbo Goblet", 1),
+        new Option("acquire crystal Crimbo platter", 2),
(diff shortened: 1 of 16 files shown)


src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
```diff
@@ -6221,6 +6221,15 @@ public static final Option findOption(final Option[] options, final int decision
         "Crimbo Train (Dining Car)",
         new Option("acquire 3 lost elf trunks", 1),
         new Option("decrease Trainbot strength", 2));
+
+    // Slagging Off
+    new ChoiceAdventure(
+        1489,
+        "Crimbo22",
+        "Crimbo Train (Coal Car)",
+        new Option("acquire crystal Crimbo Goblet", 1),
```
jaadams5, line 6230: Perhaps Crimbo goblet or Crimbo Platter?  The inconsistency jumps out at me and I think the game uses lower case.

jaadams5 (approved):

src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
Veracity0: Should be Crimbo goblet. Will fix.

PR #1365, opened 2022-12-23
Track movement for empty stations with model trainset trainsetPosition

libraryaddict:
Due to an oversight, a test wasn't added for tracking movement and updating `trainsetPosition` when a station was empty.

As the empty track does not have an image, it is caught in another section of the code then passed to the trainset managing. method.


midgleyc:
Gausie got in touch with CDM and the You / Your typo should now be fixed.

It's now:

```
Your toy train moves ahead to some empty track.
```

Veracity0:
> Gausie got in touch with CDM and the You / Your typo should now be fixed.

Huh.

I submitted this bug report a couple weeks ago:

```
Toy train combat messages:

Your toy train moves ahead to the Groin Silo. You train picks up some fresh groin!
Your toy train moves ahead to the Brawn Silo. You train picks up some fresh brawn!
Your toy train moves ahead to the Brain Silo. You train picks up some fresh brain!

Those should all be "Your train", not "You train"

Your toy train moves ahead to the Ectoplasmic Oil Refinery. Slimy, spooky sprays into the air from the stacks of the refinery, coating your body.

"slimy, spooky" what?

Your toy train moves ahead to the Water Tower, Sewage. Hot, soupy, grabage water pours out of the water tower and freezes you to the bones.
Your toy train moves ahead to the Water Tower, Frozen. Frigid, just unfrozen water pours out of the water tower and freezes you to the bones.

Did you intend to say that both of those "freeze you to the bones"?
And "grabage water" should be "garbage water".
```

The bug report remains unacknowledged and the bugs remain unfixed.

I am happy that the dev team has direct access to the devs.
Isn't it part of their job to vet user-supplied bug reports and alert the devs when things need to be fixed?
Considering that I write lots of bug reports, never get an acknowledgement from the dev team, and the bugs never get fixed, perhaps what I SHOULD be doing is simply alerting gausie and requesting that he pass them along to CDM et. al.?

midgleyc:
I don't know, honestly. I have received responses to the in-game bug report in the past.

I assumed, before I reported it, that the in-game bug reports are triaged and things like typos are low priority. I escalated this one because Mafia doing an exact match on a string with a typo in it is something that will break when the typo is fixed, and with the circumstances this would be hard for us to notice when it happens and could lead to hard-to-detect bugs, so this typo is more important than the others.

Veracity0:
> I assumed, before I reported it, that the in-game bug reports are triaged and things like typos are low priority. I escalated this one because Mafia doing an exact match on a string with a typo in it is something that will break when the typo is fixed, and with the circumstances this would be hard for us to notice when it happens and could lead to hard-to-detect bugs, so this typo is more important (to gameplay through Mafia) than the others.

Here's another I submitted:
```
I tried using my ping-pong table on a character who had not logged in during this Crimbo season.

It failed. I was told:

"That item cannot be used on a player in Ronin or Hardcore."

Since player was definitively not in either state, that was super confusing.

What was the ACTUAL failure reason?

How about if you say it in the message, rather than re-using the existing message?

Thanks.
```
I assume the actual reason is "That player does not have ping-pong paddle" or something, but KoL reports "ronin or hardcore".

I am hesitant to start writing a "pingpong" command because we pass failures from KoL straight back to the user, and we WILL get KoLmafia bug reports until KoL fixes this - and we add code to recognize the correct message.

I'll kamil gausie and ask him to escalate that one.

midgleyc:
> I assume the actual reason is "That player does not have ping-pong paddle" or something, but KoL reports "ronin or hardcore".

Going way off tangent for this PR, but I've successfully used the table against a player whose last login was October 13, 2022 and so likely doesn't have a paddle. So it might have been a temporary KoL bug? Or something more complex?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7038,7 +7045,7 @@ private static boolean handleGrimstone(
   }
 
   private static final Pattern TRAINSET_MOVE =
-      Pattern.compile("^Your toy train moves ahead to the (.+?)\\.");
+      Pattern.compile("^Your toy train moves ahead to (?:(?:the)|(?:some)) (.+?)\\.");
```
midgleyc, line 7048:
```suggestion
      Pattern.compile("^Your toy train moves ahead to (?:the|some) (.+?)\\.");
```
I don't think you need the inner capturing groups?

libraryaddict:
It's interesting about the inner capturing groups.

I was under the belief that logically, it should be a flip for only the surrounding characters.

It's why I always want to bracket when possible. Assumptions normally make a fool of me.

midgleyc (approved):
Thanks!

Veracity0:
> > I assume the actual reason is "That player does not have ping-pong paddle" or something, but KoL reports "ronin or hardcore".
> 
> Going way off tangent for this PR, but I've successfully used the table against a player whose last login was October 13, 2022 and so likely doesn't have a paddle. So it might have been a temporary KoL bug? Or something more complex?

No, it was a misunderstanding on my part. I looked again at my targets that failed, and darned if one wasn't in Hardcore (abandoned in 2019) and the other was in Ronin (abandoned in 2012).

I tried with a clannie not logged in since 2021, but with 12,000 some turns "this run" and it worked.

I submitted an "apology" bug report to KoL.

PR #1369, opened 2022-12-24
Allow aborting breakfast with escape.

phulin:
As described. In BreakfastManager, the forceContinue calls were causing userAborted to be ignored. Instead we should just set the continuation state to ignore errors incurred during the breakfast actions.

Not sure how a test could be written for this PR.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/BreakfastManager.java
+++ b/src/net/sourceforge/kolmafia/session/BreakfastManager.java
@@ -331,8 +331,6 @@ public class BreakfastManager {
       if (count > 0) {
         KoLmafiaCLI.DEFAULT_SHELL.executeLine("hermit " + count + " 11-leaf clover");
       }
-
-      BreakfastManager.ignoreErrors();
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/session/BreakfastManager.java
```diff
@@ -101,30 +122,20 @@ public static void getBreakfast(final boolean runComplete) {
...
+          if (GenericRequest.abortIfInFightOrChoice()) {
+            KoLmafia.updateDisplay(
+                MafiaState.ABORT, "Breakfast aborted; stuck in fight or choice.");
+            return;
+          }
+          if (StaticEntity.userAborted) {
+            KoLmafia.updateDisplay(MafiaState.ABORT, "Breakfast aborted by user.");
+            return;
+          }
+          ;
```
midgleyc, line 136: spare semicolon? Kind of surprised spotless didn't catch this.

phulin: Thanks. Intellij had grayed it out, so I missed it too. odd.

midgleyc (approved):
Thanks! I am sure I've closed the entire window to get it to quit breakfast before.

PR #1371, opened 2022-12-24
Transform if-then-else chain of string tests to a switch statement

Veracity0:
I noticed this when adding quest processing for defeating The Superconductor.

I include a test for a fraction of the change that uses saved HTML from fighting that monster and detecting that the expected property updates.

There are many other monsters that could have their own tests.
(Some already do - Oil Peak and A-Boo Peak, for example. Those tests continue to pass. Whew!)


src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -1682,423 +1681,469 @@ public static void updateQuestData(String responseText, String monsterName) {
...
+          "Sewage Treatment Dragon",
+          "Duke Vampire",
+          "Spider Queen",
+          "Archwizard",
+          "Ley Incursion",
+          "Ghoul King",
+          "Ogre Chieftain",
+          "Ted Schwartz, Master Thief",
+          "Skeleton Lord" -> {
+        if (!monsterName.equals("spooky ghost")
```
midgleyc, line 2008: I'd probably split out the spooky ghost case for clarity

Also do we not read the monster ID? Maybe not for historical reasons but we could read it out the HTML nowadays.

src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -1682,423 +1681,469 @@ public static void updateQuestData(String responseText, String monsterName) {
...
+        QuestDatabase.setQuest(Quest.GLACIER, QuestDatabase.UNSTARTED);
+        QuestDatabase.setQuest(Quest.MAELSTROM, QuestDatabase.UNSTARTED);
+
+        Preferences.setInteger("lastThingWithNoNameDefeated", KoLCharacter.getAscensions());
+      }
+      case "The Superconductor" -> {
+        Preferences.setBoolean("superconductorDefeated", true);
+      }
+      default -> {
+        if ((monsterName.startsWith("Jerry Bradford") && monsterName.contains("World Champion"))
```
midgleyc, line 2123: checking the Git history, I think the initial problem was that the name was given as `Jerry Bradford, Pokéfam World Champion`, so it could probably go inline as `Jerry Bradford, Pok&eacute;fam World Champion`

src/net/sourceforge/kolmafia/session/QuestManager.java
Veracity0: I split it out for clarity, including a comments.

Historical reasons, for sure - but as a result of those, we do not have MonsterPool.java.
That would end up being quite the substantial refactoring!

src/net/sourceforge/kolmafia/session/QuestManager.java
Veracity0: That makes sense. Done.

Doing so, leaves only checking the Thwaitgold bee statue in the "default" section. That's because Bees Hate You final boss is The Guy Made of Bees. So, I made a case statement for that monster and got rid of the default. With a comment.

midgleyc (approved):
Thanks!

src/net/sourceforge/kolmafia/session/QuestManager.java
midgleyc: so I don't know how `MonsterStatusTracker` works, but the final round of a fight (and every other round) includes text like:
```
<td valign="center">You're fighting <span id="monname">a Brake-Operating Trainbot</span></td><!-- MONSTERID: 2254 --></tr></tbody>
```
so I think we could extract the name (plus article) and the monster id.

I've not checked how this interacts with things like darkness.

PR #1373, opened 2022-12-25
feat: add portable steam unit to daily deeds

midgleyc:
Refactor daily deeds and add portable steam unit.

Deliberately don't make a new version, because I want it to go next to
Legendary Beat, not at the end of the daily deeds.

### DIFF
--- a/src/data/dailylimits.txt
+++ b/src/data/dailylimits.txt
@@ -244,6 +244,7 @@ Use	perfectly fair coin	_perfectlyFairCoinUsed
 Use	picky tweezers	_pickyTweezersUsed
 Use	pixel orb	_pixelOrbUsed
 Use	portable ping-pong table	_pingPongGame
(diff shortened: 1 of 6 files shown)


jaadams5 (approved):

tyrion-the-imp:
If the dailyDeedsOptions pref is reset to its default value via the property_default_value() function, it appears that this deed will be left out since "portable steam unit" wasn't added to the default value.  I guess this is related to not changing the version?

midgleyc:
Not exactly. The only thing the version does is that if the the current version (`dailyDeedsVersion`) is less than the version of a particular deed, that deed gets added to the end of your panel.

It's left out because I didn't think about it, but by default it doesn't contain all the builtin deeds either: it's missing Crimbo Tree, Mom, Fishy Pipe, both Dread flasks, Tea Party, Envyfish Egg, Locket, Jick Jar and all the terminal stuff.

PR #1378, opened 2022-12-27
Don't cache fought monsters in Locket.

phulin:
In the current code, resetting when the day turns over doesn't reset LocketManager, meaning that you won't be able to make locket requests because the monsters fought property has been reset but the cached `foughtMonsters` variable hasn't. 

This PR makes it so `getFoughtMonsters` is no longer cached, meaning the value can't get out of sync with the pref (there is one source of truth). `getFoughtMonsters` gets called only infrequently, so this is not a performance issue. The PR also adds a test for this case and cleans up the tests a little.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/LocketManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/LocketManagerTest.java
@@ -8,10 +8,17 @@ import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.hasSize;
 
 import net.sourceforge.kolmafia.persistence.MonsterDatabase;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class LocketManagerTest {
+  @BeforeAll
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/session/LocketManagerTest.java
```diff
@@ -1,21 +1,27 @@
...
 import net.sourceforge.kolmafia.persistence.MonsterDatabase;
 import net.sourceforge.kolmafia.preferences.Preferences;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class LocketManagerTest {
+  @BeforeAll
+  public static void beforeAll() {
+    Preferences.reset("ReminisceCommandTest");
```
gausie, line 19: LocketManagerTest please!

gausie (commented):
All good, just fix this

test/net/sourceforge/kolmafia/session/LocketManagerTest.java
phulin: Good catch, thanks. Fixed.

gausie (approved):

PR #1379, opened 2022-12-28
A whole bunch of code modernization suggested by IntelliJ

Veracity0:
Chris taught me (us) how to get IntelliJ to inspect code and find all sorts of ways we can improve it.

Not all are useful - and in fact there are many that are opposites:
- Do not use Class name when calling static methods from within the class vs.
- Always use Class name when doing that.

Those are "code style conventions". Hola started us out with the second, but I rather like the first, so have been using it for new code. I have occasionally converted an entire package, when I was in there changing something, but I am not interested in enforcing "coding style" changes at this time.

Instead, I am looking at code improvement/modernizations. I have selected various things that should be easy and non-controversial.

- Use <> when allocating objects
- replacing if-then-else chains by switch statements
- Using enhanced switch statement
- Using switch expressions
- Replacing (some) lambdas by method references
- issues with the Javadoc that I wrote

By the way, I really like
```
switch (x) {
  case 1 -> {
    int temp;
    stuff...;
  }
}
```
as opposed to
```
switch (x) {
  case 1: {
    int temp;
    stuff...;
    break;
  }
}
```
or
```
switch (x) {
  case 1: {
    int temp;
    stuff...;
  }
  break;
}
```
or even
```
switch (x) {
  case 1:
    int temp;
    stuff...;
    break;
}
```
all of which we have in our code base.

I declined to accept automatic conversion to new switch statements in a number of places where I, for example, had comments on the case labels; the conversion tool did not leave those in an attractive state.

Perhaps a future PR can convert the gigantic switch statements in ChoiceAdventure, ChoiceManager, ChoiceControl, etc.


jaadams5:
Thank you.  "lint" is my shorthand for doing this but I usually don't do it unless I have some other reason to be in the code.

At one point the justification for having source code in the lib tree was that the source was not ours and we did not want to maintain it.  If the owner did an up update the hope was we could use the update by doing a **replace** and not a **merge**.  I will defer to your judgement as to whether that needs to continue to be the case but if the philosophy does change maybe we add a text file to the tree and document the philosophy?

I trust your ability to run the IDE's analyze function and so, under the circumstances, I consider a self-approve as an acceptable alternative to reviewing 100+ files.

We may wish to avoid merging issues by getting this in before we do the 2022 to 2023 change in copyrights and other places or defer that change until this is committed.

src/net/sourceforge/kolmafia/session/BastilleBattalionManager.java
```diff
@@ -3,14 +3,7 @@
...
 import java.io.PrintStream;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
```
midgleyc, line 6: oops ;)

heeheehee [posted an xml file for formatting](https://kolmafia.us/threads/mafia-source-code-formatting.26013/#post-161996) a while ago. It might do with adding to the repo?

Veracity0: Irritating. It has a tool to "fix" those - on commit. I will fix all of this.

Yeah. We should have a standard file of code style, starting with heeheehee's
spotless will have to agree with it.
I also think spotless makes some things really ugly and hard to read.

Veracity0: I fixed this.

I grepped and found another 15 or so packages that have "*" imports for Java utilities, KoLmafia utilities, KoLmafia persistence, KoLmafia requests ...

I did not touch those in the PR - although I would have, had IntelliJ's automated tool to fix them been callable from the Inspections frame.

jaadams5: I tried to move some discussion to https://kolmafia.us/threads/source-code-formatting.28429/

I'm not sure the xml file for IntelliJ is still useful given the inclusion of spotless in the workflow but if spotless is not enforcing what we want we should try and "fix" spotless.

midgleyc: > I also think spotless makes some things really ugly and hard to read.

Definitely agreed for some things: I think the way it indents comments can be really dumb, and the way it spreads things across lines can also make it harder to read.

We could see if we could configure it to make it more amenable. I'm not sure there are any alternatives in this space.

lib/net/java/dev/spellcast/utilities/SortedListModel.java
```diff
@@ -139,6 +139,7 @@ public int indexOf( final Object o )
 	@Override
 	public boolean contains( final Object o )
 	{
+		//noinspection ListIndexOfReplaceableByContains
```
midgleyc, line 142: not really keen on these IDE specific comments

midgleyc (approved):
Historically we've not touched `lib`, treating them as vendored third-party code.

I think some is actually closer to our code (`SortedListModel`) and it might be worth considering them effectively managed by us, as I think the original authors may have dropped off the net.

lib/net/java/dev/spellcast/utilities/SortedListModel.java
Veracity0: Me either. But the suggested refactoring caused an infinite loop, if adopted.

I wanted to never see it again.
I can remove it.

midgleyc: If most of us are using Intellij, you can leave it. Normally I just avoid doing anything in the case that the suggestion is wrong.

There are more than just this one -- search for "//noinspection" to find them.

Veracity0:
Everything in lib/.../spellcraft came from a project that Hola worked on earlier in his career. Maybe even in college. He pulled in LockableListModel, ChatBuffer,and all sorts of other things from its "utilities" directory.

(I'm typing this on my phone. My one-fingered typing had LickableLustModel, which is not QUITE right. 😀 Easy typo, considering {uio} position on keyboard.)

I was the one who added Generics to LockableListModel and SortedListModel. We have already heavily customized those packages.

I consider the spellcraft Packages to be part of KoLmafia source, at this point. There will never be another source release of the program they came from.

PR #1381, opened 2022-12-29
Use setForeground instead of HTML for coloring equipment list.

phulin:
While executing scripts, rendering the JLabels in the gear changer takes up something like 10-15% of CPU time in my profiling runs. The menus are all fully rerendered every time there's e.g. a status update that might change what equipment is equippable.

Changing from HTML to the basic JLabel methods dramatically reduces the amount of time needed to render. It goes from 10-15% of CPU time to essentially no time at all.


midgleyc (approved):
nice!

PR #1382, opened 2022-12-29
Optimize mall price database.

phulin:
This PR optimizes one of the remaining CPU bottlenecks in the code, the mall price database. Every write of the database serializes every recorded price again, which is slow—taking as much as 30% of execution CPU time while running mall-heavy segments of garbo.

Much like in Preferences, we can just cache the encoded values and only concatenate them when the database is written to disk. The PR also removes double-recording of mall prices when `mall_price` is called from the runtime library—in the current code `getMallPrice` saves the price and calls `searchMall`, which also saves the price.

MallPriceDatabase/MallPriceManager are sufficiently well-tested that I don't think any new tests are needed.

The PR increases speed of the following synthetic benchmark by a factor of roughly 2, and the real-world impact should be better due to removal of the double-recording behavior.

```java
  @Test
  void speed() {
    int N = 300;
    MallPriceDatabase.savePricesToFile = true;
      MallPriceDatabase.recordPrice(entry.getKey(), entry.getKey(), true);
    }
    for (int i = 0; i < 2 * N / 5; i++) {
    }
    long start =
        ManagementFactory.getThreadMXBean().getThreadCpuTime(Thread.currentThread().getId());
    for (int i = 0; i < N; i++) {
    }
    long end = ManagementFactory.getThreadMXBean().getThreadCpuTime(Thread.currentThread().getId());
    new File(KoLConstants.DATA_LOCATION, "mallprices.txt").delete();
  }
```

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
@@ -12,10 +12,11 @@ import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.BodyPublishers;
 import java.net.http.HttpResponse;
 import java.net.http.HttpResponse.BodyHandlers;
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Map;
+import java.util.SortedMap;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import net.sourceforge.kolmafia.KoLConstants;
@@ -33,7 +34,7 @@ public class MallPriceDatabase {
   // If false, blocks saving of mall prices. Do not modify outside of tests.
   public static boolean savePricesToFile = true;
 
-  private static final Map<Integer, Price> prices = new HashMap<>();
+  private static final SortedMap<Integer, Price> prices = new TreeMap<>();
   private static final HashSet<String> updated = new HashSet<>();
   private static final HashSet<String> submitted = new HashSet<>();
   private static int modCount = 0;
@@ -86,12 +87,11 @@ public class MallPriceDatabase {
         if (!ItemDatabase.isTradeable(id)) continue;
         Price p = MallPriceDatabase.prices.get(id);
         if (p == null) {
-          MallPriceDatabase.prices.put(id, new Price(price, timestamp));
+          MallPriceDatabase.prices.put(id, new Price(id, price, timestamp));
           ++count;
           ++MallPriceDatabase.modCount;
         } else if (timestamp > p.timestamp) {
-          p.price = price;
-          p.timestamp = timestamp;
+          p.update(price, timestamp);
           ++count;
           ++MallPriceDatabase.modCount;
         }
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
```diff
@@ -86,12 +87,11 @@ private static int updatePricesFromSource(String filename) {
         if (!ItemDatabase.isTradeable(id)) continue;
         Price p = MallPriceDatabase.prices.get(id);
         if (p == null) {
-          MallPriceDatabase.prices.put(id, new Price(price, timestamp));
+          MallPriceDatabase.prices.put(id, new Price(id, price, timestamp));
```
jaadams5, line 90: For some reason this bothers me.  I know you want prices sorted by id, but could you get the same behavior by making Price implement Comparable and having the associated compareTo just do the expected sort on id?  Or maybe there is a  different implementation that does not include id in price?

phulin: Either Price has to know its id in order to updated the encoded field when a new price/timestamp is computed, or the id has to be passed into the Price object on each update. I prefer the former, let me know if you disagree.

jaadams5: I have no problem including id as part of price.  What I would like to explore is the data structures that contain Price.  If they have been selected, and use id as a key, for various sorts on id, an alternative might be to make price implement Comparable.  I also wonder whether you can accomplish what is needed by never storing anything but encoded prices and decoding when they are used.  I might also explore storing the encoding in a record so there is just one data structure with price info but the places that use it don't always access all fields.  Make sense?

jaadams5: Just noticed encoded data is in price so some of my comments no longer apply.

jaadams5: Looking at it, I think you don't add id as a field to price.  Instead you add it as a parameter to update.  I think every place update is called the id is already known.

phulin: I'm not sure why you would need to make Price implement Comparable - could you explain that? `MallPriceDatabase.prices` serves two purposes: it 1) tracks all prices for when we need to save/load from file, and it 2) tracks cached prices for when we do lookups. Purpose 2 means it has to function as a map from some kind of item identifier to the price object. Purpose 1 implies it should be a sorted data structure, or we should keep track of a sorted view of the data structure. The cleanest approach in my view is to make it a sorted map since we're not doing so many lookups that TreeMap vs HashMap matters.

jaadams5: OK.  I talked myself out of moving id out of price and then passing id to update.

jaadams5: And you win :-)  I tried to implement my own suggestions and decided at best my way was different but not better.

jaadams5 (approved):

PR #1385, opened 2022-12-30
refactor: prefer skill ids to skill names when checking hasSkill

midgleyc:
Avoid a map lookup where we know the skill IDs already.


Veracity0 (approved):
Thank you!

PR #1387, opened 2022-12-30
Remove HTML from CompactSidePane.

phulin:
Much like the equipment PR, rendering HTML in the side pane is unnecessary computational effort. This PR removes HTML from the side pane rendering code in favor of native Java layout tools. It's not a huge speedup but seems to be a reduction of 1-2% of CPU time in my tests.

### DIFF
--- a/src/net/sourceforge/kolmafia/swingui/panel/CompactSidePane.java
+++ b/src/net/sourceforge/kolmafia/swingui/panel/CompactSidePane.java
@@ -2,6 +2,7 @@ package net.sourceforge.kolmafia.swingui.panel;
 
 import java.awt.BorderLayout;
 import java.awt.CardLayout;
+import java.awt.Color;
 import java.awt.Dimension;
 import java.awt.FlowLayout;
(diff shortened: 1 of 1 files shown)


midgleyc (changes requested):
There looks to be a small unintended change: previously, the tooltip (hover text) on the familiar read "<Familiar type> (X lbs)". Now it reads "<Player name>'s <Familiar type> (X lbs)". Can we go back to the previous text?

phulin:
Sorry, where do you see that? Here's what mine shows, this must be somewhere else:

image

midgleyc:
No, it's exactly there. Mine has the player name before the familiar type. How odd.

midgleyc (approved):
issue unrelated to PR but still weird

Veracity0:
My hover text is fine.
However.

Screen Shot 2022-12-31 at 12 33 54 PM

phulin:
Ah, a possibility I didn't consider. WIll fix.

PR #1388, opened 2022-12-30
Use non-autoflush PrintStream rather than LogStream for mall prices.

phulin:
I meant to include this in #1382 - the price writing code should use a buffered output stream, instead of a LogStream which auto-flushes each byte array write.

This PR reduces time on the below benchmark from about 10 s to about 500 ms.

```java
  @Test
  void speed() {
    int N = 300;
    MallPriceDatabase.savePricesToFile = true;
      MallPriceDatabase.recordPrice(entry.getKey(), entry.getKey(), true);
    }
    for (int i = 0; i < 2 * N / 5; i++) {
    }
    long start =
        ManagementFactory.getThreadMXBean().getThreadCpuTime(Thread.currentThread().getId());
    for (int i = 0; i < N; i++) {
    }
    long end = ManagementFactory.getThreadMXBean().getThreadCpuTime(Thread.currentThread().getId());
    new File(KoLConstants.DATA_LOCATION, "mallprices.txt").delete();
  }
```

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia.persistence;
 
+import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


midgleyc (approved):
Confirmation: does the writer flush automatically before being closed?

This is probably a problem in many places in the codebase; could it be centralised somewhere? Doesn't have to be done as part of this PR. I think we have many classes that deal with reading from / writing to files already.

phulin:
It does flush before writing, see PrintStream.close documentation:

image

phulin:
And yeah, it should be done elsewhere. DataUtilities.getOutputStream should probably return a buffered output stream by default...

PR #1389, opened 2022-12-31
Speed improvements for maximizer.

phulin:
This PR makes a number of small changes to the maximizer. Together, they take the runtime on the below synthetic benchmark from 8-9 s down to roughly 6 s.

```java
  @Test
  public void speed() {
    int N = 40;
      KoLConstants.inventory.add(ItemPool.get(entry.getKey(), 1));
    }
    int k = 0;
    for (var entry : SkillDatabase.entrySet()) {
      if (k++ % 4 == 0) KoLCharacter.addAvailableSkill(entry.getKey());
    }
    long start = 0;
    ThreadMXBean threadMXBean = ManagementFactory.getThreadMXBean();
    for (int i = -N / 2; i < N; i++) {
      if (i == 0) start = threadMXBean.getCurrentThreadUserTime();
      Maximizer.maximize("5item,meat,0.5initiative,0.1da 1000max,dr,0.5all res,1.5mainstat," +
          "-fumble,mox,0.4hp,0.2mp 1000max,3mp regen,0.25spell damage,1.75spell damage percent");
    }
    long end = threadMXBean.getCurrentThreadUserTime();
  }
```


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1295,7 +1295,7 @@ public static final int findBooleanName(String name) {
   private final double[] doubles;
   private final int[] bitmaps;
   private final String[] strings;
-  private ModifierExpression[] expressions;
+  private ArrayList<Indexed<ModifierExpression>> expressions;
```
midgleyc, line 1298: Why is this an improvement? Was it a sparse array to begin with?

src/net/sourceforge/kolmafia/Modeable.java
```diff
@@ -139,18 +149,20 @@ public static Modeable find(final String command) {
...
+    return idToModeable.get(itemId);
   }
 
   public static Map<Modeable, String> getStringMap(Function<Modeable, String> cb) {
     return Arrays.stream(values()).collect(Collectors.toMap(m -> m, cb));
   }
 
   public static Map<Modeable, String> getStateMap() {
-    return getStringMap(Modeable::getState);
+    // This is performance-critical for the maximizer and needs to be written imperatively.
```
midgleyc, line 160: It is a real shame just how non-performant streams are when compared with for loops. I miss languages like C, where the compiler is a magic wand that can detect code that looks like it's doing a certain operation and optimise it for you.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5137,10 +5142,10 @@ public static final Modifiers recalculateAdjustments(
         Modifiers.evaluatedModifiers("Mummery", Preferences.getString("_mummeryMods")));
 
     // Add modifiers from inventory
-    if (InventoryManager.hasItem(ItemPool.FISHING_POLE)) {
+    if (InventoryManager.getCount(ItemPool.FISHING_POLE) > 0) {
```
midgleyc, line 5145: I think this technically does something slightly different. Does `hasItem` also get the item out of the closet?

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -2131,13 +2131,21 @@ public static final SpecialOutfit currentOutfit() {
...
+    // Go through any outfit that any worn item belongs to.
     int hash = SpecialOutfit.equipmentHash(equipment);
-    for (SpecialOutfit outfit : EquipmentDatabase.normalOutfits.values()) {
-      if (outfit == null) {
-        continue;
-      }
-      if (outfit.isWearing(equipment, hash)) {
-        return outfit;
+    List<Integer> checkedOutfits = new ArrayList<>();
+    for (int i = 0; i < EquipmentManager.FAMILIAR; i++) {
```
midgleyc, line 2137: this sort of thing where we treat the HAT...FAMILIAR stuff as integers always looks dodgy to me (especially as 2 is "holster").

You could start it "for (int i = EquipmentManager.HAT" to make it slightly clearer what's going on. Maybe it's okay as is.

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -2131,13 +2131,21 @@ public static final SpecialOutfit currentOutfit() {
...
   public static final SpecialOutfit currentOutfit(AdventureResult[] equipment) {
+    // Go through any outfit that any worn item belongs to.
     int hash = SpecialOutfit.equipmentHash(equipment);
-    for (SpecialOutfit outfit : EquipmentDatabase.normalOutfits.values()) {
-      if (outfit == null) {
-        continue;
-      }
-      if (outfit.isWearing(equipment, hash)) {
-        return outfit;
+    List<Integer> checkedOutfits = new ArrayList<>();
```
midgleyc, line 2136: may as well be a Set, yeah? Order doesn't matter, `add` and `contains` want to be fast.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5594,6 +5600,10 @@ public static final double getSmithsnessModifier(
     }
 
     for (AdventureResult effect : effects) {
+      int effectId = effect.getEffectId();
+      if (effectId != EffectPool.VIDEO_GAMES
```
midgleyc, line 5604: comment here please or we're going to forget to update it.

or could we pre-cache the effect ids that have smithsness by looking up modifiers at the start of the maximization session (or mafia session, when first requested)?

src/net/sourceforge/kolmafia/Modifiers.java
phulin: Right now it's a dense array, almost all of the elements of which are null. We iterate over all of them looking for expressions to evaluate. Instead, we should store a store list of pairs of (index, expression), which is what this does.

src/net/sourceforge/kolmafia/Modeable.java
phulin: Agreed, sadly. It's not horrendously slower but enough to be noticeable, and this gets called very frequently.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: Yeah, this looked wrong and I fixed it while I was there. `hasItem` checks whatever you have marked as `autoSatisfy`, which is definitely not the right model for "does this item affect me by being in my inventory right now."

src/net/sourceforge/kolmafia/session/EquipmentManager.java
phulin: I'll make that change, thanks, but yeah, this is a bad pattern.

src/net/sourceforge/kolmafia/session/EquipmentManager.java
phulin: It's going to be one or two elements the vast majority of the time, so I'd expect a list to be better.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: Sure, I'll add a comment. I wanted to hardcode the code to evaluate whether or not it's a smithsness effect so it's as fast as possible. I think it's unlikely effects will be added in the future, but not impossible of course.

midgleyc (approved):
Thanks!

PR #1392, opened 2023-01-01
chore: don't time tests any more

midgleyc:
The problem seems to have gone away, so don't time tests by default.

Leave the class in in case somebody (e.g. me) wants to time a test locally.


jaadams5 (approved):

jaadams5:
I don't mind the default being "time the tests" but we can reenable locally or globally if there are people who do care.

Veracity0:
I don't think the problem has gone away. I had a PR that was still processing (her) on Windows after 24 minutes. I restarted and it finished in 7 minutes.

Having the timings going to stdout let me look at the test output (here) and see that it WAS continuing. Slowly.

I\But, I agree - if it is easy to reenable locally when we want to see how long a test (or package of tests) is taking, we don't need it as the default.

Having it enabled or disabled will not negate having to restart tests for a platform whose instance is processing the tests really really slowing.

PR #1394, opened 2023-01-01
Store modifiers sparsely when there are only a few.

phulin:
The maximizer spends most of its time in `Modifiers::add`, so that's a prime target for improvements. Every time that function runs, it iterates over all 145 elements of `Modifiers::doubles` to add any double modifiers on the right-hand side to the `Modifiers` on the left-hand side. In the vast majority of cases, this is used to accumulate modifiers for e.g. an item onto an object that is accumulating all modifiers for the player. In these cases, the right-hand side will have an effectively sparse doubles array.

This all means that it makes more sense to store `Modifiers` as sparse, rather than dense, arrays until they reach a certain size. The intuition is that we want the right-hand side in `add` to be sparse and the left-hand side to be dense. This PR does just that.

Seems to be a 10-15% speedup on the benchmark I posted earlier.


phulin:
Not sure what is going on with the checks, I think they're okay.

midgleyc (changes requested):
What's the perf if you just always use a Map (probably a HashMap, as I don't think order matters in that case)?

I'm concerned about the apparent increase in complexity in the code. If the perf change is significant, could you make a class that stores either a map or an array, and do all the logic to get / set / iterate inside that? e.g. `getInternal`, `densify`, the `set` logic, etc.

phulin:
> What's the perf if you just always use a Map (probably a HashMap, as I don't think order matters in that case)?

The naive version using HashMap is not good, equivalently bad to the current code, unless you put in some more sophisticated logic to pick the size. There are a few issues:
* As the Modifiers object that acts as an accumulator grows, it triggers multiple resizes of the HashMap which need to copy every value. 
* The default HashMap capacity is 16, so any `add()` call has to iterate over 16 entries in the RHS to find the modifiers to add. Neither reducing nor increasing the initial capacity helps.
* It seems be spending a lot of time calling `HashMap::hash`, which should be fast for integer keys. I'm not sure exactly why. Maybe boxing/unboxing. Or the profiler might be confused.

I think you'd need some kind of densify function that upsized it when accumulating onto, which would end up looking a lot like this PR in terms of complexity.

Using a TreeMap doesn't work very well either, since it has to spend lots of time traversing the tree of the LHS.

> I'm concerned about the apparent increase in complexity in the code. If the perf change is significant, could you make a class that stores either a map or an array, and do all the logic to get / set / iterate inside that? e.g. `getInternal`, `densify`, the `set` logic, etc.

It's tricky because the `add` methods tie together other logic in `Modifiers` with the updates to the data structure itself. I'll keep thinking about it and report back.

phulin:
Okay, how's this? @midgleyc

midgleyc (approved):
> It seems be spending a lot of time calling `HashMap::hash`

Very weird!

The extraction of the data structure is just as I'd hoped! Thanks!

phulin:
Sorry, there was a test failure caused by an NPE. Needs to be looked at again :)

midgleyc (approved):

phulin:
Conflicts resolved.

midgleyc (approved):

PR #1395, opened 2023-01-01
More maximizer speed improvements.

phulin:
* Cache passive modifier modifiers to avoid extra calls to `getModifier`.
* Cache voter and mummery modifiers, as it's expensive to parse/evaluate every time `recalculateAdjustments` is called.
* Pass IDs all the way down the call tree when available to `override`, avoiding a hit to the idByName maps.


midgleyc (approved):
Nice use of a listener. We should use those more...

PR #1396, opened 2023-01-01
Renovate modifier-doublers code and add Dirty Pear.

phulin:
This PR cleans up some of the modifier-doubling code (Bend Hell, Squint, Champagne, etc).

NB that the modifier-name check in the prior code was never true, as modifiers for items, skills, and effects are written with a bracketed ID, not the actual name.


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -1,5 +1,7 @@
 package net.sourceforge.kolmafia;
 
+import static net.sourceforge.kolmafia.Modifiers.MUS_EXPERIENCE_PCT;
```
midgleyc, line 3: Might be better to either import everything statically or nothing? Just one looks a bit odd.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5321,46 +5325,27 @@ public static final Modifiers recalculateAdjustments(
...
+              Modifiers.STENCH_DAMAGE,
+              Modifiers.SPOOKY_DAMAGE,
+              Modifiers.SLEAZE_DAMAGE,
+              Modifiers.HOT_SPELL_DAMAGE,
+              Modifiers.COLD_SPELL_DAMAGE,
+              Modifiers.STENCH_SPELL_DAMAGE,
+              Modifiers.SPOOKY_SPELL_DAMAGE,
+              Modifiers.SLEAZE_SPELL_DAMAGE)) {
+        newModifiers.add(
+            modifier, newModifiers.getExtra(modifier), "Effect:[" + EffectPool.BENDIN_HELL + "]");
```
midgleyc, line 5341: as a tiny improvement you could create the `"Effect:[" + EffectPool.BENDIN_HELL + "]"` string outside the loop?

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1632,20 +1634,17 @@ public void add(final int index, final double mod, final String desc) {
...
-        if (!name.equals("makeshift garbage shirt")) {
-          this.extras[index] += mod;
-        }
+      case MUS_EXPERIENCE_PCT:
+      case MYS_EXPERIENCE_PCT:
+      case MOX_EXPERIENCE_PCT:
+        // Possibly doubled by makeshift garbage shirt, Bendin' Hell, Bow-Legged Swagger, or
+        // Dirty Pear.
+        // TODO: Figure out which ones aren't doubled and exclude them. BoomBox?
+        this.extras[index] += mod;
```
midgleyc, line 1647: sorry, I don't understand what's gone on here. I'm not familiar with this area: what are "extras"? "doubles" are modifiers which are of type `double`? Why did the previous one conditionally add to extras, but you unconditionally add?

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5404,6 +5385,19 @@ public static final Modifiers recalculateAdjustments(
           Modifiers.ITEMDROP,
           newModifiers.getExtra(Modifiers.ITEMDROP),
           "Item:[" + ItemPool.BROKEN_CHAMPAGNE + "]");
+      // Subtract fightMods back out.
+      if (fightMods != null) {
```
midgleyc, line 5389: what are `fightMods`, and why are they being subtracted here?

midgleyc (changes requested):

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: Mistake, thanks.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: I'm planning on moving it to a two-level map at some point rather than the current system of string lookups. OK if we leave this? I suspect it may get optimized away anyway.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: `fightMods` comes from the lookup string `Generated:fightMods`. It stores modifiers that can occur during a fight for the duration of the fight, a prime example being Otoscope. Squint doubles Otoscope, but Champagne doesn't, and we should reproduce that behavior. The cleanest way I found to do it was to have it counted on extras and then subtract it back out later.

src/net/sourceforge/kolmafia/Modifiers.java
phulin: The previous check was never true, as it had been written incorrectly (the lookup string passed in looks like Effect[1234], not Effect[name] which is what the check was expecting). I'll add a comment to better explain how it currently works and possibly change the name.

`doubles` is (in current code) an array of all modifiers of type double - anything you would reference in ASH as a `numeric_modifier`. When two Modifiers objects are added together, typically the array on the RHS has only a few nonzero values, and the array on the LHS is used as an accumulator. `extras` is used as an accumulator for what any doubler would add, since some sources of e.g. item drop aren't subject to doubling. Later, in `recalculateAdjustments`, as you can see the code checks to see whether you have each doubler and adds the `extras` value to `doubles` if so. This handles double-doubling cleanly too, which is a nice feature of the existing code.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: Alternatively, I could rewrite it to add fightMods in explicitly at the Squint stage. That might be better.

src/net/sourceforge/kolmafia/KoLCharacter.java
midgleyc: sure, sounds good

src/net/sourceforge/kolmafia/Modifiers.java
phulin: Changed the name of the field.

src/net/sourceforge/kolmafia/KoLCharacter.java
midgleyc: it does seem clearer that way

midgleyc (changes requested):
Looking at the uncovered code, would you mind adding a test for the garbage shirt?

phulin:
Sure.

phulin:
Done, thanks.

midgleyc (approved):

PR #1397, opened 2023-01-01
Don't add null concoctions for speakeasy/hot dog.

phulin:
This is causing an NPE.


src/net/sourceforge/kolmafia/session/ClanManager.java
```diff
@@ -155,8 +155,10 @@ public static final void changeClan(final int clanId, String clanName) {
         // such need to be re-added as concoctions for the returned-to clan
         for (AdventureResult item : ClanManager.getClanLounge()) {
           if (ClanLoungeRequest.isSpeakeasyDrink(item.getName())) {
-            ConcoctionDatabase.getUsables()
-                .add(ClanLoungeRequest.addSpeakeasyDrink(item.getName()));
+            Concoction c = ClanLoungeRequest.addSpeakeasyDrink(item.getName());
```
Veracity0, line 158: ClanLoungeRequest.addSpeakeasyDrink(itemName) returns null if:
- ClanLoungeRequest.speakeasyNameToIndex(itemName) < 0 -- i.e. it is not a speakeasy drink
OR
- ClanLoungeRequest.speakeasyNameToIndex(itemName) -- i.e. it is already on the the usables list.

If it is not on the usables list, it resets calculations for the the item

My question: Is there any reason it should not also add it to the usables list at that time, rather than depending on the caller to do it if it returns non-null. I.e., it is not already on the list?

Is there any place where addSpeakeasyDrink is called that you do NOT want it to end up on the usables list?

The only other call is in ClanManager.java - which also iterates over hotdogs - and calls addHotdog and adds them to usables.

I'd like you to look a little harder, here, rather than just doing a null check at the point of failure.

I conjecture that both "addSpeakeasyDrink" and "addHotDog" should add to usables list, if not already present, rather than depending pon caller to do it

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/session/ClanManager.java
Veracity0: Looking deeper:

Three kinds of lounge furniture give you items:

Speakeasy - actual items, complete with ID, that are consumed upon purchase
Hot Dog Stand - pseudo items, no official id, that are consumed upon purchase
Floundry - actual items, complete with ID, that appear in inventory.

hot dogs and speakeasy drinks have to be unlocked and may or may not be available in a particular clan.
hot dogs and floundry items require ingredients to be "stocked" in order to be available.

All of them have creation methods and are thus Concoctions

ClanLoungeRequest has:
resetHotDogs, resetSpeakeasy, and resetFloundry which remove all corresponding items from usablesList
availableHotDog, availableSpeakeasyDrink which returns true if on usables, false otherwise
addHotDog, addSpeakeasyDrink which returns null if on usables, a Concoction otherwise

I think a lot of refactoring/cleanup could be done.

That need not be in the scope of this PR.

phulin: Yeah, this is messy and it's partially why I haven't come up with a better solution so far. I think the code should ultimate enforce the invariant that the usablesList never changes, unless we encounter an unknown-to-Mafia item. As far as I can tell this is one of the few places in the code that's actually modifying the usablesList based on dynamic things in the game rather than a one-time change at init. I'm inclined to leave this PR the way it is for now, and someone can come back to your nicely-documented thoughts later.

Veracity0: I think so, too, now.
This can go in as is.

Veracity0 (approved):

phulin:
Unfortunately I don't have merge privileges so you'll have to do more than just approve :)

PR #1398, opened 2023-01-02
Added Golden Mr. Eh

poshpotato:
Custom Item given to Invariel (#1043600) that was not in the database,
https://kol.coldfront.net/thekolwiki/index.php/Golden_Mr._Eh%3F


midgleyc (approved):
Thanks!

You had some issues with using spaces instead of tabs in the data files; I fixed those myself.

PR #1400, opened 2023-01-02
Refactor Speakeasy support in ClanLongeRequest

Veracity0:
As discussed in https://github.com/kolmafia/kolmafia/pull/1397 the code in ClanLoungeRequest to handle Speakeasy Drink, Hot Dogs, and Floundry items is dubious. Among other things:

- Uses arrays of Record Structures and does all its work in terms of the "index" of a given structure in the array.
- Adds and deletes concoctions to the "usables" list depending on whether they are currently available in your clan.

This PR will attempt to  improve all of those.

To start, this refactors the code for the Speakeasy

1) Eliminate usage of array indices to manipulate SpeakeasyDrink objects
2) Added availableSpeakeasyDrink set which is cleared or added to when looking at joining or leaving a clan
3) Adds a test for parsing the Speakeasy.
Note that I removed the clan name from the HTML. Not because I think it's private - it's a typical clan that has every clan item ;) - but because when we parse a respone from a clan lounge visit, if the clan name is different than expected, we generate extra requests for visiting a "new clan".
4) Added ClanManager setClan(clanId, clanName) for use in testing.
This sets those variables and also resets all cached clan data.
5) Added withClan() and withClan(clanId, clanName) to use that in testing.
6) Speakeasy drinks are no longer added to and removed from the usables list. They are always present.
Their visibility (availability), however, now depends on whether your clan has them, whether you can afford them, and whether you have not drunk your fill (3) from the Speakeasy already today.

And then on to Hot Dogs and Floundry items.

Yes - but in other PRs. Each of those is sufficiently different from the Speakeasy - although both have the flaw of modifying the usables list - that it will take more than copying and modifying the Speakeasy code - although that will certainly be useful.


src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
```diff
@@ -340,94 +324,202 @@ public static final boolean isFancyHotDog(String name) {
...
+   *    and creatable = 0 if the drink isn't on the available list. If it is on that list,
+   *    creatable can be 0, 1, 2, or 3, depending on available Meat and # of speakeasy
+   *    drinks consumed today.
+   *    (This satisfies the last two checks)
+   * 4) We need to track speakeasy drinks consumed - _speakeasyDrinksDrunk - and trigger
+   *    concoction.resetCalculations() for all speakeasy drinks.
+   *
+   * *** Current implementation adds or removes speakeasy drinks to the usables list.
+   * *** That is incorrect; the list should remain untouched and visibility of the drink
+   * *** determined by the ListCellRenderer
```
midgleyc, line 377: is this a description of how it worked before you changed it, or how it still works now?

src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
```diff
@@ -340,94 +324,202 @@ public static final boolean isFancyHotDog(String name) {
...
+      concoction.speakeasy = true;
+      concoction.price = cost;
+      this.concoction = concoction;
+
+      // Lookups
+      SpeakeasyDrink.nameToData.put(name, this);
+      SpeakeasyDrink.canonicalNameToName.put(canonicalName, name);
+      SpeakeasyDrink.idToData.put(id, this);
+
+      ClanLoungeRequest.allSpeakeasyDrinks.add(this);
```
midgleyc, line 431: while this is a pattern that occurs elsewhere in the Mafia codebase, I really would much prefer it if the constructor didn't have any side effects and the static initialization block at line 463, instead of just calling each constructor, called something like `populateSpeakeasyData()` with the relevant values (for each drink) which did the setup.

src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
Veracity0: It was the original implementation, as a reminder to myself what I was fixing.
I could have sworn I removed it.
I will do so.

src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
Veracity0: This was previously a record. I rejected that since those are not allow to have private - "derived" - fields.
Both a class and an enum allow those.
Additionally, both allow private static members.
gausie had a class (or enum?) with a private static map and a public static "findInMap" method.

I quite liked that, since the maps were private to the inner class, rather than visible (public or private) in the enclosing class, and anybody with access to class could call the static class method to lookup an object.

Do you object to private static class members in inner classes?
If not, surely they have to be manipulated in the constructor?
It is a "side effect".

You can do the same thing with enums - and, in a case like this, where all the members are known at compile time, that is probably a better choice. The global map could be a EnumSet.allOf(SpeakeasyDrink).

I will try converting to an Enum and we can see what it looks like then.

If the issue is that you don't think inner classes should have static members, we will probably have to agree to disagree. :)

midgleyc: > If not, surely they have to be manipulated in the constructor?

or from methods called off the object, outside the constructor?

> It is a "side effect".

I suppose the main reason I don't like this style of construction is that it has the following code:

```
static {
  new SpeakeasyDrink("glass of &quot;milk&quot;", 1, 1, 250);
  new SpeakeasyDrink("cup of &quot;tea&quot;", 2, 1, 250);
  new SpeakeasyDrink("thermos of &quot;whiskey&quot;", 3, 1, 250);
  new SpeakeasyDrink("Lucky Lindy", 4, 1, 500);
  new SpeakeasyDrink("Bee's Knees", 5, 2, 500);
  new SpeakeasyDrink("Sockdollager", 6, 2, 500);
  new SpeakeasyDrink("Ish Kabibble", 7, 2, 500);
  new SpeakeasyDrink("Hot Socks", 8, 3, 5000);
  new SpeakeasyDrink("Phonus Balonus", 9, 3, 10000);
  new SpeakeasyDrink("Flivver", 10, 2, 20000);
  new SpeakeasyDrink("Sloppy Jalopy", 11, 5, 100000);
}
```

We create a series of `SpeakeasyDrink`s, and then throw them all away. Without knowing that the constructor adds all constructed objects to a list, this seems quite surprising.

Additionally, it means that you can't create a new `SpeakeasyDrink` object, because the constructor does something in addition to merely constructing the object.

There's [some discussion on StackOverflow](https://softwareengineering.stackexchange.com/questions/343447/should-constructors-ever-be-used-only-for-side-effects) about using constructors for side effects: the answer references SRP and the principle of least astonishment. I was never too big on SRP as I saw it massively overused to fragment classes into far-too-tiny portions, but I think the principle of least astonishment has some merit. I expect a constructor to construct an object, and find it surprising if it also does something else (e.g. adding the object to a list).

Veracity0: OK. SpeakeasyDrink is now an enum with no side effects in its constructor.
See how you like it.

midgleyc (approved):
Looks good now, thank you!

PR #1401, opened 2023-01-02
Refactor Modifiers database to use an enum type / int or string key for two-level lookups.

phulin:
Interested in comments here - this is more about making the code prettier than performance, although the old system was pretty slow. This PR creates a new ModifierType and uses that and an int or string, depending on type, to index into the modifier tables. The vast majority of the changes are in Modifiers.java.


src/net/sourceforge/kolmafia/utilities/StringUtilities.java
```diff
@@ -921,4 +922,10 @@ public static String stripHtml(String s) {
   public static String capitalize(final String s) {
     return s.substring(0, 1).toUpperCase() + s.substring(1);
   }
+
+  public static String pascalToCamelCase(final String s) {
```
heeheehee-kolmafia, line 926: This is UPPER_SNAKE_CASE, not PascalCase.

heeheehee-kolmafia:
Is this user- / script-visible in any way?

(I think passing around enums are an improvement over strings that need to be parsed and can be error-prone, between typos / case sensitivity etc. That said, I'd like to avoid breaking user scripts unless it's unavoidable in exchange for a clear win.

I might come back for a more detailed review later, just providing some initial thoughts from a very cursory glance.)

phulin:
This should not be user-visible at all, although it's always possible there are changes to edge-case behavior.

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -662,7 +662,8 @@ public Value get_noob_skill() {
      * @return The Skill granted
      */
     public Value get_skill() {
-      String skillName = Modifiers.getStringModifier("Item", (int) this.contentLong, "Skill");
+      String skillName =
+          Modifiers.getStringModifier(ModifierType.ITEM, (int) this.contentLong, "Skill");
```
heeheehee-kolmafia, line 666: This file is missing an import for ModifierType, hence test failures. Looks like your latest commit lost this when you replaced the wildcard.

heeheehee-kolmafia:
My specific concern was regarding the Generated:_spec modifiers -- see https://wiki.kolmafia.us/index.php/Whatif

From staring at the code more closely, numeric_modifier parses the type in RuntimeLibrary#getModifierType, which you've updated appropriately, so this shouldn't be an issue.

src/net/sourceforge/kolmafia/utilities/StringUtilities.java
heeheehee-kolmafia: Rather... the conversion is UPPER_SNAKE_CASE (or UPPER_CASE_WITH_UNDERSCORES if you prefer Python's naming) to PascalCase.

heeheehee-kolmafia:
I see some potential for mismatch if the old name had a space in it (e.g. Local Vote), since you're storing this as a single-word lowercase name (localvote). I can't see any way of getting at these specific modifiers directly (other than debugging output via modtrace), so that's probably fine.

src/net/sourceforge/kolmafia/session/GreyYouManager.java
```diff
@@ -486,7 +487,7 @@ public static class GooSkill extends Absorption {
     private final int level;
 
     private String enchantments = "";
-    private String modsLookup = "";
+    private Modifiers.Lookup modsLookup = new Modifiers.Lookup(ModifierType.GENERATED, "");
```
heeheehee-kolmafia, line 490: (nothing to address here; just noting that you're assigning the modifier type GENERATED unilaterally here and in a few other cases.)

heeheehee-kolmafia (changes requested):
Can you explain the motivation behind making this a two-level lookup instead of using a map keyed by a composite type (such as the convenient Lookup class that you've added)?

(I don't have time right now to finish this review. I got through about half the files.)

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1581,7 +1627,11 @@ public boolean set(final Modifiers mods) {
           "Throne",
           "UnbreakableUmbrella");
```
heeheehee-kolmafia, line 1628: Should doubledBySquintChampagne be a `Set<ModifierType>`? Set#contains takes an Object, and it's unlikely that anything casts its argument from a ModifierType to a String.

src/net/sourceforge/kolmafia/DebugModifiers.java
```diff
@@ -55,32 +55,24 @@ private static void flushRow() {
...
-      type = desc.substring(0, ind);
-      name = desc.replace(type + ":", "");
-    } else {
-      type = "";
-      name = desc;
-    }
-    if (!desc.equals(DebugModifiers.currentDesc) || DebugModifiers.adjustments.containsKey(key)) {
+    ModifierType type = lookup.type;
+    String name = lookup.getName();
+    if (!lookup.equals(DebugModifiers.currentDesc) || DebugModifiers.adjustments.containsKey(key)) {
```
heeheehee-kolmafia, line 72: currentDesc should probably be renamed currentLookup, or possibly currentName.

Should this be a comparison against `name`, since you set currentDesc to `name` below?

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5049,28 +5051,27 @@ public static final Modifiers recalculateAdjustments(
...
 
     int brimstoneMonsterLevel = 1 << newModifiers.getBitmap(Modifiers.BRIMSTONE);
     // Brimstone was believed to affect monster level only if more than
     // one is worn, but this is confirmed to not be true now.
     // Also affects item/meat drop, but only one is needed
     if (brimstoneMonsterLevel > 1) {
-      newModifiers.add(Modifiers.MONSTER_LEVEL, brimstoneMonsterLevel, "Outfit:brimstone");
-      newModifiers.add(Modifiers.MEATDROP, brimstoneMonsterLevel, "Outfit:brimstone");
-      newModifiers.add(Modifiers.ITEMDROP, brimstoneMonsterLevel, "Outfit:brimstone");
+      newModifiers.add(Modifiers.MONSTER_LEVEL, brimstoneMonsterLevel, ModifierType.OUTFIT);
```
heeheehee-kolmafia, line 5062: I'm really not a fan of discarding the information in Outfit:brimstone -> Outfit.

If you have one piece of Cloathing and one piece of brimstone equipped, modtrace will print out the resulting item drop for both in a way you can differentiate between the two. Not so in this case.

heeheehee-kolmafia (dismissed):
I don't need / expect there to be zero visible changes in modtrace output, especially for cases where it's pretty obvious what the extra info is (e.g. Ballroom:quartet -> BALLROOM), but IMO this PR errs on the side of removing too much debug info. I pointed out at least one of the more egregious cases, but there are at least a few others where the disambiguation is useful.

midgleyc:
Happy to see this change, but given the intent:

> This should not be user-visible at all, although it's always possible there are changes to edge-case behavior.

can we have some more tests, focusing on behaviour that has changed but wasn't intended and behaviour that has changed but is minor and so on? Ideally at a high enough level such that the same test passes before and after without changes needed (for ease of comparison): maybe testing through the maximizer / whatif. The tests could go in before this PR and serve as proof of intent.

Also not keen on the use of Object. Shame Java doesn't have union types, because I think it's `Integer | String`? You could make something like an `Optional` with `isString` / `isInteger` checks (or `getType == Type.STRING` for proof of exhaustion).

src/net/sourceforge/kolmafia/session/GreyYouManager.java
phulin: Yep, this should be type NONE I think.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2651,23 +2705,23 @@ public void lookupFamiliarModifiers(
       if ((tuning = this.get(Modifiers.FAMILIAR_TUNING_MUSCLE)) > 0) {
         double mainstatFactor = tuning / 100;
         double offstatFactor = (1 - mainstatFactor) / 2;
-        this.add(Modifiers.MUS_EXPERIENCE, factor * mainstatFactor, "Tuned Volleyball:" + race);
```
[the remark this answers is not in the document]
phulin: Fixed.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1154,20 +1150,68 @@ public int[] predict() {
...
+  public static final Collection<Lookup> getAllModifiers() {
+    ArrayList<Lookup> result = new ArrayList<>();
+    for (var entry : modifierStringsByName.entrySet()) {
+      ModifierType type = entry.getKey();
+      for (Object key : entry.getValue().keySet()) {
+        if (key instanceof Integer intKey) {
+          result.add(new Lookup(type, intKey));
+        } else if (key instanceof String stringKey) {
+          result.add(new Lookup(type, stringKey));
+        }
```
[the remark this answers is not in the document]
phulin: Fixed.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -58,10 +51,13 @@
 import net.sourceforge.kolmafia.utilities.FileUtilities;
 import net.sourceforge.kolmafia.utilities.LogStream;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
+import net.sourceforge.kolmafia.utilities.TwoLevelEnumHashMap;
 
 public class Modifiers {
-  private static final Map<String, String> modifierStringsByName = new HashMap<>();
-  private static Map<String, Modifiers> modifiersByName = new HashMap<>();
+  private static final TwoLevelEnumHashMap<ModifierType, Object, String> modifierStringsByName =
```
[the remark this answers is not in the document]
phulin: Fixed.

src/net/sourceforge/kolmafia/utilities/StringUtilities.java
phulin: Fixed.

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
phulin: Fixed.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -5,15 +5,7 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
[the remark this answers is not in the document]
phulin: Fixed.

src/net/sourceforge/kolmafia/Modifiers.java
phulin: Fixed.

src/net/sourceforge/kolmafia/DebugModifiers.java
phulin: Fixed.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: Fixed.

phulin:
Alright, I've pushed fixes for most of the issues addressed. @heeheehee-kolmafia , I'll do some thinking about one-level vs. two-level. And @midgleyc I will add more tests, but I think the DebugModifiersTest here is a good start.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5104,19 +5122,21 @@ public static final Modifiers recalculateAdjustments(
     }
 
     if (HolidayDatabase.getRonaldPhase() == 5) {
-      newModifiers.add(Modifiers.RESTING_MP_PCT, 100, "Moons:Ronald full");
+      newModifiers.addDouble(
+          Modifiers.RESTING_MP_PCT, 100, ModifierType.EVENT, "Moons (Ronald " + "full)");
```
midgleyc, line 5126: pretty sure this string doesn't need splitting

src/net/sourceforge/kolmafia/ModifierType.java
```diff
@@ -0,0 +1,90 @@
...
+          .collect(
+              Collectors.toMap(
+                  type -> type.name().replace("_", "").toLowerCase(), Function.identity()));
+
+  public static ModifierType fromString(String typeString) {
+    return stringToType.get(typeString.replace("_", "").toLowerCase());
+  }
+
+  public String camelCaseName() {
+    return StringUtilities.upperSnakeToPascalCase(this.name());
```
midgleyc, line 84: camel case or pascal case?

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: fixed.

src/net/sourceforge/kolmafia/ModifierType.java
phulin: Fixed.

phulin:
> Can you explain the motivation behind making this a two-level lookup instead of using a map keyed by a composite type (such as the convenient Lookup class that you've added)?
> 
> (I don't have time right now to finish this review. I got through about half the files.)

Performance of the two-level map is significantly better, although I agree it's a little more complex. I think it's because you end up with fewer collisions when most of the reads/writes are in the three separate integer-keyed maps, vs collapsing the integer- and string-keyed maps into one.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2724,14 +2804,15 @@ public void lookupFamiliarModifiers(
...
+          this.addDouble(Modifiers.FAMILIAR_WEIGHT_CAP, 1, ModifierType.FAMILIAR, "naked hatrack");
         }
         break;
       case FamiliarPool.SCARECROW:
         if (famItem == EquipmentRequest.UNEQUIP) {
-          this.add(Modifiers.PANTSDROP, 50.0, "Familiar:naked scarecrow");
-          this.add(Modifiers.FAMILIAR_WEIGHT_CAP, 1, "Familiar:naked scarecrow");
+          this.addDouble(Modifiers.PANTSDROP, 50.0, ModifierType.FAMILIAR, "naked scarecrow");
+          this.addDouble(
+              Modifiers.FAMILIAR_WEIGHT_CAP, 1, ModifierType.FAMILIAR, "naked " + "scarecrow");
```
midgleyc, line 2815: another split string

src/net/sourceforge/kolmafia/utilities/WikiUtilities.java
```diff
@@ -30,10 +31,12 @@ public static String getWikiLocation(String name, int type, boolean dataPage) {
     boolean checkOtherTables = true;
 
     if (type != ANY_TYPE) {
-      String modType =
+      ModifierType modType =
```
midgleyc, line 34: I think this can be a switch to be slightly cleaner

test/net/sourceforge/kolmafia/DebugModifiersTest.java
```diff
@@ -0,0 +1,337 @@
...
+  @Test
+  void listsPath() {
+    try (var cleanups = withPath(AscensionPath.Path.YOU_ROBOT)) {
+      evaluateDebugModifiers(Modifiers.ENERGY);
+      assertThat(output(), containsDebugRow("Path", "You, Robot", 1.0, 1.0));
+    }
+  }
+
+  // TODO: Thrall, Autumnaton, Florist, Generated, Hobo Power, Smithsness, Slime Hates It, all
+  // specific path effects/path companions, VYKEA
```
midgleyc, line 336: definitely getting there on the test coverage!

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -4964,16 +4965,18 @@ public static final Modifiers recalculateAdjustments(
...
-    newModifiers.add(Modifiers.getModifiers("Loc", Modifiers.currentLocation));
-    newModifiers.add(Modifiers.getModifiers("Zone", Modifiers.currentZone));
+    newModifiers.add(Modifiers.getModifiers(ModifierType.LOC, Modifiers.currentLocation));
+    newModifiers.add(Modifiers.getModifiers(ModifierType.ZONE, Modifiers.currentZone));
 
     // Look at sign-specific adjustments
-    newModifiers.add(Modifiers.MONSTER_LEVEL, MCD, "MCD:MCD");
-    newModifiers.add(Modifiers.getModifiers("Sign", KoLCharacter.ascensionSign.getName()));
+    newModifiers.addDouble(
+        Modifiers.MONSTER_LEVEL, MCD, ModifierType.MCD, "Monster Control Device");
```
midgleyc, line 4973: just a general comment on differences: I note that most of these are the same, some of these are slightly different (e.g. one is capitalised, like Brimstone or Cloathing), some are more different but understandably (e.g. MCD, quartet).

Most are capitalised, but some are not (e.g. fake hands). Are there intended to be rules as to how these are formed / capitalised?

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -3628,80 +3733,82 @@ else if (itemId != ItemPool.TOILET_PAPER) {
...
+    Modifiers.writeModifierCategory(writer, potions, ModifierType.ITEM, "Everything Else");
+    Modifiers.writeModifierCategory(writer, freepulls, ModifierType.ITEM);
+    Modifiers.writeModifierCategory(writer, wikiname, ModifierType.ITEM);
 
     writer.close();
   }
 
   private static void writeModifierCategory(
-      final PrintStream writer, final Set<String> set, final String type, final String tag) {
+      final PrintStream writer, final Set<String> set, final ModifierType type, final String tag) {
```
midgleyc, line 3795: a note to me or you to check before merge: there's a function somewhere that writes out `modifiers.txt`, and it would be nice to confirm it still does what it used to do

src/net/sourceforge/kolmafia/Modifiers.java
phulin: thanks, this was IntelliJ being overenthusiastic.

src/net/sourceforge/kolmafia/KoLCharacter.java
phulin: My rule was generally to capitalize unless it's the name of something, in which case use that capitalization. `fake hand` is a name, so I left it here.

src/net/sourceforge/kolmafia/Modifiers.java
phulin: I started down the road of adding a test to ensure that `writeModifiers` is writing out a subset of the lines in `modifiers.txt`, but it turns out that there are a number of ways in which it's not arranged exactly the same way. For example, the Spelunky items are all in their own section and not in the Hats/Weapons/whatever sections. And there are quite a few slight sorting errors where things are misalphabetized. We could correct this but I don't think it's a priority. With my latest commit I've verified that `writeModifiers` does essentially the right thing.

phulin: Added a disabled test for this.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -3623,120 +3737,124 @@ else if (itemId != ItemPool.TOILET_PAPER) {
...
-    Modifiers.writeModifierCategory(writer, synergies, "Synergy", "Synergies");
+    Modifiers.writeModifierCategory(writer, synergies, ModifierType.SYNERGY, "Synergies");
     writer.println();
-    Modifiers.writeModifierCategory(writer, mutexes, "Mutex", "Mutual exclusions");
+    Modifiers.writeModifierCategory(
+        writer, mutexes, ModifierType.MUTEX_I, "Mutually exclusive items");
     writer.println();
-    Modifiers.writeModifierCategory(writer, maximization, "MaxCat", "Maximization categories");
+    Modifiers.writeModifierCategory(
+        writer, mutexes, ModifierType.MUTEX_I, "Mutually exclusive effects");
```
midgleyc, line 3792: MUTEX_E for this one, right?

src/net/sourceforge/kolmafia/Modifiers.java
midgleyc: ace!

src/net/sourceforge/kolmafia/Modifiers.java
phulin: Fixed.

phulin:
This is ready to go from my perspective - obviously happy to continue addressing any comments folks have.

midgleyc (approved):
Thanks!

phulin:
@heeheehee-kolmafia I think I've addressed all your comments, but let me know if not.

midgleyc (approved):

PR #1403, opened 2023-01-03
fix: modifiers NPE

midgleyc:
Came in on #1394. A modifier can end up with particular string entries (e.g. `Modifiers.EQUALIZE`) being null.

Not yet sure what's causing it, but the only way I can see that happening is if the Modifiers constructor is called, and not all the string fields are set (for some reason), Not sure how this could happen, can't reproduce from scratch. `getModifiersInternal` can return null under some circumstances, and in that case `set` doesn't populate anything, so /maybe/ that's happening?


phulin:
Here's the cause. Some places in the code cause calls to `new Modifiers(getItemModifiers(xxx))`. getItemModifiers can return null, which causes Modifiers::set to abort and not initialize any of the fields. In that case, we should initialize strings() to null but leave everything else. I.e. if set returns false, call the Arrays.fill. That should be enough.

phulin:
Oh wait, that's not quite right re: the fix. I think your fix is good.

midgleyc:
As seen in the test, this issue occurs when:
* you have the left-hand man or disembodied hand out
* you have at least one item in the relevant slot with no enchantments (not in modifiers.txt)
* you maximize anything

PR #1407, opened 2023-01-04
feat: more data for the new year

midgleyc:
Missing the wool item concoctions and some garden items. Will trickle in.


Veracity0 (approved):
If you have the grubby wool accessories, what does checkpowers say?

PR #1412, opened 2023-01-06
feat: two more standard items

midgleyc:
It's harder to find the standard stuff this year, as they're not in Valhalla. Could just leave this open until we get a decent number.


jaadams5 (approved):

Veracity0:
After Crimbo, I will start my yearly cycle of 12 Standard runs.

I will do 14 this year, because I want to collect the "rows" in the Armory and Leggery for trading pulverized Standard rewards for the previous year's rewards - which can be iterated back to 2015...

But, sheesh - they don't even show you items with descids in Valhalla? Why in the world not?
In what possible world is that a "user friendly" thing to do?

PR #1417, opened 2023-01-08
Add set_ccs ASH function.

phulin:
I got annoyed with the standard scripting path of calling `cliExecute("ccs xxx")` that prints an extra line of "CCS set to xxx" every adventure, so this function doesn't print anything and just returns true on success. I'll note in the docs on the wiki that script authors have to know that `battleAction` needs to be set as well or it won't work. This goes along with `read_ccs` and `write_ccs` which were recently added.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.regex.Matcher;
@@ -2165,6 +2166,9 @@ public abstract class RuntimeLibrary {
 
     params = new Type[] {DataTypes.BUFFER_TYPE, DataTypes.STRING_TYPE};
     functions.add(new LibraryFunction("buffer_to_file", DataTypes.BOOLEAN_TYPE, params));
+
+    params = new Type[] {DataTypes.STRING_TYPE};
+    functions.add(new LibraryFunction("set_ccs", DataTypes.BOOLEAN_TYPE, params));
 
     params = new Type[] {DataTypes.STRING_TYPE};
     functions.add(new LibraryFunction("read_ccs", DataTypes.BUFFER_TYPE, params));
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks!

Veracity0:
I'm away from working on this project for 3 hours and it is submitted and approved before I even have a chance to look at it.

Is there something we could (or should) have done about setting battleAction so that if you intentionally change your CCS, we set things up so it is used, rather than requiring another function call?

I'm not happy that this was not called out/commented upon/addressed.

Not that I have THE answer, but no discussion?

Veracity0:
Well, shit.

The "standard scripting path" is such-and-such.

Literally none of my scripts do that. None. Not one.

What do your scripts do? Call that and then do adv1 and then do it again - even though nothing has executed between setting the CCS the first time and the second time?

Fix your fucking buggy script.

Sheesh.

Veracity0:
Or you know - if you do set_ccs to the SAME script, ignore the script error and don't print (or do) anything for that specific case.

Why no PR comments allowed?

phulin:
Hey Veracity -

Sorry, didn't mean to imply anything by calling it standard - it's just what most scripts I work on do. My description up top was a quick description of a pretty small change that I think is positive on the whole. If you would like to see further changes to this PR I'd be happy to make them.

I wrote it because I don't know of a way to check if the CCS is set correctly, so all scripts I write reset the CCS every adventure. The control flow in some scripts I've worked on has been so complicated that this pattern feels like the easiest way to ensure the CCS is always set correctly. Maybe there is some smarter architecture, but I don't know what it is.


> Or you know - if you do set_ccs to the SAME script, ignore the script error and don't print (or do) anything for that specific case.

The existing routes - the CLI command or setting the customCombatScript faux property - both use the same code path, which I think we'd want to have print no matter what rather than silently do nothing when a user runs it from the CLI. So I added a new path for script authors.

midgleyc:
Sorry @Veracity0, when reviewing I just checked `ChangeCombatScriptCommand.java`, confirmed that the proposed ASH function did something similar (the command checks exact matches, then partial matches, both case-insensitively; the ASH function just checks exact matches: seems reasonable), confirmed there was a test and approved it.

 Patrick also added notes on the wiki (https://wiki.kolmafia.us/index.php/Set_ccs) which I thought was good.

Veracity0:
I was irritated. I felt left out of the loop. My language was excessively harsh and inappropriate. I apologize for that.

I trust you to do the right thing.

I was concerned because ccs is one thing that not only does file system operations but affect play in the GUI and Relay Browser, not just the script.

Changing it seemed like something the user would want to be aware of, rather than being surprised about later.

That is perhaps a philosophical difference, but I'm not going to worry about it.

Thanks for all you (both) do.

Peace.

PR #1418, opened 2023-01-09
fix: cyrpt evilness boss trigger changed to 13

midgleyc:
Think I got everywhere.


gausie (commented):
Pretty easy to add a test for AreaCombatData if you like

gausie (approved):
Super awesome

PR #1420, opened 2023-01-09
Support booze and food fairies

gausie:
- Track Vintner and Cookbookbat as food and booze fairies
- Add Superconductor's CPU as Single Equip while we're here

Note: this adds new familiar types: item1 and item2


midgleyc (changes requested):
It might be nice to handle Peppermint Rhino while you're adding fairy types: the Rhino is a 1x fairy and a 1x candy fairy (for a total of 2x candy drops).

I note this is a case currently not handled by the code -- a familiar is either a food or booze or item fairy, not a combination. It would be nice to sort out this part of the design -- for the difficulty in adding it after the fact, see how other parts of the codebase handle drunki-bears ;)

gausie:
Hmm ok. This would mean I need to add `Booze Fairy`, `Food Fairy` and `Candy Fairy` modifiers. Does that sound like a good idea?

midgleyc:
Yeah. I can't really see another way of doing it, and I think it's better than having a Fairy type that switches based on familiar tags.

gausie:
That should be it! Thanks for the feedback, this is much better. It would be great to turn Modifiers into an enum, having this numerical key is very annoying.

src/data/familiars.txt
```diff
@@ -6,6 +6,8 @@
 # stat0 - vollyball-like
 # stat1 - sombrero-like
 # item0 - Item Drop
+# item1 - Booze Drop
+# item2 - Food Drop
```
midgleyc, line 10: need to correct these post-candy

also rather than (booze, candy, food), (food, booze, candy) might be a more intuitive order?

test/net/sourceforge/kolmafia/ModifiersTest.java
```diff
@@ -998,4 +999,26 @@ public void correctlyAppliesAmphibianSympathyToDodecapede(
   public void identifiesVariableModifiers(String skillName, boolean variable) {
     assertThat(Modifiers.getModifiers("Skill", skillName).variable, equalTo(variable));
   }
+
+  @ParameterizedTest
+  @CsvSource({
+    FamiliarPool.VAMPIRE_VINTNER + ", " + Modifiers.BOOZEDROP + ", 1.5",
+    FamiliarPool.PEPPERMINT_RHINO + ", " + Modifiers.CANDYDROP + ", 1.0",
```
midgleyc, line 1006: can we have another test confirming the rhino is a 1x fairy (for straight itemdrop), please?

midgleyc:
the enum type is in https://github.com/kolmafia/kolmafia/pull/1401, just waiting for final comments ;)

src/data/familiars.txt
gausie: Booze Candy Food is alphabetical ;)

midgleyc: yeah but if something new gets added between candy and food you'd have to renumber. Alphabetical isn't so good on lists that could change.

gausie: Legit. Ok Food, Booze, Candy it is

test/net/sourceforge/kolmafia/ModifiersTest.java
gausie: done and the test failed! good spot

midgleyc (approved):
Thanks!

test/net/sourceforge/kolmafia/ModifiersTest.java
gausie: Well, the test didn't fail, the test failed for the vintner and cbb

PR #1421, opened 2023-01-09
feat: full ceramic set

midgleyc:
Full ceramic set + trainbot halloween reward


Veracity0 (approved):
Finally!

midgleyc:
well, it's half of it ;)

PR #1422, opened 2023-01-10
feat: moleman

midgleyc:


gausie (approved):
Weird that he is Title case, but confirmed and approved

PR #1424, opened 2023-01-11
refactor: move EquipmentRequestType to enum

midgleyc:
Accept various refactorings while we're in the file.

There are several spaces in the codebase where we use `public static final int`s where enums would be better (and prevent passing ints outside the range, or from the wrong source, to functions). This is one of them (and likely the simplest).


Veracity0 (approved):
Nice!

PR #1425, opened 2023-01-11
refactor: move VykeaCompanionType to enum

midgleyc:

This one uses the ordinal, both in the exposed ASH type and the image. Perhaps it should be explicitly in the enum, but equally I can't see it changing. I added a test for the image, so we'll know by failing test if the ordinal of the last entry changes.

### DIFF
--- a/src/net/sourceforge/kolmafia/VYKEACompanionData.java
+++ b/src/net/sourceforge/kolmafia/VYKEACompanionData.java
@@ -422,11 +422,11 @@ public class VYKEACompanionData implements Comparable<VYKEACompanionData> {
     if (choice == 1120) {
       switch (decision) {
         case 1 ->
-          // Start with 5 planks -> bookshelf, ceiling fan, dresser
-          ResultProcessor.processItem(ItemPool.VYKEA_PLANK, -5);
+        // Start with 5 planks -> bookshelf, ceiling fan, dresser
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/VYKEACompanionData.java
```diff
@@ -132,22 +132,24 @@ public class VYKEACompanionData implements Comparable<VYKEACompanionData> {
...
-
-  public static final int BOOKSHELF = 1;
-  public static final int DRESSER = 2;
-  public static final int CEILING_FAN = 3;
-  public static final int COUCH = 4;
-  public static final int LAMP = 5;
-  public static final int DISHRACK = 6;
+  public enum VYKEACompanionType {
+    NONE,
+
```
jaadams5, line 137: Is there a reason for a blank line here?

jaadams5 (approved):
If we can turn off the feature - unchanged files with check annotations - that would IMO be useful.  The feature seems useful.  Appending it to a review does not.

src/net/sourceforge/kolmafia/VYKEACompanionData.java
midgleyc: there was a blank line in the original file between NONE and the rest, so I kept it.

PR #1426, opened 2023-01-11
refactor: convert priority into enum

midgleyc:

This one also used the ordinal in `compareTo`. The specific numbers have been changed (most are exactly one out, except `PSEUDO_ITEM`), and don't seem to matter.

The order has been kept the same except in the case of `EFFECT` and `BOUNTY_ITEM` which had the same priority. This may have been a bug, as it seemed unintended in most cases. Sorting a list containing both effect and bounty items should now put all the effects after all the bounty items, whereas previously it would have jumbled them together. I don't think that's behaviour we want to keep.


jaadams5 (approved):
I'd consider renaming the class Priority to SortPriority (because AdventureResultSortPriority is overkill), but that's not a blocker.

PR #1428, opened 2023-01-12
refactor: move simple request ints to enum

midgleyc:

Convert the simple requests (closet, storage, stash) at once. They're all very similar.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ClosetRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ClosetRequest.java
@@ -1,5 +1,9 @@
 package net.sourceforge.kolmafia.request;
 
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.CLOSET_TO_INVENTORY;
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.INVENTORY_TO_CLOSET;
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.REFRESH;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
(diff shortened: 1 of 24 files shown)


src/net/sourceforge/kolmafia/request/ClosetRequest.java
```diff
@@ -1,5 +1,9 @@
 package net.sourceforge.kolmafia.request;
 
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.CLOSET_TO_INVENTORY;
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.INVENTORY_TO_CLOSET;
+import static net.sourceforge.kolmafia.request.ClosetRequest.ClosetRequestType.REFRESH;
+
```
Veracity0, line 6: Are these because the previous code did not qualify the static class variable with ClosetRequest, so you couldn't qualify the enum constant with ClosetRequestType? I mean you could - and I think I'd prefer that to importing some (but not all) of tge enum constants.

src/net/sourceforge/kolmafia/request/StorageRequest.java
```diff
@@ -1,5 +1,9 @@
 package net.sourceforge.kolmafia.request;
 
+import static net.sourceforge.kolmafia.request.StorageRequest.StorageRequestType.EMPTY_STORAGE;
+import static net.sourceforge.kolmafia.request.StorageRequest.StorageRequestType.REFRESH;
+import static net.sourceforge.kolmafia.request.StorageRequest.StorageRequestType.STORAGE_TO_INVENTORY;
+
```
Veracity0, line 6: Same comment as with ClosetRequest.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/ClosetRequest.java
midgleyc: yes, they were simply to minimise changes in the refactor. I could qualify instead.

Veracity0 (approved):

PR #1431, opened 2023-01-12
Fix adding speakeasy drinks using existing AdventureResults

Veracity0:
As reported by ... somebody.

When switching clans, speakeasy/hotdogs/floundry are rebuilt from saved clan lounge item lists.
Don't need to - can't - re-add the item to the item list.


midgleyc (approved):
Thanks!

PR #1437, opened 2023-01-15
refactor: make double modifiers enum

midgleyc:
Make `Modifiers`'s DOUBLE_MODIFIERS an enum.

Initially I planned to do it all, but even this was a huge undertaking. Plus there's a lot I don't understand about the class.

Fortunately our coverage is pretty good.

Only uncovered one error: `UseSkillRequest` was indexing into boolean modifiers using a double modifier, which silently did nothing.

Fixed one likely other issue where an enum value was compared against a String.

Perf: a bit slower. ~13k with compared to ~10k without on https://github.com/kolmafia/kolmafia/pull/1389. Did remove the sparse / dense thing due to array usage being replaced with `EnumMap`. Adding that back in made things worse: 14.5k.

Other comments from when I was running around:
* we have a caseless lookup in addition to `modifierTypesByName` that looks up both name and tags. I think the caseless lookup is only used by maximizer
* in some places we start with a DoubleModifier, then convert it to a String (by taking its name), pass the String to `currentNumericModifiers` where it gets passed to `get` which converts it back into a DoubleModifier. This is (should be?) unnecessary, except that the String lookup also considers "derived" modifiers and bitmap modifiers
* the code assumes names and tags are globally unique
* raveosity is a bitmap while surgeonosity and clownosity are doubles. Not sure why.


midgleyc:
Open for comments.

Overall: slower, but more correct. I know we've mentioned in the past that we wanted to move away from array / index into array for the modifiers. Did find one bug. Makes certain sections of the code easier to understand (e.g. the ones that did arithmetic on the indices).

@phulin: any obvious perf improvements I can do with the map? I tried adding back your densify thing, but it got even slower.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -98,689 +99,15 @@ public class Modifiers {
...
-
-  public static final int DOUBLE_MODIFIERS = Modifiers.doubleModifiers.length;
-
   private static final HashSet<String> numericModifiers = new HashSet<>();
 
   static {
-    for (int i = 0; i < DOUBLE_MODIFIERS; ++i) {
-      DoubleModifier modifier = Modifiers.doubleModifiers[i];
-      modifierIndicesByName.put(modifier.getName(), i);
+    for (var modifier : DoubleModifier.values()) {
```
Veracity0, line 107: I will mention that the values() method returns an array of enums.
Java does not have immutable arrays, so that is allocated and filled every time you call that static method.
You might consider making an EnumSet.allOf(DoubleModifer.class) somewhere. Perhaps in Modifiers.java?
Although Java does have immutable Sets, I don't think so for EnumSets, although you could agree not to modify it.
That is likely to be similarly fast to iterate over as an array and should not require memory allocation.

I think you have other iterations of DoubleModifer.values(), so this same comment applies, although I will not repeat it; you can decide what, if anything, you want to do globally.

phulin:
> @phulin: any obvious perf improvements I can do with the map? I tried adding back your densify thing, but it got even slower.

I think you don't want EnumMap here. That's just an array per the [implementation](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/EnumMap.java), so we'd just be reproducing the dense representation. We should use a TreeMap below the threshold and then just set `this.doubles = new EnumMap<>(this.doubles)` when densifying. That should work.

Veracity0 (approved):
This fixes what I noticed. Can't say it addresses all the "slower" observation you made, but it is more efficient in both time and space, so it can't hurt.

I don't know if you are doing anything with phulan's comment, but this is good enough for me.

midgleyc:
I was going to try again after work to see if I can improve the benchmark using the `densify` approach. If not I'll leave it.

midgleyc:
I checked the benchmark with the densify `DoubleModifierCollection`. It was close, but overall the non-densify version was faster.

I used the following code:
```java
public class DoubleModifierCollection {
  public static final int SPARSE_DOUBLES_MAX_SIZE = 32;

  // If only a few values are set in doubles, we instead store all modifiers in a sparse TreeMap.
  // When that map gets bigger than SPARSE_DOUBLES_MAX_SIZE, we copy it over to the dense array.
  // We track whether this is dense or not by whether sparseDoubles is null.
  private final Map<DoubleModifier, Double> doubles = new EnumMap<>(DoubleModifier.class);
  private TreeMap<DoubleModifier, Double> sparseDoubles = new TreeMap<>();

  public void reset() {
    this.sparseDoubles = new TreeMap<>();
    this.doubles.clear();
  }

  public void densify() {
    // When reset, this.doubles is set to all 0s, so all we have to do is copy all the sparse
    // values over.
    var sparseDoubles = this.sparseDoubles;
    if (sparseDoubles == null) return;

    this.doubles.putAll(sparseDoubles);
    this.sparseDoubles = null;
  }

  public double get(final DoubleModifier mod) {
    var sparseDoubles = this.sparseDoubles;
    return sparseDoubles != null
        ? sparseDoubles.getOrDefault(mod, 0.0)
        : this.doubles.getOrDefault(mod, 0.0);
  }

  public boolean set(final DoubleModifier mod, final double value) {
    var sparseDoubles = this.sparseDoubles;
    Double oldValue;
    if (sparseDoubles != null) {
      oldValue = value == 0.0 ? sparseDoubles.remove(mod) : sparseDoubles.put(mod, value);

      if (sparseDoubles.size() >= DoubleModifierCollection.SPARSE_DOUBLES_MAX_SIZE) {
        this.densify();
      }
    } else {
      oldValue = value == 0.0 ? this.doubles.remove(mod) : this.doubles.put(mod, value);
    }
    // TODO: does anything use this return value, or can we save ourselves a check?
    return oldValue == null || oldValue != value;
  }

  public double add(final DoubleModifier mod, final double value) {
    // Anything being accumulated onto should be dense.
    this.densify();
    return this.doubles.merge(mod, value, Double::sum);
  }

  public void forEach(BiConsumer<? super DoubleModifier, ? super Double> action) {
    var sparseDoubles = this.sparseDoubles;
    if (sparseDoubles != null) {
      sparseDoubles.forEach(action);
    } else {
      this.doubles.forEach(action);
    }
  }
}
```

PR #1438, opened 2023-01-15
feat: support rock garden in garden command

midgleyc:
Support harvesting rock garden. The normal `action=garden` destroys your garden but doesn't give you anything.

Tested with an actual rock garden.

Don't support harvesting individual slots, as that would be more complex and I think most of the time you want to harvest all the slots anyway. A user can easily do this by just calling the URL.

Fix a likely bug with harvesting grass / non-grass: the comment says that anything none-grass should have 1 request, but then sets grass (or anything with count 8) to count 1.

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/command/GardenCommandTest.java
+++ b/test/net/sourceforge/kolmafia/textui/command/GardenCommandTest.java
@@ -1,0 +1,124 @@
+package net.sourceforge.kolmafia.textui.command;
+
+import static internal.helpers.HttpClientWrapper.getRequests;
+import static internal.helpers.Networking.assertPostRequest;
+import static internal.helpers.Player.withCampgroundItem;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasSize;
+
+import internal.helpers.Cleanups;
+import internal.helpers.HttpClientWrapper;
+import net.sourceforge.kolmafia.KoLConstants.MafiaState;
+import net.sourceforge.kolmafia.StaticEntity;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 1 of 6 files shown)


Veracity0:
Heh. When I looked at this an hour ago, it said "not tested with an actual Rock Garden; I hope it doesn't require a GET".

"a" tags submit a GET, and that's what's in campground.php, as I am sure you know.

I have a 3 day rock garden and, having freed the king, I will ascend tomorrow, so I was debating whether to harvest today or tomorrow. I was willing to harvest today and test for you, but now I will do it tomorrow. :)

I'll review this, by and by. After lunch.

midgleyc:
I realised I could use a fertilizer to get a day 1 rock garden to test with, so I did that. It harvested fine.

test/net/sourceforge/kolmafia/textui/command/GardenCommandTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+  @BeforeEach
+  public void initializeState() {
+    HttpClientWrapper.setupFakeClient();
+    StaticEntity.setContinuationState(MafiaState.CONTINUE);
+  }
+
+  @BeforeAll
+  public static void beforeAll() {
+    // there is a leak somewhere that makes every test return "you have 0 peppermint sprouts"
```
Veracity0, line 34: test_campground_fuel_asdon.html contains "otherimages/campground/pepperpatch_0.gif"
That is in AsdonMartinCommandTest.fuelValidSendsRequest()

test_visit_el_vibrato_campground.html contains "otherimages/campground/pepperpatch_1.gif"
That is in ElVibratoManagerTest.canRecogniveActivePortalAndEnergy

The latter uses withEmptyCampground.
The former does not.

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/textui/command/GardenCommandTest.java
midgleyc: thanks

Veracity0 (approved):

PR #1440, opened 2023-01-15
chiffon chaps

Veracity0:


midgleyc:
Chapeau is descid 900963107, while you're there.

Veracity0:
> Also if you have one, and I assume you do because of the power, you could check the skill it offers?
What skill? Wearing the chiffon item does damage at the start of each combat round after the initial one without needing a skill.

With the chaps:
```
A chill wind blows your chaps aside, revealing a spooky tomb where the sun doesn't shine. Your foe is terrified, and takes 11 damage.
```

midgleyc (approved):
Thanks!

I assumed it gave a skill because the last set did, but if it doesn't that explains why nobody's mentioned it yet.

PR #1441, opened 2023-01-16
CalendarFrame Test

jaadams5:
I'm interested in a minor cosmetic change concerning time.  I realized it would have the greatest impact on CalendarFrame and there were not tests for that.  So this is a test for CalendarFrame.  I'm not really interested in the GUI actions so the interesting part of the test is getting the data from the tabs.  That data is derived from "today" so I made a way to pass the date to CalendarFrame so the test was not time dependent.  I did something similar to avoid having the test depend upon a random number.  On the test itself there are some assertions that are probably not useful.  I'm not convinced making a parametrized test of tabs and data but I could cave if someone else feels strongly.


jaadams5:
So my Swing test needs to not throw java.awt.HeadlessException.  Will investigate.

jaadams5:
#1454 seems to fix the test failure without refactoring the code.

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
```diff
@@ -0,0 +1,102 @@
...
+      w.write();
+      return sw.toString();
+    } catch (IOException | BadLocationException e) {
+      e.printStackTrace();
+      return "";
+    }
+  }
+
+  public String getDataFromTab(Component aTab) {
+    assertTrue(aTab instanceof RequestPane);
```
midgleyc, line 41: You can get better failure messages using `assertThat`, e.g.

```
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.instanceOf;

assertThat(aTab, instanceOf(RequestPane.class));
```

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
```diff
@@ -0,0 +1,102 @@
...
+    assertEquals("Farmer's Almanac (Not Logged In)", testFrame.getTitle());
+    Component[] components = testFrame.getComponents();
+    assertEquals(1, components.length);
+    assertTrue(components[0] instanceof JRootPane);
+  }
+
+  @Test
+  public void itShouldCalculateExpectedDataForKnownTime() {
+    String expectedKOLDay =
+        "drawn by SpaceMonkeySeptember 1, 2010Boozember 2Ronald: waxing crescentGrimace: new moonStats: 3 days until Mysticism.Grue: bloodlustyBlood: +38%Baio: +20%Jekyllin: +7 stats, 25% items";
```
midgleyc, line 66: this is a bit run-on. Can you not strip the HTML, and test directly against it (including the tags)?

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
```diff
@@ -0,0 +1,102 @@
...
+          {
+            setLineLength(999_999);
+          }
+        };
+    try {
+      w.write();
+      return sw.toString();
+    } catch (IOException | BadLocationException e) {
+      e.printStackTrace();
+      return "";
```
midgleyc, line 36: may as well `fail()` here, or even just let the error propagate to fail the test.

.github/workflows/check.yml
```diff
@@ -25,6 +25,13 @@ jobs:
         with:
```
Veracity0, line 25: So this all from Rinn's draft PR, right?

How about if we get that merged first, on its own, rather than pulling it into this PR?

Veracity0 (dismissed):

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
jaadams5: OK

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
jaadams5: OK

.github/workflows/check.yml
jaadams5: I merged here to establish that it solved the problem.  I will not push this until Rinn's has been approved and main merged back here.

jaadams5:
Got concurrent mod errors from GearChangePanelTest.  Hypothesis is that the Script menu was getting initialized more than once.  Simulated login and broke up tests.  Need to revise the expected data, possibly make a parameterized test and consider using Cleanups instead of beforeAll/afterAll.

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
jaadams5: OK.

jaadams5:
Draft until I figure the errors out.  I think there is an EOL issue on Ubuntu and the concurrent mod errors are because the failed tests aren't cleaning up.

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
```diff
@@ -76,139 +76,269 @@ public String getDataFromTab(Component aTab) {
...
-          + "                  </td>\n"
-          + "                </tr>\n"
-          + "              </table>\n"
-          + "            </center>\n"
-          + "          </td>\n"
-          + "        </tr>\n"
-          + "      </table>\n"
-          + "    </center>\n"
-          + "  </body>\n"
+      "<html>"
```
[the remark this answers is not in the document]
jaadams5: Do you know an IntelliJ command that would do that?  spotless didn't.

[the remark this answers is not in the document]
jaadams5: nevermind.  figured it out.

.github/workflows/check.yml
jaadams5: And Rinn's was pushed.  Dismissing review since can't figure out how to resolve.

midgleyc (approved):
Thanks!

Are those "\s"s both real and intended, though?

jaadams5:
> Thanks!
> 
> Are those "\s"s both real and intended, though?

I copied the string as displayed in the debugger and pasted it into the code.  The calendar does display differently based upon the moon phases so maybe they would make more sense when the moon data was different?

midgleyc:
This change seems to have broken the Jenkins build. I suppose a similar change needs making to the `jenkinsfile` after the server is updated to install `xvfb` as was installed here.

PR #1443, opened 2023-01-18
warn about overdrinking stillsuit distillate

horrible-little-slime:
I'm not entirely sure how to set up tests for this, given that the functions I want to test involve an input field

### DIFF
--- a/src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
+++ b/src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
@@ -161,7 +161,8 @@ public class DrinkItemRequest extends UseItemRequest {
     if (inebriety == 1
         && !ConcoctionDatabase.queuedMimeShotglass
         && InventoryManager.getCount(ItemPool.MIME_SHOTGLASS) > 0
-        && !Preferences.getBoolean("_mimeArmyShotglassUsed")) {
+        && !Preferences.getBoolean("_mimeArmyShotglassUsed")
+        && itemId > 0) {
       shotglass = 1;
     }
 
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
```diff
@@ -396,12 +393,18 @@ private boolean allowBoozeConsumption() {
     return DrinkItemRequest.allowBoozeConsumption(itemName, count);
   }
 
+  private static final boolean shotglassApplies(String itemName) {
+    int inebriety = ConsumablesDatabase.getInebriety(itemName);
+    return inebriety == 1
+        && InventoryManager.getCount(ItemPool.MIME_SHOTGLASS) > 0
+        && !Preferences.getBoolean("_mimeArmyShotglassUsed")
+        && ItemDatabase.getItemId(itemName) > 0;
```
gausie, line 401: Are you sure this is an accurate limitation? Are there drinkables other than the distillate which are 1-size and are fake items?

src/net/sourceforge/kolmafia/request/StillSuitRequest.java
```diff
@@ -52,7 +52,7 @@ public static boolean canMake() {
       return false;
     }
 
-    return true;
+    return DrinkItemRequest.allowBoozeConsumption("stillsuit distillate", 1);
```
gausie, line 55: I don't know if it's right to have a user prompt here since canMake is used in Concoction.canMake

horrible-little-slime: Should that move to `run()`, then?

src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
horrible-little-slime: That's a good question. It feels hacky to explicitly exclude the stillsuit instead of using a more general rule, but I'd be willing to bet there are some fake items that use the regular drink code for kol. I don't appear to be able to use Item.all() to test this through mafia; do you know an efficient way to check for other fake items?

src/net/sourceforge/kolmafia/request/StillSuitRequest.java
gausie: I knew at the time but have forgotten now. Maybe!

PR #1445, opened 2023-01-19
refactor: convert string modifiers to enum

midgleyc:
Much easier than doubles. Might try some perf improvements, now that this is detracting:
* I think we can inline every function call that only calls a different function
* We can probably avoid doing the null checks most of the time, especially for constants
* stat-modifying intrinsic effects on equipment were removed in 2012, so code dealing with that (e.g. in the maximizer) can also be removed

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -35,7 +35,6 @@ import net.sourceforge.kolmafia.modifiers.BooleanModifier;
 import net.sourceforge.kolmafia.modifiers.DerivedModifier;
 import net.sourceforge.kolmafia.modifiers.DoubleModifier;
 import net.sourceforge.kolmafia.modifiers.DoubleModifierCollection;
-import net.sourceforge.kolmafia.modifiers.Modifier;
 import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.modifiers.StringModifierCollection;
 import net.sourceforge.kolmafia.objectpool.EffectPool;
@@ -148,9 +147,8 @@ public class Modifiers {
 
     for (int i = 0; i < BITMAP_MODIFIERS; ++i) {
       BitmapModifier modifier = Modifiers.bitmapModifiers[i];
-      int index = i;
-      modifierIndicesByName.put(modifier.getName(), index);
-      modifierIndicesByName.put(modifier.getTag(), index);
+      modifierIndicesByName.put(modifier.getName(), i);
+      modifierIndicesByName.put(modifier.getTag(), i);
     }
   }
 
@@ -493,20 +491,12 @@ public class Modifiers {
     Modifiers.modifiersByName.remove(lookup.type, lookup.getKey());
   }
 
-  public static final String getModifierName(final DoubleModifier modifier) {
-    return modifier.getName();
-  }
-
   public static final String getBitmapModifierName(final int index) {
     return Modifiers.bitmapModifiers[index].getName();
   }
 
   public static final String getBooleanModifierName(final int index) {
     return Modifiers.booleanModifiers[index].getName();
-  }
-
-  public static final String getStringModifierName(final StringModifier mod) {
-    return mod.getName();
   }
 
   public static final String getDerivedModifierName(final int index) {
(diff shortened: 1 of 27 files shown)


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -103,8 +106,8 @@ public class Modifiers {
 
   private static final HashSet<String> numericModifiers = new HashSet<>();
 
-  public static final EnumSet<DoubleModifier> DOUBLE_MODIFIERS =
-      EnumSet.allOf(DoubleModifier.class);
+  public static final Set<DoubleModifier> DOUBLE_MODIFIERS =
+      Collections.unmodifiableSet(EnumSet.allOf(DoubleModifier.class));
```
Veracity0, line 110: Not as efficient in execution or memory, but fair, given that you can't get an unmodifiable EnumSet. Yet.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -861,12 +792,7 @@ public double getDoublerAccumulator(final DoubleModifier modifier) {
   public double getDoublerAccumulator(final String name) {
     // doublerAccumulators uses the same indexes as doubles, so the same lookup will work
```
Veracity0, line 793: It uses the same "indexes", eh? Old comment...

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -1138,12 +1060,7 @@ public boolean setModifier(final Modifier mod) {
     }
 
     index -= BITMAP_MODIFIERS;
-    if (index < BOOLEAN_MODIFIERS) {
-      return this.setBoolean(index, mod.getValue().equals("true"));
-    }
-
-    index -= BOOLEAN_MODIFIERS;
-    return this.setString(index, mod.getValue());
+    return this.setBoolean(index, mod.getValue().equals("true"));
```
Veracity0, line 1063: I foresee a forthcoming PR addressing BooleanModifiers...

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -2629,10 +2509,6 @@ private static <T extends net.sourceforge.kolmafia.modifiers.Modifier> String pa
 
         String value = matcher.group(1);
 
-        if (tag.equals("Class")) {
```
Veracity0, line 2632: This is no longer needed because there is no longer a "table" of StringModifiers any more, right? Just Booleans? So this particular method no longer applies to StringsModifiers.

Veracity0 (dismissed):
I pointed out an old comment that is no longer applicable, but other than that, this looks good.

src/net/sourceforge/kolmafia/Modifiers.java
midgleyc: yes: Class is a string modifier, so there's no point doing the check when you know you've a modifier of different type

src/net/sourceforge/kolmafia/Modifiers.java
midgleyc: going to do derived shortly, then I'm not sure. The complexity is that at the moment the implementation is very memory-efficient, in that it shoves all the booleans in the first bitmap modifier. The downside of doing that is that I don't really understand the code around it.

src/net/sourceforge/kolmafia/Modifiers.java
midgleyc: `unmodifiableSet` just creates a proxy that overrides methods that change the set to throw. In theory, I think the overhead should be minimal.

src/net/sourceforge/kolmafia/Modifiers.java
Veracity0: bitmap modifiers are restricted to bits in a single int, aren't they?
Looks like we have 31 booleans.

Making code easier to understand & maintain is a plus.

You made a comment in another PR wondering why raveosity was a bitmap, but clownosity (built on clowniness) and surgeonosity are numeric. Considering that clowninessa can be 25, 50, or 75 - corresponding to Clownosity +1, +2, or +3 - and raveosity can be +1 or +2, it seems that bitmaps are obscure.

Perhaps it has something to do with Clowniness and Surgeonosity are the only two "Unique" modifiers - no more than one worn item that supplies them contributes to the score.

Investigating all that would be its own project,, I think.

PR #1446, opened 2023-01-19
Chiffon outfit

Veracity0:
From the Wiki.

I gave it Power 50 because all Standard Normal equipment has that power.
Similarly, all Standard Hardcore equipment has power 120.


midgleyc (approved):
Grand. Who will be the first to smash one?

Veracity0:
> Grand. Who will be the first to smash one?

Might be me. I'm intending to do 7 each Standard Normal/Hardcore runs before I start whatever the spring challenge path is, specifically to see what chiffon/ceramic pulverize into - allowing me to figure out the ROWS in the Armory and Leggery, all the way back to Bakelite and Polyester.

PR #1450, opened 2023-01-21
Add sparse/dense optimization to DoubleModifierCollection.

phulin:
This gets back some of the performance gains from sparse modifier collections, but not all for some reason. Might just be the increased indirection from using an enum rather than an int.


midgleyc (approved):
Thanks!

I'm guessing the extra slowness comes from:
* more indirection (methods calling other methods)
* maps are generally slower than array access

I would have thought this would be offset by needing to do fewer lookups, but it wasn't.

PR #1451, opened 2023-01-21
Initial support for The 8-Bit Realm

Veracity0:
1) The 8-bit Realm is a zone
2) The Fungus Plains, Hero's Field, Vanya's Castle, and Megalo-City are adventures
3) Three new monsters: Blader, Met, Tackle Fire
4) digital key (and key lime and key lime pie) are no longer concoctions
5) pixel bread and pixel whiskey are new pixel items which give new status effects
6) New properties to track your "score"
8BitScore - your "score" for use in the treasure house
8BitColor - the color of the score in the charpane, which indicates which adventure area has double score and bonus

I considered making that the adventure name, but you still have to know which adventure benefits from which modifier.
color -> adventure -> modifier
vs.
color -> adventure
color -> modifier

Whatever.

7) When you automate with the continuum transfunction equipped, get charpane.php, not api.php, so we can pick up the score.

Other things we could (maybe) do:

1) Update score from fight respone
2) Decorate the 8-bit map, somehow, to show needed color/modifier for double points
3) Something for the Treasure House? Log visit?


src/data/combats.txt
```diff
@@ -307,6 +308,7 @@ The Edge of the Swamp	-1	huge mosquito	mega frog	swamp entity
 The Electric Lemonade Acid Parade	100	Essence of Interspecies Respect	Essence of Soy	Essence of Tofu	Neptune, the Dog that Is a Respected Equal and Not a Pet	Neptune, the Dog that Is a Respected Equal and Not a Pet cast member	Susie Soyburger	Susie Soyburger cast member	Timmy Tofurkey	Timmy Tofurkey cast member
 The Enormous Greater-Than Sign	80	swarm of lowercase As	lowercase B	lowercase H	lowercase K	uppercase Q
 The Evil Cathedral	100	plywood cultists
+The Fungus Plains	100	Blooper	Bullet Bill	Buzzy Beetle	Koopa Troopa
```
midgleyc, line 311:
```suggestion
The Fungus Plains	100	Blooper	Bullet Bill	Buzzy Beetle: 1e	Goomba: 1o	Koopa Troopa
```

midgleyc: ah, is this one you're picking from the other PR? Sorry, I'll wait.

midgleyc:
One other thing to remove is that `UseLinkDecorator.java` no longer needs to track the RGB pixels to make white pixels (keeping the white pixels in Exploathing is probably helpful)

Veracity0:
Yeah. You can still make white pixels, but not worth reminding about that. We don't remind you for other pixel concoctions?

src/data/combats.txt
Veracity0: > Suggested change
> The Fungus Plains	100	Blooper	Bullet Bill	Buzzy Beetle	Koopa Troopa
> The Fungus Plains	100	Blooper	Bullet Bill	Buzzy Beetle: 1e	Goomba: 1o	Koopa Troopa

 I got a Goomba today, so, yeah - that is a correct fix.

Veracity0:
> One other thing to remove is that `UseLinkDecorator.java` no longer needs to track the RGB pixels to make white pixels (keeping the white pixels in Exploathing is probably helpful)

Done.

phulin:
One other piece: per cf40 Vanya's Castle is the only zone with a stat req. The rest have none so I had that wrong in my PR.

Veracity0:
OK. And I may as well check out DiffLevels while I am here.

midgleyc (dismissed):
merge when ready ;-)

I assume 8-Bit Realm is no longer accessible on the autumn-aton screen, so there's no need to move it to obsolete instead of deleting it?

Veracity0:
> I assume 8-Bit Realm is no longer accessible on the autumn-aton screen, so there's no need to move it to obsolete instead of deleting it?

Good catch.

It IS on my Automaton's dropdown.

I'll move it to Obsolete.

Kasekopf:
Thanks for the quick support! One thing I noticed during testing now is that the score is being saved to the preference with commas:
```
Preference 8BitScore changed from 9,600 to 10,000
```

Veracity0:
> Thanks for the quick support! One thing I noticed during testing now is that the score is being saved to the preference with commas:
> 
> ```
> 
> Preference 8BitScore changed from 9,600 to 10,000
> 
> ```

That's OK because if your script converts it to an integer, it goes through a StringUtilities method which handles that fine.

Kasekopf:
Well, not when I tried to use `parseInt` in javascript to handle the setting value (since e.g. `parseInt("12,200") === 12`), which is how I noticed. But you're right that it's easy enough to handle on the script side.

Veracity0:
> Well, not when I tried to use `parseInt` in javascript to handle the setting value (since e.g. `parseInt("12,200") === 12`), which is how I noticed. But you're right that it's easy enough to handle on the script side.

OK.

I do ASH and Java.

ash to_int("12,200") -> 12000
StringUtilities.parseInt("12,200") -> 12000

I have no idea what JS "parseInt" calls.

Is this a Kolmafia JS bug?

Not my wheelhouse.

midgleyc:
As V suggests, you should use `toInt` instead.

```
> js toInt("12,200")

Returned: 12200

> js parseInt("12,200")

Returned: 12.0
```

phulin:
I don't like that - I had been under the impression that the "contract" exposed by Mafia was that integer preferences are not supposed to be stored with commas. That's why typical JS code (e.g. the `libram` library that lots of folks use in JS scripts) uses parseInt when reading out of a preference. I don't know of another preference that works this way. Maybe I am just uninformed.

Veracity0:
Well, I don't do JS, so am unaware of coding conventions.

It's easy enough to do the integer parsing when we store the property.

No change in behavior for KoLmafia internal use and for ASH scripts, but, whatever.

I have an open PR which will do that for this property.

PR #1452, opened 2023-01-21
Add basic data for new 8-Bit Realm.

phulin:
Still need to spade item drop rates for the new monsters and add progress tracking, and confirm that the buzzy beetle/goomba thing still works the same way.


Veracity0:
A subset of https://github.com/kolmafia/kolmafia/pull/1451 although your monster data is more accurate.
How about if I just take that and incorporate it into my PR?

phulin:
Sounds good!

Veracity0:
OK, I have pulled it in.

By the way - I got a Goomba today and no Buzzy Beetles - and KoLmafia's location details predicts I'll see Goombas - so that should be good.

PR #1453, opened 2023-01-21
refactor: make derived modifiers an enum

midgleyc:
Convert "derived" modifiers to enum.

I think derived modifiers are doubles which have to be computed after all the other doubles are computed. But then, prismatic damage is in a similar boat, but that's a double.

Looking at the function `Modifiers.get`, line 681, we compute all the derived modifiers, then throw most of them away and only take the one we're interested in. Ha! But then, all the buffed stats require each other if anything is equalised, and HP and MP require muscle and mysticality respectively, so perhaps most of it washes out.

`SpeculateCommand.doNumeric` is a good example of something we can update once `bitmaps` is moved to an enum: we're starting with a modifier, converting to a string, passing through to a method that doesn't know what the type of modifier is any more, but wants to do different things in different cases...

Curious how `isNumericModifier` is true only for doubles, despite `doNumeric` receiving doubles, derived and bitmaps.

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.EnumMap;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


Veracity0 (approved):

Veracity0:
I like this. I spent a lot of time working on mods.predict - and I am seeing big prediction discrepancies in the CompactSidePane when I am in a 2023 Standard run - but cleaning this up before I investigate more and make a bug report - is a good thing.

midgleyc:
One other thing I have noticed is that in some places in the code we assume HP / MP is a long, but in `predict()` we assume it is an int.

PR #1454, opened 2023-01-21
Run ubuntu checks through xvfb

Rinn:
This should allow headless ubuntu to run tests in swing components

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -25,6 +25,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Setup xvfb
+        if: matrix.os == 'ubuntu-latest'
+        uses: awalsh128/cache-apt-pkgs-action@latest
+        with:
+          packages: xvfb
+          version: 1.0
+
       - name: Setup Java
         uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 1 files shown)


jaadams5:
I merged this into #1441 and the failing test now runs and passes on Ubuntu.

Veracity0:
So can we get this made a non-draft and therefore open for review?

Veracity0:
My question is, what about non-Ubuntu platforms? Do we need to mark tests that depend on this as disabled for Mac and Windows? If so, how? And what can we do about it?

jaadams5:
This was proposed as an environmental change to fix a failing test in #1441,  It works.  Locally Windows tests have passed with and without this. It would be reassuring to have someone confirm that this works locally on non-Windows systems.

For reference xvfb implements an X11 display server virtually.  https://en.wikipedia.org/wiki/Xvfb

jaadams5:
> My question is, what about non-Ubuntu platforms? Do we need to mark tests that depend on this as disabled for Mac and Windows? If so, how? And what can we do about it?

This works on Windows.  It is a change to the environment and I don't have any expectation that code or tests would have to be enabled or disabled because of this change.  Before this GitHub ubuntu made no provisions for an X11 display in a test environment.  Now code that uses an X11 display (i.e. creating some top level Swing components) can be run and tested.

jaadams5:
I'm happy to approve and merge but I would like some  confirmation that @Veracity0 's concerns have been addressed.

midgleyc (approved):
We can see that this works for Windows and Mac already because it was used to run a Swing test on #1441, where the Windows and Mac tests already passed.

Veracity0 (approved):
Go for it.

PR #1456, opened 2023-01-22
pixel rocks and other tweaks

Veracity0:
1) Log when you visit the Treasure House
2) <s>pixel rock grows into a Pixel Rock</s>
3) Save 8BitScore property as an integer.

Somebody pointed out that it was being stored with commas.
I responded that all properties are strings but the method/function to get it as an integer, either in KoLmafia itself or in a script, handles that just fine.
They responded that JavaScript parseInteger() (or something) does not.
Well, that function is not part of KoLmafia's RuntimeLibrary - although toInt() is, surely - but, whatever.

It's easy enough to save the property already parsed - even though scripts will have to parse it again.


midgleyc (approved):
Thanks!

On commas: elf gratitude, when read from the quest log, is the other numeric preference I recall being saved with commas.

Veracity0:
I remembered that too and looked a bit at elfGratitude. It's found via consequences.txt
```
QUEST_LOG	elfGratitude	You earned ([\d,]+) Elf Gratitude during Crimbo 2022.	elfGratitude=$1
```
can we simply use ```elfGratitude=StringUtilities.parseInt($1)```?
That's pretty ugly, so I decided not to dive into how consequences work, for now, in favor of getting this case working sooner.

PR #1459, opened 2023-01-23
refactor: convert bitmap modifiers to enum

midgleyc:
Keep doing the bit operations for now. I'm still not sure what the point of them is. It seems to be treating every int as equivalent to a list of booleans, but sometimes it counts the bits. I don't know what the "mask" is for, either. Number of relevant slots? But I don't see where this is stored.

It has bitmaps and "raw" bitmaps. Bitmaps are a count of the bits. "Raw" bitmaps are using for masking?

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -25,6 +25,7 @@ import net.sourceforge.kolmafia.listener.NamedListenerRegistry;
 import net.sourceforge.kolmafia.listener.PreferenceListenerRegistry;
 import net.sourceforge.kolmafia.modifiers.BitmapModifier;
 import net.sourceforge.kolmafia.modifiers.BooleanModifier;
+import net.sourceforge.kolmafia.modifiers.DerivedModifier;
 import net.sourceforge.kolmafia.modifiers.DoubleModifier;
 import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.moods.HPRestoreItemList;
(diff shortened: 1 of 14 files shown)


Veracity0 (approved):
I sure wish I understood how these are supposed to work.

Raveosity is a bit map. Equipment that provides it gives +1 or +2.
3 items from the quest give +2, 3 items give +1, 1 item from the castle gives +1, 5 items from SSPD give +1.
You need +7 or more from distinct items being worn.
The two accessories from the quest are marked "cannot equip more than one at a time".

Why not just make it a double, like clowniness and surgeonosity?

Perhaps because the 5 accessories from SSPD are not so-marked.

If you somehow mapped a particular piece of equipment to a particular bit or two of a mask, you could look at the mask and deduce which items provided it. But it doesn't do this.

Perhaps it all came about because Jason wanted to store booleans as bits.

But all the other ones could just be counts in a double modifier. Notice that when KoL published "Clowniness" - a percentage - we kept Clownosity visible but implemented it in terms of Clowniness.

You want to say "only one of each distinct item?".
We now have Unique specification:

```
# Unique items section of modifiers.txt.
# These are items that all contribute to a modifier, but only once per unique item.
# Consider them to be SIngle Equip, insofar as the specified modifier is concerned.
```
(That should be "Single" not "SIngle".)

I am in favor of accepting this PR for now, since it finally gets rid of the "tables" of modifiers.

But perhaps a subsequent PR could look closely at the "bitmaps" - and how they are used in the Maximizer - and perhaps make them all into DoubleModifiers.

midgleyc:
I do plan to look at this more. The "numeric_modifiers" are doubles, "derived" and bitmaps. Some of the doubles could be derived; equally all of the derived could be (special-cased) doubles. I don't see why raveosity can't be a double if surgeonosity can.

"Evaluated Modifiers" is a special-case string modifier that isn't part of the enum (it's only a string). I want to see what it's for and if I can add it.

PR #1460, opened 2023-01-24
Hover text over next platform in Volcano Maze is now 'Next Platform'

Veracity0:
1) DiffLevel of the Outer Compund is mid
2) You can walk away from the Treasure House choice
3) In the Volcano Maze, once we have seen all the maps and can calculate a solution (always the case, if using cached maps - the default), the hover text over the "correct" next platform (where we will go if you click the Step button) now says "Next Platform".

I considered actually changing the image to have an "X" on it or something, but I don't seem to have access to a gif editor that lets me change the pixels.

### DIFF
--- a/src/relay/volcanomaze.2.js
+++ b/src/relay/volcanomaze.2.js
@@ -28,7 +28,7 @@ function adjustPlatforms(res) {
 				t.removeClass('no').addClass('you');
 				setTitle(t, 'You');
 			}
-		        t.removeClass('next');
+			t.removeClass('next');
 			if (t.attr('rel') == res.next) {
 				t.removeClass('no').addClass('yes').addClass('next');
 				setTitle(t, 'Next Platform');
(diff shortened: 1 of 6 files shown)


src/relay/volcanomaze.2.js
```diff
@@ -28,16 +28,24 @@ function adjustPlatforms(res) {
 				t.removeClass('no').addClass('you');
 				setTitle(t, 'You');
 			}
+		        t.removeClass('next');
+			if (t.attr('rel') == res.next) {
```
midgleyc, line 32: curious looking indentation here (tabs + space) -- if this wasn't copied from KoL, might be nice to make equivalent?

Veracity0: That was my editor. I will fix.

Veracity0: Done.

midgleyc (approved):

PR #1461, opened 2023-01-24
tests: add more tests for complex modifiers / maximizer code

midgleyc:
More learnings on modifiers:
* "uniques" only works in the maximizer -- if you maximize "surgeonosity" with 3 masks, it'll only recommend you equip one; but if you equip all three it'll still give you a score of 3 instead of 1
* `add(final Modifiers mods)` has some code dealing with watches, but it adds in the modifier anyway if it's not "adventures". However, you can't equip watches at the same time, so they don't stack even outside of adventures. The maximizer does deal with this correctly (separately).
* added some comments on what three top-level fields were. Notably, in all three cases "String" means something different.
* the class itself could (should?) be split into `ModifiersDatabase`, which like `EffectDatabase` does the parsing of the text file and serves as a store for the abstract modifiers, and `Modifiers`, which handles player modifiers as it currently does

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -8602,7 +8602,7 @@ MutexE	Broken Heart/Fiery Heart/Cold Hearted/Sweet Heart/Withered Heart/Lustful 
 
 # Unique items section of modifiers.txt.
 # These are items that all contribute to a modifier, but only once per unique item.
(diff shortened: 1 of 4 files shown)


Veracity0 (approved):
Crazy that someBoostIs was'nt actually being checked in all those places.

PR #1462, opened 2023-01-25
Handle day passes correctly when we already have access

libraryaddict:
## What we do

(As an example)
A player acquires an absentee voter ballot and uses it, expecting to be sent to the voting booth. However, the attempt failed as they already own the voter registration or have voted already.

## What we expect to happen

Mafia to see that we already own said iotm as per the response and thus the pass was not used. Nothing happens.

## What happens instead

Mafia assumes the pass was used, and removes it from inventory while setting `_voteToday` to true, which is invalid as we do own the iotm.

This results in a preference being wrongly set, with mafia having a mismatch in inventory counts for the day pass.

## Tests

The first commit are failing tests


midgleyc (approved):
Thanks!

PR #1463, opened 2023-01-25
refactor: modifiers changes

midgleyc:
1. Extract inner static classes from Modifiers.
2. Rename Modifier to ModifierValue.
3. Inline two methods doing the same thing as other methods.

I'm looking at extracting `ModifierDatabase` from `Modifiers`: this essentially amounts to moving almost all the static methods out. This PR contains the other refactoring, and can go in first to make things a bit easier.


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -506,7 +506,7 @@ private void parse(String expr) {
         continue;
       }
 
-      index = Modifiers.findName(keyword);
+      index = DoubleModifier.byCaselessName(keyword);
```
jaadams5, line 510: Was Modifiers.findName case insensitive?

jaadams5 (approved):

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
midgleyc: yes

PR #1465, opened 2023-01-26
Timezone Constant

jaadams5:
I decided GMT-0330 occurred in enough places that it, or actually TimeZones derived from it, could be a static constant.

This does that.  Note that CalendarFrameTest was created before this so that this change could be tested.  Note also that DateTimeManager already had 100% coverage so I did not make a new test.


gausie:
It's already a static constant in Holiday database, where it probably belongs I think

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -267,11 +270,7 @@ public static final void main(final String[] args) {
       }
     }
 
-    // All dates are presented as if the day began at rollover.
-
-    TimeZone koltime = TimeZone.getTimeZone("GMT-0330");
-
-    KoLConstants.DAILY_FORMAT.setTimeZone(koltime);
+    KoLConstants.DAILY_FORMAT.setTimeZone(kolTimeZone);
```
jaadams5, line 273: The variable koltime was only used to set the format TimeZone and so it is no longer needed.

jaadams5:
> It's already a static constant in Holiday database, where it probably belongs I think
> 
> Edit: maybe I'm wrong. I had a timezone in there for a while but perhaps I ended up not actually needing it

I just looked for usages of the string GMT-0330.  The constant ended up in KoLmafia because that was where the things that used it first were, but maybe KoLConstants or somewhere else are a better place.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -161,6 +161,9 @@ public static void ensureContentTypes() {
   public static MafiaState displayState = MafiaState.ENABLE;
   private static boolean allowDisplayUpdate = true;
 
+  // All dates are presented as if the day began at rollover.
+  public static final TimeZone kolTimeZone = TimeZone.getTimeZone("GMT-0330");
```
midgleyc, line 165: Might be nicer as `KOL_TIME_ZONE`. We're very inconsistent, but in general I think having constants in SHOUTY_SNAKE_CASE is a good thing.

jaadams5: Agreed.

midgleyc (approved):

PR #1467, opened 2023-01-26
refactor: split ModifierDatabase from Modifiers

midgleyc:
Split "Modifiers" into "Modifiers" and "ModifierDatabase". I also shifted a bunch of functions around and tried to separate them into "regions" of similarity. Also, I moved all the fields to the top, and all the static initialisation together.

Roughly, ModifierDatabase tries to handle those actions which are not character-specific: reading from the database in modifiers.txt, writing to it, parsing it, parsing KoL descriptions.

There is one exception: `getNumericModifier` for FamiliarData sets the user's current familiar on the Modifiers object as a static field. There are a few of these fields. This is used in `Expression` when calculating familiar-specific modifiers. I feel like we shouldn't do it this way -- for example, we could instead pass in "optional overrides" to `eval` stored on the `Modifiers` object, non-statically, so they apply only while the given object is in use.

One function that was only used in `BasementRequest` was moved there. I felt based on how it worked it wasn't usable in general.

I also added a bunch of TODOs throughout the code that will hopefully be answered as I learn more, or with the benefit of more tests. For example, in `KoLCharacter` we call `resetModifiers` on ascension and on character reset, which resets the database of modifiers, and I think we may not need to do this (unless moving to / from TCRS). This is mostly important for tests, as `reset()` is called a lot.


midgleyc:
I've gone back to initialising the ModifierDatabase in the Modifiers class due to the test failures. If anyone has an idea for a better way of doing this, please let me know.

jaadams5 (approved):
Perhaps longer than my attention span...

src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
```diff
@@ -0,0 +1,1524 @@
...
+
+      newMods.setString(mod, value);
+    }
+
+    newMods.setString(StringModifier.MODIFIERS, string);
+
+    return newMods;
+  }
+
+  // TODO: what's the difference between these are the above?
```
Veracity0, line 662: I believe that these are not in the blue "enchantments" section of a desc file. But by all means, check it out. 🙂

Veracity0 (approved):
What a big job. I like your organization.

Mostly. I like static Patterns that are used in a single method to be right in front of the method, since you easily see what it is while inspecting the method.

But your way is also a widely used code style, so, fine.

Veracity0:
Wow. It got merged while I was performing MY code review.

Well, I approved it, but will you even see my comments now?

midgleyc:
I created the PR and am watching it, so I see all the comments posted, whether the PR is open or not.

Thanks for the review :-D

midgleyc:
> Mostly. I like static Patterns that are used in a single method to be right in front of the method, since you easily see what it is while inspecting the method.

I also like that style... when the file is short. If I can't keep it all in my head at once, I find it easier to have "fields", "constant fields", "methods" etc. sections.

For example, in the original Modifiers code there was a static initializer ~1000 lines in. I really don't like that!

PR #1468, opened 2023-01-26
refactor: avoid string references for known modifiers

midgleyc:
1. Make "Evaluated Modifiers" a proper enum value instead of a special-cased string
2. Use functions taking Modifiers over functions taking Strings when looking up by enum.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -27,6 +27,7 @@ import net.sourceforge.kolmafia.modifiers.BitmapModifier;
 import net.sourceforge.kolmafia.modifiers.BooleanModifier;
 import net.sourceforge.kolmafia.modifiers.DerivedModifier;
 import net.sourceforge.kolmafia.modifiers.DoubleModifier;
+import net.sourceforge.kolmafia.modifiers.Modifier;
 import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.moods.HPRestoreItemList;
 import net.sourceforge.kolmafia.moods.MPRestoreItemList;
@@ -1615,7 +1616,7 @@ public abstract class KoLCharacter {
   }
 
   public static final int calculateMaximumPP() {
-    return 1 + (int) KoLCharacter.currentModifiers.get(DoubleModifier.PP);
+    return 1 + (int) KoLCharacter.currentModifiers.getDouble(DoubleModifier.PP);
   }
 
   public static final void resetCurrentPP() {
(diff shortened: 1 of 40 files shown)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5377,7 +5380,7 @@ record StatExp(DoubleModifier exp, DoubleModifier pct) {}
           };
       Function<DoubleModifier, Double> calc =
           (DoubleModifier statPct) ->
-              (finalBaseExp + finalExp) * (1 + newModifiers.get(statPct) / 100.0f);
+              (finalBaseExp + finalExp) * (1 + newModifiers.getDouble(statPct) / 100.0f);
```
Veracity0, line 5383: I realize this is modified existing code, but, 100.0f? As in a single float?

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -145,34 +146,34 @@ public Map<DerivedModifier, Integer> predict() {
...
-            + (int) Math.ceil(this.get(DoubleModifier.MUS_PCT) * mus / 100.0));
+            + (int) this.getDouble(DoubleModifier.MUS)
+            + (int) Math.ceil(this.getDouble(DoubleModifier.MUS_PCT) * mus / 100.0));
     rv.put(
         DerivedModifier.BUFFED_MYS,
         mys
-            + (int) this.get(DoubleModifier.MYS)
-            + (int) Math.ceil(this.get(DoubleModifier.MYS_PCT) * mys / 100.0));
+            + (int) this.getDouble(DoubleModifier.MYS)
+            + (int) Math.ceil(this.getDouble(DoubleModifier.MYS_PCT) * mys / 100.0));
```
Veracity0, line 171: Note that this one is not "f".

Veracity0 (changes requested):
I saw a lot of those string names in your last PR but didn't comment because it was outside the scope of that PR.

src/net/sourceforge/kolmafia/KoLCharacter.java
midgleyc: To clarify: the change you want is to remove the `f`?

midgleyc: There are more `f` in the stat lines ~5390 -- do you want them gone as well?

Veracity0: Yes. when you are calculating a double, why force some of the factors to be simple floats?

Veracity0 (dismissed):
Looks good. Thanks!

Veracity0 (approved):

PR #1473, opened 2023-01-28
Concurrent Modification Exception

jaadams5:
I saw this while testing CalendarFrame.  I added to the test and didn't see it again.  This PR removes the code I added that made the problem go away.  I have run the tests locally on Windows and have had 100% pass once and two concurrent modification failures. 

Maybe it will fail (again) on GitHub and someone (@Veracity0 ?) will have some ideas.

I do not think this should be committed but this is a way to both collaborate and access the environments GitHub uses.


jaadams5:
@Veracity0 - most recent failure is a concurrent modification although this is still intermittent.

jaadams5:
Closing because #1480 will address.

PR #1475, opened 2023-01-29
Correctly track usage of Molehill Mountain & Strange Stalagmite

libraryaddict:
## What happens

The player attempts to use either molehill mountain or strange stalagmite.

After using either item, the player is sent to the choice/fight.

## What we expected

The daily preference of molehill mountain & strange stalagmite to be set to true.

## What happens instead

The property is not updated when item is used.

After resolving the fight/choice, attempting to use the item again will have kol tell us that we can't use it again today.

Mafia now updates the property to true.


midgleyc (approved):
Thanks!

PR #1476, opened 2023-01-29
Add missing ascension resets for the properties `8BitColor`, `8BitScore` and `commerceGhostCombats`

libraryaddict:
## What happens

We ascend

## What do we expect

The three named preferences to be reset to default

## What happens instead

They're not reset


midgleyc (approved):
Thanks!

PR #1478, opened 2023-01-29
When a consequence sets a property with a value that is numeric, use setInteger

Veracity0:
1) Convention Hall Lobby is DiffLevel mid
2) consequences.txt sets properties based on text in quest logs, item descriptions, effect descriptions, skill descriptions, and so on. Many of them are numeric, and the Pattern will include something like ```(\d+)```. A handful of them have ```([d,]+)```.

In particular:
```
QUEST_LOG	elfGratitude	You earned ([\d,]+) Elf Gratitude during Crimbo 2022.	elfGratitude=$1
QUEST_LOG	pingpongSkill	You have achieved a skill level of ([\d,]+) at playing ping-pong	pingpongSkill=$1
QUEST_LOG	royalty	You have accumulated ([\d,]+) Royalty	royalty=[$1]
DESC_ITEM	bone abacus	defeated ([\d,]+) opponent	boneAbacusVictories=$1
DESC_ITEM	familiar scrapbook	Scraps Collected: <b>([\d,]+)</b>	scrapbookCharges=[stripcommas($1)]
```

I notice that the last two are in [], which means they are evaluated as a ModifierExpression.
One of them uses "stripcommas" and the other does not.
The first three are simply stored as parsed.

That confused several programmers, one of whom declared that "it is in the contract of Preferences that a numeric property's value has no commas". Bold statement, and not true, to my knowledge, but, whatever. We can do it that way to make JS coders happier.

This PR will look at the (evaluated) value and if it isNumeric (defined as "an integer, possibly with commas"), parse it as an integer (which strips commas), and store as an integer. If you later retrieve it as a String, there will be no commas.


jaadams5 (approved):

Veracity0:
Note that I am willing to consider forcing individual consequences to specifically require stripping out commas.

Veracity0:
Not that I don't trust frono, but I was hoping for another review to confirm this is a reasonable solution?

Whatever. And now to bed, and I will merge this when I get up unless there is discord among my reviewers.

midgleyc:
> Note that I am willing to consider forcing individual consequences to specifically require stripping out commas.

I think that's a good idea. They are numbers, and straight numbers is nicer than numbers with commas.

PR #1479, opened 2023-01-30
Release held session file lock before acquiring a new one

heeheehee-kolmafia:
Previously, we tied this to LogoutManager. However, given that a session can be ended without going through LogoutManager (e.g. sleeping through rollover, session timeout, creating a new session) and LoginManager can be invoked again, this breaks an existing [use case](https://kolmafia.us/threads/prevent-multiple-instances-of-kolmafia-from-accessing-the-same-account.28084/#post-170956).


jaadams5 (approved):

BadHorseMonkey:
My no-rollover and rollover tests all worked as expected.  I was both correctly locked out and the lock was restablished after rollover on reconnection.

src/net/sourceforge/kolmafia/KoLmafia.java
```diff
@@ -192,12 +192,23 @@ public static String imageServerPath() {
...
 
       if (KoLmafia.SESSION_FILE.exists()) {
         KoLmafia.SESSION_CHANNEL = new RandomAccessFile(KoLmafia.SESSION_FILE, "rw").getChannel();
         KoLmafia.SESSION_HOLDER = KoLmafia.SESSION_CHANNEL.tryLock();
+        if (KoLmafia.SESSION_HOLDER == null) {
+          KoLmafia.updateDisplay(
+              MafiaState.ABORT,
+              "Could not acquire file lock for "
+                  + suffix
+                  + ". Most likely, another instance of KoLmafia is logged in.");
```
BadHorseMonkey, line 210: This is not wrapping in my login window and cutting off part of the text.  We could make it two lines, or more condensed.  Not a showstopper.

BadHorseMonkey (approved):
This is working for me with minimal testing with two accounts and after rollover.

You might clean up the failure message, but otherwise a nice, small fix to a problem.

PR #1480, opened 2023-01-30
Address race condition in ScriptMenu.

heeheehee-kolmafia:
List.copyOf seems adequate to address the frequent ConcurrentModificationExceptions, but it most likely just makes the underlying data race less frequent.

Replacing these with synchronized lists is insufficient, as even synchronizedList requires explicit synchronization when using an Iterator.


jaadams5:
Walk me through, please.

The Concurrent Modification exception comes from iterating over a list.  copyOf makes it so that iteration is over a copy of the list and thus no exception should be possible.  However that is not acceptable if the list can change while it is being copied.

The list has possible sources.

One source is now a Collections.synchronizedList which in practice means it won't be changed while it is being copied.

The other source (scriptMRU) is now a ConcurrentLinkedDeque which in practice also means it won't be changed while it is being copied.

Not using the same object type in both cases kind of makes sense since the scriptMRU adds to the head and deletes from the tail.  Adding and removing from scriptMRU are already synchronized so this change in object may mean the synchronization can be removed.

What actually gets used in the ScriptMenu.init if the MRU is used is a new list that is created from a synchronized copy and reformat of scriptMRU.  This suggests to me that the race condition is not caused by the scriptMRU getting updated since the iteration is over a copy.  Nevertheless the object change may make it so the ScriptMenu operations do not need to be synchronized which sucked up a lot of wall clock time.

Anyway, I think this will at least reduce the frequency of occurrence.

heeheehee-kolmafia:
The previous ConcurrentModificationException as we saw in tests came from
- iterating over KoLConstants.scripts in ScriptMenu.init() while
- GenericFrame.compileScripts() rewrote the same list, element-by-element.

Java Iterators in particular are rather fickle, and backing them by a synchronized collection is insufficient as they aren't atomic -- if you modify a list, any extant Iterators would enter into an inconsistent state, so they detect this failure mode and throw an exception.

By using copyOf, no other thread has access to the resulting object, and so we can iterate over it without fear of data races.

You're correct that I made the switch to thread-safe collections specifically so copyOf would be safe (even though that wasn't a failure mode I personally saw in testing). We theoretically could have both structures as ConcurrentLinkedDeques, but certain methods have different performance characteristics (e.g. size() isn't a constant-time operation).

I'm not sure whether we can drop synchronization of ScriptMRUList entirely, since we would lose atomicity of various operations, and concurrent accesses could expose intermediate inconsistent states. If the class is appropriately synchronized, then we could probably revert the ScriptMRUList changes I had here.

jaadams5:
Thank you.  I think we should try this as is.  Synchronization of ScriptMRUList can be dealt with later.  There are also issues with building the menu as noted at https://kolmafia.us/threads/scriptmenu-init.28535/#post-171054 so it may be that dealing with that will change the race condition.

jaadams5 (approved):

PR #1481, opened 2023-01-30
Set timezone in CalendarFrameTest.

heeheehee-kolmafia:
new GregorianCalendar() uses the local time / timezone. This is problematic when the KoL date does not match the local date.


midgleyc (approved):
Thanks!

jaadams5:
Whoops.  Thanks.

PR #1485, opened 2023-01-31
Add oasisAvailable property

Veracity0:
Since a milestone can add desertExploration without opening the oasis, some adjustments are needed.

1) Boolean "oasisAvailable" property tracks whether The Oasis is available
2) If we see it on the desert map, true
3) If adventure in the desert and it opens up after a fight, true
4) if we have not adventured in the desert, but have used a milestone, desertExploration > 0, but oasisAvailable is false
5) canAdventure() in The Oasis depends on oasisAvailable, not desertExploration
6) unHydratedDesert Relay Warning depends on oasisAvailable, not desertExploration


midgleyc (approved):
Thanks!

PR #1487, opened 2023-02-01
Fix CalendarFrameTest failures, take two.

heeheehee-kolmafia:
This makes several changes:

- Explicitly specify the hour / minute to remove non-hermiticity of tests.
- Replace instances of local zone with the KoL "timezone" in CalendarFrame.
- Move timezone instantiation of KoLmafia formatters from main() to a
  static block, to ensure that they're executed during tests.


jaadams5:
Thank you.  It is my mess but I don't always hear about it.

src/net/sourceforge/kolmafia/swingui/CalendarFrame.java
```diff
@@ -85,7 +84,7 @@ private void buildCalendarFrame(Calendar calendarToUse) {
     CalendarFrame.selectedDate = calendarToUse;
 
     CalendarFrame.calculatePhases(
-        CalendarFrame.selectedDate.toInstant().atZone(ZoneId.systemDefault()));
+        CalendarFrame.selectedDate.toInstant().atZone(KoLmafia.KOL_TIME_ZONE.toZoneId()));
```
jaadams5, line 87: I am somewhat vindicated because the the .systemdefault() was a change from 2022 and not mine.  I am, however, distressed that my test was inadequate.

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
```diff
@@ -243,8 +244,8 @@ public void calendarFrameShouldHaveBasicFunctionality() {
 
   @Test
   public void itShouldCalculateExpectedCalendarDataForKnownTime() {
-    Calendar useTime = new GregorianCalendar();
-    useTime.set(2010, Calendar.SEPTEMBER, 1);
+    Calendar useTime = new GregorianCalendar(KoLmafia.KOL_TIME_ZONE);
```
jaadams5, line 247: We have a GitHub or merge issue.      Calendar useTime = new GregorianCalendar(KoLmafia.KOL_TIME_ZONE); is already in main according to what my system thinks is main.

jaadams5 (approved):
Comments are not blocking but I am concerned about possible confusion about branching.  I will let someone else resolve the conflict but I agree with the solution.

test/net/sourceforge/kolmafia/swingui/CalendarFrameTest.java
heeheehee-kolmafia: Manually merged. I think I must have just based this off an older commit.

src/net/sourceforge/kolmafia/swingui/CalendarFrame.java
heeheehee-kolmafia: Your test was fine for most of the day :)

It was only between rollover and midnight (server local time) that it ran into issues.

PR #1491, opened 2023-02-02
feat: diff levels for stupors

midgleyc:


midgleyc:
This is already in, I was just fooled by the PR interface into thinking it wasn't.

PR #1493, opened 2023-02-03
sit skills and oasis tests

Veracity0:
1) When you examine the S.I.T. Course Completion Certificate, set _sitCourseCompleted setting
2) When you choose a course, which grants you a skill, remove the other two skills
(Note that the skill you choose persists across rollover)
3)For completion, add the other two cases for detecting (or not) that the Oasis is available:
- Before the Level 11 Quest opens the desert
- After exploring but before Gnasir talks to you.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -2119,6 +2119,7 @@ user	_shortOrderCookCharge	0
 user	_shrubCharge	0
 user	_shrubDecorated	false
 user	_silverDreadFlaskUsed	false
+user	_sitCourseCompleted	false
 user	_skateBuff1	false
 user	_skateBuff2	false
 user	_skateBuff3	false
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -11141,7 +11141,7 @@
 11113	pixel whiskey	262338395	pixelwhiskey.gif	drink	t,d	30
 11114	pixel rock	827961712	pixelrock.gif	grow	d	16384
 11115	S.I.T. Course Voucher	362646202	sitvoucher.gif	usable	t	0
-11116	S.I.T. Course Completion Certificate	495156664	sitcert.gif	usable		0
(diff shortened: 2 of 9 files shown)


midgleyc (approved):
Thanks!

Veracity0:
I will have a followup tomorrow. After rollover at least.

Add a reset-on-ascension setting which is your current SIT passive skill.
(Perhaps)when we parse charsheet, any of the three which appears goes into that property
When you select a new skill, clear that one and set the property to the new one.
Verfiy that you cannot walk away from this choice.

It will require testing in KoL - and collecting html for writing some tests.

For now, this agrees with this morning's experimentation.

First turn of the day:

[code][302] The Smut Orc Logging Camp
Encounter: smut orc pipelayer
Round 1: Veracity uses the rock band flyers!
Round 2: You lose 2 hit points
Round 2: Veracity casts FIRE EXTINGUISHER: ZONE SPECIFIC!
lump of loyal latite is multiusable, but KoLmafia thought it was not
Round 3: Chef Bookicus speaks carefully, "My recollection is that Yeast of Boris was often collected from an A.M.C. gremlin. If I recall correctly, you can hunt them in Out by that Rusted-Out Car."
You acquire an item: lump of loyal latite
[/code]

Psychogeology was still in effect from previous day.
Note also the multi-usability.
I would not be surprised if all the potions had that.

[code]use 1 S.I.T. Course Completion Certificate
Encounter: Examine S.I.T. Course Certificate
Took choice 1494/3: Write in Cryptobotany
choice.php?pwd&whichchoice=1494&option=3
You learned a new skill: Cryptobotanist
[/code]

midgleyc:
You also get a drop every 11 combats -- it would be nice to track that. I think free runs count. Like the train set?

However, there's no in-game counter, and we only know when it resets by when we get a drop.

Veracity0:
Notice that blasting the zone with the fire extinguisher did not cost a turn - but I got the drop.

PR #1494, opened 2023-02-03
Check coat of paint's modifiers after exiting avatar path

libraryaddict:
## Steps to reproduce

With coat of paint in inventory or equipped, break the prism to exit an avatar path.

As the player's class has changed, coat of paint has its modifiers changed as well.

## What happens

Mafia does not recognize coat of paint has changed and the equipment modifiers are now different.

The description was not visited as it should have been by mafia.
Manually checking the description will correct the modifiers.

## What we expected instead

Mafia to check coat of paint and update the equipment modifiers without player intervention required.


libraryaddict:
Oh dear, just realized that the test that should've been failing, was succeeding.

src/net/sourceforge/kolmafia/session/InventoryManager.java
```diff
@@ -1733,14 +1733,18 @@ public static void checkVampireVintnerWine() {
...
 
-  public static void checkCoatOfPaint() {
+  public static void checkCoatOfPaint(boolean playerClassChanged) {
     AdventureResult COAT_OF_PAINT = ItemPool.get(ItemPool.COAT_OF_PAINT, 1);
-    String mod = Preferences.getString("_coatOfPaintModifier");
+
     if (!KoLCharacter.hasEquipped(COAT_OF_PAINT, EquipmentManager.SHIRT)
-        && !KoLConstants.inventory.contains(COAT_OF_PAINT)) {
+        && COAT_OF_PAINT.getCount(KoLConstants.inventory) == 0
+        && COAT_OF_PAINT.getCount(KoLConstants.closet) == 0) {
```
midgleyc, line 1741: `InventoryManager.getAccessibleCount` instead (or one of the related methods)? Or do we really want exactly equipped, inventory + closet?

midgleyc (approved):
Thanks!

PR #1496, opened 2023-02-05
refactor: convert panel actions to enums

midgleyc:


src/net/sourceforge/kolmafia/swingui/panel/ItemManagePanel.java
```diff
@@ -370,7 +372,7 @@ protected int getDesiredItemAmount(
       final String itemName,
       final int itemCount,
       final String message,
-      final int quantityType) {
+      final QuantityType quantityType) {
     int quantity = 0;
     switch (quantityType) {
```
Veracity0, line 377: Probable out of scope for this simple change, but making this an enum might allow this to be a switch expression.

There are other issues - break at the end of new-style switch block, instanceof followed by casts, etc. - but those sre not related to enumitude.

Veracity0 (approved):

midgleyc:
may as well while we're here.

I think all the switches are already switch expressions?

Veracity0:
> I think all the switches are already switch expressions?

Not quite. The method you just fixed is like this:
```
    method {
      int quantity = 0;
      switch (quantityType)
      ... case statements to calculate/set quantity
      }
      return quantity;
    }
```

midgleyc:
I've converted more switch expressions, but I've left that one. It is a switch expression (`case ->` instead of `case:`), just the expression isn't used. It's a nontrivial conversion as `quantity` changes in the middle of some expressions, and there aren't tests around there.

Veracity0:
> I've converted more switch expressions, but I've left that one. It is a switch expression (`case ->` instead of `case:`), just the expression isn't used. It's a nontrivial conversion as `quantity` changes in the middle of some expressions, and there aren't tests around there.

Fair enough.

Veracity0 (approved):

PR #1497, opened 2023-02-05
refactor: make clan lounge requests enum

midgleyc:
The ones below "actions" are "options", which are meaningful ints.


Veracity0 (approved):
Timely. I still have refactoring Floundry and Hot Dog Stand, a la what I did with the speakeasy, on my to do list. 😃

PR #1498, opened 2023-02-05
refactor: replace top menu style with enum

midgleyc:


src/net/sourceforge/kolmafia/request/AccountRequest.java
```diff
@@ -275,10 +273,8 @@ private static void parseAction(final String location, final String responseText
       // account.php?pwd&action=menu&value=normal&ajax=1
       GenericRequest.topMenuStyle =
           valueString.equals("fancy")
-              ? GenericRequest.MENU_FANCY
-              : valueString.equals("compact")
-                  ? GenericRequest.MENU_COMPACT
-                  : GenericRequest.MENU_NORMAL;
+              ? TopMenuStyle.FANCY
+              : valueString.equals("compact") ? TopMenuStyle.COMPACT : TopMenuStyle.NORMAL;
       return;
```
Veracity0, line 278: This could be a switch expression, because it uses equals(), but the parallel example above uses contains(), so can't. I think consistency is good, so this us fine.

Veracity0 (approved):

PR #1500, opened 2023-02-05
refactor: remove some locked lists from equipment manager

midgleyc:
1. `equipment` is never used in a GUI so doesn't need to be Lockable.
2. `newOutfits` is a local variable passed to a function that only uses list methods, so it can be a plain list too.

Looking at `historyLists` not sure there's benefit in storing every piece of equipment we've equipped this session, but no major harm I guess.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/EquipmentManager.java
+++ b/src/net/sourceforge/kolmafia/session/EquipmentManager.java
@@ -36,6 +36,7 @@ import net.sourceforge.kolmafia.request.EquipmentRequest.EquipmentRequestType;
 import net.sourceforge.kolmafia.swingui.panel.GearChangePanel;
 import net.sourceforge.kolmafia.textui.command.ConditionsCommand;
 import net.sourceforge.kolmafia.utilities.LockableListFactory;
+import net.sourceforge.kolmafia.utilities.SortedList;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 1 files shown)


Veracity0 (approved):
I was intending to (soon?) look into a GearChanger issue: for accessories, at least, if you acquire one while that frame is open, the dropdowns do not update until you close and reopen the frame. I assume the GUI elements need a listener to be signaled, or something.

In any case, your comment lines are useful.

One terminology comment: a LockableListModel is not actually "lockable"; hola implemented that data structure for Spellcast - a previous project - and removed "Lockability" because it caused issues (deadlocks?) in the AWT event thread - before bringing it to this project. Or so I glean from comments.

So, the aspect that GUI elements need is the "ListModel" interface - not Lockable.

This is a nice simple fix.

And if we don't actually use or expose equipment historyLists, as far as I'm concerned, they can go away.

midgleyc:
We use `historyLists` only for replacing a "breakable" piece of equipment with whatever we had equipped before. We just store everything we ever equipped because it tries pieces in order in case we no longer have the last thing we equipped (say we sold it, perhaps).

The `accessories` list is a strange one: EquipmentManager doesn't do anything special for the other duplicates: folder 1/2/3/4/5, sticker 1/2/3. Also, it gets created as a ListModel, then each `equipmentList` entry is another ListModel that "mirrors" (which seems to just be wrapping) the first one. Then the gear panel ignores all that and just creates a series of internal models for the main equipment anew.

Veracity0:
"Mirroring" allows each ListModel to have its own Selected Index.

midgleyc:
ah thanks, that's what I was missing.

PR #1503, opened 2023-02-06
refactor: convert candy type, bad moon adventure type to enum

midgleyc:


src/net/sourceforge/kolmafia/session/BadMoonManager.java
```diff
@@ -7,41 +7,41 @@
...
-    DAMAGE2,
-    RESIST1,
-    RESIST2,
-    ITEM_DROP,
-    MEAT_DROP,
-    DAMAGE_REDUCTION,
-    MEAT,
-    ITEMS,
-  };
+  public enum Type {
```
Veracity0, line 10: Whenever (somebody) converted the class(?) or Object array(?) into "record Encounter", they declared a field named "type" for the "type of the encounter".

I sort of wish they'd named it "encounterType" - in which case this enum could have been "EncounterType".

Which is a lot more informative than a top-level enum named "Type".

This is named consistently with the existing code, so I will accept it - but I will still grumble about the existing code. :)

Veracity0 (approved):

src/net/sourceforge/kolmafia/session/BadMoonManager.java
midgleyc: it could well be `EncounterType`, but in this case as it's an inner enum, I see it as being called `BadMoonManager.Type` -- perhaps not entirely descriptive, but you can infer what it means.

PR #1505, opened 2023-02-07
The correct platform to step to in the volcano maze now has a white plus on it

Veracity0:
I'm going to write s comment explaining exactly how KoL handles the maze:

- Initial visit gives HTML with a "div" for each square specifying a "class" and an HTML link.
- Each square is rendered via a "style" - classes "you", "goal", "yes", "no"
- Javascript handles onClick for each square and submits a request.
- KoL responds with JSON and the JS changes the "yes"/"no" status of each square as visibility changes and re-renders.

I will also get the raw HTML of visiting the maze - including line breaks - and will write a simple test for decorating that page.

And that is done. Hope it is helpful!


midgleyc (approved):
I will trust that you've tested this in-game :)

Veracity0:
Not only in-game, but in-run.

I love the Nrmrsis quest, so always have the volcanic island open by day two at the latest. I never claimed to be optimal. 😃 

This time I got it open on day one and made the gif/wrote the code last night - and after rollover, spent less than 10 turns total on the island to reach the maze.

After getting my 5 Breathitin; I'm not COMPLETELY sub optimal.

The code worked fine first time, but I initially had actual animated gifs for the platformup{1234} images, and, try as I might, literally none of the online gif editing tools actually set the loop parameter to once only, even though they said they did. Instead, infinite looping - which is unacceptable.

But it looks great even if the "next" platform does not animate rising from the lava.

I am very happy with how this all worked out. I think it's all I want.

PR #1507, opened 2023-02-07
refactor: maximizer list indexed by slot and familiar

midgleyc:
I'm looking at using enums in EquipmentManager, and as part of this this particular Maximizer construct needs to change. It is a list of lists, indexed by slot (and Evaluator's pseudo-slots), with optional familiars stuck on the end. After changing to an enum, it can no longer work that way.

Extract it into its own object to make it easier to change, but don't actually make the full change yet.


jaadams5 (approved):
Other folks should note that SlotList is always big enough to hold the equipment slots which initially confused me when looking at ArrayList being replaced by SlotList.

PR #1508, opened 2023-02-07
tests: add tests for parsing status + folders

midgleyc:
As part of converting EquipmentManager to use enums, I'd like parts of the codebase to have more tests.

This covers parsing `status.php` for equipment and parsing the folder holder page. It also fixes a bug in the latter.


jaadams5 (approved):

Veracity0:
FWIW, it's ```api.php```, not ```status.php```.

midgleyc:
Right, yes. `api.php?what=status`. I even went there to get the data for the test :P

PR #1509, opened 2023-02-07
tests: (un)equip command

midgleyc:
Tests for unequipping and equipping things, especially folders, as this involves arithmetic with slots.


gausie (approved):
Awesome, and definitely get merged before the slot enum changes as I imagine was your intention!

PR #1510, opened 2023-02-08
refactor: make EquipmentManager slots an enum

midgleyc:
Pieces of code using 
```java
(int slot = EquipmentManager.HAT; slot <= EquipmentManager.FAMILIAR + 1; ++slot)
``` 
were changed to omit the crown of thrones. I don't think it's necessary (and I don't know why they were written this way to begin with).

UnequipCommand was hat -> sticker3. It was also changed to omit the crown.

There are a couple of outstanding things to check on the Maximizer (e.g. CLI calls that equip as they go along) but this should be mostly done.


src/net/sourceforge/kolmafia/SpecialOutfit.java
```diff
@@ -102,24 +104,22 @@ public boolean isWearing() {
     return this.isWearing(-1);
   }
 
-  public boolean isWearing(AdventureResult piece, int type) {
-    if (type == EquipmentManager.ACCESSORY1
-        || type == EquipmentManager.ACCESSORY2
-        || type == EquipmentManager.ACCESSORY3) {
+  public boolean isWearing(AdventureResult piece, Slot type) {
+    if (type == Slot.ACCESSORY1 || type == Slot.ACCESSORY2 || type == Slot.ACCESSORY3) {
```
gausie, line 108: This could use `EquipmentManager.ACCESSORY_SLOTS`

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -42,83 +49,110 @@
...
+      return phpNameToSlot.getOrDefault(name.toLowerCase(), Slot.NONE);
+    }
+
+    public static Slot byOrdinal(int ordinal) {
+      return VALUES[ordinal];
+    }
+  }
+
+  // Count of slots visible on equipment page: HAT to FAMILIAR
+  public static final EnumSet<Slot> SLOTS = EnumSet.range(Slot.HAT, Slot.FAMILIAR);
```
gausie, line 128: Can we give these EnumSets javadocs so the difference is clearer in an IDE? e.g. the difference between SLOTS and ALL_SLOTS isn't immediately obvious.

src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -42,83 +49,110 @@
...
-  public static final int HOLSTER = 2;
-  public static final int OFFHAND = 3;
-  public static final int CONTAINER = 4;
-  public static final int SHIRT = 5;
-  public static final int PANTS = 6;
-  public static final int ACCESSORY1 = 7;
-  public static final int ACCESSORY2 = 8;
-  public static final int ACCESSORY3 = 9;
-  public static final int FAMILIAR = 10;
+  public enum Slot {
```
gausie, line 52: I feel like we could spin this out into its own file with great benefit to the readability of the two resulting files, but maybe you tried this

midgleyc: perhaps I've just looked at these files too long so it's familiarity, but I do like seeing it at the top of `EquipmentManager`

Could move it and all the SLOTS sets out.

gausie: That's what I was thinking. It's stylistic obviously but it works for me partly because of Slot being used as an enumerated value / class in our scripting world.

src/net/sourceforge/kolmafia/EquipmentSlot.java
```diff
@@ -0,0 +1,114 @@
+package net.sourceforge.kolmafia;
+
+import java.util.Arrays;
+import java.util.EnumSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class EquipmentSlot {
```
gausie, line 10: Why contain slot within equipment slot? Can you store the sets as statics on Slot?

midgleyc: I could do, but I think it looks better to only have CAPITALS be the enum values, and anything else to be out of it. So e.g. `Slot.NONE` is a slot, but `Slot.SLOTS` would not be, so I've put the latter at `EquipmentSlot.SLOTS`.

gausie: I think there are ways to achieve that goal without making Slot a subclass. It's annoying DX that I want to move away from in (for example) AscensionPath.Path

midgleyc: Why do you think it's annoying DX? I rather like it.

gausie (approved):
This is a huge improvement to the codebase. I can't be certain that nothing breaks here, but I think our users can let us know pretty quickly :)

PR #1512, opened 2023-02-08
Strip choice number readout from NCs on dev server

gausie:
You'll have to take my word for the test HTML! devsters get a readout of the choice number when visiting a choice, which mafia records as part of lastEncounter. It's convenient for me to automate testing in mafia just as much as it's convenient for me to automate playing, and stripping this makes it more compatible with various library scritps I use that check the string value of lastEncounter


src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -390,7 +405,8 @@ public static final String registerEncounter(final GenericRequest request) {
           };
     }
 
-    String prettyEncounter = StringUtilities.getEntityDecode(encounter);
+    String prettyEncounter = stripDevReadout(StringUtilities.getEntityDecode(encounter));
```
jaadams5, line 408: Conceptually I would rather only do this if mafia knows it is running on the dev server.  But there are very few (two?) places where mafia knows that and, in the absence of a profiler, I can certainly understand that checking might be as expensive as just running stripDevReadout unconditionally.  A comment, not a blocker.

jaadams5 (approved):
I expect the number of users who will benefit can be counted on one hand but given who those users are, I'm OK with this.

src/net/sourceforge/kolmafia/request/AdventureRequest.java
gausie: I can see if it's easy to check for that!

midgleyc (changes requested):
Both `useDevProxyServer` and `loginServerName` should be able to be used to check whether you're on the dev server or not.

Can you check before calling `stripDevReadout`?

gausie:
> Both `useDevProxyServer` and `loginServerName` should be able to be used to check whether you're on the dev server or not.
> 
> Can you check before calling `stripDevReadout`?

@midgleyc done

midgleyc (approved):
Thanks!

PR #1513, opened 2023-02-08
refactor: extract equipscope to enum

midgleyc:
EquipScope is an int used by the maximizer taking values -1 to 2.

-1 is used for the CLI call to indicate the maximization isn't just speculation, but the equipment returned should be equipped immediately. The others (0/1/2) are from the GUI and indicate that the maximization is speculative and whether the maximization should consider inventory / creatable / pullable or buyable items.

It's multi-purpose. Make it slightly clearer what's going on by making it an enum.

Taking feedback on the names; they're not great.


gausie:
Maybe ANY, SPECULATE_INVENTORY, SPECULATE_CREATABLE, SPECULATE_ANY?

Veracity0:
EQUIP_NOW, SPECULATE_INVENTORY, SPECULATE_CREATE, SPECULATE_ANY

midgleyc:
Changed to Veracity's suggestions. The first can't be ANY because it only checks inventory, not all possible creations. I think EQUIP_NOW fits nicely "equip what is available to equip".

jaadams5 (approved):

PR #1515, opened 2023-02-09
Use getAction / getPreaction in CampgroundRequest.registerRequest.

heeheehee-kolmafia:
GenericRequest.ACTION_PATTERN changed back in September to no longer match preactions, which in turn broke handling of campground.php-originated skill casts. This is especially problematic for librams, as it caused Mafia's tracking of libram casts to desynchronize from reality, which quickly compounded when trying to cast libram skills in bulk.

(This doesn't have tests yet. If anyone wants to contribute those before I get around to it, that'd be appreciated. Otherwise, I'll try to remember to come back to this over the weekend.)


heeheehee-kolmafia:
Added a test and flipped this to be ready for review.

gausie (approved):

gausie:
Nice

PR #1518, opened 2023-02-09
WOOL is a creation method that uses grubby wool

Veracity0:
I decided to go creation method, rather than coinmaster.
Because, meh. Who cares if you see the "create from grubby wool" method in CoinmastersFrame or the Purchases frame?

I modeled it on the METEOROID method. There are a couple of differences from MeteoroidRequest

1) grubby woolball has a yield of 3. Therefore, if you want to create 1-3, it takes one grubby wool. Creating 4-6 takes 2. And so on.
2) That takes care, since the subclass needs to do a "inv_use.php" -> "choice.php" redirection before doing its call to super.run (CreateItemRequest) which will invoke ITS super.run(), which will submit the choice.php request selecting the option. CreateItemRequest has its own "createLoop" which will do that until desired amount has been created.
3) Therefore, since GrubbyWoolRequest has to have its own loop (calling inv_use each time), it has to force CreateItemRequest to NOT iterate more than once.
4) *** Other creation methods that use choice.php need to do this too.
5) After fixing that, the code I copied from MeteoroidRequest still failed because it created the request to "inv_use.php" outside of the loop and attempted to simply reuse it. The issue is that the initial request followed a redirect to choice.php, so reusing that request resubmits the redirected URL - and you are no longer in a choice. Therefore, needed to make a new request to "inv_use.php" each time through the loop. And now it works.
6) *** Other creation methods that use choice.php need to be aware of this too.
7) I fixed MeteoroidRequest
8) I fixed BurningNewspaperRequest
9) I fixed WaxGlobRequest. This had a number of issues, which I addressed.
- choice 1218 - Wax On - cannot be walked away from
- Therefore, expose and run super.runCreateItemLoop() method, rather than calling super.run() directly, which aborts if we are handling a choice
- Therefore exclude choice 1218 from logging the Encounter for each option taken
- Both WaxGlobRequest and CreateItemRequest wanted to loop "count" times. It worked by accident, since after CreateItemRequest finished its loop, quantityNeeded was zero, and subsequent calls to it became no-ops.
- Therefore just let CreateItemRequest handle the looping; unlike the other three creation methods, you do NOT need to let the subclass call inv_use.php each time; you stay in the choice.


midgleyc (approved):
Thanks!

PR #1519, opened 2023-02-10
Track capped adventure absorbs in Grey You properly

libraryaddict:
## Brief

In Grey You you can no longer absorb monsters after reaching an adventure cap of 500. In this scenario we have reached the cap.

In both tests we are fighting a tomb asp which gives 10 adventures each absorb.

## Reprocessing an absorb

The first test, we are trying to reprocess the tomb asp. The reprocess fires as expected, however adventures are not given.

The reprocess can not be triggered on the tomb asp again, even if we recover the familiar's weight.
As such, we should track the monster as reprocessed, but as we did not absorb the monster, it should not be tracked as absorbed.

Mafia however, will track it as reprocessed, but also track it as absorbed.

## Defeating an absorb

Secondly, we defeat the tomb asp normally. It normally would give 10 adventures, but as we have reached the cap it will instead give +10 stats to a random stat.

This "absorb" is not tracked by kol, and you will always qualify for the +10 stats.

Mafia however will see it as a successful absorb, and track it as such. However, this is only acknowledged by mafia, and only in the current session.

If you restart mafia, the monster will no longer be considered by mafia as absorbed, as the charsheet does not mention it.

## Tracking adventures from absorbs

Seeing as we have the 500 adventure cap, tracking the adventures given by an absorb in a preference makes sense as otherwise it requires players to know how many adventures each absorb would give and tally up the result themselves.

This is obviously not very feasible for a human.


src/net/sourceforge/kolmafia/session/GreyYouManager.java
```diff
@@ -124,6 +124,32 @@ public static GooSkill learnSkill(int skillId) {
...
+    }
+
+    if (absorb.getType() == AbsorptionType.ADVENTURES) {
+      Preferences.increment("_greyYouAdventures", ((GooAbsorption) absorb).value);
+    }
+
+    absorbMonster(monster);
+  }
+
   public static void absorbMonster(MonsterData monster) {
```
midgleyc, line 153: as this is now only called from the above method + tests, it can be package private.

src/net/sourceforge/kolmafia/session/GreyYouManager.java
```diff
@@ -124,6 +124,32 @@ public static GooSkill learnSkill(int skillId) {
...
+    }
+
+    Absorption absorb = allAbsorptions.get(monster.getId());
+
+    if (absorb == null) {
+      return;
+    }
+
+    // If absorb is expected to give adventures, but neither message for adventure absorbs is found
+    if (absorb.getType() == AbsorptionType.ADVENTURES
```
midgleyc, line 140: might be slightly cleaner to split this one out into its own `if()` then indent the rest under it, so the check only gets called once and combines with the preference increase below

midgleyc (approved):
Thanks!

Yeah, removing the other method entirely is a nicer simplification.

PR #1521, opened 2023-02-10
refactor: remove liminal maximizer method

midgleyc:
A while back, when the filter levels were changed, an old version of the method taking an int was left around to give time to migrate the preferences.

I think it's been long enough now. Additionally convert the filters from a Map<filterType, boolean> to a Set<filterType> for simplicity. We only care about whether the values are true, not whether they're in the map.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
@@ -2,6 +2,7 @@ package net.sourceforge.kolmafia.maximizer;
 
 import java.util.Collections;
 import java.util.EnumMap;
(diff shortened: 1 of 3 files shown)


heeheehee-kolmafia:
Hm. (Just writing down my thought process.)

filter.getOrDefault(filterType, false) is equivalent to filterSet.contains(filterType), but the cases where the default value is true is not.

So, this is safe for the case where filterLevel is 0 (where we manually added all the elements to the map) and the case in MaximizerFrame which also adds all possible values to activeFilters (and never removes it).

And, since those are the only two places where filters or filterLevel are specified, this shouldn't change any behavior in practice.

PR #1523, opened 2023-02-10
Fix rock garden tracking

libraryaddict:
## What we do

The player harvests an item from the rock garden, then uses `get_campground()` to check what their garden status is.

After that, they use their packet of tall grass seeds, then call `get_campground()` again for some reason.

## What happens

Their first `get_campground()` will claim that there was no harvest done, because mafia does not parse campground on `rgarden[123]`. So whetstone will report `1`

The second `get_campground()` will claim that rock garden and the harvests are still available.

## What we expected

The first check will properly see whetstone as `0` in get_campground

The second check will remove garden seeds and harvests.

## Disclaimer

So I realized after I made the test that its not to my knowledge possible to have a garden, then get rid of said garden to have no garden. So the test may not be valid.

That said, it seems obvious that the crop clearing should be performed just before we update what crops are available.


midgleyc (approved):
Thanks!

PR #1528, opened 2023-02-12
Buying hat from firework shop only requires autoSatisfyWithNPCs

Veracity0:
A fireworks hat is an NPC-only quest item.
It should only require autoSatisfyWithNPC.
KoLmafia.makePurchases erroneously discarded the NPCPurchaseRequest since there was not an unlimited stock.

Fix: have it actually look for an NPCPurchaseRequest to decide it found one, rather than requiring an unlimited inventory to signal that.


test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -137,4 +144,153 @@ public void willNotLookAtDescriptionWithValidSkill() {
       }
     }
   }
+
+  @Nested
+  class LimitedNPCItems {
+    @Test
+    public void willRetrieveIfCanUseMall() {
```
midgleyc, line 151: If `autoSatisfyWithMall` were true and `autoSatisfyWithNPCs` false, would purchasing work (I assume not)? Does `willRetrieveIfCanUseMall` use a different code path to `willRetrieveIfCanUseNPCStore`? Should it, given that you can't buy the fireworks hats in the mall?

midgleyc (approved):
One question, but I think this is correct.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
Veracity0: It does not use a different code path.

[code]
    boolean shouldUseMall = !forceNoMall && InventoryManager.canUseMall(item);
...
      boolean onlyNPC = forceNoMall || !InventoryManager.canUseMall();
      List<PurchaseRequest> results =
          onlyNPC ? MallPriceManager.searchNPCs(item) : MallPriceManager.searchMall(instance);
...
  public static boolean canUseMall(final int itemId) {
    return ItemDatabase.isTradeable(itemId) && InventoryManager.canUseMall();
  }
[/code]

Since the item is non-tradable, it always goes to the NPC store and the only PurchaseRequests are from the NPC store.

PR #1530, opened 2023-02-13
Add 2CRS Ring parsing

BadHorseMonkey:
Adds 2CRS ring to individually parsed items.


src/net/sourceforge/kolmafia/session/InventoryManager.java
```diff
@@ -1859,6 +1860,18 @@ public static void checkDesignerSweatpants() {
     }
   }
 
+  public static void checkRing() {
+    // checks the Two Crazy Random Summer Ring, which has up to 37 little enhancements on it.
+    AdventureResult RING = ItemPool.get(ItemPool.RING, 1);
+    if (!KoLCharacter.hasEquipped(RING)
```
midgleyc, line 1866: `getAccessibleCount` is probably nicer here

midgleyc: or `itemAvailable`

jaadams5:
Any chance of a test?

BadHorseMonkey:
> Any chance of a test?

heh.  I've asked enough people to write them, I should.

src/net/sourceforge/kolmafia/session/InventoryManager.java
```diff
@@ -1859,6 +1860,17 @@ public static void checkDesignerSweatpants() {
     }
   }
 
+  public static void checkRing() {
```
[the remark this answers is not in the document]
BadHorseMonkey: done.

BadHorseMonkey:
> I also agree that a test would be nice.

test added.  I could test more properties if we want to test them all.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -293,4 +299,29 @@ public void willNotRetrieveIfAlreadyBoughtFireworkHat() {
...
+  class PerUserVariableItems {
+    @Test
+    public void RingWillSetModsForUser() {
+      var builder = new FakeHttpClientBuilder();
+      var client = builder.client;
+
+      var cleanups = new Cleanups(withHttpClientBuilder(builder), withItem(ItemPool.RING, 1));
+      try (cleanups) {
+        String descId = ItemDatabase.getDescriptionId(ItemPool.RING);
+        var req = new GenericRequest("desc_item.php?whichitem=" + descId);
```
gausie, line 313: You can prep the HttpClientBuilder to respond with certain strings to requests. Since you've set one up here, it probably makes more sense to use it rather than set req.responseText directly.

ZD-MichaelCroft: sweet.  will update that after work.

BadHorseMonkey: done

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -296,6 +301,34 @@ public void willNotRetrieveIfAlreadyBoughtFireworkHat() {
...
+      var builder = new FakeHttpClientBuilder();
+      builder.client.addResponse(200, html("request/test_desc_item_ring_2crs.html"));
+
+      var cleanups = new Cleanups(withHttpClientBuilder(builder), withItem(ItemPool.RING, 1));
+      try (cleanups) {
+        Modifiers mods = ModifierDatabase.getModifiers(ModifierType.ITEM, ItemPool.RING);
+        assertEquals(
+            0, mods.getDouble(DoubleModifier.MONSTER_LEVEL), "Value should not be set before desc");
+
+        String descId = ItemDatabase.getDescriptionId(ItemPool.RING);
```
Veracity0, line 317: How about just calling InventoryManager.checkRing()?
That will do the same thing - fetch and parse the desc_item - but you won't have to duplicate the implementation.

As a bonus, it increases coverage. :)

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
BadHorseMonkey: yep, that works.

BadHorseMonkey:
Hey reviewers,

Looking at my code, I notice that we don't call checkMods() when we remove items from storage, just on launch and when we refresh status.  This should affect all of the variable items.  Should we?  I think it should be a separate PR, but if you want it in this one, I can work on it.

Veracity0:
InventoryManager.itemAvailable would have checked that if InventoryManager.canUseStorage.
If you have disabled that, I suppose pulling from storage could call checkModes.
Or ResultProcessor.gainItem (which is called whenever an item arrives in inventory) could check for the ring (and other items that checkMods would check) and call the particular item-checking method from InventoryManager.
It does that for 1950 Vampire Vintner wine - which is a quest item, and thus cannot be pulled, but same idea.

I agree that could/should be a separate PR.

Veracity0 (approved):

BadHorseMonkey:
I'll take a look at it again after I ascend and it's back in storage.  Maybe I want to know the mods while it's still in storage so the the maximizer can find it.

PR #1531, opened 2023-02-13
Shadow Rift item drop penalty

gausie:
Shadow Rift divides item drop bonuses by 5 for all fights that take place there, which needs support by mafia. We previously only dealt in "doublers" which I have renamed to "multipliers" to reflect this.

Also added some spaded item drops to the shadow monsters, and some missing elemental attacks to the shadow prism


test/net/sourceforge/kolmafia/ModifiersTest.java
```diff
@@ -377,13 +378,24 @@ public void bendinHellAndDirtyPearStack() {
         assertThat(mods.getDouble(DoubleModifier.SLEAZE_SPELL_DAMAGE), equalTo(200.0));
       }
     }
+
+    @Test
+    void shadowRiftQuartersItemDrop() {
```
midgleyc, line 383: `fifths`?

midgleyc (changes requested):

test/net/sourceforge/kolmafia/ModifiersTest.java
gausie: whoopsy

midgleyc (approved):
I assume you have confirmed that this appears in `modtrace` as expected?

gausie:
> I assume you have confirmed that this appears in `modtrace` as expected?

I have! I even accidentally left some code to that effect in an earlier commit.

PR #1532, opened 2023-02-13
refactor: convert familiar training goals to enum

midgleyc:


src/net/sourceforge/kolmafia/textui/command/TrainFamiliarCommand.java
```diff
@@ -22,14 +23,14 @@ public void run(final String cmd, final String parameters) {
 
     String typeString = split[0].toLowerCase();
 
-    int type;
+    Goal type;
 
     if (typeString.equals("base")) {
```
gausie, line 28: Might as well do an enhanced switch here that returns null as default and then error out on null. Only if you feel like it

src/net/sourceforge/kolmafia/swingui/FamiliarTrainingFrame.java
```diff
@@ -1116,20 +1117,20 @@ private static void statusMessage(final MafiaState state, final String message)
...
-  private static void printFamiliar(final FamiliarStatus status, final int goal, final int type) {
+  private static void printFamiliar(final FamiliarStatus status, final int goal, final Goal type) {
     FamiliarData familiar = status.getFamiliar();
     String name = familiar.getName();
     String race = familiar.getRace();
     int weight = familiar.getWeight();
     String hope = "";
 
-    if (type == FamiliarTrainingFrame.BASE) {
+    if (type == Goal.BASE) {
```
gausie, line 1127: Enhanced switch here!

gausie (approved):

PR #1539, opened 2023-02-14
Sot's Parcel Tracking

jaadams5:
Added preference _sotParcelLocation.  It is set, if possible when visiting the Sot.  Remove the parcel from inventory when it is turned in for a milk cap.

Added tests.

There are probably better ways to implement these but this starts the discussion.

PlaceRequest has some lint but that will be submitted separately, probably after this is committed or abandoned.


jaadams5:
And if anyone can tell me an easy way to resolve the conflicts that would be helpful.  A side by side diff display that makes it easy to keep one side or the other and cat tag white space differences would be wonderful.

Veracity0:
> And if anyone can tell me an easy way to resolve the conflicts that would be helpful. A side by side diff display that makes it easy to keep one side or the other and can tag white space differences would be wonderful.

Somehow, you need to pull in later commits - which made drastic modifications to a routine you made minor modifications to - and simply merge the tiny handful of lines you changed that were conflictin.

Veracity0:
I submitted a merge commit which inserts the 3 new lines into the new-style  switch statement.
One hopes it is correct.

Veracity0:
It wasn't. Perhaps it is now?

Veracity0:
It is.

Basically, there was a big if-then elseif chain that you added a 3-line elseif block to.
Somebody changed that if-then-elseif chain int a switch statement.
You needed to take your 3 lines and make them into a case clause in that switch statement.

I did that.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
```diff
@@ -339,6 +342,34 @@ else if (responseText.contains("a giant jagged hole in the wall")) {
     SpadingManager.processPlace(urlString, responseText);
   }
 
+  private static void parseSotVisit(String responseText) {
+    String part1 = "What he means is that he lost something over in ";
```
gausie, line 346: Can you not just use a regex here? It's much simpler and we do it everywhere

jaadams5:
Thank you @Veracity0   I figured out what had happened and remembered how I dealt with it in my pre-Git days but I chose to go to bed and deal with it later.  But it looks correct, so thank you.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
jaadams5: We can.  I can't.  I had concerns about whether this code was even in the right place and whether the test structure did things in the current preferred way.  I figured I'd get that out of the way first.  Since it looks like this is going to "survive" I'll force myself to relearn enough regex to redo this.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
```diff
@@ -339,6 +344,31 @@ else if (responseText.contains("a giant jagged hole in the wall")) {
...
+    String location = "";
+    // What he means is that he lost something over in <location> and he'd like you to get it for
+    // him.
+    Pattern visit = Pattern.compile("(?<=something over in )(.*)(?= and he'd like)");
+    Matcher m = visit.matcher(responseText);
+    if (m.find()) {
+      location = m.group(0);
+    } else {
+      // He's just sitting there, waiting for you to bring his package back from <location>.
+      visit = Pattern.compile("(?<=back from )(.*)(?=\\.</td>)");
```
gausie, line 357: You don't need a positive lookbehind here, just lose the brackets. also pull out the regexes into statics!

jaadams5: See previous comment about me not being especially proficient with regex's.  Do you mean what I call parentheses when you say bracket?  Does your comment apply if  the interesting part of the response is 

`<tr><td>He's just sitting there, waiting for you to bring his package back from The Haunted Storage Room.</td></tr>`

with HTML before and after?  I had a simpler regex but I then had to find a period and throw it away and every after it.  Do you suggest making both of my pattern's static, final and then having a much larger scope?  Is compiling a pattern that expensive or are there other reasons you prefer that?

I'll gladly do what is asked but I do want to understand "why?" so there is at least a chance I will do things differently next time.  Thanks.

midgleyc: if the only possible message is that one, then:
```
back from (.+?)\\.</td>
```
should do it.

We may as well compile patterns we use more than once for efficiency resasons. If they're not static / final they'll be recompiled every time they get used.

jaadams5: The patterns are static final now.  The suggested changes are not working, or being made properly, so I need to try and figure out what isn't going on.

`Pattern.compile("back from (.+?)\\.</td>");`

did not work which is my understanding of the suggestion.

midgleyc: I've changed it to what I intended. I suspect you didn't change `group(0)` to `group(1)`. Group 1 is the first captured group, while group 0 is the entire matched string (which doesn't include the parts from lookaheads or lookbehinds).

jaadams5: Thank you.  I am somewhat chagrined that my biggest issue in simplifying the regex was using the wrong group.  Maybe I will remember that next time.  I tested this "live" before this last change and saw what I expected.  I'm going to consider this resolved.  Thanks again.

jaadams5:
Added a preference that is true if the parcel has been turned in today.  I think there is enough to script the parcel acquisition if anyone wants to.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
```diff
@@ -339,6 +348,29 @@ else if (responseText.contains("a giant jagged hole in the wall")) {
     SpadingManager.processPlace(urlString, responseText);
   }
 
+  private static void parseSotVisit(String responseText) {
+    String location = "";
```
Veracity0, line 352: I mean, whatever. I'd just use null and != rather than !xxx.equals(). Why force a method call?

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
-}
 
-// Generated with love by TestMe :) Please report issues and submit feature requests at:
-// http://weirddev.com/forum#!/testme
+  @Nested
+  class Speakeasy {
+    @BeforeEach
+    public void initializeCharPrefs() {
+      KoLCharacter.reset("PlaceRequestTestFakeSotVisit");
+      KoLCharacter.reset(true);
```
Veracity0, line 88: Why?

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+    }
+
+    @AfterEach
+    public void resetCharAndPrefs() {
+      KoLCharacter.reset("");
```
Veracity0, line 93: Why any of this?
Usually, the BeforeEach method would reset Preferences only.
We have many many examples of that.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+      KoLCharacter.reset(true);
+      KoLCharacter.setUserId(0);
+    }
+
+    @Test
+    public void itShouldGetParcelLocationFromFirstVisit() {
+      String prefName = "_sotParcelLocation";
+      assertEquals("", Preferences.getString(prefName), "Preference already set.");
```
Veracity0, line 101: What? Why are these tests not using withProperty and ensuring the property is set the way you expect, rather than guarding against leaks from other tests? You can't count on the order tests are run. They must be independent.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+      var req = new GenericRequest("place.php?whichplace=speakeasy&action=olivers_sot");
```
Veracity0, line 102: As coded, why are you making a GenericRequest?
You set a responseText and then remove it.
You never run the request.

We have many examples of setting up a fake http server and actually running it, but if you simply want to call a responsetext parser which is not an instance method, you are not using tge request.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot",
+          req.responseText);
+      assertEquals(
+          "The Haunted Storage Room", Preferences.getString(prefName), "Preference not set.");
+    }
+
+    @Test
+    public void itShouldGetParcelLocationFromSubsequentVisit() {
+      String prefName = "_sotParcelLocation";
```
Veracity0, line 113: Again, use a withProperty Cleanup. In all of these tests.

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +79,75 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
     assertEquals(
         -1, freePulls.indexOf(toolbelt), "toolbelt should not be in Free Pulls after TTT fades");
   }
-}
 
-// Generated with love by TestMe :) Please report issues and submit feature requests at:
-// http://weirddev.com/forum#!/testme
+  @Nested
+  class Speakeasy {
+    @BeforeEach
```
Veracity0, line 85: Why are these @BeforeEach and @AfterEach classes here?
They were copied right from the top of the file - where they apply to all tests, even Nested ones.

That said, they are dubious. They are commented:
```
  // These need to be before and after each because leakage has been observed between tests
  // in this class.
```

Before this, there was a single test in this class.
The correct fix would have been to use Cleanups.

Make the BeforeAll do a single KoLCharacter.reset.
Make the BeforeEach do a single Preferences.reset

Veracity0 (changes requested):

Veracity0:
OK. Your tests are overcomplicated. Here is how I would simplify one of them:

```
    @Test
    public void itShouldRemoveParcelWhenTurnedIn() {
      var cleanups = new Cleanups(withItem(ItemPool.THE_SOTS_PARCEL, 1),
                                  withProperty("_sotParcelReturned", false));
      try (cleanups) {
        var URL = "place.php?whichplace=speakeasy&action=olivers_sot";
        var response = html("request/test_visit_sot_to_return.html");
        PlaceRequest.parseResponse(url, response);
        assertEquals(0, InventoryManager.getCount(ItemPool.THE_SOTS_PARCEL));
        assertThat("_sotParcelReturned", isSetTo(true));
      }
    }
```
If you wanted to actually run the request, you could do this:
```
    @Test
    public void itShouldRemoveParcelWhenTurnedIn() {
      var client = builder.client;
      var cleanups = new Cleanups(withHttpClientBuilder(builder),
                                  withItem(ItemPool.THE_SOTS_PARCEL, 1),
                                  withProperty("_sotParcelReturned", false));
      try (cleanups) {
        client.addResponse(200, html("request/test_visit_sot_to_return.html"));
        var request = new GenericRequest("place.php?whichplace=speakeasy&action=olivers_sot");
        request.run()
        assertEquals(0, InventoryManager.getCount(ItemPool.THE_SOTS_PARCEL));
        assertThat("_sotParcelReturned", isSetTo(true));
      }
    }
```
There is another way to set up a FakeHttpClient.
```
    @Test
    public void itShouldRemoveParcelWhenTurnedIn() {
      setupFakeClient();
      var cleanups = new Cleanups(withItem(ItemPool.THE_SOTS_PARCEL, 1),
                                  withProperty("_sotParcelReturned", false),
                                  withNextResponse(200, html("request/test_dakota_1.html")));
      try (cleanups) {
        var request = new PlaceRequest("speakeasy", "olivers_sot");
        request.run()
        assertEquals(0, InventoryManager.getCount(ItemPool.THE_SOTS_PARCEL));
        assertThat("_sotParcelReturned", isSetTo(true));
      }
    }
```
That is popular with everybody except me.
I don't like it because it uses a hidden global FakeHttpClient.
That's why it needs to set responses in Cleanups.

Doing that preclude running tests in parallel
Considering that KoLmafia' globals already preclude that, meh, I guess.

gausie:
Fair point on the last one! We could maybe let you attach vars to the
parent instance of Cleanups or something somehow.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
jaadams5: You're right.  Using a string was leftover from when I thought I needed to compare the location just found to the current value of the preference.  I don't do that so yes, null is a better check.

jaadams5:
Thanks.  I never remember what tests implement the latest understanding of how to do things or what that understanding is.  So I copy/paste/edit a lot and in this case I started from a bad example.  But since I did so to solicit feedback and improve my understanding and/or memory there are several opportunities to improve things that I will take.  Stand by for another iteration and thank you.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
jaadams5: This wasn't a deliberate guard.  I thought of this test as testing a change in state and my generic method has been to check the old state, do something, check the state and confirm that it changed.  That said, I think reworking to use withProperty is a better way.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
     assertEquals(
         -1, freePulls.indexOf(toolbelt), "toolbelt should not be in Free Pulls after TTT fades");
   }
-}
 
-// Generated with love by TestMe :) Please report issues and submit feature requests at:
-// http://weirddev.com/forum#!/testme
+  @Test
```
Veracity0, line 81: Why did you un-Nest these tests?
It was much cleaner to have them in a Nested class just for whichplace=speakeasy

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+  public void itShouldGetParcelLocationFromFirstVisit() {
+    String prefName = "_sotParcelLocation";
+    String responseText = html("request/test_first_visit_sot_to_get_location.html");
+    var cleanups = new Cleanups(withProperty(prefName, ""));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
```
Veracity0, line 88: No. The URL passed around never includes the server.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+  }
+
+  @Test
+  public void itShouldGetParcelLocationFromSubsequentVisit() {
+    String prefName = "_sotParcelLocation";
+    var cleanups = new Cleanups(withProperty(prefName, ""));
+    try (cleanups) {
+      String responseText = html("request/test_next_visit_sot_to_get_location.html");
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
```
Veracity0, line 101: ditto

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+
+  @Test
+  public void itShouldRemoveParcelWhenTurnedIn() {
+    String prefName = "_sotParcelReturned";
+    String responseText = html("request/test_visit_sot_to_return.html");
+    var cleanups =
+        new Cleanups(withProperty(prefName, false), withItem(ItemPool.THE_SOTS_PARCEL, 1));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
```
Veracity0, line 115: ditto.

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+  }
+
+  @Test
+  public void itShouldDetectParcelAlreadyTurnedIn() {
+    String prefName = "_sotParcelReturned";
+    String responseText = html("request/test_visit_sot_parcel_done.html");
+    var cleanups = new Cleanups(withProperty(prefName, false));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
```
Veracity0, line 128: ditto

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
-// http://weirddev.com/forum#!/testme
+  @Test
+  public void itShouldGetParcelLocationFromFirstVisit() {
+    String prefName = "_sotParcelLocation";
+    String responseText = html("request/test_first_visit_sot_to_get_location.html");
+    var cleanups = new Cleanups(withProperty(prefName, ""));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
+      assertEquals(
```
Veracity0, line 89: Cleaner to use ```isSetTo(prefName, "The Haunted Storage Room")```

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+
+  @Test
+  public void itShouldGetParcelLocationFromSubsequentVisit() {
+    String prefName = "_sotParcelLocation";
+    var cleanups = new Cleanups(withProperty(prefName, ""));
+    try (cleanups) {
+      String responseText = html("request/test_next_visit_sot_to_get_location.html");
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
+      assertEquals(
```
Veracity0, line 102: Cleaner to use ```assertThat(prefName, isSetTo("The Haunted Storage Room"))```

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+  public void itShouldRemoveParcelWhenTurnedIn() {
+    String prefName = "_sotParcelReturned";
+    String responseText = html("request/test_visit_sot_to_return.html");
+    var cleanups =
+        new Cleanups(withProperty(prefName, false), withItem(ItemPool.THE_SOTS_PARCEL, 1));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
+      assertEquals(0, InventoryManager.getCount(ItemPool.THE_SOTS_PARCEL));
+      assertTrue(Preferences.getBoolean(prefName), "Preference not set.");
```
Veracity0, line 117: Cleaner to use ```assertThat(prefName, isSetTo(true))```

test/net/sourceforge/kolmafia/request/PlaceRequestTest.java
```diff
@@ -74,7 +77,56 @@ void itShouldSetTheToolbeltAsAFreePullInTTT() {
...
+
+  @Test
+  public void itShouldDetectParcelAlreadyTurnedIn() {
+    String prefName = "_sotParcelReturned";
+    String responseText = html("request/test_visit_sot_parcel_done.html");
+    var cleanups = new Cleanups(withProperty(prefName, false));
+    try (cleanups) {
+      PlaceRequest.parseResponse(
+          "http://server.fakepath/place.php?whichplace=speakeasy&action=olivers_sot", responseText);
+      assertTrue(Preferences.getBoolean(prefName), "Preference not set.");
```
Veracity0, line 129: Cleaner to use ```assertThat(prefName, isSetTo(true))```

Veracity0 (changes requested):

jaadams5:
Restored Nested.  Tweaked assertion style.  No fake servers in URL.  Using cleanups.

src/data/defaults.txt
```diff
@@ -2142,6 +2142,8 @@ user	_snowglobeDrops	0
 user	_snowSuitCount	0
 user	_sobrieTeaUsed	false
 user	_softwareGlitchTurnReceived	false
+user	_sotParcelLocation
+user    _sotParcelReturned  false
```
Veracity0, line 2146: Looking at this and the immediately preceding line, is there exactly one tab (and no spaces) between "user" and "_"?

Veracity0 (approved):
Much improved. Aside from my question about defaults.txt (which I can't answer for myself, since I'm reviewing on my phone, and the Github app is not always good with whitespace), I am satisfied and am approving.

Now, if there really is a whitespace issue, please fix. 😏

src/data/defaults.txt
jaadams5: I think there was and I tweaked it.

jaadams5:
I think I have addressed everything.

I also applied some lessons learned to what had previously been the only test in the file.  If I learned the wrong lessons or introduced more issues I'd rather just leave that alone in this PR.

Veracity0:
There is more that could be done in the other tests, but that is fine.

I told you that defaults.txt needed tabs, not spaces, so you adjusted the property that had a tab to agree with the one that had spaces.

I just submitted a fix for that into this PR.

You're good to go, I think.

Veracity0 (approved):

jaadams5:
> I told you that defaults.txt needed tabs, not spaces, so you adjusted the property that had a tab to agree with the one that had spaces.

That was my ide applying settings that I apparently have to restore every time I upgrade.  or something.

Thank you for your patience.

PR #1547, opened 2023-02-15
feat: add filter for modref

midgleyc:
It can be hard to remember all the types of modifier KoL offers (e.g. for a hat drop, there is item drop, gear drop and hat drop).

Add a filter for modref so you can type `modref @Drop` to get all the drop modifiers (and some other stuff where the name contains "drop", but this is good enough for me).

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/ModRefCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/ModRefCommand.java
@@ -12,6 +12,7 @@ import net.sourceforge.kolmafia.modifiers.DoubleModifier;
 import net.sourceforge.kolmafia.modifiers.Modifier;
 import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
 
 public class ModRefCommand extends AbstractCommand {
   public ModRefCommand() {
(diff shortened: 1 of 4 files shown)


gausie:
Maybe we want to use * and treat it as a wildcard instead? @ feels unintuitive to me. Great change though

midgleyc:
I picked `@` because `modref` already takes an item name with no extra flags and I figured `@` was the least likely character to appear in an item name.

My initial hope was that `modref Drop` could give you all the drops, without having to add wildcards. Because of the item name thing, you also can't provide more than one word (or I have to figure out how to do the flags for that, but I don't want to make it inconvenient to use).

gausie:
So why not use *?

Veracity0:
"Item name" thing? Example, please?

So, you are worried there might be an item name that starts with a *?

I think * is more intuitive than @.

src/net/sourceforge/kolmafia/textui/command/ModRefCommand.java
```diff
@@ -1,85 +1,111 @@
...
 import net.sourceforge.kolmafia.modifiers.DoubleModifier;
+import net.sourceforge.kolmafia.modifiers.Modifier;
 import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
 
 public class ModRefCommand extends AbstractCommand {
   public ModRefCommand() {
-    this.usage = " [<object>] - list all modifiers, show values for player [and object].";
+    this.usage = " [filter>] [<object>] - list all modifiers, show values for player [and object].";
```
Veracity0, line 19: &lt;filter&gt;

Veracity0 (dismissed):

src/net/sourceforge/kolmafia/textui/command/ModRefCommand.java
midgleyc: thanks, good spot

Veracity0 (approved):

PR #1548, opened 2023-02-15
fix: use getOrDefault on all turnsRemaining calls

midgleyc:
Should fix https://kolmafia.us/threads/trying-to-equip-a-sticker-from-the-cli-fails-and-prints-an-error.28638/


Veracity0 (approved):

gausie:
Yeah I also made this change in a local branch I stopped, wanting to try to
make a test case for it though

PR #1550, opened 2023-02-15
feat: aosol skills, effects, some items

midgleyc:
Throw a bunch of data into Mafia.


midgleyc:
Also added the path / class enum values. Images are still unknown as nobody has finished an ascension yet. Not sure if it's a proper avatar or if the class stays in aftercore, but we can change it if needed.

Veracity0 (approved):

midgleyc:
Folk have now finished ascensions -- but the images are broken on live for non-devs.

Need to wait for an update to fix.

PR #1551, opened 2023-02-15
validation for Shadow Rift adventuring

Veracity0:
1) When registering KoLAdventure recognizes adventuring locations (which now includes all the place.php Shadow Rift Ingresses), log that you are entering the Shadow Rift via the specific ingress. This used to be in PlaceRequest.registerRequest, but that is never called for these URLs any more, now that they are KoLAdventures.
2) canAdventure has tests for all 14 adventures in the Shadow Rift zone.
The one that is just a direct call to adventure.php requires that you have a non-empty shadowRiftIngress property. That means we had to have seen you go through a rift, but if you have never been through a rift this life, you'll get nothing but tumbleweeds, which is unlikely to be what you want.
3) Unit tests for canAdventure() using all 14 ways to go through a Shadow Rift
4) prepareForAdventure will plant enchanted bean for Beanstalk Shadow Rift, if necessary
5) prepareForAdventure will equip Continuum Transfunctioner for 8-Bit Realm Shadow Rift, if necessary
(It will get the Transfunction from the Crackpot Mystic, first, if necessary, but I have no test for that - because I DO have that test for the othe Pixel Realm adventuring locations, and it uses, literally, the identical code path.)


gausie (dismissed):
Looks great. FYI once the Shadow Rift "event" is over, these zones will remain permanently accessible in the AoSOL path.

Veracity0:
> Looks great. FYI once the Shadow Rift "event" is over, these zones will remain permanently accessible in the AoSOL path.

Thanks!

That's good to know!  You and I did a fair amount of work on this - as we would have, even if they were JUST and "event" - but it's nice to know they'll still be available.

By the way - I figured out how to create a test for rest of code which changes shadowRiftIngress.
It should be easy. I'll add it and then merge without further approval, since it's new test code.

PR #1553, opened 2023-02-16
Consumption notes for Vampire Vintner wine update

Veracity0:
When you log in, we look at your 1950 Vampire Vintner wine, if any.
When you acquire one, we do it again, since the modifiers change for each one.

Doing so updates the modifiers for that item.

With this PR, it now updates the Notes for that item to include the effect it provides and their modifiers.
That shows up in the Booze panel of the Item Manager.

I notice that the Maximizer recommends the wine for every one of the effects that 1950 Vampire Vintner wine is capable of providing. It SHOULD recommend only the mods for the effect that your currently available wine provides.

That's a bug - and should go in a new PR.


gausie:
Good spot on the other bug. I think this could use a quick test

Veracity0:
OK, I added a couple of tests:
- One for the InventoryManager method called when you logged in.
- One for ResultProcessor, when you gain a 1950 Vampire Vintner wine
The diff coverage is now 100%. Hopefully that's good enough? ;)

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -293,4 +295,88 @@ public void willNotRetrieveIfAlreadyBoughtFireworkHat() {
       }
     }
   }
+
+  @Nested
+  class VampireVintnerWIne {
```
jaadams5, line 300: WIne instead of Wine deliberate?

jaadams5 (dismissed):

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
Veracity0: Typo. Fixed.

PR #1557, opened 2023-02-17
Maximizer suggests vintner wine only if you have one with the correct effect

Veracity0:
My character was overdrunk when I coded this, so, although debug logging showed the right code paths being executed, I have not actually seen the end result in-game.

However, I wrote tests - and they all seem to indicate it will work, so I look forward to trying it live after rollover - using the Wine-Fortified wine I generated with Drunkula's wineglass while overdrunk. :)

I want to see the tests succeed here, so, I opened the PR.

Edit: it worked perfectly.

Maximize for init, it says to drink the wine for +80.
Maximize for other wine effects, they are not considered.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
@@ -577,6 +577,21 @@ public class Maximizer {
               continue;
             }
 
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks for testing!

PR #1560, opened 2023-02-18
Add spend Milk Cap to Ascension reminders

jaadams5:


midgleyc (approved):
Seems reasonable. It might be more helpful to link directly to the shop at `shop.php?whichshop=olivers`.

jaadams5:
>It might be more helpful to link directly to the shop at shop.php?whichshop=olivers.

Done, although untested with a real character yet.

midgleyc (approved):

PR #1563, opened 2023-02-19
update more aosol monsters

Veracity0:
Yendor gave us a bunch of Avatar of Shadows over Loathing monsters.
We are grateful. We took his data and integrated it.
Yendor then edited his post in-place to give us the rest.
I happened to notice it, and brought the new data in

I want confirmation that the name "Dark Noël" contains UTF-8 characters, as opposed to HTML character entities, but other than that, we're cool.

I think in addition to path monster maps, we need class monster maps.

I did that. Observations:

- The only place these are used is in ASH functions:
get_monsters(location) returns an array: index -> monster
get_location_monsters(location) returns a map: monster -> boolean
Both of these functions used to map monsters through the path map. Now they also run them through the class map.
- get_monster_mapping(string pathName) returns monster -> monster
That could be upgraded to take either a "path" object or a "class" object.
For now, I left it untouched.
- It'd be nice if the Area Combat Data as displayed in the GUI also took that into account, so you could look in Throne Room and see the actual version of the Goblin King that is there.
- The Naughty Sorceress' Chamber lists only Naughty Sorceress.
It omits Naughty Sorceress (2) and Naughty Sorceress (3).
Fair, I guess, since not every path has all of those. Unfortunately, AoSoL does have at least two. And even though I have class mappings for both, you don't get to see it.
- Speaking of which, do the "true form" versions replace (2) or (3)? Or what? Do you need a wand? I'lll see tomorrow.

Improving how we handle combat data when there are mapped monsters would be  a different project.

I also think we could have all three sorceress forms in combats.txt for the Sorceress's Chamber and map them to null (or something) if they are not present in a given path/class. All three with a 100% chance? Tricky.
AreaCombatData would map monsters and omit those that map to null.
That would also be a different project.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -8359,12 +8359,16 @@ public abstract class RuntimeLibrary {
     return DataTypes.makeMonsterValue(MonsterStatusTracker.getLastMonster());
   }
 
-  private static MonsterData mapMonster(MonsterData mon, Map<MonsterData, MonsterData> mapping) {
-    if (mapping == null) {
+  private static MonsterData mapMonster(
+      MonsterData mon,
+      Map<MonsterData, MonsterData> classMapping,
+      Map<MonsterData, MonsterData> pathMapping) {
+    if (classMapping == null && pathMapping == null) {
       return mon;
     }
-    MonsterData mapped = mapping.get(mon);
-    return mapped != null ? mapped : mon;
+    MonsterData classMapped = classMapping.get(mon);
+    MonsterData pathMapped = pathMapping.get(mon);
+    return classMapped != null ? classMapped : pathMapped != null ? pathMapped : mon;
   }
 
   public static Value get_monsters(ScriptRuntime controller, final Value location) {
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -8359,12 +8359,16 @@ public static Value last_monster(ScriptRuntime controller) {
     return DataTypes.makeMonsterValue(MonsterStatusTracker.getLastMonster());
   }
 
-  private static MonsterData mapMonster(MonsterData mon, Map<MonsterData, MonsterData> mapping) {
-    if (mapping == null) {
+  private static MonsterData mapMonster(
+      MonsterData mon,
+      Map<MonsterData, MonsterData> classMapping,
+      Map<MonsterData, MonsterData> pathMapping) {
+    if (classMapping == null && pathMapping == null) {
```
midgleyc, line 8366: if either of these two are null, it will still NPE.

I would check for a path mapping first, as almost all the time there won't be a class mapping.

Veracity0: Right. Fixed.

midgleyc (approved):

PR #1567, opened 2023-02-20
AoSoL has points per class, like AoWoL

Veracity0:


midgleyc (approved):
Using a shadowy cheat sheet should also increase all of these by 1.

It may want to be a different preference too, as others can be calculated from ascension history. Don't know how WOL does it.

Veracity0:
Whoah. This is the first I heard of a "shadowy cheat sheet".
OK. AWoL had something like that, too:
```
8986	Memories of Cow Punching	854594023	pamphlet.gif	usable	t	0
8987	Memories of Beanslinging	239656627	pamphlet.gif	usable	t	0
8988	Memories of Snake Oiling	613587003	pamphlet.gif	usable	t	0
```
Which could be bought and used only while you were a particular character class.

We have 
```
user	awolDeferredPointsBeanslinger	0
user	awolDeferredPointsCowpuncher	0
user	awolDeferredPointsSnakeoiler	0
```

I guess using the cheat sheet increments whatever your current class is?

midgleyc:
By the description, I think it's +1 point to all classes.

Veracity0:
Gelatinous Noob also had Gummi-Memory, which sent you to a choice to confirm you wanted to use it.
I guess I'll buy a shadowy cheat sheet and try it out.

It's not as if it will really "spoil" anything for me; with the model train set, I am flyering monsters for the island war on day 1, anyway, so I easily had all the skills quickly.

Veracity0:
I bought and used one. It just says "You learn some advanced shadowy cheating techniques!" and is consumed.

midgleyc (approved):

PR #1572, opened 2023-02-20
More Shadows Over Loathing path teaks

Veracity0:
1) Avoid NPE in Valhalla due to not having an Ascension Class
2) Fondeluge gives 50 turns of Everything Looks Yellow. Disable the Stationary Button for that skill when that effect is active
3) AoSoL skills are not permable
4) No Relay Browser Boss warnings for AoSoL


midgleyc (approved):
Thanks!

I think perhaps `KoLCharacter.getPath()` (and `KoLCharacter.getAscensionClass()`) should never return `null`. The values are used by `my_path()` and `my_class()`, and if a new class or path is released that Mafia doesn't know about, these functions crash. It would be nicer to have an "unknown" value for these cases.

PR #1574, opened 2023-02-21
Add daycount function

libraryaddict:
Mafia does not natively present a way for the player to determine what day it is, and how many days its been since X

We have the timestamp which gives us the year, month, day. But that isn't programmatically easy to parse because months vary from day to day. We could parse the current time, but that doesn't tell us about rollover.

Sure, we can already do this by doing some fancier date parsing and whatnot. But why put in all that work when there's a simple solution?

Instead, it's nicer to have a simple incremental counter that goes up by one every rollover.
Which kol happens to tell us.

As we already have `my_daycount()` which tells us how many days in this ascension, adding `daycount()` which tells us how many days kol has been a thing seems to make sense.


gausie:
You can use date_to_timestamp or now_as_timestamp, subtract and divide by 60*60*24? What does this do that that function does not?

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -237,6 +237,7 @@ public enum Gender {
   private static int turnsPlayed = 0;
   private static int currentRun = 0;
   private static long rollover = 0;
+  private static int globalDaycount = 0;
```
gausie, line 240: While I appreciate that this is done everywhere, I don't love it being a static property.

libraryaddict: I don't either, but other than a large refactor to change the "normal way", doesn't seem like there's a real alternative.

Probably best to be discussed as its own change.

gausie: I wasn't suggesting changing it everywhere else I was suggesting doing it differently here.

gausie (approved):

PR #1575, opened 2023-02-22
Add session_log(int) : string

libraryaddict:
This introduces a new function `session_log(int)` and `session_log(name, int)` which differs from `session_logs()` in that it will retrieve a single session log.

Another way it differs is that it starts from 0, where `session_logs` starts from 1, meaning how many logs `session_logs` will return.

This could be changed, but my train of thought is that `session_logs` asks for the number of days / session logs to return.

`session_log` is asking how many days ago should it go to find the session log, where 0 is obviously today as 1 day ago is not today.

So the main use case of this PR is that you can retrieve a single session log from a specified day, without parsing every session log up to said day.

Such as a script that parses a run and compares it against another run, if you want to do a large number of sessions you're not forced to load every file into memory at once.

Of course, `session_logs` does support adding a calendar date to start from, but it is far simpler to just ask for the session log for a specified `day ago`.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -668,6 +668,12 @@ public abstract class RuntimeLibrary {
     params = new Type[] {};
     functions.add(new LibraryFunction("moon_light", DataTypes.INT_TYPE, params));
 
+    params = new Type[] {DataTypes.INT_TYPE};
+    functions.add(new LibraryFunction("session_log", DataTypes.STRING_TYPE, params));
+
+    params = new Type[] {DataTypes.STRING_TYPE, DataTypes.INT_TYPE};
+    functions.add(new LibraryFunction("session_log", DataTypes.STRING_TYPE, params));
+
     params = new Type[] {};
     functions.add(new LibraryFunction("stat_bonus_today", DataTypes.STAT_TYPE, params));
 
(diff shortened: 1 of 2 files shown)


jaadams5:
I'm not following your use case.  Specifically I cannot find an ash function named session_logs that takes no parameters so I don't know how the new work is supposed to be different.

The two parameter version of session_logs accepts a base date and a count.  So you should be able to call that and get exactly one log for a specific date.  

I'm not understanding the new functionality here.

gausie:
Looks like it avoid you having to subtract dates to get your desired date.

libraryaddict:
Hmm, I don't think there's actually an easy way to understand how to use `session_logs(name : string, date : string, days : number)`

It looks like `date` is supposed to be `yyyyMMdd`

So if I want to get the session log from say, 60 days ago. And this is a hardcoded constant.
Also assume I'm using ash.

So the recommended way to do this is to get the date by

```
int date = date_to_timestamp("yyyyMMdd", today_to_string());
date -= 60 * 24 * 60 * 60 * 1000;
string datestr = timestamp_to_date(date, "yyyyMMdd");
string session = session_logs("Irrat", datestr, 1);
```

Instead of

`session_logs("Irrat", 60)`

Mind you, it took a little longer than it should've to figure that out. Because I got the date formatting wrong, then the wrong ash functions.

I'm not saying this is a solid reason to add session_log, the whole PR is more about making it easier. But anyone that gets this far is probably wise enough to know how to manipulate a day.

Just the short code snippet seems about the right way to be handling this date, and it feels a bit clumsy doing it repeatably. That said, I'm happy enough to close this PR.

Perhaps editing kolmafia's wiki with this code snippet as session_logs(date) isn't actually written down.

jaadams5:
Feel free to edit the wiki.  The people who were passionate about keeping it up to date have faded from the game.  You might also be interested in the discussion at https://kolmafia.us/threads/ability-to-specify-a-range-of-session-logs-to-open.22993/

PR #1576, opened 2023-02-22
Equip a club even if tiebreaker is used

eegeeZA:
This test will fail until the maximiser is adjusted to use explicit equipment specified. This should still work even if a tiebreaker argument is added.

_Logging this in lieu of creating a bug report on the forums._


eegeeZA:
If I get some time over the weekend, I can look at making the adjustment to the maximiser logic.

PR #1577, opened 2023-02-23
fix: check familiars even in avatar paths

midgleyc:
I haven't actually ascended into a path that can't use familiars to test this, but I think it should be correct.

As mentioned in #1081, there is benefit to owning familiars, even if you can't use them. This applies also to paths where you can't see your terrarium (but can `/goto` it).

Most relevant to most users will be the ability to drink the stillsuit distillate through Mafia, having equipped it onto a familiar (using `/goto familiar.php` or Mafia itself).

Pokefam replaces the whole terrarium screen. I don't know how Quantum works in detail, but I assume you only "own" the familiar you have out.


Veracity0 (approved):
II think KoL's only visible link to the terrarium is in the charpane.
On a path where you can't equip familiars, it will not be there.
I've never used the /goto chat command, so was unaware you could use it to go there.
And of course, KoLmafia is happy to inject any URL it wants.

It'd be nice to have some actual testing that it does, in fact, work in non-Pokefam/non-Quantum paths, but I'm not going to hold you to that right now.

Although - somebody alerted you to this? in a no-familiar path and wants to access a stillsuit?

I'll approve, but can you get some feedback that it actually behaves as expected?

midgleyc:
I have an on-ascension script that equips the stillsuit to a familiar giving the appropriate stat bonus, e.g. `equip($item[tiny stillsuit], $familiar[pet rock]);`. This works fine (by the look of it, Mafia only checks `ownedFamiliars` when parsing the response). However, `familiarSweat` remains at 0 and `drink stillsuit distillate` fails due to the lack of sweat.

`/goto` just takes you directly to a given PHP page. It's useful for chat macros or top menu icons (which can be chat commands). For example, I have a `/sales` macro that calls `/goto backoffice.php?which=3`, and greycat's old `/repeat` macro calls `/goto '?'':function() {for(i=$1;i;i--) {submitchat('$2-');} throw{};}();z='`.

midgleyc:
I went into Boris, and `familiarSweat` is ticking up like I expect.

PR #1579, opened 2023-02-23
getawayCampsiteUnlocked should not change on visiting locked woods

s-k-z:
The Getaway Camp is accessible even when the distant woods are locked, therefore there's no reason for KoLmafia to assume it locked when the woods are.

These changes ensure the `getawayCampsiteUnlocked` property remains unchanged after visiting the woods through either a PlaceRequest and GenericRequest.


Veracity0 (approved):

gausie:
I think this needs to be accompanied with the correct check - i.e. what
does it look like when you try to visit the getaway campsite and actually
don't have it

s-k-z:
Those already exist. See https://github.com/kolmafia/kolmafia/blob/main/test/net/sourceforge/kolmafia/request/CampAwayRequestTest.java

gausie:
Then we in business boys.

PR #1580, opened 2023-02-24
Find GitHub hosted scripts that are updated by SVN

jaadams5:
[This post](https://kolmafia.us/threads/implement-warnings-when-using-svn-with-github-projects.28617/#post-171266) observes that support for SVN access to GitHub will be eliminated.  Since the gCLI command already traversed the local svn directory looking for "duplicates" I extended it to check for repositories were on GitHub.  (The installation protocol determines whether the local repository is under svn or git).

At this point the user action for such scripts is unclear but at least this indicates action will be needed.

I'm on the fence about whether I should print a partial path name instead of just the file name but since I don't have time to do that now...

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
@@ -50,6 +50,7 @@ import net.sourceforge.kolmafia.request.MonsterManuelRequest;
 import net.sourceforge.kolmafia.request.StorageRequest;
 import net.sourceforge.kolmafia.request.StorageRequest.StorageRequestType;
 import net.sourceforge.kolmafia.request.ZapRequest;
+import net.sourceforge.kolmafia.scripts.svn.SVNManager;
 import net.sourceforge.kolmafia.session.DisplayCaseManager;
 import net.sourceforge.kolmafia.session.EquipmentManager;
 import net.sourceforge.kolmafia.session.InventoryManager;
@@ -62,6 +63,8 @@ import net.sourceforge.kolmafia.utilities.WikiUtilities;
 import net.sourceforge.kolmafia.utilities.WikiUtilities.WikiType;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.tmatesoft.svn.core.SVNException;
+import org.tmatesoft.svn.core.SVNURL;
 import org.w3c.dom.Document;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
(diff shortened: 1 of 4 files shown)


jaadams5:
Test failure is interesting.  Making Draft until I resolve it.

jaadams5:
The tests are failing because I wasn't mocking everything the new code needed in a "SVN file system".   I'm going to investigate but hope that the disabled tests are not the sow stopper :-)

jaadams5:
@midgleyc - I'm looking at GitManagerTest and if I understand it you are actually checking out a repository.  That isn't going to help me unless we want a test that requires SourceForge to be available.  Any ideas or suggestions?  Should I work harder on mocking a file system or is there some other approach I should switch to?  I also ran GitManagertest on Windows and it worked once.  Was it intermittent or should it be reenabled?

midgleyc:
> I'm looking at GitManagerTest and if I understand it you are actually checking out a repository.

Yes, I am actually checking out a repository. I think a test that requires SourceForge to be available is fine. The GitManagerTest does checkout some repositories using SVN (to test the dependencies feature).

The reason why I'm checking out an actual repository is to do with confidence. In my mind, the purpose of a test is to make you confident that the feature you've written works. If I mock responses, then either:
* the code works
* my responses are wrong in some way, and the code appears to work although it actually doesn't
and I don't really have any way of knowing which it is.

One thing I considered and would have liked to do was cloning a local repository, instead of an networked one. However, JGit returned an error when trying to clone a `file://` repository. Whatever we use for SVN might work doing this?

>  I also ran GitManagertest on Windows and it worked once.

GitManagerTest [fails intermittently on Windows when deleting repositories](https://github.com/kolmafia/kolmafia/blob/05b4d72636e011a81fb79ef6843f81db2da847ea/src/net/sourceforge/kolmafia/scripts/git/GitManager.java#L352). The exact error conditions are:
* Windows fails to remove a file in `.git/pack/`, saying that it's in use
* Moving the file to `%TMP%` also fails because `%TMP%` is on a different network drive to Mafia
It's fine on Linux because Linux can remove files which are being used. `%TMP%` being on a different network drive appears to be true on GitHub CI.

> [actual PR]

I think given that you've had to disable some tests (or will have to extensively change them), this command should be distinct from the existing command checking for duplicates. Additionally, `checkrepo` works on the file level, whereas a command checking for SVN repos installed from GitHub would be more useful if it worked from the repo level -- as in, it told you which repos were installed using SVN, instead of which files belong to repos installed using SVN.

jaadams5:
Thanks.

My philosophy on testing is that a test must be deterministic and repeatable.  That certainly cannot be guaranteed if the test accesses resources not 100% controlled by the environment.   But I also recognize that attitude is rooted in an environment where most tests just tested a class (or "unit") and when a test on checked in code failed the CEO got an email.  So my first approach is not going to involve a real network call.  I do agree if the role of a test is to generate confidence and the test code infrastructure is more massive than the code being tested then avoiding the network brings no benefits.  At some point needing to "test the test" is not a good thing.  

So I am going to close this once I know you have seen the response.

I am going to make a new command so that existing tests work unchanged.  I am going to defer automated testing of the new command because it is primarily a diagnostic (and not something that effects game state).  As a diagnostic it should be unnecessary after January 2024 since no new instances of SVN access to GitHub will be allowed or created.  Since there are tests in DebugDatabaseTest that are disabled because they access an external resource I will at least raise the question of whether they should be enabled.

Thank you.

midgleyc:
I think a manual test is perfectly reasonable. I had `GitManagerTest` as an automated integration test because I didn't want to do a manual test every time I made a change.

I think the ideal version wouldn't involve a network call -- if you could clone a local repo that would be ideal (create it with JGit + appropriate files, clone it, delete it after the test). As that didn't work I think I'd be down to "docker container that serves a real git repo in a way JGit accepts". Once you have the testing infrastructure in place it's worth it but it can be a right slog to get to that point.

jaadams5:
See #1582.

PR #1581, opened 2023-02-25
return 64 bit Long to JavaScript so that we don't get weird 2s complement values in large Ints

BadHorseMonkey:
This is a patch to change the ValueConverter.asJava method to return BigInteger instead of long (only if Math.abs(value) is greater than Integer.MAX_VALUE).  Rhino/Javascript is interpreting that as a 2s complement value and returning negative numbers.  I'm seeing this specifically in nowToInt(), where the value is coming back as a negative number, since EpochTime in milliseconds is a 42 bit number.

[image](https://user-images.githubusercontent.com/58645293/221340388-2406803a-8b22-47d1-841c-78f307c03300.png)

This patch causes that to work as expected. I included a test for nowToInt() and a test for Value.asJava().  I didn't find any other places where I saw any difference, although scripts that rely on bad values may now suddenly have correct values.


BadHorseMonkey:
awww, nope!  This breaks Garbo.

gausie:
What does it do?

BadHorseMonkey:
> What does it do?

This: 
> garbo
JavaScript error: Cannot convert BigInt to a number
at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:10945
at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:23470 (__webpack_require__)
at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:23525
at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:1

I'm currently thinking about only making it a BigInteger in asJava() when it's above 32bits.

But I don't know if that's the best answer.

BadHorseMonkey:
> > What does it do?
> 
> This:
> 
> > garbo
> > JavaScript error: Cannot convert BigInt to a number
> > at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:10945
> > at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:23470 (**webpack_require**)
> > at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:23525
> > at file:/Users/mcroft/Library/Application%20Support/KoLmafia/scripts/garbage-collector/garbo.js:1
> 
> I'm currently thinking about only making it a BigInteger in asJava() when it's above 32bits.
> 
> But I don't know if that's the best answer.

This is what I've got, and it works for Garbo and nowToInt(), but I'm still leery of it.  
 public Object asJava(Value value) {
...else if (value.getType().equals(DataTypes.INT_TYPE)) {
      if (value.contentLong > Integer.MAX_VALUE) {
        return BigInteger.valueOf( value.contentLong );
      }
      return value.contentLong;

gausie:
The issue is that JS numbers can between -(2^53 – 1) and 2^53 – 1 before needing to go to BigInt

BadHorseMonkey:
> The issue is that JS numbers can between -(2^53 – 1) and 2^53 – 1 before needing to go to BigInt

OK, so maybe some variant on my idea above Is correct.

```
if (abs(value.contentLong) > Integer.MAX_VALUE) {
return BigInteger.valueOf( value.contentLong );
}
return value.contentLong;
```

But it still might fail if something got a BigInt and was expecting a Long.  It would not-fail before, but would have the wrong value (this was the nowToInt() problem).  

And the function is nowToInt(), so it seems to be wrong to return a non-Int...

BadHorseMonkey:
Added test for asJava with BigInt and Integer

gausie:
> > The issue is that JS numbers can between -(2^53 – 1) and 2^53 – 1 before needing to go to BigInt
> 
> OK, so maybe some variant on my idea above Is correct.
> 
> ```
> if (abs(value.contentLong) > Integer.MAX_VALUE) {
> return BigInteger.valueOf( value.contentLong );
> }
> return value.contentLong;
> ```
> 
> But it still might fail if something got a BigInt and was expecting a Long. It would not-fail before, but would have the wrong value (this was the nowToInt() problem).
> 
> And the function is nowToInt(), so it seems to be wrong to return a non-Int...

Hmm I have an issue here with things becoming randomly polymorphic. I'm not sure what the solution is but I don't want to have to type everything returning a number as `number | bigint`. Also note that `Integer.MAX_VALUE` in Java is much smaller than `Number. MAX_SAFE_INTEGER` in JS

Veracity0:
> Also note that `Integer.MAX_VALUE` in Java is much smaller than `Number. MAX_SAFE_INTEGER` in JS

These are the fields in a Value object:
```
  public Type type;

  public long contentLong = 0;
  public String contentString = null;
  public Object content = null;
```

Long.MAX_VALUE is therefore the largest "int" in a script.

It's a shame that a BigInt is not a Number.

Veracity0:
Huh. Does this mean that as-currently-coded, an ASH script that depends on 64 bit ints cannot be reproduced in JS?

BadHorseMonkey:
I am confused enough that I'm starting a thread on the forum.  I think I'm addressing this in the wrong place.  Also now_to_int is returning a 41 bit not-an-int and ASH doesn't blink.

Veracity0:
Executive summary:

An ASH "int" is a Java "long".
An ASH "float" is a Java "double".

The ASH Wiki claims otherwise, but it is out of date.

BadHorseMonkey:
Aha!  much clarity added. Thank you, @Veracity0 !  

This is a new, more targeted fix and a totally re-written test.  It should address @gausie 's concerns as well.
I will also fix the wiki, if no one else gets to it.

jaadams5 (approved):
I'll approve since it addresses the issue as I now understand it.  I might wait, however, to see if @Veracity0 or @gausie have an opinion before actually pushing this.

test/net/sourceforge/kolmafia/textui/javascript/AshInteropTest.java
```diff
@@ -71,4 +72,25 @@ void getClassFromId() {
...
+    var js = new JavascriptRuntime("nowToInt()");
+    assertNotNull(js, "JavascriptRuntime returned as null.");
+    Value ret = js.execute(null, null, true);
+    assertNotNull(ret, "Javascript execute returns null instead of a result to be tested.");
+    long retVal = Long.parseLong(ret.toString());
+    assertTrue(retVal >= 0, "Now should not be negative.");
+
+    long javaNow = Instant.now().toEpochMilli();
+    long diff = Math.abs(javaNow - retVal);
+    // there's no asertCloseTo() ...
```
midgleyc, line 87: There is `assertThat(x, closeTo(y, z));`, though.

If you search the codebase for `closeTo` you'll find some tests using it.

BadHorseMonkey: I do see that.  But the only implementation we have is for doubles, so I'd have to worry about that conversion.

And when I look at the implementation from another project of `closeToInteger()`, it's the same function I have (abs (val-val) > margin).

Do you think I should use closeTo()?  It's very readable, that's for sure.

midgleyc: oh right, I didn't see what you were testing.

Can you use `dateToTimestamp` with a fixed date instead? That would obviate the need for a "close to".

BadHorseMonkey: that wouldn't really test nowToInt(), though.  I'm open to changing it, just not sure it's viable.

midgleyc: It wouldn't, but your change is "timestamps that are too high are correctly converted to longs, not negative numbers" -- nothing specific about `nowToInt()`.

e.g. at the moment we have
> > js print(dateToTimestamp("yyyy-MM-dd", "2023-03-03"));
>
> -1530612736

but this should not be negative.

gausie:
Haha @heeheehee-kolmafia said "resolve #this-pr's-id" in #1597 and it closed this PR. Re-opened!

test/net/sourceforge/kolmafia/textui/javascript/AshInteropTest.java
BadHorseMonkey: well, it specifically came up in the context of nowToInt, but you're right.  The fix is larger in scope than dates

> > js print(dateToTimestamp("yyyy-MM-dd", "2023-03-03"));

1677823200000
Returned: null

BadHorseMonkey: done.

midgleyc (approved):
Thanks!

BadHorseMonkey:
> Thanks!

Thanks for helping make my test better!

PR #1582, opened 2023-02-25
Find GitHub hosted repositories that are updated by SVN (Take 2)

jaadams5:
A new command "checksvngit" that scans the local SVN directory for repositories that are hosted on GitHub.  These repositories will not updated after January 2024.

Test code for this is deliberately not included because this could be removed from the codebase circa January 2024 as being no longer necessary,  The complexity of mocking a repository seems like it is more effort than it is worth especially when "testing the test" might be appropriate.  Checking out a repository from GitHub is feasible but if the test checks out the repository via SVN the test will fail in January 2024.  If it is checked out via git then the test setup will still have to provide information that would be created by SVN.

See #1580 for some discussion that lead to this approach.

Sample Output
Local installation of coandco-mafia-NewYou-trunk uses SVN to update from GitHub.
Local installation of Ezandora-Choice-Override-branches-Release uses SVN to update from GitHub.
Local installation of Ezandora-Comb-Beach-trunk-Release uses SVN to update from GitHub.
Local installation of Ezandora-FantasyRealm-branches-Release uses SVN to update from GitHub.
Local installation of Ezandora-Gain-branches-Release uses SVN to update from GitHub.
Local installation of Ezandora-PirateRealm-trunk-Release uses SVN to update from GitHub.
Local installation of Ezandora-Voting-Booth-trunk-Release uses SVN to update from GitHub.
Local installation of gausie-excavator-trunk-RELEASE uses SVN to update from GitHub.
Local installation of matt-chugg-manuel_progress.git-trunk uses SVN to update from GitHub.
Local installation of phulin-shorts-ui-branches-master-build uses SVN to update from GitHub.
Local SVN repos scanned for possible GitHub access via SVN.


midgleyc (approved):
Seems reasonable.

PR #1583, opened 2023-02-26
refactor: convert confirms to enum

midgleyc:
These are "enum-like", in that they're all alike, and the numbers don't matter, only that they're distinct.

Generate them automatically instead. Most numbers do change, but the specific values don't matter.

This should make adding new ones slightly easier (as now you don't have to care about the string value).

### DIFF
--- a/src/net/sourceforge/kolmafia/request/RelayRequest.java
+++ b/src/net/sourceforge/kolmafia/request/RelayRequest.java
@@ -116,37 +116,37 @@ public class RelayRequest extends PasswordHashRequest {
   public static String redirectedCommandURL = "";
 
   public enum Confirm {
-    CONFIRM_COUNTER,
-    CONFIRM_MCD,
-    CONFIRM_FAMILIAR,
-    CONFIRM_RECOVERY,
-    CONFIRM_SORCERESS,
-    CONFIRM_WOSSNAME,
-    CONFIRM_TOKENS,
-    CONFIRM_SEAL,
-    CONFIRM_ARCADE,
-    CONFIRM_KUNGFU,
-    CONFIRM_POOL_SKILL,
-    CONFIRM_WINEGLASS,
-    CONFIRM_COLOSSEUM,
-    CONFIRM_GREMLINS,
-    CONFIRM_HARDCOREPVP,
-    CONFIRM_DESERT_UNHYDRATED,
-    CONFIRM_MOHAWK_WIG,
-    CONFIRM_CELLAR,
-    CONFIRM_BOILER,
-    CONFIRM_DIARY,
-    CONFIRM_BORING_DOORS,
-    CONFIRM_SPELUNKY,
-    CONFIRM_ZEPPELIN,
-    CONFIRM_OVERDRUNK_ADVENTURE,
-    CONFIRM_STICKER,
-    CONFIRM_DESERT_OFFHAND,
-    CONFIRM_MACHETE,
-    CONFIRM_RALPH,
-    CONFIRM_RALPH1,
-    CONFIRM_RALPH2,
-    CONFIRM_DESERT_WEAPON;
+    COUNTER,
+    MCD,
+    FAMILIAR,
+    RECOVERY,
+    SORCERESS,
+    WOSSNAME,
+    TOKENS,
+    SEAL,
+    ARCADE,
+    KUNGFU,
+    POOL_SKILL,
+    WINEGLASS,
+    COLOSSEUM,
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -869,15 +876,15 @@ public boolean sendBreakPrismWarning(final String urlString) {
       // If you don't have any coins, nothing to spend
       if ( InventoryManager.getCount( ItemPool.COIN ) > 0 )
       {
-        if ( this.getFormField( CONFIRM_RALPH2 ) == null )
+        if ( this.getFormField( Confirm.CONFIRM_RALPH2 ) == null )
```
Veracity0, line 879: Why is this not Confirm.RALPH2?

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -869,15 +876,15 @@ public boolean sendBreakPrismWarning(final String urlString) {
...
         {
           StringBuilder warning = new StringBuilder();
           warning.append( "When you stop being a plumber, you will lose access to the Mushroom District." );
           warning.append( " Since you have unspent coins, perhaps you would like to visit some shops?" );
           warning.append( " If you are sure you don't want to spend your coins, click the icon on the left. " );
           warning.append( " If you wish to visit the Mushroom District, click on icon on the right." );
           this.sendOptionalWarning(
-            CONFIRM_RALPH2,
+            Confirm.CONFIRM_RALPH2,
```
Veracity0, line 887: Ditto.

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -1240,7 +1251,7 @@ private boolean sendHardcorePVPWarning(final String urlString) {
     }
 
     // Don't remind a second time in a session if you decide not to do it.
-    if (urlString.contains(CONFIRM_HARDCOREPVP)) {
+    if (urlString.contains(Confirm.HARDCOREPVP.toString())) {
```
Veracity0, line 1254: Curious why it is necessary to explicitly call toString here.

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -2602,7 +2613,7 @@ private boolean sendSorceressWarning(final String urlString) {
       return false;
     }
 
-    if (urlString.contains(CONFIRM_SORCERESS)) {
+    if (urlString.contains(Confirm.SORCERESS.toString())) {
```
Veracity0, line 2616: Why explicit toString()?

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -2830,34 +2841,37 @@ private boolean sendSpelunkyWarning(final KoLAdventure adventure) {
...
       return true;
     }
 
     return false;
   }
 
-  public void sendGeneralWarning(final String image, final String message, final String confirm) {
-    this.sendGeneralWarning(image, message, confirm, null, false);
+  public void sendGeneralWarning(final String image, final String message, final Confirm confirm) {
+    this.sendGeneralWarning(image, message, confirm.toString(), null, false);
```
Veracity0, line 2861: OK. These are all turning into calls on the 5-argument version of sendGeneralWarning.
Why not change the third argument of that from ```String confirm``` to ```Confirm confirm```?

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: it's commented-out code, so didn't get picked up by the refactor. I'll adjust it.

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: `Confirm` doesn't implement `CharSequence`, so it has to be converted to a string. I could implement it, but I thought it was simpler to just convert it.

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: I thought the last `sendGeneralWarning` was implemented on `GenericRequest`, but I was wrong. Change made.

Veracity0 (approved):

PR #1585, opened 2023-02-27
fix: break early when iterating over modifiers if matched

midgleyc:
Throw exceptions instead of silently continuing if error.


Veracity0 (approved):
Whoa. A label. I don't think I've ever seen this syntax before.
Well, it's OK.

midgleyc:
I don't think I've ever used it before. I think it's more elegant than a `boolean matched = false;`, then checking the match after every loop.

PR #1588, opened 2023-03-01
feat: add `equipped_amount(item, boolean)`.

midgleyc:
Takes a boolean parameter on whether to check all familiars (including those not currently out).

Works like `available_amount`, but just the equipped bit.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -71,7 +71,6 @@ import net.sourceforge.kolmafia.combat.CombatActionManager;
 import net.sourceforge.kolmafia.combat.Macrofier;
 import net.sourceforge.kolmafia.combat.MonsterStatusTracker;
 import net.sourceforge.kolmafia.equipment.Slot;
-import net.sourceforge.kolmafia.equipment.SlotSet;
 import net.sourceforge.kolmafia.maximizer.Boost;
 import net.sourceforge.kolmafia.maximizer.Maximizer;
 import net.sourceforge.kolmafia.maximizer.PriceLevel;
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5518,15 +5520,18 @@ public static Value closet_amount(ScriptRuntime controller, final Value arg) {
 
   public static Value equipped_amount(ScriptRuntime controller, final Value arg) {
     AdventureResult item = ItemPool.get((int) arg.intValue(), 0);
-    int runningTotal = 0;
+    int amount = InventoryManager.getEquippedCount(item);
```
jaadams5, line 5523: Both flavors of getEquippedCount are public and static.  I could make a case that calling the two parameter version here is clearer, especially in an IDE that will helpfully show that the parameter is named includeAllFamiliars.  I might go farther and wonder whether the one parameter version should be deleted and all calls to it add false as a second parameter.   I'm not certain some kind of parallelism with available_amount is beneficial but that is my opinion.  I checked and there are about 10 calls to the one parameter version.  So that change might be too intrusive.  This is not intended as a blocker.

jaadams5 (approved):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
midgleyc: I think it's a reasonable case to make. There are a few cases in the codebase where searching all familiars would be potentially desired (if a bit slower than not searching).

For example, there's a "do we have Thor's Pliers" check that checks equipped items. We still have pliers even if they're on a Disembodied Hand we can't use. All these cases are fringe, though.

PR #1595, opened 2023-03-03
New Preference - 8BitBonusTurns

JamesDowney:
The idea here is that a user might want to see how many turns are left before the bonus in the 8bit realm shifts to the next location. I welcome any and all feedback, this is my first real mafia PR and my first attempt at writing tests.

Mafia forum thread: https://kolmafia.us/threads/track-remaining-turns-before-8bit-realm-bonus-shifts.28707/


gausie:
Thank you! I think the way I imagine doing this is to just have a counter for turns spent in the zone and leave it to users to mod 5.

I'll take a deeper look later today

gausie:
Also we will need unit tests to get this past the line

JamesDowney:
> Thank you! I think the way I imagine doing this is to just have a counter for turns spent in the zone and leave it to users to mod 5.

I've been adding up the turns spent in all 4 of the 8-bit zones and mod 5ing it, but if turns_spent is off for whatever reason, there's no good way for a user to remedy the issue. This happened to fronobulax [here](https://kolmafia.us/threads/8bit-relay.28676/post-171491) for example.


> Also we will need unit tests to get this past the line

Can you clarify on this at all? I'm just genuinely not sure what else needs testing and I couldn't find any documentation on testing as it relates to mafia (though that's probably on me!)

midgleyc:
So a dumb mechanics question, but does the counter increment at the start or end of combat? That is, if you go into a combat then saber-force (or rollover runaway) it, does the counter increment?

gausie:
Oh my apologies, I didn't see you had already added tests! Shows me for checking from my phone in bed.

On the mod5, what I mean is that you can just have your pref count up from 0 tracking turns spent in all the 8-bit realm, and the user can mod5 to see how far in to one bonus they are. That sort of feels more in line with the way we offer tracking to me.

JamesDowney:
> So a dumb mechanics question, but does the counter increment at the start or end of combat? That is, if you go into a combat then saber-force (or rollover runaway) it, does the counter increment?

I had no idea, so I gave it a shot! From testing, I'm pretty confident saber/rollover adventures don't increment KoL's counter. https://pastebin.com/cjSvP9B4

JamesDowney:
> On the mod5, what I mean is that you can just have your pref count up from 0 tracking turns spent in all the 8-bit realm, and the user can mod5 to see how far in to one bonus they are. That sort of feels more in line with the way we offer tracking to me.

Oh, gotcha. Sorry, I misunderstood initially. If that is more in line with other preferences then I'm on board to swap to that. The only thing I'm wondering is what I'd set that property to if it gets out of sync with KoL - I suppose I could just set it to the next multiple of 5?

gausie:
> Oh, gotcha. Sorry, I misunderstood initially. If that is more in line with other preferences then I'm on board to swap to that. The only thing I'm wondering is what I'd set that property to if it gets out of sync with KoL - I suppose I could just set it to the next multiple of 5?
> 
> Editing to clarify: If we see that the color changes, but that the property is out of sync (mod 5 != 0), what's the best way to handle that?

Yeah great question. The maths is a bit silly but to me the solution to that would be to set the pref to the next number that is a multiple of 5 and also gives the correct zone when you mod the turns spent by 20 - does that make sense? I'd also rename the number to 8BitZoneTurns or something

midgleyc:
I don't think the extra complexity is worth it. I don't think it adds anything.

For an example of one that works like this (maybe it counts up to 11? not sure) there's the cursed magnifying glass.

src/data/defaults.txt
```diff
@@ -170,6 +170,7 @@ global	_gitUpdated	false
 global	_svnRepoFileFetched	false
 global	_svnUpdated	false
 
+user	8BitBonusTurns	5
```
jaadams5, line 173: I like descriptive preference names.  I'd prefer a name like 8BitTurnsBeforeBonusChange even though I will get annoyed every time I have to type it.  We might want to revisit the name once we know exactly what we are keeping track of.  I wondering whether 8BitTurnsSpent is a better name if the user has to handle the mod 5 calculation.  Not a blocker though.

jaadams5:
Once we can describe exactly what this is doing - counting up or down, mod 5 or not, I would like to reconsider the name to be something more descriptive.

If my hypothesis that the scripts counter got off because I visited the URL for the "old" zone then we need to confirm that needs to be included in the count.  I'll try and gather more data.

Thanks.

jaadams5:
It appears that I visited http://127.0.0.1:60080/adventure.php?snarfblat=73 at some point.  Perhaps that location also needs to be included in incrementing counts?

gausie:
> It appears that I visited http://127.0.0.1:60080/adventure.php?snarfblat=73 at some point. Perhaps that location also needs to be included in incrementing counts?
> 
> Maybe not - I adventured there once(and received no "points") but still had five turns in the bonus zone until KoLmafia changed the color...

I don't think that zone has anything to do with anything anymore

JamesDowney:
> Yeah great question. The maths is a bit silly but to me the solution to that would be to set the pref to the next number that is a multiple of 5 and also gives the correct zone when you mod the turns spent by 20 - does that make sense? I'd also rename the number to 8BitZoneTurns or something

I made this change locally, but (and I might be inventing a situation that's unlikely to occur) I was thinking about the possibility of someone relying on a preference that counts the turns in all of the 8-bit zones rather than adding up each of the turns_spents, and getting somewhat surprising behavior if the preference jumped by as much as 19.

An example off the top of my head would be if a script had a failsafe where it aborted when more than x turns in 8bit realm had passed without whatever goal being reached.

Incrementing from 0 and resetting when 8BitColor changes is a possibility as well.

gausie:
> An example off the top of my head would be if a script had a failsafe where it aborted when more than x turns in 8bit realm had passed without whatever goal being reached.

I think if that number gets out of sync significantly there is a problem, however it surfaces. Because it's counting *up* an error is always going to result in undercounting not over.

> Incrementing from 0 and resetting when 8BitColor changes is a possibility as well.

this is fine I think too. Also if they were to do something like add an iotm that let you repeat a zone multiplier for another 5 turns or something it would maybe be less confusing.

JamesDowney:
> Do we know for sure that the zone bonus is determined by turns spent and not, for example, by fights won? I was doing this manually and thought I saw color changes six or seven turns apart when I was losing fights. I can revisit and try and confirm but I wat to be out of run when I try that.

Oh no. You're definitely right, some spading by player `breakingamber` confirms that getting beaten up does not advance the color.

JamesDowney:
Free runs also do not increment it.

gausie:
Great! Looks like we're almost done here. So @JamesDowney does this code respect that it is not advanced on a fight loss? Can we add a test for that?

Also I can see your HTML isn't coming from running mafia with debug mode on, which is my personal preference for HTML fixture generation but not a blocker

jaadams5:
I think I am Ok with this but I would prefer @gausie approve and push it.

JamesDowney:
I rewrote my tests to use FightRequest which I think is an improvement, and I replaced html I got from the relay browser with debug log html, which made good sense to me and is just a much nicer way to access the information on my end - so thanks for this suggestion!

@gausie I did also go ahead and add tests to make sure it wasn't incremented on loss or free run.

@jaadams5 You had previously mentioned considering a different, more descriptive name, and I'm totally on board with that, but nothing immediately came to mind for me. Do you have any suggestions?

gausie:
I think the name is fine. Lets merge

gausie (approved):

PR #1596, opened 2023-03-03
fix: bird of the day doesn't show up in maximizer

midgleyc:
Issue introduced in #1307.

The problem is that `getAllModifiers` checks `modifierStringsByName` for keys, but `overrideModifier` only modifies `modifiersByName`. So if we override a modifier that isn't in `modifiers.txt`, the maximizer won't see it.

I wrote the test in r26967 to confirm this was the issue, then fixed it.

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -5291,6 +5291,7 @@ Effect	Blessing of Serqet	Monster Level: +20
 Effect	Blessing of She-Who-Was	Mysticality: +5, Spooky Damage: +5
 # Blessing of Squirtlcthulli: Recover HP and MP after battle
 Effect	Blessing of Squirtlcthulli	HP Regen Min: 2, HP Regen Max: 10, MP Regen Min: 2, MP Regen Max: 10
+Effect	Blessing of the Bird	none
 Effect	Blessing of the Creepy Pasta	Spooky Spell Damage: +20, Sleaze Resistance: +2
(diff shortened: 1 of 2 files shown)


gausie (approved):
Is `none` a thing or is that failing silently?

gausie:
Can we show a warning if modify `modifiersByName` for something that's not in `modifierStringsByName`?

midgleyc:
"none" is a thing. It's used in mutexes (and the scratch 'n' sniff weapons). It means "this has modifiers, but nothing material"

midgleyc:
> Can we show a warning if modify modifiersByName for something that's not in modifierStringsByName?

We could, but I think the fact this is a problem indicates the data structure isn't right. Ideally (I think?) we want the keys to be the same. Perhaps it should be one object, which contains both the modifiers and a string, if we have a string.

Plus... I think most of our warnings just get ignored by the user. They're not really visible enough.

midgleyc:
Will do a follow-up if I can get something working well.

I was hoping this same approach would make distillate show up in the maximizer, but it doesn't seem to.

midgleyc:
I think the trick is actually that `getAllModifiersOfType` should be checking `modifiersByName`, not `modifierStringsByName`. From reading the file, the latter is "the contents of `modifiers.txt`" which never gets reset, whereas the former is "modifiers as they affect the current character", which does get reset.

Same for `getAllModifiers` which is only used by the `modifies` command, which I wasn't aware of.

midgleyc:
Okay, we can't do that because we get an NPE. Presumably the modifiers are lazily loaded.

PR #1601, opened 2023-03-04
feat: dump of items / effects from the new iotm

midgleyc:


Veracity0 (approved):
OK.

I'm doing this too, but also including support for The Shadow Forge, so we will keep inventory up to date when purchasing items. Also, I've been fighting the shadow entities and collecting info for them.

You can go first. :)

midgleyc:
very good :D

PR #1605, opened 2023-03-05
numeric_modifier(thrall, string) exists

horrible-little-slime:
This function already exists in both js and ash; `RuntimeLibrary.getModifierType` handles thrall perfectly. This will make `ashref` and `jsref`, as well as the typescript definitions that get published alongside each mafia revision, reflect this reality.


midgleyc (approved):
I have checked, and you are right.

Thanks!

PR #1606, opened 2023-03-05
Bathole and Tower Ruins tests

JamesDowney:
Tests for the Wizard of Ego quest and the Level 4 quest QuestManager handling.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/QuestManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/QuestManagerTest.java
@@ -3550,4 +3550,54 @@ public class QuestManagerTest {
       }
     }
   }
+
+  @Nested
+  class WizardOfEgo {
+    @ParameterizedTest
+    @CsvSource({"staring_into_nothing, 4", "into_the_maw_of_deepness, 5", "take_a_dusty_look, 6"})
+    public void canParseTowerRuinsChoices(String html, int expectedResult) {
+      var cleanups = new Cleanups(withQuestProgress(Quest.EGO, QuestDatabase.UNSTARTED));
+
+      try (cleanups) {
+        String path = "request/test_adventure_tower_ruins_" + html + ".html";
+        var request = new GenericRequest("adventure.php?snarfblat=22");
+        request.responseText = html(path);
+        QuestManager.handleQuestChange(request);
+        assertThat(Quest.EGO, isStep(expectedResult));
+      }
+    }
+
+    @Test
+    public void doNotUnfinishQuest() {
+      QuestDatabase.setQuest(Quest.EGO, QuestDatabase.FINISHED);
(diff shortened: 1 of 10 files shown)


Veracity0 (approved):
I like this a lot.

I manually tested this a while back when I was working on KoLAdventure.canAdventure for these zones, and I convinced myself that it worked, but supplying tests supported by actual html responseTexts is the cat's meow.

Veracity0:
One comment is that I like setting up an HTML client, setting the response to the submitted URL, and doing s run() on an actual request with the submitted URL.

You are creating the request, setting the responseText, and then verifying that QuestManager will do the right thing.
OK, for 99% of requests that is fine - but my preferred technique tests that QuestManager is actually called for the request that you set up.

Your method is fine, but my preferred technique takes one step back (and higher) in the request processing and verifies that submitting the request actually calls the specific QuestManager method that you expect to be doing the work you require to be done.

test/net/sourceforge/kolmafia/session/QuestManagerTest.java
```diff
@@ -3550,4 +3550,54 @@ public void doNotProgressQuestPyramidOnUhOh() {
...
+      try (cleanups) {
+        String path = "request/test_adventure_tower_ruins_" + html + ".html";
+        var request = new GenericRequest("adventure.php?snarfblat=22");
+        request.responseText = html(path);
+        QuestManager.handleQuestChange(request);
+        assertThat(Quest.EGO, isStep(expectedResult));
+      }
+    }
+
+    @Test
```
Veracity0, line 3570: OK. I think I will insist that you use our standard "Cleanups" method.

This test could/should be this:

```
     @Test
     public void doNotUnfinishQuest() {
      var client = builder.client;
      var cleanups =
          new Cleanups(
              withHttpClientBuilder(builder),
              withQuestProgress(Quest.EGO, QuestDatabase.QuestDatabase.FINISHED));
      try (cleanups) {
        client.addResponse(200, html("request/test_adventure_tower_ruins_staring_into_nothing.html"));
        var request = new GenericRequest("adventure.php?snarfblat=22");
        request.run();
        assertThat(Quest.EGO, isStep(QuestDatabase.FINISHED));
      }
    }
```

Note that it:
1) actually submits the request, and verifieds that the expected quest change happens, rather than calling the magic QuestManager method that does that.
2) uses withQuestProgress to set the property and clean up the property after the quest is finished.

This is our standard test convention. Do it for all your tests. :)

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/session/QuestManagerTest.java
JamesDowney: I really appreciate this feedback! Everything you said makes perfect sense to me and I can see how it's a clear improvement.

Veracity0 (approved):

PR #1610, opened 2023-03-06
When get Professor What Garment, give navigation link to Melvign

Veracity0:


midgleyc (dismissed):
Optional: instead of making methods `public` for tests, you can make them package-private, as the test has the same namespace as the source file.

Veracity0:
> Optional: instead of making methods `public` for tests, you can make them package-private, as the test has the same namespace as the source file.

Assuming that means `protected`, it works, and I have done so.

I just learned something about how the testing environment works. Thanks!

midgleyc:
ah sorry, not `protected`. package-private is the "default" visibility: as in, if you put nothing at all, it's package-private.

https://stackoverflow.com/a/215505/2591803

PR #1611, opened 2023-03-07
AoSoL council quests text

Veracity0:
I am missing two:

- Visiting the council again after they have told you to start the war but before you have done so.
- Visiting the council after you have won the war by defeating the frat boys.

The first was an oversight. The second was inevitable, since I defeated the hippies.

I'll be starting my 9th (and final?) AoSoL run this afternoon, to finish off getting the factoids.
I'll make another PR tomorrow, I expect, with the two missing messages.

This is good enough for now, I think.

One observation:

The council text opening the lair starts with "Veracity, I've got to say, it's been a pleasure working with you."
Obviously, that won't work for anyone else. I entered it as "I've got to say, it's been a pleasure working with you."

Looking at other quest texts in this same file, I see:

Wow! You actually brought back the Holy MacGuffin, Darzil?

I have a feeling that will not match for anyone else. :)

I don't think we need the FULL text when doing text matching. I looked at how it was done a few months ago and it is pretty complicated. I included the full text for the messages from this path because they are quite amusing.


midgleyc (approved):
If you look at ones nearer the top, it seems that
```
quest<tab>state<tab>start text<tab>return text
```
works for setting tab to state, instead of 
```
quest<tab>state<tab>start text
quest<tab>state<tab>return text
```
but either is fine I think.

Veracity0:
I saw that, but intentionally used the two-line version, since the saved texts are quite long.
I figured it would be easier to read - and laugh - if each message was alone on the line.

PR #1615, opened 2023-03-08
update usable offhands to 2

htremura:
Given that the other `Usable-` values assume a player has all the relevant slot-increasing items, by my understanding the default `UsableOffhands` should be higher to account for Sinistral Homunculus/Left-Hand Man.


gausie:
So this is just how many offhands are retained when you select "all but usable" in the Equipment manager? I am surprised anyone uses this but agree that this is possibly a more sensible default? I suppose really it should be set on ascension based on what familiars and skills you have/can access. But that doesn't need to be done now, and I'm not in a rush to do it ever!

gausie (approved):

htremura:
> So this is just how many offhands are retained when you select "all but usable" in the Equipment manager? I am surprised anyone uses this but agree that this is possibly a more sensible default? I suppose really it should be set on ascension based on what familiars and skills you have/can access. But that doesn't need to be done now, and I'm not in a rush to do it ever!

It's a very niche setting, but I noticed it when I was pulverizing standard equipment (using that "all but usable" button). I manually changed it on my mafia, but it seemed reasonable to have it match the functionality of all the other defaults. 

Even if someone were to implement that which you describe, it would also require a way to lock it to a number. I like the defaults as I see little reason _**not**_ to hang onto a few more of some item types even if I do not have the necessary familiars/skills/items to increase my slots. As I think many a KoL player thinks, I **_might_** someday have them (I will likely not, but I'd like to keep my options open).

PR #1618, opened 2023-03-08
feat: remove validate gradle wrapper action

midgleyc:
This action is failing far too often (as a false positive: failing to access some website) to be worth doing.

The aim is to avoid being vulnerable to a social engineering attack where an outsider upgrades gradle; to avoid that we can just always ensure that we only update gradle ourselves.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,9 +29,6 @@ jobs:
           java-version: "17"
           cache: "gradle"
 
-      - name: Gradle Wrapper Validation
-        uses: gradle/wrapper-validation-action@v1
-
       - name: Build
         env:
           ORG_GRADLE_PROJECT_commit: ${{ github.sha }}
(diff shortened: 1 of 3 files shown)


Veracity0 (approved):
We don't have any sort of "auto upgrade" for gradle.
So, when (and why) do we decide to upgrade gradle?

jaadams5:
I haven't noticed the failures.

I do not do anything to maintain Gradle on my Windows system.  I do "gradlew <some task>" and sometimes that operation decides there is a newer version of gradle and AFAIK fetches it and "installs" it.  So as a user I see gradle as automatically updating with no user action.  I would like that to continue without any change in my behavior.  I can't tell if this still supports that or not.

midgleyc:
This only affects a GitHub workflow and shouldn't affect local behaviour at all.

Previously I've updated gradle when I've seen there's been a new version. I don't think we have any schedule for checking, and I've not been able to find a local tool, though I expect Renovate could handle it if we wanted to.

jaadams5:
OK my updates were caused by "your" action and this doesn't change that.  Thank you.

PR #1620, opened 2023-03-09
Update monsters.txt res for Shadows

ckb11:


gausie (approved):
Happy to merge if you're confident these are correct!

PR #1622, opened 2023-03-09
Add feasted proxy record to expose info

s-k-z:
This can tell script authors whether KoLmafia knows if a current familiar has the familiar weight bonus from the [Moveable Feast](https://kol.coldfront.net/thekolwiki/index.php/Moveable_feast). Currently it's inconvenient to check, since it isn't a modifier and isn't counted as such.

Examples:

```java
> ash $familiar[pocket professor].feasted

Returned: false

> ash $familiar[frumious bandersnatch].feasted

Returned: true
```

image
image

Do proxy record fields have tests? I'm not sure what they look like.


midgleyc:
I think we're just throwing those tests onto `RuntimeLibraryTests` (or `CustomScriptTests` for the complicated ones).

s-k-z:
Thanks, found it with the CustomScriptTests

gausie:
This is a great PR but I think it makes more sense to just expose this via the modifier API. Does anyone who has been around for longer than me know why that wasn't done in the first place?

gausie:
I suppose we can do both this PR and that also!

Veracity0:
> This is a great PR but I think it makes more sense to just expose this via the modifier API. Does anyone who has been around for longer than me know why that wasn't done in the first place?

It's like a hidden buff that lasts for 20 combats. I lobbied CDM to put it into api.php, since otherwise, you could only tell it was in effect from the charpane.

There are other hidden things that you can only detect indirectly. Getting the Be Gregarious combat skill from Extrovermectin, for example. But I can't think of other "hidden" buffs. Or any other buffs that apply to a particular familiar, rather than to whichever one us currently equipped.

I'm curious to see how you would model it as a Modifier.

gausie:
I suppose I would just make it appear in your familiar weight calculation in the scenario that you have that familiar out with an active feast, just like location modifiers that appear when your current location is set to a certain place

gausie:
The example that comes to me is extra xp when you're adventuring in the same location as your autumnaton quest

s-k-z:
The snowing effect from the Ghost of Crimbo Cheer is probably another hidden modifier, increasing cold damage in the zone for the rest of the day.

Veracity0:
> I suppose I would just make it appear in your familiar weight calculation in the scenario that you have that familiar out with an active feast, just like location modifiers that appear when your current location is set to a certain place 

Sure, but the issue on the forums was that "modtrace" didn't show "feasted" as a source of +10 familiar weight.

I assume that applies to location-specific location modifiers.

Do we, somehow, add generated modifiers for Familiar and Location that show up in modtrace rather than being secretly applied? Perhaps this is (hidden).

Stay tuned; this is part of your followup PR. 😃

gausie:
They do show in modtrace! Yeah I'll do it in a follow-up PR

midgleyc (approved):
I think this alone is good to go in as-is, and the rest can happen as a follow-up.

PR #1624, opened 2023-03-10
feat: add modifiers for passive damage

midgleyc:
Add modifiers for damage aura / thorns, with a value of the number of damage sources.

This, rather than damage, was chosen for two and a half reasons:
* the primary use-case for this is fighting bosses, which all have such physical / elemental resistance such that the number of sources is almost all you care about (wall of skin, Frosty, Unkillable Skeleton)
* I didn't want to add 48 new modifiers for min / max in each element + physical
* it's unclear what to do about modifiers like "Feeling Nervous" which deal a percentage of enemy HP

The other use-case is being able to check, with `modtrace`, before a fight with a boss that heals if you hit it with any non-group-damage source (e.g. wall of bones, Zombie Homeowners' Association)

Further notes:
* double-ice cap is once per fight and so omitted
* fiberglass items are class-only, but only for the passive damage bit. Similarly omitted for complexity.
* underwater-only passive I don't care about, so was omitted
* bottle opener belt buckle implies it should be thorns, but also works on enemy miss by wiki. Don't know if it works if the enemy is stunned. One to check.


gausie:
Hey super cool. Did you check this against batfactors.txt? IIRC it contains a lot of this information, encoded in zarqon's DSL

midgleyc:
I did not. I forgot about that. I took them off the wiki.

midgleyc:
I took a glance through batfactors, and I don't see any new ones I didn't have. The wiki has some that batfactors doesn't have, too: SMOOCH / marble / chiffon gear, for instance.

gausie (approved):
It would be useful to document this in the mafia wiki. Plus then we can link zarqon to it and he can consider relying on mafia for these data instead.

midgleyc:
I added them to the modifiers page, but I think zarqon's aim is a bit different to mine -- his data contains details about the damage amounts / elements, whereas mine is mostly just that it exists. He could use it to check whether there are any sources he's missed, though.

PR #1631, opened 2023-03-13
Don't refill the June Cleaver queue with data from previous ascension

s-k-z:
If you have a KoLmafia session that has encountered at least one [June Cleaver](https://kol.coldfront.net/thekolwiki/index.php/June_cleaver) noncombat, ascend, then encounter another June cleaver noncombat, the `juneCleaverQueue` preference will do something like this:
```
Preference juneCleaverQueue changed from  to 1475,1473,1470,1467,1474,1467
```
Instead of:
```
Preference juneCleaverQueue changed from  to 1467
```

Currently, the backing queue data is not emptied because the `reset` function is not hooked up.

This change will remove the static data structure, instead use a new one each time the function is called, and populate it with data from the preference if there is any.


gausie (approved):
I was going to say that we could just reset JuneCleaverManager on ascension but actually you're right, it looks like there's no reason to persist this in memory.

PR #1632, opened 2023-03-13
shadow rift fixes

Veracity0:
1) Shadow Rift bosses are marked BOSS
2) If your current Shadow Rift Ingress is from The 8-Bit Realm, you do not need to have the continuum transfunctioner equipped to adventure there.
3) Item Drop is reduced by 80% when adventuring in the Shadow Rift zone, not just the location of that name.

The existing code that did that for the Location reduced the Item Drop bonus EXCLUDING that from your familiar by 80%.
Is that correct? It is trivial to INCLUDE the familiar's contribution - and I have a commented out line that does it that way.

Which is correct? -80% not including familiar or including familiar?

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -8574,6 +8574,7 @@ Event	Mysticality Day	Experience Percent (Mysticality): +25
 # -100% is the most common penalty in The Sea; use it as the default
 
 Zone	The Sea	Initiative Penalty: -100, Item Drop Penalty: -100, Meat Drop Penalty: -100
+Zone    Shadow Rift	Item Drop: [-0.8*mod(Item Drop))
 
 # Location-specific section of modifiers.txt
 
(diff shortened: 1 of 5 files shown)


midgleyc (approved):
Thanks for checking!

src/data/modifiers.txt
```diff
@@ -8574,6 +8574,7 @@ Event	Mysticality Day	Experience Percent (Mysticality): +25
 # -100% is the most common penalty in The Sea; use it as the default
 
 Zone	The Sea	Initiative Penalty: -100, Item Drop Penalty: -100, Meat Drop Penalty: -100
+Zone    Shadow Rift	Item Drop: [-0.8*mod(Item Drop))
```
midgleyc, line 8577: oops, no: this is not spaced correctly

midgleyc (changes requested):

src/data/modifiers.txt
Veracity0: Fixed.

midgleyc (approved):
alright, now good :)

PR #1633, opened 2023-03-14
to_wiki_url in the RuntimeLibrary

gausie:
- Add tests for to_wiki_url function
- Write to_wiki_url functions

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -186,6 +186,7 @@ import net.sourceforge.kolmafia.utilities.HTMLParserUtils;
 import net.sourceforge.kolmafia.utilities.InputFieldUtilities;
 import net.sourceforge.kolmafia.utilities.LogStream;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
+import net.sourceforge.kolmafia.utilities.WikiUtilities;
 import net.sourceforge.kolmafia.webui.RelayServer;
 import org.htmlcleaner.HtmlCleaner;
 import org.htmlcleaner.SimpleXmlSerializer;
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -3654,6 +3670,20 @@ public static Value to_url(ScriptRuntime controller, final Value value) {
         : new Value(adventure.getRequest().getURLString());
   }
 
+  public static Value to_wiki_url(ScriptRuntime controller, final Value value) {
+    var type =
+        switch (value.getType().getName()) {
```
midgleyc, line 3675: can you not switch on `getType()` directly?

midgleyc (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
gausie: No, switches have to be primitives or enums!

gausie: And before you go in and try to enum types... the issue is that we use the same class to make arbitrary types at runtime

midgleyc: I thought `getType` was a TypeSpec.

midgleyc: okay, that's `value.getType().getType()`. Can you switch on that instead for slightly less stringly-typed code?

gausie: yeah ok

midgleyc (approved):
Nice :D

PR #1636, opened 2023-03-14
feat: remove langserver

midgleyc:
Remove the langserver due to lack of interest in maintaining it.

Keep the parser improvements (e.g. the error messages that tell you which line / character the error was on) because I like those.


gausie (approved):
Fantastic. This was long due and I appreciate you doing it

PR #1637, opened 2023-03-14
Automation for Labyrinth of Shadows

Veracity0:
I moved the ShadowTheme enum and spoiler generation code from ChoiceAdventures to RufusManager.
The Labyrinth of Shadows appears whether or not you are working on a quest from Rufus, but how we handle it changes if you are seeking an artifact.

So, all Rufus-related stuff is now in RufusManager.
And hundreds of lines were removed from ChoiceAdventures, which is already bloated. :)

I added support for automating the Labyrinth of Shadows choice adventure (1499)

1) In the Choice Adventures panel of the AdventureFrame, there is now a location-specific configuration control for it, attached to the Shadow Rift zone.
```
    this.shadowLabyrinthSelect.addItem("show in browser");
    this.shadowLabyrinthSelect.addItem("90-100 Muscle substats");
    this.shadowLabyrinthSelect.addItem("90-100 Mysticality substats");
    this.shadowLabyrinthSelect.addItem("90-100 Moxie substats");
    this.shadowLabyrinthSelect.addItem("+3 turns to 3 random effects");
    this.shadowLabyrinthSelect.addItem("Maximum HP +300%");
    this.shadowLabyrinthSelect.addItem("Maximum MP +300%");
    this.shadowLabyrinthSelect.addItem("+5 Spooky, Hot, Sleaze resistance");
```
These correspond to the value of the shadowLabyrinthGoal property:
"browser", "muscle", "mysticality", "moxie", "effects", "maxHP", "maxMP", "resistance"

Obviously, these are not the choice decisions we will submit to KoL.
Those are calculated based on character state and the actual options that are offered.

2) When automating - GUI, script, or "auto" button on the choice page in the Relay Browser - we look at what you desire and what is on offer and tell KoL to reroll the options (by submitting decision "1") until the desired result is available, at which point, we get it.
- If you are on an artifact quest, that overrides your configured result
- Otherwise, we get our configured result 

Something I noticed:

The Labyrinth of Shadows does not consume a turn.
Therefore, if you automate via GUI or script for N turns, it does not count - and it will automate an additional fight.
In order to avoid that, set "1 choiceadv" as your goal and it will stop after it runs it.
```
Request 11 of 20 (Shadow Rift: Shadow Rift (The Nearby Plains)) in progress...

[1515] Shadow Rift (The Nearby Plains)
Encounter: A Labyrinth of Shadows
Encounter: A Labyrinth of Shadows
Encounter: A Labyrinth of Shadows
You acquire an effect: Shadow's Heart (30)

Conditions satisfied after 11 adventures.
```


Veracity0:
I'm going to change this to having a "shadowLabyrinthGoal" property with a descriptive keyword.
No more complex, but easier for the user to understand.

Veracity0:
It is intended that if you are on an "artifact" quest that you have to get the artifact first.
If you are on an "entity" quest, the NC is replaced by the boss.
If you are on an "items" quest, you can configure the NC any way you want.

Aside from the fact that I can't imagine a use case where you'd want to say "I specifically signed up for the artifact quest, but I want to get this or that other effect first and then spend anther 11 turns getting the artifact", if you are getting the artifact, there is one effect that is unavailable. Plus, nothing says that you have to turn in the artifact right away.

PR #1640, opened 2023-03-18
Tracking for next Shadow Rift NC

Veracity0:
Three new properties:

shadowRiftTotalTurns - fights and NCs in the Shadow Rift that count as turns (monsters & lodestone)
shadowRiftLastNC - the turn of the last encounter that counts as an NC (Labyrinth and bosses)
encountersUntilSRChoice - how many turns until the next NC

Info in [This forum post](https://kolmafia.us/threads/new-shadow-rift-ncs-buffs-and-buff-extension.28712/post-171580) by Malabu Stacey citing info from canonnfire40.


src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -488,4 +489,48 @@ private static String shadowLabyrinthChoiceDecision(String responseText) {
...
+    switch (choice) {
+      case 1499 -> {
+        // The Shadow Labyrinth
+        Preferences.setInteger("shadowRiftLastNC", currentTurns);
+        lastNC = currentTurns;
+      }
+      case 1500 -> {
+        // Like a Loded Stone
+        ResultProcessor.removeItem(ItemPool.RUFUS_SHADOW_LODESTONE);
+        currentTurns = Preferences.increment("shadowRiftTotalTurns", 1);
```
midgleyc, line 531: Like a Loded Stone is a free NC. Does it definitely increment the turns until the next NC?

midgleyc (approved):
One thing to check, but otherwise looks fine.

It's odd to have "SR" not be "semi-rare"!

src/net/sourceforge/kolmafia/session/RufusManager.java
Veracity0: Yes. I tested all these cases in the browser.

According to cannonfire40, Like a Loded Stone is a "super likely", not an "NC".

Veracity0:
Ha ha ha.

Further study of my logs indicate two misunderstandings:

- Shadow boss increments turn in zone - AFTER saving current turn as last NC.
That advances turn towards next NC.
- Therefore, it was not the lodestone which did that.

I opened a new PR to fix this.

I'm considering just approving it, so that our users don't get upset and file a bug report on this PR. :)

PR #1642, opened 2023-03-19
feat: add git scripts to Script Manager UI

midgleyc:
Add git scripts to the Script Manager GUI.

I don't use the GUI myself (just the CLI), so open for any comments on things that should work differently.

There are also, as yet, no git scripts in `svnrepo.json`, though I could change the github scripts that work over from SVN?


Veracity0:
I am experimenting. I didn't have any git scripts installed. So I found [8bit-relay](https://kolmafia.us/threads/8bit-relay.28676/).

I installed it it via ```git checkout loathers/8bit-relay```.
```
> git list

loathers-8bit-relay

> git update loathers-8bit-relay

Updating project loathers-8bit-relay
Pull
No changes
```

In Script Manager, I can see loathers-8bit-relay in the Manage tab.
Right click for options:
Delete script -> ```Cannot find unique match for loathers-8bit-relay-main```
Open forum thread -> nothing. Well, maybe if we had it in the repo, it would have a pointer to the forum thread.
Update this script -> ```Cannot find unique match for loathers-8bit-relay-main```
Update all scripts -> Looked at all SVN scripts and also - successfully - at this one.

midgleyc:
> Open forum thread -> nothing. Well, maybe if we had it in the repo, it would have a pointer to the forum thread.

That's what happens for the existing SVN scripts too.

> Cannot find unique match for loathers-8bit-relay-main

I see what's happening here. `main` is the default branch, so it's optional, and it's not present in the installed version. We do know what the folder is, though, so we should be able to figure it out.

jaadams5:
>Open forum thread -> nothing. Well, maybe if we had it in the repo, it would have a pointer to the forum thread.

That is expected behavior.  At least it is what happens with SVN scripts that I have installed that are not referenced in svnrepo.json.

I would be concerned about changing the protocol on existing scripts until the local changes issue mentioned at https://kolmafia.us/threads/excavator-gausies-spading-script.25076/page-3#post-171587 is resolved.

Veracity0 (approved):
Thank you for doing this. I updated and tested the GUI again. Looks good!

I notice that GitManager catches - and logs - exceptions, returning false or null or whatever to indicate failure, whereas SVNManager just throws the exception, requiring the caller to catch and handle them.

I realize you wrote GitManager and it's just a cleaner implementation.
I think it is beyond the scope of this PR to sanitize SVNManager similarly.

Maybe someday.

PR #1644, opened 2023-03-20
Fix formula for eight_bit_points locs with base != 300

Veracity0:
Experimentation in The 8-Bit Realm has shown me a couple of things:

1) All four areas have a base value for the modifier which you must exceed in order to get any bonus:
    - Vanya's Castle and Megalo-City = 300
    - Hero's Field = 100
    - The Fungus Plains = 150
2) All four areas have a range of 0-300 ABOVE that minimum which provide a bonus.
    - Vanya's Castle and Megalo-City = 301-600
    - Hero's Field = 101-400
    - The Fungus Plains = 151-450

Which means my previous conclusion that only mods > 300 gave a bonus was incorrect for Hero's Field and The Fungus Plains.

3) All bonuses are in increments of 10; Init 305 in Vanja's Castle does NOT give a bonus of 5.
4) Bonuses round. 300-304 -> 0 and 305-314 -> 10.
5) I did a lot of experimenting. 8bit-relay has what looks like an accurate algorithm. I adapted it.

Now, what I have here has an overloaded version of eight_bit_points which allows testing the algorithm via script. ```eight_bit_points(loc, color, modvalue)``` doesn't depend on current character state.

Here's a little program I wrote to see how modifier bonuses work out.
It takes bonuses from 0 - 300 and sees how many points you get when either the area is favored or not.

The program:
```
location VANYAS_CASTLE = $location[Vanya's Castle];
location MEGALO_CITY = $location[Megalo-City];
location HEROS_FIELD = $location[Hero's Field];
location FUNGUS_PLAINS = $location[The Fungus Plains];

void loop8(location loc, string color)
{
    boolean bonus = false;
    int base = 0;
    switch (loc) {
    case VANYAS_CASTLE:
	base = 300;
	bonus = color == "black";
	break;
    case MEGALO_CITY:
	base = 300;
	bonus = color == "blue";
	break;
    case HEROS_FIELD:
	base = 100;
	bonus = color == "green";
	break;
    case FUNGUS_PLAINS:
	base = 150;
	bonus = color == "red";
	break;
    }
    print(loc + " base=" + base + " color=" + color);
    print("--------------");

    // Modifier value
    int minX = 0;
    int maxX = base + 300;

    // Points
    int minY = bonus ? 100 : 50;
    int maxY = bonus ? 400 : 200;

    // So we can detect when 7 value (points) increases
    int curX = minx;
    int curY = miny;

    void print_range(int low, int high, int y)
    {
	int lowx = low == 0 ? 0 : low - base;
	int highx = high - base;
	print(lowx + "-" + highx + " -> " + y);
    }

    for (int x = minX; x <= maxX; ++x) {
	int points = eight_bit_points(loc, color, x);
	if (points < curY) abort(x + " -> " + points);
	if (points > curY) {
	    print_range(curX, x-1, curY);
	    curX = x;
	    curY = points;
	}
    }
    if (curX < maxX) {
	print_range(curX, maxX, curY);
    }
}

void main()
{
    void loop8(location loc, string color1, string color2)
    {
	loop8(loc, color1);
	print();
	loop8(loc, color2);
	print();
    }

    loop8(VANYAS_CASTLE, "white", "black");

    // loop8(MEGALO_CITY, "white", "blue");
    // loop8(HEROS_FIELD, "white", "green");
    // loop8(FUNGUS_PLAINS, "white", "red");
}
```
Results:
```
> test8bit

Vanya's Castle base=300 color=white
--------------
0-9 -> 50
10-29 -> 60
30-49 -> 70
50-69 -> 80
70-89 -> 90
90-109 -> 100
110-129 -> 110
130-149 -> 120
150-169 -> 130
170-189 -> 140
190-209 -> 150
210-229 -> 160
230-249 -> 170
250-269 -> 180
270-289 -> 190
290-300 -> 200

Vanya's Castle base=300 color=black
--------------
0-4 -> 100
5-14 -> 110
15-24 -> 120
25-34 -> 130
35-44 -> 140
45-54 -> 150
55-64 -> 160
65-74 -> 170
75-84 -> 180
85-94 -> 190
95-104 -> 200
105-114 -> 210
115-124 -> 220
125-134 -> 230
135-144 -> 240
145-154 -> 250
155-164 -> 260
165-174 -> 270
175-184 -> 280
185-194 -> 290
195-204 -> 300
205-214 -> 310
215-224 -> 320
225-234 -> 330
235-244 -> 340
245-254 -> 350
255-264 -> 360
265-274 -> 370
275-284 -> 380
285-294 -> 390
295-300 -> 400
```
These values agree with what my manual testing observed.

The debugging function which allows this program is labeled DEBUG. We don't have to submit it - although who knows? It might actually be useful for scripts who want to speculate without actually equipping anything.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -741,6 +741,10 @@ public abstract class RuntimeLibrary {
     params = new Type[] {DataTypes.LOCATION_TYPE};
     functions.add(new LibraryFunction("eight_bit_points", DataTypes.INT_TYPE, params));
 
+    // DEBUG
+    params = new Type[] {DataTypes.LOCATION_TYPE, DataTypes.STRING_TYPE, DataTypes.FLOAT_TYPE};
+    functions.add(new LibraryFunction("eight_bit_points", DataTypes.INT_TYPE, params));
+
     params = new Type[] {DataTypes.INT_TYPE};
     functions.add(new LibraryFunction("get_ccs_action", DataTypes.STRING_TYPE, params));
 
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
I think that either the debug version should be present and tested, or it should be completely removed from the codebase. So this is fine.

PR #1645, opened 2023-03-21
feat: try to sync local changes when updating with git

midgleyc:
I wasn't able to test this because the scripts I tried updated fine without needing this change.

@jaadams5, you've had issues with some scripts, right? Can you post the script, the starting commit, and a change that fails, so I can test this?

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
@@ -138,7 +138,14 @@ public class GitManager extends ScriptManager {
 
       RequestLogger.printLine("Updating project " + folder);
       try {
-        git.pull().setProgressMonitor(new MafiaProgressMonitor()).setRebase(true).call();
+        if (!rebase(folder, git)) {
+          KoLmafia.updateDisplay(
+              MafiaState.ERROR,
+              "Failed to update project "
+                  + folder
+                  + ": rebase error. Perhaps there are local changes we are unable to automatically reconcile. Consider deleting and re-installing project");
+          return false;
+        }
       } catch (GitAPIException e) {
         KoLmafia.updateDisplay(MafiaState.ERROR, "Failed to update project " + folder + ": " + e);
         return false;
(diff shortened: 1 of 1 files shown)


jaadams5:
I have a script that as it happens was installed by both SVN and GIT.  I have local changes in the SVN version and those changes are propagated to my scripts/.  I would like to migrate my local changes to the GIT install, delete the SVN install and have future GIT updates respect my local changes - merge rather than replace in SVN terms.  It didn't work as expected several weeks ago.  If something has change then I will gladly retry.   And if you want files I can share.

midgleyc:
Yes, that's what works for me -- if I go to the folder in `git/`, reset it to an older commit, make a change, and then pull, git automatically reconciles my changes. So I don't have a test case where it breaks.

So I'd like to know:
* which script are you making changes to?
* which commit hash are you on (i.e. if you do `git show` in the `git/whatever` folder, what does it show)?
* an example of a change you've made that makes the merge not work

jaadams5:
Tangentially I feel like my question about something working has morphed into a claim that it doesn't work.

I built what my system called 27296-M which I think was PR without a couple of changes to main.

In a fresh directory I ran mafia and opened the gCLI but did not log in.

I installed autoscend via git checkout https://github.com/Loathing-Associates-Scripting-Society/autoscend.git 

That worked as expected.

It installed excavator as a dependency, as expected.

I then replaced excavator.ash in .git with a version with my local changes.  I added x_beat_me.ash in projects.  That is adding a new file.  I did both using Windows OS commands and not a git command that "added" or "committed".

I then tried git update https://github.com/Loathing-Associates-Scripting-Society/autoscend.git and got an error.  I tried git update and the repository was checked but nothing was fetched from it and my versions of the files were not in scripts.

Then I had to stop and deal with real life for a few hours.

Strip the .txt which was added because github does not like.ash and I'm in a hurry.

Thanks.


[excavator.ash,txt](https://github.com/kolmafia/kolmafia/files/11041886/excavator.ash.txt)

[x_beat_me.ash.txt](https://github.com/kolmafia/kolmafia/files/11041882/x_beat_me.ash.txt)

midgleyc:
> I then tried git update https://github.com/Loathing-Associates-Scripting-Society/autoscend.git and got an error

I will look into the error. If you mean that you typed exactly that command, the error was probably something like "nothing matched", and that's because you need to type the name of the script as it appears in `git list`, not the url. So `git update Loathing-Associates-Scripting-Society-autoscend`, or as it takes unique partial matches, `git update autoscend`.

If there are no updates (i.e. you're on the latest version of autoscend / excavator already), the behaviour you saw matches what I coded to happen -- if there are no updates but you have local changes, nothing happens. If you want your local changes to excavator to be copied to the scripts folder, you want to type `git sync excavator`. Also, if you `git update autoscend` and there's a new version of excavator, nothing will happen there either.

So in detail, I expect:
* if you type `git sync X`, the local files from X are copied to the main folders
* if you type `git update X` and X has no updates, nothing happens
* if you type `git update X` and X has updates, the updates are pulled into the version in `git/` (merging with your local changes if present), then copied to the main folders. Dependencies of X are installed if absent, but not updated if already present.

midgleyc:
Thanks: I can confirm the failure to update with those files, and also that this PR fixes the issue.

Veracity0 (approved):
Frono reported a bug and gave you sufficient info to reproduce it.
You verified this PR fixed his bug.
Surely that is good enough?

PR #1647, opened 2023-03-23
Track currently active latte ingredients

horrible-little-slime:
This information is currently exposed through the pref `latteModifier`, which users can in theory parse and work backwards from to find which ingredients are currently installed. But I think it would be easier to access that directly.

~~This is currently incomplete; I have code designed to parse the latte's name, but I don't know where to put it and couldn't think offhand of any items whose names we parse (my first instinct was umbrella, but it looks like we parse the enchantment rather than the name).~~

We now parse the item name whenever we check the item description, which is imperfect but definitely good enough; we also will update the preference any time we refill the latte


test/net/sourceforge/kolmafia/request/LatteRequestTest.java
```diff
@@ -0,0 +1,35 @@
...
+  public void beforeEach() {
+    KoLCharacter.reset("LatteRequest");
+    Preferences.reset("LatteRequest");
+  }
+
+  private LatteRequestTest() {}
+  ;
+
+  @ParameterizedTest
+  @ValueSource(
```
midgleyc, line 22: may as well do `CsvSource` with a custom delimiter (e.g. `|`, which you already have set up).

test/net/sourceforge/kolmafia/request/LatteRequestTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class LatteRequestTest {
+  @BeforeEach
+  public void beforeEach() {
+    KoLCharacter.reset("LatteRequest");
+    Preferences.reset("LatteRequest");
+  }
+
+  private LatteRequestTest() {}
+  ;
```
midgleyc, line 19: loose ;

gausie (approved):

PR #1652, opened 2023-03-28
we can only adventure in piraterealm islands that we can currently adventure in

horrible-little-slime:
Right now, to check to see if we can adventure in a piraterealm location, we just naively check whether we have access to piraterealm today; this will also verify that we have unlocked the island.

Right now, this will continue to assert that you can access Red Roger's Fortress even if you've completed piraterealm. I don't see any tracking that lets us handle this the way things are currently set up

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -2132,7 +2132,8 @@ public class KoLAdventure implements Comparable<KoLAdventure>, Runnable {
 
     if (this.zone.startsWith("PirateRealm")) {
       // *** You have limited turns available per day.
-      return Preferences.getBoolean("prAlways") || Preferences.getBoolean("_prToday");
+      return (Preferences.getBoolean("prAlways") || Preferences.getBoolean("_prToday"))
+          && Preferences.getString("_LastPirateRealmIsland").contains(this.adventureName);
     }
 
     if (this.zone.equals("Speakeasy")) {
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -2133,7 +2133,7 @@ public class KoLAdventure implements Comparable<KoLAdventure>, Runnable {
     if (this.zone.startsWith("PirateRealm")) {
       // *** You have limited turns available per day.
       return (Preferences.getBoolean("prAlways") || Preferences.getBoolean("_prToday"))
-          && Preferences.getString("_LastPirateRealmIsland").contains(this.adventureName);
+          && Preferences.getString("_LastPirateRealmIsland").equals(this.adventureName);
     }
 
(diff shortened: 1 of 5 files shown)


gausie:
Is this not reflected in native canAdventure? Make a change upstream!

horrible-little-slime:
> Is it not also possible to visit an island that you've just unlocked?

I'm not sure I understand the question; the property we check gets set upon selecting the appropriate choice adventure, so this should make `canAdventure` return true once you select the choice and return false once you select the next choice. One issue with this is that it doesn't perfectly reflect reality--there are times when you've finished adventuring on a particular island but haven't selected the next island, and there are times when you've finished adventuring in the very last island. I don't think mafia piraterealm tracking is precise enough to return a better answer here, though. I personally consider improving piraterealm tracking out of scope of this PR--I'm trying to improve the situation at hand, not perfect it

gausie:
Legit, I had no idea how this pref worked

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -2132,7 +2132,8 @@ public boolean canAdventure() {
 
     if (this.zone.startsWith("PirateRealm")) {
       // *** You have limited turns available per day.
-      return Preferences.getBoolean("prAlways") || Preferences.getBoolean("_prToday");
+      return (Preferences.getBoolean("prAlways") || Preferences.getBoolean("_prToday"))
+          && Preferences.getString("_LastPirateRealmIsland").contains(this.adventureName);
```
midgleyc, line 2136: is there a case where `equals` wouldn't work?

midgleyc:
also I kind of feel like we want to fix that leading L. Not sure how many scripts are using this thing. If we assume they're on GitHub might be "literally just yorick".

https://github.com/search?q=_LastPirateRealmIsland&type=code&p=1

src/net/sourceforge/kolmafia/KoLAdventure.java
horrible-little-slime: Not that I'm aware of, but I'm yet to check logs nontrivially

horrible-little-slime:
> also I kind of feel like we want to fix that leading L. Not sure how many scripts are using this thing. If we assume they're on GitHub might be "literally just yorick".
> 
> https://github.com/search?q=_LastPirateRealmIsland&type=code&p=1

Is PirateRealm recently enough that we think most scripts that touch it would be on github rather than sourceforge?

src/net/sourceforge/kolmafia/KoLAdventure.java
horrible-little-slime: update: @DamianDominoDavis just scraped some logs and `equals` is good

taltamir:
a workaround until this is merged is to use mafia version 27304
https://github.com/kolmafia/kolmafia/releases/tag/r27304

gausie (approved):

PR #1657, opened 2023-04-01
Wine Glass works in Shadow Rift locations

Veracity0:
Shadow Rift locations are place.php that redirects to adventure.php?snarfblat=567 (that further redirects to fight.php or choice.php).

Drunkula's wineglass works fine there.


midgleyc (approved):
Thanks!

PR #1661, opened 2023-04-04
parse Rufus quest completion details from questlog

Veracity0:


midgleyc (approved):
Thanks! Will be useful next April 1 :D

PR #1663, opened 2023-04-06
Detect speakeasyName correctly given KoL change to town_wrong map

Veracity0:
I wondered why speakeasyName was set to "Speakeasy", not "Veracity's Place".
Turns out that KoL changed the town_wrong map and the name is in its own div now.


midgleyc (approved):
Thanks!

PR #1666, opened 2023-04-07
Track cursed monkey paw wishes used

Veracity0:
Property _monkeyPawWishesUsed starts at 0 and goes up to 5 as you cast wishes.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/MonkeyPawRequest.java
+++ b/src/net/sourceforge/kolmafia/request/MonkeyPawRequest.java
@@ -77,7 +77,7 @@ public class MonkeyPawRequest extends GenericRequest {
 
   // You acquire an effect: <b>Cursed by a Monkey</b><br>(duration: 7 Adventures)
   private static final Pattern CURSE_PATTERN =
-      Pattern.compile("<b>Cursed by a MOnkey</b>.*?(duration: (\\d+) Adventures)");
+      Pattern.compile("<b>Cursed by a Monkey</b>.*?(duration: (\\d+) Adventures)");
 
   public static void postChoice(final String responseText, final String wish) {
     if (!responseText.contains("Wish granted.")) {
--- a/src/net/sourceforge/kolmafia/request/MonkeyPawRequest.java
+++ b/src/net/sourceforge/kolmafia/request/MonkeyPawRequest.java
@@ -77,7 +77,7 @@ public class MonkeyPawRequest extends GenericRequest {
 
   // You acquire an effect: <b>Cursed by a Monkey</b><br>(duration: 7 Adventures)
   private static final Pattern CURSE_PATTERN =
-      Pattern.compile("<b>Cursed by a Monkey</b>.*?(duration: (\\d+) Adventures)");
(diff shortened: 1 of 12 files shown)


src/net/sourceforge/kolmafia/request/MonkeyPawRequest.java
```diff
@@ -0,0 +1,112 @@
...
+    Matcher matcher = MonkeyPawRequest.WISH_PATTERN.matcher(responseText);
+    if (matcher.find()) {
+      int wishesLeft = Integer.parseInt(matcher.group(1));
+      Preferences.setInteger("_monkeyPawWishesUsed", 5 - wishesLeft);
+    }
+  }
+
+  // You acquire an effect: <b>Cursed by a Monkey</b><br>(duration: 7 Adventures)
+  private static final Pattern CURSE_PATTERN =
+      Pattern.compile("<b>Cursed by a MOnkey</b>.*?(duration: (\\d+) Adventures)");
```
midgleyc, line 80:
```suggestion
      Pattern.compile("<b>Cursed by a Monkey</b>.*?(duration: (\\d+) Adventures)");
```

Perhaps a test for this one?

Veracity0: It should actually be tested, although not explicitly:

1) when we visit the monkey paw, we parse the number of fingers and set _monkeyPawWishesUsed
2) when we submit a successful wish, we increment it.
3) and then, we double check by parsing the curse and (we assume) setting the property to the derived value.

If the parsing the visit didn't work, the increment would be inaccurate.
- But the test before I fixed the typo in the pattern gave the expected results.
Now that I fixed the typo, the pattern should match.
- The tests still pass, since the property ends up as expected.

I suppose there could be a test which skips visitChoice and just calls postChoice and verifies it parses the curse effect.

Veracity0: Or, perhaps both of my parameterized tests could visit + submit a wish and the first one, which uses GenericRequest could zero out the property before submitting making the wish.

The second one can't be split like that, since it uses an actual MonkeyPawRequest which does both requests. I.e. it tests the automation, not just choice handling.

midgleyc: I don't understand how `usingMonkeyPawDetectsWishesUsed` managed to pass while the regex was wrong. It set the preference to 0, then hit the HTML, and then the preference was right -- but looking at the code, it should have just always been 1?

Veracity0: Because it parsed the number of fingers in the visit of the intial request.
After submitting the second request to make the wish, it just incremented.

midgleyc (dismissed):
As it parses on the choice screen, this is likely fine.

Veracity0:
> As it parses on the choice screen, this is likely fine.

Yeah, parsing the Curse is redundant.

Regardless, I updated the test to verify that we successfully parse Cursed by a Monkey.

PR #1671, opened 2023-04-13
the citadel quest grants access to the distant woods

horrible-little-slime:
This is mostly relevant for paths that don't follow the traditional council quests, like Community Service or Grey Goo.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -1102,6 +1102,7 @@ public class KoLAdventure implements Comparable<KoLAdventure>, Runnable {
         case AdventurePool.BARROOM_BRAWL -> QuestDatabase.isQuestStarted(Quest.RAT);
           // prepareForAdventure will visit the Crackpot Mystic to get one, if needed
         case AdventurePool.PIXEL_REALM -> QuestDatabase.isQuestStarted(Quest.LARVA)
+            || QuestDatabase.isQuestStarted(Quest.CITADEL)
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -888,7 +888,8 @@ public boolean canAdventure() {
           // Right Side of the Tracks and The Nearby Plains
         case TOWN, PLAINS -> true;
           // The Distant Woods, Forest Village, The 8-Bit Realm
-        case WOODS, VILLAGE, REALM -> QuestDatabase.isQuestStarted(Quest.LARVA);
+        case WOODS, VILLAGE, REALM -> QuestDatabase.isQuestStarted(Quest.LARVA)
+            || QuestDatabase.isQuestStarted(Quest.CITADEL);
```
midgleyc, line 892: given this appears in a bunch of places (that you've just added), could we extract to a function like `isWoodsOpen` (that checks LARVA or CITADEL) and call that?

midgleyc (dismissed):
Thanks!

src/net/sourceforge/kolmafia/KoLAdventure.java
```diff
@@ -539,6 +539,10 @@ public boolean tooDrunkToAdventure() {
     return true;
   }
 
+  private static Boolean woodsOpen() {
```
Veracity0, line 542: No need to box the return value; a simple "boolean" is fine.

Veracity0 (changes requested):

PR #1672, opened 2023-04-14
feat: monkey's paw banish / track

midgleyc:
Pointing at a monster adds 2 copies, crosses rollover.

Slapping is a banish, resets at rollover. Track both.

Reading the pointed monster from the description doesn't work perfectly because some monsters can contain dots, and we can't be sure where the sentence ends. But I think it's probably better than not doing that.

### DIFF
--- a/src/data/consequences.txt
+++ b/src/data/consequences.txt
@@ -86,7 +86,7 @@ DESC_ITEM	industrial fire extinguisher	The extinguisher's tank is currently <b>(
 DESC_ITEM	designer sweatpants	Your sweatpants are currently (\d+)% sweat-logged.	sweat=$1
 DESC_ITEM	Powerful Glove	The Glove's battery is (\d+)% charged.	_powerfulGloveBatteryPowerUsed=[100-$1]
 DESC_ITEM	Crimbo training manual	all that's left is Chapter (\d+).	crimboTrainingSkill=$1
-DESC_ITEM	cursed monkey's paw	monkey's paw recently pointed at a ([^.]+).	monkeyPointMonster=$1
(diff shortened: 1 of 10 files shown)


src/data/consequences.txt
```diff
@@ -86,6 +86,7 @@ DESC_ITEM	industrial fire extinguisher	The extinguisher's tank is currently <b>(
 DESC_ITEM	designer sweatpants	Your sweatpants are currently (\d+)% sweat-logged.	sweat=$1
 DESC_ITEM	Powerful Glove	The Glove's battery is (\d+)% charged.	_powerfulGloveBatteryPowerUsed=[100-$1]
 DESC_ITEM	Crimbo training manual	all that's left is Chapter (\d+).	crimboTrainingSkill=$1
+DESC_ITEM	cursed monkey's paw	monkey's paw recently pointed at a ([^.]+).	monkeyPointMonster=$1
```
Veracity0, line 89: What is the full html for how this appears on the page? Assuming it is a sentence ending in a period, is it followed by a space (or two, as is Jick's wont) or an HTML tag starting with a <?

I.e. perhaps a stronger regexp could be more accurate.

Veracity0 (changes requested):
Tests?

src/data/consequences.txt
midgleyc: unfortunately it can be followed either by the item id (if you haven't banished anything) or by the banish message (if you have).

midgleyc: wait, I'm wrong: the banish message comes first. So this should be possible.

horrible-little-slime: Can you look for the newline instead of looking for the period?

Veracity0 (approved):

PR #1673, opened 2023-04-16
non-C.H.U.M. chum items are no-pickpocket

Shiverwarp:
From information on the wiki, as well as my own testing via autumnaton

https://kol.coldfront.net/thekolwiki/index.php/C._H._U._M.

### DIFF
--- a/src/data/monsters.txt
+++ b/src/data/monsters.txt
@@ -1038,7 +1038,7 @@ X-dimensional horror	592	dimhorror.gif	NOCOPY Atk: [2*ceil(BL^1.4)+ML] Def: [2*c
 
 # Hobopolis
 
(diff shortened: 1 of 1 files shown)


Veracity0:
Aren't they actually conditional (based on the image of the C.H.U.M. - i.e. holding a knife or holding a lantern)? That by itself makes them no pickpocket.

It also means that since there are 5 images, if the lantern (1/5 images) and knife (3/5 images) are base 1% each, the knife is effectively 0.6% and the lantern is effectively 0.2%.

Veracity0 (changes requested):
The equipment, at least, is conditional, not simply no pickpocket.

Shiverwarp:
Awesome, just checked with Grey Goose to confirm that all the other items aren't duped by drones either. 

(I think this is the best way to determine both conditional and no-pickpocket. Autumnaton will get items that are conditional, but not nopp. Goose will only copy non-conditional items)

I'm not 100% sure of the syntax for items that have multiple types, I don't seem to see others that have multiple so maybe this is not the correct way for me to identify them. According to CF40 on the ASS discord they are seperate : [here](https://discord.com/channels/466605739838930955/474736566246572042/1027287719497441441)

Veracity0 (approved):

PR #1675, opened 2023-04-17
decrement encountersUntilSRChoice at end of combat, not start

horrible-little-slime:
This is failing tests right now. I need to get the two tests it's failing (and any potential other tests I add) to process the end of combat with all the monsters in question, but I don't immediately see a way to do that, so this is a draft for now.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/RufusManager.java
+++ b/src/net/sourceforge/kolmafia/session/RufusManager.java
@@ -525,7 +525,6 @@ public class RufusManager {
 
   public static void handleShadowRiftFight(MonsterData monster) {
     int currentTurns = Preferences.getInteger("shadowRiftTotalTurns");
-    int lastNC = Preferences.getInteger("shadowRiftLastNC");
     switch (monster.getName()) {
       case "shadow spire",
           "shadow orrery",
@@ -535,14 +534,12 @@ public class RufusManager {
           "shadow matrix" -> {
         // Bosses are NCs
         Preferences.setInteger("shadowRiftLastNC", currentTurns);
-        lastNC = currentTurns;
+        Preferences.resetToDefault("encountersUntilSRChoice");
       }
     }
 
     // All fights - including bosses - advance turns in zone
     currentTurns = Preferences.increment("shadowRiftTotalTurns", 1);
-    // Next NC comes if (turns in zone - last nc) >= 11
-    Preferences.setInteger("encountersUntilSRChoice", 11 - Math.max(currentTurns - lastNC, 0));
   }
 
   public static void handleShadowRiftNC(int choice, String responseText) {
@@ -553,13 +550,12 @@ public class RufusManager {
       case 1499 -> {
         // The Shadow Labyrinth
         Preferences.setInteger("shadowRiftLastNC", currentTurns);
-        lastNC = currentTurns;
+        Preferences.resetToDefault("encountersUntilSRChoice");
       }
       case 1500 -> {
         // Like a Loded Stone
         ResultProcessor.removeItem(ItemPool.RUFUS_SHADOW_LODESTONE);
       }
     }
-    Preferences.setInteger("encountersUntilSRChoice", 11 - Math.max(currentTurns - lastNC, 0));
   }
 }
--- a/src/net/sourceforge/kolmafia/request/FightRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FightRequest.java
@@ -3375,6 +3375,10 @@ public class FightRequest extends GenericRequest {
       Preferences.decrement("encountersUntilNEPChoice", 1, 0);
     }
 
+    if (adventure == AdventurePool.SHADOW_RIFT) {
+      Preferences.decrement("encountersUntilSRChoice", 1, 0);
+    }
+
     if (monsterName.equals("unusual construct")) {
       ResultProcessor.removeItem(ItemPool.STRANGE_DISC_WHITE);
       ResultProcessor.removeItem(ItemPool.STRANGE_DISC_BLACK);


Veracity0:
[This post](https://kolmafia.us/threads/new-shadow-rift-ncs-buffs-and-buff-extension.28712/#post-171580) from Malibu Stacey quotes cannonfire40 (a dev):

> cannonfire said:
>> fwiw
>> it's not turnsinzone % 11 == 0 precisely, it's (turnsinzone - lastnc) >= 11
>> which may help your intuitions
>> (this is how all scheduled NC forcing works)
>> also, the bosses are NCs.

> and
>> the lodestone is a superlikely
>>not an NC

> when asked if the "Like a Loded Stone" encounter resets the schedule too.

> Also
>>because the NC does not advance turnsinzone
but bosses do

I chose to implement the counter using exactly the formula that cannonfire40 said was used by KoL itself.

I am surprised that you chose to eliminate that.

src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -535,14 +534,12 @@ public static void handleShadowRiftFight(MonsterData monster) {
           "shadow matrix" -> {
         // Bosses are NCs
         Preferences.setInteger("shadowRiftLastNC", currentTurns);
-        lastNC = currentTurns;
+        Preferences.resetToDefault("encountersUntilSRChoice");
       }
     }
 
     // All fights - including bosses - advance turns in zone
     currentTurns = Preferences.increment("shadowRiftTotalTurns", 1);
```
Veracity0, line 542: Since you threw away the calculation that uses currentTurns, you do not need to set currentTurns.

src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -525,7 +525,6 @@ private static String shadowLabyrinthChoiceDecision(String responseText) {
 
   public static void handleShadowRiftFight(MonsterData monster) {
     int currentTurns = Preferences.getInteger("shadowRiftTotalTurns");
-    int lastNC = Preferences.getInteger("shadowRiftLastNC");
```
Veracity0, line 528: This eliminates the first "get" of shadowRiftLastNC.

src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -535,14 +534,12 @@ public static void handleShadowRiftFight(MonsterData monster) {
           "shadow matrix" -> {
         // Bosses are NCs
         Preferences.setInteger("shadowRiftLastNC", currentTurns);
```
Veracity0, line 536: This is the first "set" of shadowRiftLastNC.

src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -553,13 +550,12 @@ public static void handleShadowRiftNC(int choice, String responseText) {
       case 1499 -> {
         // The Shadow Labyrinth
         Preferences.setInteger("shadowRiftLastNC", currentTurns);
```
Veracity0, line 552: This is the second "set" of shadowRiftLastNC.

src/net/sourceforge/kolmafia/session/RufusManager.java
```diff
@@ -553,13 +550,12 @@ public static void handleShadowRiftNC(int choice, String responseText) {
...
-        lastNC = currentTurns;
+        Preferences.resetToDefault("encountersUntilSRChoice");
       }
       case 1500 -> {
         // Like a Loded Stone
         ResultProcessor.removeItem(ItemPool.RUFUS_SHADOW_LODESTONE);
       }
     }
-    Preferences.setInteger("encountersUntilSRChoice", 11 - Math.max(currentTurns - lastNC, 0));
```
Veracity0, line 563: Since you no longer use lastNC, you no longer need to get shadowRiftLastNC.

This PR eliminates all "gets" of shadowRiftLastNC.
Therefore, you no longer need either of the "sets".
Therefore, you no longer need the property.

Veracity0 (changes requested):
Consider this:

You eliminated all "gets" of shadowRiftLastNC.
Therefore, you no longer need to set it.
Therefore, you no longer need the property.

Every "get" of shadowRiftTotalTurns is used in a calculation you removed.
Therefore, you no longer need to set it.
Therefore, you no longer need the property.

You have replaced cannonfire40's "KoL's internal" calculation and replaced it with something else.

I guess I am not surprised that the new implementation does not pass the existing tests, which use actual response texts from KoL to track the properties which this PR eliminates.

Since you eliminated both properties, you could remove the lines from the tests that test those properties, and look only at the "encountersUntilSRChoice" property.

I'd like to see a test that illustrates the reported issue and a formerly failed but works, with your fix.

Veracity0:
I think PR https://github.com/kolmafia/kolmafia/pull/1676 does this with a much less invasive change.
I happened to have DEBUG logs lying around with the end of a fight with a shadow monster and with a shadow boss.
With those, I could test that we are now incrementing turns at the end of the fight, not the start.

I suppose that DEBUG logs of ending a fight via Using the Force could also verify that the "end of fight" code was not running and the counters changing, but I had no such DEBUG logs available today.

Maybe tomorrow.

horrible-little-slime:
Looks like your PR is handling this in a much more elegant way--great work! I have no idea why I didn't think to just move the handleRufusCombat function to be end of combat instead of start, that's an extremely good solution. 

Are we able to use CLEESH or Macrometeor against bosses? Handling the NC reset at end of combat could be troublesome if we can, but I expect that as bosses they're immune to monster replacement. Either way, that's extremely surmountable

With respect to this:
> it's not turnsinzone % 11 == 0 precisely, it's (turnsinzone - lastnc) >= 11
I don't currently see a meaningful difference between that and merely decrementing the property; the quote there seems to clarify that you can't "miss" a noncombat by hitting a wanderer or analogous disruption at the right moment. Decrementing the property and resetting it to 11 when you hit an appropriate NC should also replicate that behavior, but without the intermediary properties. But there could very well be something big that I'm missing here that makes these more different than I assumed.

Veracity0:
You have a point about "no meaningful difference".

And it would be an elegant code change to eliminate the intermediate properties and just do it the way we do "encountersUntilNEPChoice", say.

We have tests for "encountersUntilSRChoice".

Should be a "simple matter of coding" to remove the lines from the tests that deal the intermediate properties, change to the simpler implementation of "encountersUntilSRChoice", and verify that the tests still pass.

That could be a separate PR.

PR #1678, opened 2023-04-19
Simplify calculation of encountersUntilSRChoice

Veracity0:
Per PR https://github.com/kolmafia/kolmafia/pull/1675 by @horrible-little-slime we don't need to save intermediate counters in properties in order to accurately track when the next Shadow Labyrinth or shadow boss will occur.

I removed the two obsolete properties and added an array of such properties to Preferences. We could use Preferences.removeProperty to get rid of those - at some point. I'm not sure when. Startup? Rollover? Ascension?

I'm concerned about people upgrading, having those properties removed, running for a while, and then downgrading to a version that still uses them. Now, if they depended on the downgraded version having correct values for properties which are no longer updated, tough luck. But it would be nice to unclutter the prefs files.

I expect there are scores of other similarly unused properties that KoLmafia has abandoned.

After we decide on our policy, another PR might like to add all those properties to the new array and clean things up.


horrible-little-slime:
At the very least, I know `feelNostalgicMonster` has been made redundant by `lastCopyableMonster` and has been discontinued as such

jaadams5 (approved):

PR #1679, opened 2023-04-19
Fix set_location($location[none])

Veracity0:
Per [This report](https://kolmafia.us/threads/allow-set_location-location-none-to-set-your-location-to-none.28807/), ASH set_location(location[none]) does not set KoL's idea of the current location.

I can't think of any reason a script would depend on that behavior, so I changed it to actually do (more or less) what KoL itself will do with KoLAdventure.setNextAdventure("None") - which KoL itself calls when it determines that using an item will lead to a fight; the fight is (usually) not deemed to occur in a particular location.


midgleyc (approved):
The PR title should presumably refer to `set_location`, not `get_location`.

I think this is good. I used to use `$location[none]` before finding it didn't work, and it would be nice to use it again.

PR #1681, opened 2023-04-23
Track ghost pepper and Gets-You-Drunk timers for bonus adventures

gausie:
These two consumables from dread are consumed as normal and then if you can survive 3 end-of-fight elemental onslaughts you get a bonus pile of adventures. You can't consume another one of these while the last one is still being digested. If the after-combat damage (which happens after everything it seems - no HP regen still left me with 0 HP) defeats you, the timer ends. If you have the appropriate elemental form, the timer ends.

- Write tests for eating ghost pepper
- Add another test for maximum uses
- Track ghost pepper consumption
- Add tests for tracking ghost pepper in inventory
- Copy logic from ghost peppers to gyd and iterate
- Complete fight request tracking of dread consumables


midgleyc (approved):
Thanks!

One thing to check: you reset the preference to 0 if we don't see the "The mulled wine you drank earlier" text. However, does this appear in case of haiku / anapests?

gausie:
> Thanks!
> 
> One thing to check: you reset the preference to 0 if we don't see the "The mulled wine you drank earlier" text. However, does this appear in case of haiku / anapests?

Trying to work out if I fixed this problem before or after your comment! But it's sorted

PR #1682, opened 2023-04-23
feat: new halloween candies

midgleyc:
Add item records for shadow candy, extra-hard jawbreaker and chiffon lemon.

The latter two give two effects: Delayed Gratification with descid `2faeaf79c4b716a2d9fd7ac2e8fffdc7` and Bulging Jaw Muscles with descid `cd5ee49e0f2be8eae700a2ef1fddee0a`, but I haven't added those because neither the candies nor the effects are public yet, so we can't see what they do.


jaadams5 (approved):
But if the candies and the effects are not public then why are we doing this now?

midgleyc:
You can get the candies, even though they're not public. They appear in the CLI as unrecognised items.

PR #1684, opened 2023-04-27
feat: some combat rates from wiki

midgleyc:
Marked zones as 100% combat where the non-combats are scheduled (e.g. every X adventures) and non-combat doesn't help to get them faster.

Pretty sure Madness Bakery, Ice Hotel, VYKEA should be in this set but I can't find any proof of it.

Mostly just add data where there was no data but also change Inside the Palindome and The Black Forest.


Veracity0 (approved):
Thanks!

PR #1685, opened 2023-05-02
Basic Cincho de Mayo tracking

gausie:
Done:
- [x] Skills
- [x] Items
- [x] Effects
- [x] Track salt and lime being dispensed
- [x] Track cinch being restored via resting, and number of restores that day
- [x] Add free rests

We want to support tracking this NC forcing too, though that probably makes sense to be tied in to the new "Adventure Modifiers" section in the charpane


src/net/sourceforge/kolmafia/request/UseSkillRequest.java
```diff
@@ -583,19 +585,6 @@ public long getMaximumCast() {
           int birds = Preferences.getInteger("_birdsSoughtToday");
           return (birds < 7 && option != 1) ? 6 - birds : Long.MAX_VALUE;
         }
-
-      case SkillPool.INVISIBLE_AVATAR:
```
midgleyc, line 587: this block has gone away and doesn't seem to be replaced with anything: does what it did still work?

midgleyc (dismissed):
the main comment is addressed; the `getMaximumCasts` comment still applies. I'm not actually sure what that function is used for, though.

horrible-little-slime:
As a user, I think this would make a little more sense counting `_cinch` down instead of `_cinchUsed` up, both because I feel like I'm more interested in "do I have enough cinch to cast this skill", and because there's something that feels off about the idea of un-using cinch when you rest (as opposed to replenishing cinch).

At the end of the day it doesn't matter--I'm fully capable of doing subtraction--it just strikes me as a little more intuitive to make this like `sweat` instead of `_powerfulGloveBatteryPowerUsed`

gausie:
> free resting at campsite gives the message "While you rest, your cincho loosens 30%." and should probably adjust `_cinchUsed`? I don't think it can go over 100% though.

I didn't even see your comments while I was working away at solving this!

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
Veracity0: Was there a before and after test demonstrating this?

Can there be?

PR #1686, opened 2023-05-02
Support cincho expansion in other resting zones

gausie:


midgleyc (approved):
At first I thought "this should only trigger on free rests" but because we check for the message it doesn't matter.

Thanks!

gausie:
It triggers for all rests!

PR #1689, opened 2023-05-05
track cincho used as we cast combat skills

horrible-little-slime:
This sort of extends the stuff we have in consequences.txt; it lets us track cinch when you run out of cinch (and thus the skills no longer appear for consequences.txt) or when you submit combat in one big macro (and thus you never actually get to look at the skills dropdown to see what your cinch is).

The fight html used is one that brought someone to maximum cinch, making it so the skills no longer appear in the dropdown.


horrible-little-slime:
I'll get started on that once I have test html. That may be this weekend rather than today.

midgleyc (approved):
Thanks!

PR #1690, opened 2023-05-06
Combat data rates for The Battlefield

Rinn:
- Handle monsters that only appear after a fixed number of kills
- Handle heroes that appear in ranges, return -1 as the rate to denote that the hero is possible but otherwise doesn't affect normal encounter rates (superlikelies with a very low chance) 
- Return boss as the only encounter when battlefield is cleared

### DIFF
--- a/src/net/sourceforge/kolmafia/AreaCombatData.java
+++ b/src/net/sourceforge/kolmafia/AreaCombatData.java
@@ -1399,8 +1399,8 @@ public class AreaCombatData {
               ? 1
               : 0;
             // After specific number of hippies defeated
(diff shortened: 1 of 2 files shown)


midgleyc (dismissed):
By wiki (via Aen via bobson) nearly all the frat war monsters appear in certain ranges:

https://kol.coldfront.net/thekolwiki/index.php/The_Battlefield_(Frat_Warrior_Fatigues)#Monster_Ranges_.28Non-Heroes.29

May as well add these too while you're there?

Rinn:
man they did not make this data easy to find and it looks like there isn't fratboy data either

Rinn:
someone sent me a spading spreadsheet, I'll implement the data using that

midgleyc (approved):
Thanks!

PR #1691, opened 2023-05-07
Crimbo Training Skills have a ":" in the name (Crimbo Training: Barte…

BadHorseMonkey:
…nder).  The ash command numeric_modifier relies on getModifierName, which takes a substring of "name" if there's a ":".  This is a very specific change to allow Crimbo Training to return properly.

If KoL adds additional Skills or modifiers with ":" in the name, we'll need a more general solution (or at least to move our exception in the opposite direction).

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -9204,11 +9204,13 @@ public abstract class RuntimeLibrary {
     Type type = arg.getType();
     String name = arg.toString();
     int id = (int) arg.intValue();
-    if (type.equals(DataTypes.ITEM_TYPE) || type.equals(DataTypes.EFFECT_TYPE)) {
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -9208,7 +9208,7 @@ private static String getModifierName(final Value arg) {
       return "[" + id + "]";
```
midgleyc, line 9208: instead of special-casing, you can add `SKILL_TYPE` into this `if` to return the `id` directly.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
BadHorseMonkey: good catch,

midgleyc (approved):

PR #1693, opened 2023-05-08
feat: more git scripts

midgleyc:
Add some more scripts that are only installed through git.

Change some GitHub scripts which are installed through SVN to use git.

### DIFF
--- a/data/SVN/svnrepo.json
+++ b/data/SVN/svnrepo.json
@@ -1,6 +1,7 @@
 [
  {
-   "repo": "https://github.com/Rinn/kolmafia-scripts/trunk/AcquireBuff/",
+   "repo": "https://github.com/Rinn/kol-acquirebuff",
+   "type": "git",
    "author": "Rinn",
    "name": "acquireBuff",
    "forumThread": "https://kolmafia.us/showthread.php?4048-acquireBuff-Get-a-buff-from-a-buffbot",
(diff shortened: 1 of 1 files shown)


jaadams5 (changes requested):
Could we hold the svn to git for a moment?  I have local changes to the svn version of Excavator and no one has been able to explain to me exactly how to migrate those changes to a local git repository using the subset of git commands that I am actually comfortable with.  The more general issue of migrating local changes from svn to git probably should be documented by someone with more comfort with the process than me.

midgleyc:
This doesn't actually migrate any svn repositories to git -- it just changes our list of repositories so that we "know about" the git versions instead of the SVN versions. Any future installs will be git, but current installs aren't affected.

To migrate your local changes (all file paths are relative to Mafia root):
* you currently have an excavator folder in `svn/`. It has a `scripts/` folder, and files in this folder are copied to the main `scripts/` folder.
* you have local changes in the `svn/gausie-excavator` folder
* you should checkout the git version of excavator in the Mafia CLI: `git checkout gausie/excavator`. This will download files to `git/gausie-excavator` and override the version in `scripts/`
* you should copy your local changes from `svn/gausie-excavator`. Just dump the entire `svn/gausie-excavator/scripts` folder on top of `git/gausie-excavator/scripts`
* run `git sync excavator` in the Mafia CLI. Confirm that your `scripts/` folder looks the way you expect and that excavator behaves the way you expect
* run `svn delete excavator` to remove the SVN version

jaadams5:
Whatever I remembered was not that easy but OK.

jaadams5 (approved):

PR #1695, opened 2023-05-10
Cincho tracking fixes

horrible-little-slime:
- Confetti extravaganza is once per fight and so it's not an optimal candidate for consequences.txt
 - Testing that cinch went from 95 to 100 wasn't reliable because it didn't account for overshooting our cinch counts
 - setting `skillSuccess` to true meant we incremented cinch as part of our dailylimit handling as well as the manual preference tracking previously included; this has been rectified.


midgleyc (approved):
Thanks!

PR #1696, opened 2023-05-10
Validate location of Sot's Parcel

jaadams5:
This will confirm that the string extracted for the Sot's Parcel's location is a valid adventure location via exact match or a unique location via a fuzzy match.  If this is not the case then the location will not be set.

If this ever happens it is likely to be an error in the regex processing or KoL and KoLmafia not agreeing on adventure names.

The test is made by editing captured HTML so that the extracted string is not an exact adventure name but has a unique fuzzy match.


Veracity0:
Looking at your test, if KoL says "The Hippy Camp", this PR maps it into "The Hippy Camp (Bombed Back to the Stone Age)".

Considering that the original report of seeing "The Hippy Camp" was post-war where the hippies won, that will be incorrect in that case - and, in fact, for all the other situations where KoL might refer to "The Hippy Camp".

I suggest that fuzzy matching is incorrect.

Since the Sot's text does not include a snarfblat, perhaps you should disambiguate via special code that deduces which snarfblat is actually available?

jaadams5:
I added a canAdventure check and tweaked tests.

I confirmed to my satisfaction that getAdventure's fuzzy matching is the issue.  

As suggested I will explore a way to go from a string to an available snarfblat.

jaadams5:
I decided the form was "The <location name>" or <location name> if the location name already included a "The".  I look for an exact match since the string is coming from KoL, not a user, and a fuzzy match but not an exact one could be a problem with KoLmafia's data.  If I get an exact location, perhaps with a leading "The" removed then I confirm that it can be adventured in before setting the preference.

I tweaked some tests and added a couple more.

midgleyc:
I feel like this isn't the right way to go about this. The issue is that some of Mafia's names are different to KoL's names, not that sometimes a starting "The" needs stripping.

I think what you want to be doing is something like:
* look through AdventureDatabase.allAdventures for ones with matching KoL name.
* if we can adventure in them right now, it's that one. Note that the current `canAdventure` also deals with "stuff we can prepare for" -- we don't want that, we want exactly the ones we can adventure with the current character state

and you need to find the ones with different KoL names (which you can check by seeing whether the autumn-aton dropdown matches the Mafia name). I know Hippy Camp, Arid Desert and Video game levels are affected.

jaadams5:
I disagree with the issue as defined in the comment.

We have a regex that sometimes returns more than the name.  My regex ability is limited and I cannot figure out a regex that sometimes includes "The" because it is part of the name but sometimes excludes it because it was added just to make the English read better.

So if I don't have an exact match on the extracted text and the extracted text starts with "The" I strip it off and try for an exact match.  At one point there was concern that the issue was with different names in KoL and KoLmafia  but that is no longer believed to be the case.

I'm not doing fuzzy matching anymore so either the extracted string is an exact match or the extracted string without the leading "the" is an exact match.

I perhaps misunderstand canAdventure().  I thought all of the side effects had been moved to prepare_for_adventure (or whatever it is named).  It is only being called for an exact location.  Calling canAdventure() is arguably a defensive move.  KoL thinks the location can be adventured in and so canAdventure() will always be true unless there is a KoL bug or an error in KoLmafia's adventure database.

If canAdventure has side effects for an exact match to a location that KoL already thinks can be adventured in then that is a reason not to call it.

Thanks for commenting.

midgleyc:
I don't think the regex returns more than just the name. I think the regex returns exactly what KoL thinks the zone is called.

For example, Mafia thinks the five Hippy Camp locations are called:
* Hippy Camp
* Hippy Camp (Hippy Disguise)
* Wartime Hippy Camp
* Wartime Hippy Camp (Frat Disguise)
* The Hippy Camp (Bombed Back to the Stone Age)

However, KoL thinks the five Hippy Camp locations are called:
* The Hippy Camp
* The Hippy Camp
* The Hippy Camp
* The Hippy Camp
* The Hippy Camp (Bombed Back to the Stone Age)

(as you can see in the autumn-aton list)

So it's not that KoL sometimes adds a "The" -- it's that the Mafia name incorrectly omits the "The". "The Hippy Camp" might be the wartime or non-wartime Hippy Camp -- you can't tell from the face of it. However, I think the Sot will only send you to locations you can currently adventure in, and only one Hippy Camp exists at a time.

Mafia's `canAdventure` tells you whether a request to adventure in a particular location in Mafia will succeed -- but this includes things that `prepareForAdventure` will do, that the player hasn't done yet. If you have 3 sonar-in-a-biscuit in inventory and haven't yet opened the Boss Bat's Lair, `canAdventure` will return true even though you can't adventure there yet because trying to adventure there will use the sonars.

jaadams5:
OK.  I stand enlightened and corrected.  I'll go back to draft while I ponder and wish we had had this conversation when I did the parsing originally.  thank you.

jaadams5:
I'm going to close this without pushing.  It solves the wrong problem and that problem is of my own making.

I expect to be able to construct a failing test which will identify an issue when KoLmafia is not using the exact location name.

PR #1697, opened 2023-05-11
Support more quirks that the comma chameleon has with familiars it can become

gausie:
User Heulen has been doing a bunch of comma spading and confirmation of exactly how it behaves for different familiars, which he has been transferring to the wiki but which I would also like to reflect in mafia.

Comma code simplified a little here, but also added some tests confirming that mafia behaves in line with behavior observed in KoL around how the comma works. I'm sure there will be more to support here in future but this is all accurate at least


midgleyc (approved):
Thanks! I suspect the libram people will be happy to not maintain this any more :P

PR #1698, opened 2023-05-12
Include chateau items in modifiers

gausie:
Despite being in modifiers.txt, artificial skylight and antler chandelier mods  were never actually applied. This fixes that


Veracity0 (approved):
Thanks!

PR #1701, opened 2023-05-14
Chateau items keep their mods in TCRS, just like Campground items

gausie:
Plus a small refactor to the latter to use a Set


jaadams5 (dismissed):
Approved.  But for extra credit you might confirm that when you start with an empty test/root/data/ and run tests that you end up with an empty one.  TCRS_Seal_Clubber_Mongoose.txt and two other related files persist.

gausie:
> Approved. But for extra credit you might confirm that when you start with an empty test/root/data/ and run tests that you end up with an empty one. TCRS_Seal_Clubber_Mongoose.txt and two other related files persist.

I'm going to remove these files entirely soon! How's that for extra credit ;)

PR #1705, opened 2023-05-16
feat: can only access shadow rifts with payphone or ASoL path

midgleyc:
> May 14 - The shadow rifts have faded into the... darker shadows?

The Shadow Rifts are still available in ASoL, or if you have the payphone in inventory. Closet doesn't count. I presume other places also don't count.


Veracity0 (approved):
To be clear: freeing the king in AoSoL (which leaves you with the path-specific class, but leaves the path) closes the rifts?

midgleyc:
Based on a second-hand comment in Discord: yes.

If it turns out not, we can change it.

PR #1707, opened 2023-05-16
feat: initial legacy of loathing support

midgleyc:
Add path, dump initial items.

Enable Cincho skills even when we only have the replica.

I've disabled the consistency check for familiar hatchlings for replica familiars, because they "grow" to one of the existing familiars, and I think we want the "hatchling" of those familiars to still be the original one.


gausie:
Could we customise the consistency check to account for these specific replica items?

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
```diff
@@ -730,34 +734,50 @@ private static boolean canSwitchToItem(final AdventureResult item) {
         && InventoryManager.hasItem(item, false);
   }
 
-  private static void equipForSkill(final AdventureResult item, final int skillId) {
-    if (!KoLCharacter.hasEquipped(item)) {
-      if (!InventoryManager.retrieveItem(item)) {
-        KoLmafia.updateDisplay(MafiaState.ERROR, "Cannot acquire " + item.getName() + ".");
+  private static void equipForSkill(final int skillId, final AdventureResult... items) {
```
gausie, line 737: I would love to see a test for this function (or an appropriate one that calls it) if you have time

midgleyc:
Updated. Also added what I hope is a perf improvement: we can only get the replica if we're in the Legacy of Loathing path, so don't even check for it outside.

I'm not sure if we can get the regular one inside -- can you break ronin and then pull equipment? -- so I've left checking even if we're in path.

midgleyc:
Also updated to fix the familiar situation. Legacy of Loathing's interaction with standard is a bit tricky for us to handle -- it seems that various (replica) IotM-derived items are usable, but they still show up in standard.php as being unusable.

gausie (approved):

midgleyc:
I ran this in LoL yesterday and it worked fine. I think it's good to get it in the hands of autoscenders so we can identify everything else this path still needs ;)

PR #1708, opened 2023-05-17
ChibiBuddy CLI command

gausie:
For now the command can only do two things:

- Without any params, it displays the current status of the ChibiBuddy
- The param "chat" will get you the ChibiChanged effect, ensuring that if your ChibiBuddy has died it is reactivated. For fun, by default it will take a name from your contacts list as the base of the name for your ChibiBuddy, but that can be customised by setting the choiceAdventure633 appropriately


src/net/sourceforge/kolmafia/session/ChibiBuddyManager.java
```diff
@@ -19,15 +23,87 @@ public class ChibiBuddyManager {
...
+      RequestThread.postRequest(
+          new GenericRequest("inv_use.php?whichitem=" + ItemPool.CHIBIBUDDY_ON));
+      RequestThread.postRequest(new GenericRequest("choice.php?whichchoice=627&option=7"));
+    }
+
+    if (!haveChibiBuddyOn()) {
+      var req = new GenericRequest("inv_use.php?whichitem=" + ItemPool.CHIBIBUDDY_OFF);
+      RequestThread.postRequest(req);
+      if (ChoiceManager.currentChoice() == 633) {
+        // Users can provide their own names using the choice adventure prefs.
```
midgleyc, line 48: does this involve setting it to `1&chibiname=NAME`? If so, does that mean that `choiceAdventure633` is the first which isn't necessarily an integer? Does this interface well with everything else?

gausie: It's not the first, this is a thing across mafia already! See ChoiceManager.automateChoice for example

midgleyc (approved):
Thanks!

PR #1710, opened 2023-05-18
feat: undo special LoL standard familiar handling

midgleyc:
Now that standard.php no longer includes familiars, we don't need to special case this.

Probably worth waiting a little to confirm standard.php is stable before merging.


gausie (approved):
I have it on good authority that this is now stable!

PR #1711, opened 2023-05-18
Failing Test - Location Mapping

jaadams5:
The Hippy Camp is a location name used by KoL.  It should resolve to an adventurable KoLmafia location.  It doesn't.

These failing tests demonstrate the problem.  At the moment I would like feedback on the two new tests although if anyone has a suggested fix, I'll be glad to use it.

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
+++ b/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
@@ -15,6 +15,7 @@ import static internal.helpers.Player.withNextMonster;
 import static internal.helpers.Player.withNextResponse;
 import static internal.helpers.Player.withPath;
 import static internal.helpers.Player.withProperty;
+import static internal.helpers.Player.withQuestProgress;
 import static internal.helpers.Player.withValueOfAdventure;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
@@ -39,6 +40,7 @@ import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.persistence.MallPriceDatabase;
 import net.sourceforge.kolmafia.persistence.MonsterDatabase;
 import net.sourceforge.kolmafia.persistence.NPCStoreDatabase;
+import net.sourceforge.kolmafia.persistence.QuestDatabase;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.ApiRequest;
 import net.sourceforge.kolmafia.request.CharSheetRequest;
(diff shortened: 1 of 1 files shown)


gausie:
If you're testing big chunks of ASH, probably do so using the pre-existing rig. Here I would try to write a proper unit test that identifies and tests the minimal unit of code that produces the issue.

I'm also interested on how you decided on the "desired" output. To my mind "The Hippy Camp" without context is inherently non-differentiable - it would be possible to determine which version is the one that is currently adventurable, but that might not be what the user is looking for.

jaadams5:
I went for ash because it was the quickest way to isolate the problem.  If I am the only one who thinks there is a problem then I've saved myself some effort :-)  If this warrants a fix my hope would be to write additional tests that focused on the code that was changed.

In context "The Hippy Camp" is a location parsed from KoL for the Sot's parcel.   In order for scripted adventuring to get the parcel the string has to be mapped to something that KoLmafia thinks is a valid location.  The tools I have figured out to do that do not do it correctly.  The code I keep finding just does fuzzy string matching and thinks "The Hippy Camp" matches "The Hippy Camp (Bombed Back to the Stone Age)".  That is the wrong answer much of the time.  I would expect something to tell me there are multiple options or disambiguate them based upon the character and game state.  In the absence of parsing errors and KoL bugs, we know, a priori, that the string indicates a unique location and that character can adventure at the location.

The first test sets character state so that "The Hippy Camp" from KoL should map to "Hippy Camp" in KoLmafia because the character had finished the war and done so as a hippy.

The second test just checks can_adventure for the mapped location.  It is expected to fail as long as the location is mapped incorrectly.

gausie:
> the character can adventure at the location.

My point is that I don't agree with this when applied to to_location - you could be parsing anything! If you want to make a java function that explicitly translates a location name into a zone and disambiguates based on player state I'm sure that would be useful, but generic location utilities should not do that

midgleyc:
Mafia puts the parsed Sot's parcel location in `_sotParcelLocation`. If this is a location that parses uniquely under `to_location`, then the script doesn't need to be able to determine which Hippy Camp "The Hippy Camp" refers to.

That is, I think the right point to look at locations and whether we can adventure in them is on the Mafia side, just before we put something in `_sotParcelLocation`.

gausie:
I would process the location at the point it is parsed from the Sot and
save a disambiguated location to the pref.

jaadams5:
> Mafia puts the parsed Sot's parcel location in `_sotParcelLocation`. If this is a location that parses uniquely under `to_location`, then the script doesn't need to be able to determine which Hippy Camp "The Hippy Camp" refers to.
> 
> That is, I think the right point to look at locations and whether we can adventure in them is on the Mafia side, just before we put something in `_sotParcelLocation`.

This started because the location was "The Hippy Camp", to_location mapped it to the Bombed Out version and adv1 failed because the Bombed Out version was not available.  

There is a withdrawn PR that tried to handle the "no match" case by fuzzy matching and a wrong beilef that the failure was due to the extraction parsing.

If the consensus is to make the preference the KoLmafia name of an available area than that can be done at the time of extraction.  It might break if, for example, the parcel is on the island and the War is started or finished between extraction and attempt to fetch but that is pretty niche.

It does stick in my craw that to_location doesn't do what I want it to but I accept that I can't always get what I want.

I put this test out there to see whether anyone else thought there was a problem and I think the answer is No which means the problem should be solved when the location is extracted.

jaadams5:
>I put this test out there to see whether anyone else thought there was a problem and I think the answer is No which means the problem should be solved when the location is extracted.

We're almost all typing at the same time but I think we are about to agree on this.

Thanks.

PR #1713, opened 2023-05-19
automatic backups for <user>_prefs.txt

BadHorseMonkey:
So, as a safety net for users who accidentally do things that delete their prefs, I want to put in some automatic backup/restore functionality as a resiliency measure.

I've got a working bare bones implementation (no test cases yet, no messages to the gCLI, etc.) and I'd like some other devs to comment on the approach, suggested test cases, Should this be controlled by a pref (so that users could turn it off if it somehow goes wrong), etc.


BadHorseMonkey:
I used NIO for the file copy rather than using streams, because it feels like it's more atomic and therefore safer.  I didn't wholesale replace java.io with java.nio, but I considered it.  I am not sure what the implications of that for prefs writing in general would be.

BadHorseMonkey:
I did this for <user>_prefs.txt, but not for GLOBAL_prefs.txt or <user>_moods.txt.  Those would be easy enough, but I didn't know if it was necessary.

BadHorseMonkey:
If the user deletes the _prefs.txt but not the _prefs.bak on purpose, this will restore them.  They'll have to be more aggressive.  This may be why we would want a toggle for this (but it would probably have to be global, since if the prefs were deleted...).

BadHorseMonkey:
Tests have been broadly sketched out, but I would like comments on the ideas as well.
test1: env: _prefs.txt exists, _prefs.bak does not.  after loadUserPreferences(), _prefs.bak should exist and be identical to _prefs.txt
test2: env: _prefs.txt does not exist, _prefs.bak does.  after loadUserPreferences(), _prefs.txt should exist and be identical to _prefs.txt, a value from the _prefs.bak should be set.
test3: env: __prefs.txt exists, _prefs.bak exists.  before loadUserPreferences() they are different, after loadUserPreferences(), they are the same.

Rinn:
Last time this happened to me it didn't actually blank out my settings, mafia overwrote my prefs file while it was still running with default values. Might be some interaction with the preference that updates the file on every change? I can't recall if that's the default or not.

libraryaddict:
So I have several thoughts about this.

## Are backups actually needed here?

So we have the issue of file corruption, wouldn't it be simpler to just have all preference saving now write to a temp file, then rename it when the writing is complete?

## Backup at Login/Logout

I'm not sure why you don't do a backup at logout instead of login, you'd want to separate the backup creating from the preference writing, so that it's only invoked at logouts. 

## Multiple backups

I feel like if we're going to go down the approach of backups, we should probably do multiple backups. Though perhaps have that as an opt-in. So we have the simple .bak, then the opt in folders as described below.

A new folder, such as `settings/Gerald_The_Sealclubber/20230519.bak`, And an option to keep only the last X backups.

My main thought on multiple backups is that if you encounter a niche corruption, you should be able to recover from it, even if from an old backup, instead of hoping that your only backup didn't get overwritten when you started mafia.

gausie:
Are backups needed here- yes, and I think mcrofts solution covers more
potential issues than temp file writing

Backup at login/logout - login is always better because logouts don't
always happen.

Multiple backups - out of scope, don't distract this PR on it

BadHorseMonkey:
I think there's more work to do, but this is probably a start of a safety net.  I added a test, but I think we need more work to discover why or even how these failures are happening.

BadHorseMonkey:
This is also being discussed on the forum, of course.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -671,11 +673,46 @@ private static void loadGlobalPreferences() {
...
+    Preferences.userPropertiesFile = userPrefsFile;
 
-    Properties p = Preferences.loadPreferences(file);
+    Properties p = Preferences.loadPreferences(userPrefsFile);
+
+    if (p.size() == 0) {
+      // Something went wrong reading the preferences.
+      if (backupFile.exists()) {
+        KoLmafia.updateDisplay(
+            userPrefsFile + " oould not be read and backup exists, trying backup...");
```
midgleyc, line 688:
```suggestion
            userPrefsFile + " could not be read and backup exists, trying backup...");
```

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -671,11 +673,46 @@ private static void loadGlobalPreferences() {
...
+                backupFile.toPath(), userPrefsFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+
+          } catch (IOException ex) {
+            System.err.format("I/O Error when copying file");
+          }
+        }
+      } else {
+        KoLmafia.updateDisplay(
+            userPrefsFile
+                + " could not be read and no backup exists!  Please exit KoLmafia and check your preference files.");
```
midgleyc, line 706: I suspect you may not want to exit KoLmafia in this case -- or rather, I think you might want to check the directory first. I'm not sure what exiting does, but it might try to save preferences.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -671,11 +673,46 @@ private static void loadGlobalPreferences() {
...
+      }
+    } else {
+      try {
+        java.nio.file.Files.copy(
```
midgleyc, line 710: could you import `Files` instead of leaving it like this?

midgleyc (changes requested):
I was also curious about potential perf impact, but if we save a backup only when we refresh the character on login this is minimal.

It does mean that any backup is a day out of date, but this does seem like the safest way to do it -- the preferences are close, and trying at any other point runs the risk of corrupting the file.

src/net/sourceforge/kolmafia/preferences/Preferences.java
BadHorseMonkey: I'll try that.  I think that was due to java.io and java.nio being mixed, but I'll see if those actually get confused or if it was just me.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -671,11 +673,46 @@ private static void loadGlobalPreferences() {
...
+
+        p = Preferences.loadPreferences(backupFile);
+
+        if (p.size() > 0) {
+          try {
+            java.nio.file.Files.copy(
+                backupFile.toPath(), userPrefsFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+
+          } catch (IOException ex) {
+            System.err.format("I/O Error when copying file");
```
jaadams5, line 700: Is the specific IOException reported anywhere?  Is there a reason for using format rather than println?

BadHorseMonkey: No, I changed that to println.  

Surfaced with errors via updateDisplay.  Is there a different surfacing method you'd recommend?

src/net/sourceforge/kolmafia/preferences/Preferences.java
BadHorseMonkey: Updated, please see if this is more helpful.

midgleyc (changes requested):
You're using `updateDisplay` with a really long message. In my experience, this doesn't display correctly (it displays the whole thing on one line and cuts it off).

Can you log the detailed message to the session log (or console etc.), and print a shorter message in the UI for the user?

BadHorseMonkey:
> You're using `updateDisplay` with a really long message. In my experience, this doesn't display correctly (it displays the whole thing on one line and cuts it off).
> 
> Can you log the detailed message to the session log (or console etc.), and print a shorter message in the UI for the user?

yep! I'll make that change.

BadHorseMonkey:
oh fsck, I thought I was closing a comment

midgleyc (dismissed):
Thanks

midgleyc (approved):

PR #1718, opened 2023-05-20
allow ASH main function to have a vararg

Veracity0:
This script:
```
void main(string op, int... args)
{
    int result = 1;
    print("op = '" + op + "' args = " + count(args));
    foreach i, arg in args {
	switch (op) {
	case "+":
	    result += arg;
	    break;
	case "-":
	    result -= arg;
	    break;
	case "*":
	    result *= arg;
	    break;
	}
    }
    print("result = " + result);
}
```
Has a main function with a required "string" arg and an optional array of ints.
It yields:
```
> call vargi -

op = '-' args = 0
result = 1

> call vargi (+, 2, 4, 6)

op = '+' args = 3
result = 13

> call vargi (*, 2, 4, 6, 8, 10)

op = '*' args = 5
result = 3840
```
Note that if I omit the required "op" argument it will prompt me for a string.
It will not prompt me for the array of ints.

This script:
```
void main(string... params)
{
    foreach i, param in params {
	print("param = " + param);
    }
}
```
has no required arguments. It yields:
```
> call vargi2 ten twenty thirty zero

param = ten twenty thirty zero

> call vargi2

> call vargi2 (foo, bar, baz)

param = foo
param = bar
param = baz
```
It will not prompt for the string array

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafiaCLI.java
+++ b/src/net/sourceforge/kolmafia/KoLmafiaCLI.java
@@ -549,7 +549,7 @@ public class KoLmafiaCLI {
         .register("call")
         .register("run")
         .register("exec")
-        .register("exececute")
+        .register("execute")
         .register("load")
         .register("start")
         .register("profile");
--- a/src/net/sourceforge/kolmafia/textui/AshRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/AshRuntime.java
@@ -325,12 +325,16 @@ public class AshRuntime extends AbstractRuntime {
 
       for (int i = index; parameters != null && i < parameters.length; ++i) {
         Object input = parameters[i];
+        Value value = null;
         try {
-          varValues.add(DataTypes.coerceValue(paramType, input, false));
+          value = DataTypes.coerceValue(paramType, input, false);
         } catch (Exception e) {
+        }
+        if (value == null) {
           RequestLogger.printLine("Bad " + paramType.toString() + " value: \"" + input + "\"");
           return false;
         }
+        varValues.add(value);
       }
 
       // Put them into an Array value
--- a/src/net/sourceforge/kolmafia/textui/AshRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/AshRuntime.java
@@ -265,7 +265,7 @@ public class AshRuntime extends AbstractRuntime {
     return result;
   }
 
-  private boolean requestUserParams(
+  boolean requestUserParams(
       final Function targetFunction, final Object[] parameters, Object[] values) {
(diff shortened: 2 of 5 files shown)


gausie:
Well done! Can you add unit tests?

Veracity0:
I'll see what I can do.

When thinking about this yesterday, I considered testing from the "call" command, but, considering that looks up script files in the file system, it was a can of worms.

I certainly think that code can be improved - and a thorough set of tests would be appropriate before doing that - but this PR does nothing but add comments to it.

All the code here is in the AshRuntime method that prompts for user input. Again, that prompting is a can of worms. Can I mock the InputUtilities method?

That would be necessary for fully testing required args, but since this PR strictly adds code to support optional args, I can test that code without releasing the worms. 😄

midgleyc:
I think `CustomScriptTest` is a good place to add a test -- e.g. make an ASH file taking many parameters and printing them, then add a `.out` file that confirms the correct thing was printed.

Veracity0:
I wasn't aware that CustomScriptTest allowed scripts to be invoked with parameters. I considered that, but thought that a full test would invoke the same script multiple times with different inputs and verify the corresponding different outputs.

I'll look at it, but I don't want to spend a lot of work adding testing infrastructure right now for what is, essentially, an improvement to the "call" command.

Veracity0:
Yeah. ```CustomScriptTest``` does the following:

For each (ash, js, cli, txt) file in "scripts":
```
      // Inject custom output stream.
      RequestLogger.openCustom(out);

      CallScriptCommand command = new CallScriptCommand();
      command.run("call", script);

      RequestLogger.closeCustom();
```
It uses the "call" command to invoke the script with no arguments and saves the output to compare to what is in "expected".

Support for calling a script with parameters - perhaps multiple times with multiple parameter strings - without breaking the existing code - would be a Project.

Especially if you want to be able to test the script calling DataUtilities.promptForValue (which calls InputFieldUtilities.input). THAT method will either pop up a dialog box - or, if headless, will return the result of ```KoLmafiaCLI.DEFAULT_SHELL.getNextLine(" > ")```

Again, a Project. Not a bad idea, but more than I want to deal with right now.

midgleyc:
Still, `CustomScriptTest` shows you how to place a script somewhere where your script can call it.

I don't think you need the extras -- prompting for values, even calling with different sets of parameters -- just a single script that takes varargs and prints them, and a test to confirm this simple case works.

Veracity0:
It was simpler, for now, to create AshRuntimeTest -> RequestUserParameters -> VarArgs with a suite of 5 unit tests for specifically varargs.

midgleyc (approved):

PR #1719, opened 2023-05-21
feat: add replica mr store

midgleyc:
Copied Dinostaur, as this is another store with one currency.

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
@@ -3569,6 +3569,7 @@ public class ItemPool {
   public static final int SHADOW_PILL = 11184;
   public static final int SHADOW_NEEDLE = 11185;
   public static final int CURSED_MONKEY_PAW = 11186;
+  public static final int REPLICA_MR_ACCESSORY = 11189;
   public static final int REPLICA_DARK_JILL = 11190;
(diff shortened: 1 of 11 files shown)


src/net/sourceforge/kolmafia/swingui/CoinmastersFrame.java
```diff
@@ -1432,6 +1432,12 @@ public DinostaurPanel() {
     }
   }
 
+  private class ReplicaMrStorePanel extends CoinmasterPanel {
```
Veracity0, line 1435: Good. I don't see where you instantiate the panel and add it t to the frame, though.

src/net/sourceforge/kolmafia/objectpool/ItemPool.java
```diff
@@ -3569,6 +3569,7 @@ public class ItemPool {
   public static final int SHADOW_PILL = 11184;
   public static final int SHADOW_NEEDLE = 11185;
   public static final int CURSED_MONKEY_PAW = 11186;
+  public static final int REPLICA_MR_ACCESSORY = 11189;
```
Veracity0, line 3572: You need to add this to ResultProcessor.gainItem where gaining/losing a currency item signals (coinmasters) to alert the CoinmastersFrame to update.

Veracity0 (changes requested):

Veracity0 (approved):

PR #1720, opened 2023-05-21
No alert when find defective token

Veracity0:
I fired off a script and left my computer for a couple of hours.
When I returned, I found the script paused because KoLmafia just had to post a dialog box informing me I'd found a defective Game Grid token, requiring an "OK" confirmation.

Years ago, that might have been exciting. Once.

Now, it is, dare I say, annoying.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ArcadeRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ArcadeRequest.java
@@ -16,7 +16,6 @@ import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.session.ChoiceManager;
 import net.sourceforge.kolmafia.session.ResultProcessor;
 import net.sourceforge.kolmafia.utilities.ChoiceUtilities;
-import net.sourceforge.kolmafia.utilities.InputFieldUtilities;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
 
 public class ArcadeRequest extends GenericRequest {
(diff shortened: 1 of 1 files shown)


gausie:
You could fire a desktop notification instead, which at least works on mac

Veracity0:
Maybe. I'm not married to this change - although I will certainly keep it as a perma local change if not approved.

We used to pop up a dialog to show the text when you "used" a seal tooth, 15+ years ago. We removed it when it ceased to be interesting.

This is the only "alert" left of (what I see as) similar things.

If I'm alone st seeing it this way, I'll shrug and close this PR.

gausie:
No no definitely merge this

jaadams5:
Real life got in the way.  I was wondering whether "You could fire a desktop notification instead, which at least works on mac" meant there might be a suggested change and and wanted to confirm that there would be an acquisition message in the log but neither was intended to not approve.

jaadams5 (approved):

Veracity0:
```
Visiting Jackass Plumber
You acquire an item: defective Game Grid token
```
The defective game grid token was introduced with the clan hot dog stand in July 2013; it unlocks the video game hot dog. Back then, it seemed worthwhile to alert the user when they found one, in case they wanted to unlock that hot dog.

These days, it's just not that interesting. My characters have found 15 of them. They do sell for 8.7 million + in the mall, but why select only that item for special notice? primitive alien {blowgun, loincloth, necklace, spear, totem} are vastly rarer; they sell for 50 million+ and there are from 2-6 of each - total - in the mall. My characters have actually found 16 such items. My Spacegate script does list them as special "goodies" found, but even without that, I'd not want an alert to pop up and pause the script.

tl;dr:
1) There is an acquistion message
2) The reason we originally had the alert popup has long since expired.

PR #1721, opened 2023-05-22
New ASH functions: join_strings() is inverse of split_string()

Veracity0:
Just as we have:

string[] array = split_string(string input);
=> split input on new-lines and return array

string[] array = split_string(string input, string pattern)
=> split input on pattern and return array

Now we have:

string output = join_strings(string[] array)
=> Join strings in array with new-lines

string output = join_strings(string[] array, string joiner)
=> Join strings in array with chosen separator

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks! Useful and tested :D

Veracity0:
Thanks! In case you were wondering, this makes handling my new main() with optional parameter string easier:
```
void main(string... parameters)
{
...
    // Parameters are optional. Depending on how the script is invoked,
    // there may be a single string with space-separated keywords, or
    // multiple strings. Whichever, turn into an array of keywords.
    string[] params = parameters.join_strings(" ").split_string(" ");
```

PR #1722, opened 2023-05-23
Track Noncombat Forcers

horrible-little-slime:
This was added to api.php recently (thanks @gausie !)

I'm getting ready to move, so I may not be able to revisit this in the near future; at the very least, I hope that this helps identify the parts of the codebase we need to touch.


gausie:
I'll open a PR to sort this, thank you for the early look - it helped a lot

horrible-little-slime:
At this point this should "just" need tests for each of the dozen NC forcers, making sure that a successful use sets the flag and an unsuccessful use doesn't.

src/net/sourceforge/kolmafia/objectpool/ItemPool.java
```diff
@@ -2994,6 +2994,11 @@ public class ItemPool {
   public static final int SPOOKY_JELLY = 9293;
   public static final int SLEAZE_JELLY = 9294;
   public static final int STENCH_JELLY = 9295;
+  public static final int HOT_TOAST = 9297;
+  public static final int COLD_TOAST = 9298;
+  public static final int SPOOKY_TOAST = 9299;
+  public static final int SLEAZE_TOAST = 92300;
```
Veracity0, line 3000: 9230

horrible-little-slime: good catch!

PR #1723, opened 2023-05-23
feat: more familiar tags

midgleyc:
Ghost Pickle on a Stick is missing one that gives food drops.

Pet Rock, Pair of Ragged Claws, Bad Vibe, Pet Coral, Toothsome Rock and Synthetic Rock should be exact.


Veracity0 (approved):
Polygonal? Phallic?
I'm trusting you on this.

PR #1725, opened 2023-05-24
feat: treat replicas through NEP like real ones

midgleyc:
ValhallaManager is unchanged because replica items go away on prism break.

Extra preferences for chateau and NEP because they also go away on prism break, but when the path is out of standard I think you can use your originals after you break ronin. We can also use the replica chateau and NEP even if the normal ones are out of standard.

If you're in LoL, we don't set Chateau status when viewing mountains, because either:
* you have the real Chateau and may or may not have the replica
* you've just used the replica

and we can't distinguish the cases.

Use the reusable books in breakfast if you have them.


jaadams5 (approved):

src/net/sourceforge/kolmafia/request/PlaceRequest.java
```diff
@@ -240,7 +240,9 @@ else if (responseText.contains("a giant jagged hole in the wall")) {
       }
       case "mountains" -> {
         if (responseText.contains("chateau")) {
-          Preferences.setBoolean("chateauAvailable", true);
+          if (!KoLCharacter.inLegacyOfLoathing()) {
+            Preferences.setBoolean("chateauAvailable", true);
```
gausie, line 244: Do you perhaps want to set `replicaChateauAvailable` here?

src/net/sourceforge/kolmafia/session/BreakfastManager.java
```diff
@@ -882,4 +885,29 @@ private static void useBookOfEverySkill() {
       RequestThread.postRequest(UseItemRequest.getInstance(book));
     }
   }
+
+  private static void useLegacyBooks() {
```
gausie, line 889: Why legacy over replica here and in the prefs?

src/net/sourceforge/kolmafia/request/PlaceRequest.java
midgleyc: no: 

>If you're in LoL, we don't set Chateau status when viewing mountains, because either:
>
> * you have the real Chateau and may or may not have the replica
> * you've just used the replica
>
>and we can't distinguish the cases.

src/net/sourceforge/kolmafia/session/BreakfastManager.java
midgleyc: because Legacy of Loathing. Could be replica I guess 🤷

src/net/sourceforge/kolmafia/request/PlaceRequest.java
gausie: Oh i see, you can see it but can't access in LoL if you have the real one but not the replica? That's a bug

src/net/sourceforge/kolmafia/session/BreakfastManager.java
gausie: I'd like replica just so I can prefref replica, for example.

src/net/sourceforge/kolmafia/request/PlaceRequest.java
midgleyc: Yes. It's the same for NEP too.

src/net/sourceforge/kolmafia/session/BreakfastManager.java
midgleyc: okay, sure

PR #1727, opened 2023-05-26
Improve campground dwelling parsing

Veracity0:
This is inspired by the observation that get_campground() will return "ginormous pumpkin (1)" in two ways:
1) You have a pumpkin house as your dwelling
2) You have a pumpkin garden with at least 11 days growth. I.e., a ginormous pumpkin

A script can distinguish by checking which condition holds:
1) get_dwelling() returns "ginormous pumpkin (1)"
2) get_campground also contains "packet of pumpkin seeds (11)".

It's been proposed that if both conditions hold, it should return "ginormous pumpkin (2)". Mayyyyybe. But since if it returns (1), you STILL have to disambiguate as described, I'm not sure that is useful.

This PR adds tests to verify that KoLmafia internally recognizes that a dwelling is a pumpkin house and a garden crop being a ginormous pumpkin are not connected.

I also made simply visiting the campground parse your dwelling type, rather than only doing that when you inspect the interior of your dwelling.


Veracity0:
This exposed some buggy tests.

If your workshed item is a Cold Medicine Cabinet, when you visit campground.php?action=workshed, KoL redirects to choice.php. Other workshed items also redirect. For example, CampgroundRequest.shouldHandleRedirect says "Workshed may be redirected to Shop if Mayo Clinic installed". As a result, CampgroundRequest will follow redirects if action=workshed.

Which is to say, if the workshed request DOES redirect, CampgroundRequest is not involved in the parsing.

ColdMedicineCabinetResponse issues campground.php?action=workshed. It parses the response to the redirected choice.php request.

Tests involving the cold medicine cabinet call the ColdMedicineCabinetCommand and set the expected response from HTML collected from the redirected choice request. However, they do not set up the responses to include the expected redirection, with the result that when KoL submits the request, it lets CampgroundRequest process the response - which is something it will never actually see and is ill-prepared to parse.

I changed a couple of the tests in RuntimeLIbraryTest to have two responses: a redirection to choice.php and the response to that choice request. This means that CampgroundRequest is no longer erroneously trying to parse the response.

I did not make similar changes to ColdMedicineCabinetTest, other than to not get fooled by the (newly added) error message when CampgroundRequest was being able to find your dwelling in the response text. Which is to say, those tests continue to be inherently buggy. For example, as coded, if you had a garden or a bookshelf, CampgroundRequest will clear those because it can't see them in the choice.php response text.

I choose to not spend more time on ColdMedicineCabinetCommandTest at this time.

jaadams5 (approved):

PR #1730, opened 2023-05-27
feat: witchess set is still visible in lol even if replica unused

midgleyc:
We need to add a preference for actually using the replica Witchess set, because if you have one already it shows in your campsite, but you can't use it.

Also correctly show Witchess and Source Terminal in Daily Deeds if present.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/WitchessRequest.java
+++ b/src/net/sourceforge/kolmafia/request/WitchessRequest.java
@@ -15,11 +15,16 @@ public class WitchessRequest extends GenericRequest {
 
   @Override
   public void run() {
-    if (!(KoLCharacter.inLegacyOfLoathing()
-            && Preferences.getBoolean("replicaWitchessSetAvailable"))
-        && !StandardRequest.isAllowed(RestrictedItemType.ITEMS, "Witchess Set")) {
-      KoLmafia.updateDisplay("Witchess is too old to use in your current path.");
-      return;
+    if (KoLCharacter.inLegacyOfLoathing()) {
+      if (!Preferences.getBoolean("replicaWitchessSetAvailable")) {
+        KoLmafia.updateDisplay("You need to use a replica Witchess Set first.");
+        return;
+      }
+    } else {
+      if (!StandardRequest.isAllowed(RestrictedItemType.ITEMS, "Witchess Set")) {
+        KoLmafia.updateDisplay("Witchess is too old to use in your current path.");
+        return;
+      }
     }
     if (Preferences.getBoolean("_witchessBuff")) {
       KoLmafia.updateDisplay("You already got your Witchess buff today.");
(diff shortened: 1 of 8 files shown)


src/net/sourceforge/kolmafia/request/WitchessRequest.java
```diff
@@ -15,7 +15,8 @@ public WitchessRequest() {
 
   @Override
   public void run() {
-    if (!KoLCharacter.inLegacyOfLoathing()
+    if (!(KoLCharacter.inLegacyOfLoathing()
+            && Preferences.getBoolean("replicaWitchessSetAvailable"))
         && !StandardRequest.isAllowed(RestrictedItemType.ITEMS, "Witchess Set")) {
       KoLmafia.updateDisplay("Witchess is too old to use in your current path.");
```
Veracity0, line 21: I guess. I'm not going to insist, but there could be different messages to distinguish "not in standard" from "in LoL but have not bought a replica Witchess set yest".

src/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanel.java
```diff
@@ -3704,14 +3708,16 @@ public DeckOfEveryCardDaily() {
...
+          KoLCharacter.inLegacyOfLoathing()
+              && InventoryManager.getCount(ItemPool.REPLICA_DECK_OF_EVERY_CARD) > 0;
       boolean nocards = Preferences.getInteger("_deckCardsDrawn") >= 15;
-      boolean allowed = StandardRequest.isAllowed(RestrictedItemType.ITEMS, "Deck of Every Card");
+      boolean allowedNormal =
+          StandardRequest.isAllowed(RestrictedItemType.ITEMS, "Deck of Every Card");
       boolean limited = KoLCharacter.getLimitMode().limitItem(ItemPool.DECK_OF_EVERY_CARD);
-      this.setShown((!bm || kf) && (have || nocards) && allowed && !limited);
+      this.setShown(
+          (!bm || kf) && (((haveNormal || nocards) && allowedNormal) || haveReplica) && !limited);
```
Veracity0, line 3720: I don't understand why "nocards" applies only to the normal Deck but not the replica Deck.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanel.java
Veracity0: There is a special message if you have already drawn all your cards.
If you have not drawn all your cards, you have to have a (usable) deck available in order to see the Daily Deed.

Seems like setShown could be
```(!bm || kf) && ((haveNormal && allowedNormal) || haveReplica) && !limited)```
although I don't see how you could ```haveNormal``` without ```allowedNormal```, unless you are under Standard restrictions and and pulled it anyway.  Can you do that?

midgleyc: I don't think so -- you'd get a "this item is too old to use on this path" error.

Closer to the time, I think maybe you could hang out in Trendy (with the Deck) until the Deck expired, and then it would be out of standard. But you can't do that any more.

Veracity0 (approved):

PR #1732, opened 2023-05-29
[Refactor] choice utilities

kimty103:
### Issue
parseChoices method of choice utilities.java file have duplicate code
extractChoice method of choice utilities.java file have many if code
choice utilities.java file have many duplicate literal

### Check
- [x] Execute `Gradle run` and clear
- [ ] Execute Test code and clear

### Task
- [x] extract new method for parseChoices 
- [x] make enum Class for extraChoice method
- [x] make 2 Constant for duplicate literal


kimty103:
oh i make new push with running task "spotlessapply"!  Thank you for letting me know

src/net/sourceforge/kolmafia/utilities/ChoiceUtilities.java
```diff
@@ -134,43 +149,14 @@ public static Map<Integer, String> parseChoices(final String responseText) {
...
-      rv.put(key, text);
-    }
-
-    m = LINK_PATTERN.matcher(responseText);
-    while (m.find()) {
-      String form = m.group();
-      if (!form.contains("choice.php")) {
-        continue;
-      }
-      Matcher optMatcher = OPTION_PATTERN2.matcher(form);
```
midgleyc, line 152: `LINK_PATTERN` uses `OPTION_PATTERN2`, not `OPTION_PATTERN1`

src/net/sourceforge/kolmafia/utilities/ChoiceUtilities.java
```diff
@@ -31,6 +32,23 @@ public class ChoiceUtilities {
     Pattern.compile("whichchoice=(\\d+)"),
   };
 
+  private static final Pattern[] PARSE_PATTERNS = {FORM_PATTERN, LINK_PATTERN};
+
+  private static final String CHOICE_PHP = "choice.php";
```
Veracity0, line 37: I don't see the value of having this private constant. I'd prefer to actually see "choice.php" being tested in the code rather than having to look up the constant, obvious as it might be.

src/net/sourceforge/kolmafia/utilities/ChoiceUtilities.java
kimty103: ohh... i miss it thank you! i will change it ASAP

src/net/sourceforge/kolmafia/utilities/ChoiceUtilities.java
```diff
@@ -31,6 +32,23 @@ public class ChoiceUtilities {
...
+  public static final String SECRET_CHOICE = "(secret choice)";
```
Veracity0, line 38: On the other hand, having a public constant which is a KoLmafia-assigned name (rather than KoL-assigned) seems OK. It is used twice in this file.

I'm assuming it's public because you intend to include tests which will validate that that String is included when it is appropriate. :)

Veracity0 (changes requested):

gausie:
See other PR

PR #1735, opened 2023-05-31
Write debug logs to a folder "debug" instead of root directory

libraryaddict:
Given the simplicity of this change, not sure a test should be added.
Regardless, did test the code and it works fine as expected.


gausie:
Some explanation of why would be nice though!

libraryaddict:
> Some explanation of why would be nice though!

If you're using the debug a lot, it fills up your root kolmafia folder with logs which looks messy.

gausie:
"Using" debug is for developers, who are a very small group. Debug files are primarily for getting users to assist us in addressing unexpected errors. They're in the root directory so they can more easily be found by those users. Going to close this.

midgleyc:
I agree with gausie with exactly the same reasoning!

jaadams5:
agree++

PR #1738, opened 2023-06-01
Remove comments

jaadams5:
Remove comments about "visible for testing" which was a requirement elsewhere but not for KoLmafia.


midgleyc (changes requested):
If they're truly only visible for testing, can you make them protected or package private instead of public?

jaadams5:
>If they're truly only visible for testing, can you make them protected or package private instead of public?

Done but can't seem to mark the request as resolved.

midgleyc (approved):
right, nice

PR #1741, opened 2023-06-02
Fix nickname for Mr. Store 2002 coinmaster

Veracity0:
How did I not notice this?
Where do we use a coinmaster's "nickname", anyway?


Shiverwarp:
We use the nickname in the `coinmaster` CLI command I believe, so any nickname with spaces won't work for the CLI command

PR #1743, opened 2023-06-03
Change 2002 Mr. Store Catalogue from 'usable' to 'reusable'

libraryaddict:
Fixes a bug where clicking "order" or using said item, will remove it from your tracked inventory as mafia thinks it was used up.


midgleyc (approved):
Thanks!

PR #1744, opened 2023-06-03
feat: use the 2002 catalogue in breakfast

midgleyc:
Going to the store directly gets you nothing. Credits carry over between days. Visit in breakfast.

Also warn at gash if you have credits remaining.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1602,6 +1602,7 @@ user	youRobotTop	0
 user	zeppelinProtestors	0
 user	zigguratLianas	0
 user	zombiePoints	0
+user	_2002MrStoreCreditsCollected	false
 user	_absintheDrops	0
 user	_abstractionDropsCrown	0
 user	_affirmationCookieEaten	false
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -11012,6 +11012,7 @@ Item	Celsius 233	Skill: "Eternal Flame"
 Item	Celsius 233 (singed)	Skill: "Eternal Flame"
 # chainsaw chain: Deals 60-80 Physical Damage and weakens enemies a lot
 # chaos butterfly: Is Unpredictable
+Item	Charter: Nellyville	Effect: "Hot in Herre", Effect Duration: 30
 # Chateau Mantegna room key: Permanently unlocks a new residence in the Big Mountains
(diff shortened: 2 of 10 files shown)


Veracity0 (commented):
Interesting. So, visiting the shop directly does not get you your daily credits, but using the item - which redirects to the shop - does?

CoinmastersFrame does this when you press the "visit" button:
```
    public void check() {
      RequestThread.postRequest(this.getRequest());
      if (this.buyPanel != null) {
        this.buyPanel.filterItems();
      }
    }
```
and
```
    public CoinMasterRequest getRequest() {
      return this.data.getRequest();
    }
```
CoinmasterData.getRequest gets the zero-argument constructor for the CoinmasterRequest class associated with the CoinmasterData, which is the "buy" URL with no action.

Seems MrStore2002 request could munge that to go to inv_use.php, instead, to make sure you fetch credits.

jaadams5 (approved):

midgleyc:
Correct.

I feel like the coinmaster thing isn't that important -- after you've been to the shop through the item once, you have your daily credits, and can go there directly afterwards.

It also leads to a bit of complexity in that you can get there either through the actual catalogue or through the replica, and I think it's easier to avoid that and just do it through breakfast instead.

Veracity0:
Considering my own use cases:

- I never call breakfast in-run
- In or out of run, I never visit Coinmasters in the Relay Browser.
- In aftercore, I do run breakfast, so having the extra credits show up will be helpful.

I have no idea, yet, how I will use this item, either in-run or in aftercore.

I expect that some people will want a daily deed. Some will be happy with breakfast.

For myself, in-run, I want the CoinmastersFrame to update for me when I click visit so I can spend credits on whatever I eventually decide will be useful.

I'll do it myself, I guess.

That said, thanks for doing this. It's at least half what most people will want. A Daily Deed will satisfy the others. 😀

midgleyc:
Hah, well I never considered people not running breakfast in-run! I use that to ensure I get my clovers each day.

Veracity0:
Perhaps I need to rethink that claim. I certainly don't get breakfast on log in, and if/when I do run it, it is in the gCLI. I definitely don't remember to do it every day.

I finished my six ASoL runs and won't try my first (of six) LoL run for another nine days, at least, so I will have to remember how to ascend, again, as always happens after a hiatus.

PR #1745, opened 2023-06-03
feat: add loathingidol command

midgleyc:
I was hoping that if you had "satisfy with coinmasters" it would say "acquire a loathing idol microphone & loathingidol WHATEVER" but it doesn't.

Still, this is good enough for me.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/LoathingIdolCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/LoathingIdolCommand.java
@@ -15,14 +15,15 @@ public class LoathingIdolCommand extends AbstractCommand {
         " [pop | moxie | init | ballad | combat | rhyme | item | country | exp | res] - get the indicated buff";
   }
 
-  private int getUsableMicrophone() {
-    var potentials =
-        List.of(
-            ItemPool.LOATHING_IDOL_MICROPHONE_25,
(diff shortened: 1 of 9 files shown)


src/net/sourceforge/kolmafia/textui/command/LoathingIdolCommand.java
```diff
@@ -0,0 +1,65 @@
...
+      choice = 4;
+    }
+    if (choice < 1) {
+      KoLmafia.updateDisplay(MafiaState.ERROR, parameters + " is not a valid option.");
+      return;
+    }
+
+    int previousChoice = Preferences.getInteger("choiceAdventure1505");
+    Preferences.setInteger("choiceAdventure1505", choice);
+    UseItemRequest useBoomBox = UseItemRequest.getInstance(item);
```
Veracity0, line 61: useMicrophone?

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/textui/command/LoathingIdolCommand.java
midgleyc: yes, whoops

Veracity0 (approved):

Shiverwarp:
This seems to work great, the only problem is Mafia doesn't recognize that the microphone is consumed in the process, so need to refresh inventory between each command

PR #1746, opened 2023-06-03
MrStore2002Request will use a catalog if needed to check credits

Veracity0:
MrStore2002Request is used for the CoinmastersFrame and when acquiring items from Coinmasters.
If we have not checked credits today, it will now use the appropriate catalog to redirect to the shop.
When used in this way, it markes credits as being checked today, so subsequent calls go straight to the shop.

This means that if you go to the CoinmastersFrame, look at Mr. Store 2002, and click "visit", we'll update your daily credits, if needed.


src/net/sourceforge/kolmafia/request/MrStore2002Request.java
```diff
@@ -21,8 +22,24 @@ public class MrStore2002Request extends CoinMasterRequest {
...
   public MrStore2002Request() {
     super(MR_STORE_2002);
+    // If we have not yet obtained today's store credits, "visit" by
+    // using the item (which redirects to the shop), rather than going
+    // to the shop directly, since the latter will not collect credits.
+    if (!Preferences.getBoolean("_2002MrStoreCreditsCollected")) {
+      int itemId =
+          KoLCharacter.inLegacyOfLoathing()
+              ? ItemPool.REPLICA_MR_STORE_2002_CATALOG
+              : ItemPool.MR_STORE_2002_CATALOG;
```
midgleyc, line 39: It's possible to be in LoL and have the main catalog and not the replica:
* don't buy the replica
* break ronin
* pull the main catalog

Fringe, but...

Veracity0: Can you have both? If so, does it matter which you use?
Once you free the king in LoL, all replica items are removed, correct?
```
    if (InventoryManager.hasItem(ItemPool.MR_STORE_2002_CATALOG)) {
      return null;
    }
    if (KoLCharacter.inLegacyOfLoathing()
        && InventoryManager.hasItem(ItemPool.REPLICA_MR_STORE_2002_CATALOG)) {
      return null;
    }
    return "You need a 2002 Mr. Store Catalog in order to shop here.";
  }
```
Seems like simple possession of either catalog gives you access and we should use whichever one you actually have.

Note that "visit" doesn't actually check "accessible".

midgleyc: Correct. You can have both, I expect, but if you have the replica you're definitely in Legacy of Loathing.

midgleyc (approved):

PR #1750, opened 2023-06-04
fix: prefer replica genie bottle to pocket wish

midgleyc:
This made breakfast make a pocket wish, then use the pocket wish to make 2 other pocket wishes. This isn't quite right.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/GenieRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenieRequest.java
@@ -19,12 +19,15 @@ public class GenieRequest extends GenericRequest {
   private static boolean usingPocketWish = false;
   private static final Pattern WISH_PATTERN = Pattern.compile("You have (\\d) wish");
 
-  private static final int COMBAT = 0;
-  private static final int EFFECT = 1;
-  private static final int ITEM = 2;
-  private static final int MEAT = 3;
-  private static final int STATS = 4;
-  private static final int TROPHY = 5;
+  private enum GenieWishType {
+    UNKNOWN,
+    COMBAT,
+    EFFECT,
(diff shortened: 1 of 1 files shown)


jaadams5 (approved):

src/net/sourceforge/kolmafia/request/GenieRequest.java
```diff
@@ -112,11 +115,11 @@ public void run() {
     if (InventoryManager.hasItem(ItemPool.GENIE_BOTTLE)
         && Preferences.getInteger("_genieWishesUsed") < 3) {
       itemId = ItemPool.GENIE_BOTTLE;
-    } else if (InventoryManager.hasItem(ItemPool.POCKET_WISH)) {
-      itemId = ItemPool.POCKET_WISH;
     } else if (KoLCharacter.inLegacyOfLoathing()
```
Veracity0, line 118: General observation: all over the place, we check for "in LoL and have replica item". Since freeing the king takes you out of LoL and immediately destroys all replica items and familiars, why bother checking the path? Surely possession of the replica item suffices?

Or do we not yet remove all replicas upon freeing the king?

Since you just moved the pocket wish, rather than adding replica checking here, I'm not requesting a change in this PR. We should fix everything when we add code to remove all replicas.

Veracity0 (approved):

src/net/sourceforge/kolmafia/request/GenieRequest.java
midgleyc: It was for performance: as you can only have the replica item in LoL, we don't need to check the entire inventory for the presence of an item unless you're in LoL.

PR #1754, opened 2023-06-06
feat: allow folding replica garbage tote without items

midgleyc:
Only thought of this one after #1753 went in, but it's both simpler and allows prompting folding without a generated item.

### DIFF
--- a/src/data/foldgroups.txt
+++ b/src/data/foldgroups.txt
@@ -26,5 +26,5 @@
 0	Jarlsberg's pan (Cosmic portal mode)	Jarlsberg's pan
 0	Sneaky Pete's leather jacket	Sneaky Pete's leather jacket (collar popped)
 0	toggle switch (Bounce)	toggle switch (Bartend)
-0	January's Garbage Tote	deceased crimbo tree	broken champagne bottle	tinsel tights	wad of used tape	makeshift garbage shirt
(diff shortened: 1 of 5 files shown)


src/data/foldgroups.txt
```diff
@@ -26,5 +26,5 @@
 0	Jarlsberg's pan (Cosmic portal mode)	Jarlsberg's pan
 0	Sneaky Pete's leather jacket	Sneaky Pete's leather jacket (collar popped)
 0	toggle switch (Bounce)	toggle switch (Bartend)
-0	January's Garbage Tote	deceased crimbo tree	broken champagne bottle	tinsel tights	wad of used tape	makeshift garbage shirt
+0	January's Garbage Tote	deceased crimbo tree	broken champagne bottle	tinsel tights	wad of used tape	makeshift garbage shirt	replica January's Garbage Tote
```
Veracity0, line 29: I have to ask: does this mean that Kolmafia now believes you can fold an item into a replica January's Garbage Tote?

Do we want to prevent that attempt?

midgleyc: It does, but I do not know why the user would do this. The replica tote does not do anything, and I can only see it occurring if:
* a user asks whether the replica tote is creatable when they have a normal tote
* a user tries to fold something into the replica tote

So I don't feel like it's something that would come up in normal play. Of course, I may be proven wrong ;)

Veracity0: Oh, I agree. It's not useful behavior - since it doesn't actually work - but users are full of surprises. ;)

Veracity0 (approved):

PR #1755, opened 2023-06-08
Track Spooky VHS Tape

gausie:


test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -2059,4 +2060,82 @@ public void otherSituationsEndCounter(String pref, String fixture) {
...
+        FightRequest.updateCombatData(null, null, html);
+        assertThat("Spooky VHS Tape", isInInventory());
+        assertThat("spookyVHSTapeMonster", isSetTo(""));
+        assertThat("spookyVHSTapeMonsterTurn", isSetTo(-1));
+        assertThat(
+            TurnCounter.isCounting("Spooky VHS Tape unknown monster window begin"), is(true));
+        assertThat(TurnCounter.isCounting("Spooky VHS Tape unknown monster window end"), is(true));
+      }
+
+      TurnCounter.stopCounting("Spooky VHS Tape unknown monster window begin");
```
Veracity0, line 2137: Ditto. I'm not sure how much of this there are, so assume I'm commentng on all of them. 😀

test/net/sourceforge/kolmafia/session/EncounterManagerTest.java
```diff
@@ -208,198 +209,247 @@ void underTheKnifeIsConditionalAutostop(boolean skipping) {
...
+    }
   }
 
-  @Disabled("Need some HTML")
-  @Test
-  void isEnamorangEncounterBasedOnResponseText() {
-    String html = html("request/test_fight_enamorang_monster.html");
+  @Nested
+  class Enamorang {
+    @Disabled("Need some HTML")
```
Veracity0, line 249: What HTML is needed? Doesn't this test use some?

test/net/sourceforge/kolmafia/session/EncounterManagerTest.java
```diff
@@ -208,198 +209,247 @@ void underTheKnifeIsConditionalAutostop(boolean skipping) {
...
+    }
   }
 
-  @Test
-  void isSaberForceMonsterFromNextMonster() {
-    Preferences.setInteger("_saberForceMonsterCount", 3);
-    Preferences.setString("_saberForceMonster", "Demoninja");
-    var monster = MonsterDatabase.findMonster("Demoninja");
-    MonsterStatusTracker.setNextMonster(monster);
+  @Nested
```
Veracity0, line 418: I love nested test suites and am very happy you did this.

test/net/sourceforge/kolmafia/session/EncounterManagerTest.java
```diff
@@ -208,198 +209,247 @@ void underTheKnifeIsConditionalAutostop(boolean skipping) {
...
 
-    boolean actual = EncounterManager.isSaberForceMonster();
+      boolean actual = EncounterManager.isSpookyVHSTapeMonster(html, false);
 
-    assertThat(actual, equalTo(true));
+      assertThat(actual, equalTo(true));
+    }
+
+    @Test
+    void isSpookyVHSTapeEncoutnerBasedOnMonster() {
```
Veracity0, line 430: Encounter, not Encoutner.

And how about Cleanups?

(I should retrofit my early tests to have Cleanups, but these days, all new tests should use them.)

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/session/EncounterManagerTest.java
gausie: These are just copied and pasted from Enamorang support. I will update them both!

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -2059,4 +2061,79 @@ public void otherSituationsEndCounter(String pref, String fixture) {
...
     }
   }
+
+  @Nested
+  class SpookyVHSTape {
+    @Test
+    public void canRecogniseSpookyVHSTapeMonster() {
+      var cleanups =
+          new Cleanups(
+              withProperty("spookyVHSTapeMonster", "ghost"),
```
Veracity0, line 2071: also withoutCounters()?

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/request/FightRequestTest.java
gausie: This one inherently tidies itself because if any of the three counters it makes are still active, the assertions will fail. Still, I'll move this to using `withCounter`

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -8830,6 +8836,30 @@ else if (responseText.contains("You throw the handful") || itemSuccess) {
         }
         return false;
 
+      case ItemPool.SPOOKY_VHS_TAPE:
+        if (responseText.contains("grow wide as they view the contents") || itemSuccess) {
```
Veracity0, line 8840: Not a request for a change, but an observation on all the similar usages in this method:
itemSuccess is a boolean that says whether haiku or anapest messages appeared.
Which is to say, at this point, it's rare - but isn't a haiku katana a LoL item?

Even so, checking itemSuccess first would short circuit all the string.contains checks in this method.
In other words, but the expensive checks after the trivial boolean check, rather than before it.

Veracity0 (approved):

src/net/sourceforge/kolmafia/request/FightRequest.java
gausie: That's a good point and a good idea

PR #1757, opened 2023-06-10
fix: allow 'create 3 spooky vhs'

midgleyc:
Make `create` work with Mr. Store 2002.

Talking about different people's way of using Mafia -- this is how I normally buy items from NPC stores (when I'm not using `/buy`).

### DIFF
--- a/src/net/sourceforge/kolmafia/CoinmasterData.java
+++ b/src/net/sourceforge/kolmafia/CoinmasterData.java
@@ -12,6 +12,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.sourceforge.kolmafia.AdventureResult.AdventureLongCountResult;
 import net.sourceforge.kolmafia.objectpool.Concoction;
+import net.sourceforge.kolmafia.objectpool.ConcoctionPool;
 import net.sourceforge.kolmafia.objectpool.ItemPool;
 import net.sourceforge.kolmafia.persistence.CoinmastersDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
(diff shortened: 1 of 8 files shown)


Veracity0:
Huh. Is this because the token is virtual - tracked by a preference? How do Coinmasters with actual item tokens work? Is there something that could be done in CoinmasterRegistry, say, to automatically register Concoctions like this? I believe that's where we register NPCpurchaseRequests, for example. (I'm on my phone, not my computer, so I can't look at code at the moment.)

Veracity0:
Yeah.

```
./request/QuartersmasterRequest.java:52:    ConcoctionPool.set(new Concoction("quarter", "availableQuarters"));
./request/GameShoppeRequest.java:40:    ConcoctionPool.set(new Concoction("store credit", "availableStoreCredits"));
./request/DimemasterRequest.java:52:    ConcoctionPool.set(new Concoction("dime", "availableDimes"));
./request/SwaggerShopRequest.java:141:    ConcoctionPool.set(new Concoction("swagger", "availableSwagger"));
./request/FunALogRequest.java:39:    ConcoctionPool.set(new Concoction("FunPoint", "availableFunPoints"));
```
CoinmasterRegistry:
```

  static {
    for (int i = 0; i < COINMASTERS.length; ++i) {
      CoinmasterData cm = COINMASTERS[i];
      MASTERS[i] = StringUtilities.getCanonicalName(cm.getMaster());
      NICKNAMES[i] = StringUtilities.getCanonicalName(cm.getNickname());
      COINMASTERS[i].registerPurchaseRequests();
    }
  }
```
Well, we could certainly modernize a bunch of stuff in that file, but perhaps CoinmasterData could have a registerConcoction() method which will do the right thing with token and property, iff there is a property.

One less thing you have to remember to do for this kind of Coinmaster.

Veracity0 (approved):
Thank you!

PR #1759, opened 2023-06-10
Fix Divide by zero consumption errors.

jaadams5:
Failing tests to find and fix divide by zero for some consumption.

See https://kolmafia.us/threads/divide-by-zero-error-eating-quantum-taco-consuming-astral-energy-drink.28941/


jaadams5:
Look at the new portion of UseItemRequestTest and validate my claim that it correctly tests cases where the divide by zero was observed.

Trust me when I say that I had to comment out the TCRS tests because they did not restore the normal consumption data and thus the above test failed because it was operating on unexpected data.

Next Steps

- Figure out how to restore the TCRS tests so that they do not leak.  I am looking at that.
- Find a zero drunk item and if one exists write a consumption test and implement the same fix in DrinkItemRequest
- Consider a more comprehensive test based upon zero consumption items.

jaadams5:
Replicated static initialization and restored TCRS tests.

I'm done, with the possible exception of more tests.

test/net/sourceforge/kolmafia/request/UseItemRequestTest.java
```diff
@@ -302,6 +308,67 @@ void maxUsesWorksForHtmlSpleenItemsWithConsumptionType(int spleenUsed, int maxUs
...
+      try (cleanups) {
+        assertThat(UseItemRequest.maximumUses(itemID, ConsumptionType.DRINK), is(maxUses));
+      }
+    }
+
+    @Test
+    void itShouldHandleAstralEnergyDrinkAppropriately() {
+      ConsumablesDatabase.reset();
+      var cleanups = new Cleanups(withLevel(1));
+      try (cleanups) {
```
BadHorseMonkey, line 364: Not sure why wer're testing UseItemRequest.maximumUses() for food and drink, but getRawSpleenHit() for Spleen.  Is there a reason these are different?

jaadams5: Just removed it.

src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
```diff
@@ -189,7 +189,9 @@ public static final int maximumUses(
       allowOverDrink = true;
     }
 
-    int maxNumber = inebrietyLeft / inebriety + shotglass;
+    int temp = inebriety + shotglass;
+    // Don't overflow if overdrink is allowed
+    int maxNumber = temp == 0 ? Integer.MAX_VALUE - 1 : inebrietyLeft / temp;
```
midgleyc, line 194: I don't think this is right.

Suppose a drink is 1 inebriety, you have 15 drunk left, and you have shotglass. You should be able to drink 16.

This calculation puts `temp` at 2, and answers 7.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -152,6 +152,18 @@ public static void reset() {
     ConsumablesDatabase.reset();
   }
 
+  // Used for testing to replicate static initialization
+  public static void forceRestore() {
```
midgleyc, line 156: can you make two methods: `clear` and `populateAverage`, and call `populateAverage` in the static init? I think this more clearly shows your intent.

You could have `forceRestore` call `clear`, then `populateAverage` to keep the same external interface.

test/net/sourceforge/kolmafia/request/UseItemRequestTest.java
```diff
@@ -302,6 +306,56 @@ void maxUsesWorksForHtmlSpleenItemsWithConsumptionType(int spleenUsed, int maxUs
...
+    /**
+     * The list ot tested items was derived from items with zero fullness in fullness.txt at the
+     * time the test was written. Magical sausage 10060 has zero fullness but it is not tested here
+     * because the fullness is controlled by a preference. Similarly for the glitch season reward
+     * 10207. Given that the point of this test is to show a divide by zero does not occur, leaving
+     * them out is deemed acceptable.
+     */
+    @ParameterizedTest
+    @CsvSource({"572", "1266", "1432", "2149", "2188", "4412"})
+    void itShouldNotDivideByZeroWhenConsumingZeroFullnessItems(int itemID) {
```
midgleyc, line 336: I would say you only need 1 zero fullness item to prove it works. The logic is the same for all of them.

test/net/sourceforge/kolmafia/request/UseItemRequestTest.java
```diff
@@ -302,6 +306,56 @@ void maxUsesWorksForHtmlSpleenItemsWithConsumptionType(int spleenUsed, int maxUs
...
+      int maxUses = Integer.MAX_VALUE;
+      var cleanups =
+          new Cleanups(
+              withSpleenUse(spleenUsed),
+              withHP(0, 10, 10),
+              withMP(0, 10, 10),
+              withPath(AscensionPath.Path.STANDARD));
+
+      try (cleanups) {
+        assertThat(UseItemRequest.maximumUses(itemID, ConsumptionType.SPLEEN), is(maxUses));
```
midgleyc, line 324: it shouldn't be max uses for the astral energy drink, though? If it's the 8 one it should be 1, if it's the 5 one it should be 3.

test/net/sourceforge/kolmafia/request/UseItemRequestTest.java
```diff
@@ -302,6 +306,56 @@ void maxUsesWorksForHtmlSpleenItemsWithConsumptionType(int spleenUsed, int maxUs
...
+      var cleanups = withFullness(fullness);
+
+      try (cleanups) {
+        assertThat(UseItemRequest.maximumUses(itemID, ConsumptionType.EAT), is(maxUses));
+      }
+    }
+
+    @ParameterizedTest
+    @CsvSource({"1047", "1276", "4413", "7491"})
+    void itShouldNotDivideByZeroWhenConsumingZeroInebrietyItems(int itemID) {
```
midgleyc, line 348: I would say you only need 1 zero drunk item to prove it works. The logic is the same for all of them.

src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
Veracity0: It definitely does not agree with how math works. :)
```
int maxNumber = A / B + C;
```
equals
```
int maxNumber = (A / B) + C;
```
not
```
int maxNumber = A / (B + C);

**Review by Veracity0 — CHANGES_REQUESTED** [review]

File: test/net/sourceforge/kolmafia/request/UseItemRequestTest.java (hunk 6 above)
Comment (jaadams5) [reply test/net/sourceforge/kolmafia/request/UseItemRequestTest.java:348]: I was responding to a suggestion by @BadHorseMonkey at KoLmafia.us.  If I misunderstood or it is no longer a good idea then I will gladly reduce the test to one case,

File: src/net/sourceforge/kolmafia/request/DrinkItemRequest.java (hunk 2 above)
Comment (jaadams5) [reply src/net/sourceforge/kolmafia/request/DrinkItemRequest.java:194]: Yeah.  You're right.  I missed that.  Fixing.

File: src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java (hunk 3 above)
Comment (jaadams5) [reply src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java]: Renamed and repackaged a bit.  I think I got your intent and stopped duplication.

File: test/net/sourceforge/kolmafia/request/UseItemRequestTest.java (hunk 5 above)
Comment (jaadams5) [reply test/net/sourceforge/kolmafia/request/UseItemRequestTest.java:324]: The test as written always looks for the raw spleen hit of "astral energy drink".  That is null and thus zero is used upstream.  Since the test was designed to show a div by zero no longer occurs, it shows that the error has been fixed.

That said there is a different test that would meet your expectations.  I'll look at writing it.

Comment (jaadams5) [reply test/net/sourceforge/kolmafia/request/UseItemRequestTest.java:324]: I'm having a problem with that test.

UseItemRequest has
`  public static final int maximumUses(String itemName) {
    int itemId = ItemDatabase.getItemId(itemName);
    if (itemId > 0) {
      itemName = ItemDatabase.getItemDataName(itemId);
    }
    return UseItemRequest.maximumUses(itemId, itemName, ConsumptionType.NONE, false);`

Using the debugger, if it is passed "[10883]astral energy drink" it correctly determines the itemId of 10883. ItemDatabase.getItemDataName(itemId) then returns "astral energy drink".  The end result is zero maximum uses because "astral energy drink" is not the name of a known, existing consumable, and you can't consume what doesn't exist.

**Comment by jaadams5** [comment]
There are several things going here and I'm not sure what the next step should be.

TCRS tests leaked modified consumption data.  That has been addressed.  The tests do not directly demonstrate that.

An opportunity for dividing by zero when determining the maximum uses for an item while eating, drinking or spleening has been fixed.  Failing tests that triggered the division no longer fail.

In the process of writing the tests several issues were observed.  It is not clear whether these were pre-existing issues that were masked by the division by zero or whether they are introduced in this PR.

UseItemRequest.maximumUses does not return expected values for the astral energy drink regardless of how it is specified.  Before any consumption data is fetched both [5140]astral energy drink and [10883]astral energy drink are mapped to astral energy drink which has ambiguous consumption data.

What should maximumUses mean when applied to a quantum taco?  Is it [0,1] and needs to be recomputed after every taco is consumed or is it fullness divided by 2 in which case it is possible that a loop trying to consume the maximum will fail sooner than expected?  The current fullness is zero.  Does changing it to two make things better or worse?  It is not clear that there is any special handling in KoLmafia for the taco as determined by usages of QUANTUM_TACO or 4412.

I'm not going to get back this for at least a day so if anyone has any suggestions I'm open to them.  Otherwise I am going to make a failing test (here) related to maximumUses, run it against code that doesn't have these fixes and determine whether this broke something or merely uncovered it.

Thanks.   I feel like I am being more high maintenance than usual.

**Comment by BadHorseMonkey** [comment]
> There are several things going here and I'm not sure what the next step should be.
> 
> TCRS tests leaked modified consumption data. That has been addressed. The tests do not directly demonstrate that.
> 
> An opportunity for dividing by zero when determining the maximum uses for an item while eating, drinking or spleening has been fixed. Failing tests that triggered the division no longer fail.
> 
> In the process of writing the tests several issues were observed. It is not clear whether these were pre-existing issues that were masked by the division by zero or whether they are introduced in this PR.
> 
> UseItemRequest.maximumUses does not return expected values for the astral energy drink regardless of how it is specified. Before any consumption data is fetched both [5140]astral energy drink and [10883]astral energy drink are mapped to astral energy drink which has ambiguous consumption data.
 
I don't understand why UseItemRequest.maximumUses even uses the name at all.  We pass both the name and the ID. Since we require the ID and the ID is unique, why not use it (and maybe remove the name parameter completely)?  Would that clear this up?

> What should maximumUses mean when applied to a quantum taco? Is it [0,1] and needs to be recomputed after every taco is consumed or is it fullness divided by 2 in which case it is possible that a loop trying to consume the maximum will fail sooner than expected? The current fullness is zero. Does changing it to two make things better or worse? It is not clear that there is any special handling in KoLmafia for the taco as determined by usages of QUANTUM_TACO or 4412.
> 
> Thanks. I feel like I am being more high maintenance than usual.

This is a tricky issue and I think you're right to try to work out what it should be...

Case Fullness:
14-15 : Cannot eat Quantum Taco
13: Can eat QT.  1-100 fullness consumed. Can't eat another QT
0-12: Can eat QT.  1-100 fullness consumed. Amount of maximum uses left depends on the value of the first one.

Maximum uses is there to tell me how many I can consume, mostly seen in the food panel and in scripting, IIRC.  So what is the best answer to help someone figure out what to plan for a diet?  My immediate thought is "never use this food in planning a diet", but I'm sure there's some case where it's optimal that I can't think of.  I'd want (the maximum uses to be 0 for 14-15 and 1 for all other fullness, so I can pick it, but I know that it may preclude all other food.

The other option is to say "the maximum possible uses under ideal circumstances is to treat this as a 1 fullness food, except it has a two fullness threshold to eat."  It's technically the maximum uses, but it is likely to be wrong.

Tangentially related: Does KoL allow you to eat multiple QTs?  "Eat 2 Quantum Taco"?   Does that eat all the number you put in, regardless of the fullness?  Does it randomize the food 2 times or does it randomize once?  Do those answers have anything to do with the problem/solution?

**Comment by Veracity0** [comment]
> Tangentially related: Does KoL allow you to eat multiple QTs? "Eat 2 Quantum Taco"? Does that eat all the number you put in, regardless of the fullness? Does it randomize the food 2 times or does it randomize once? Do those answers have anything to do with the problem/solution?

A multi just bought 2 quantum tacos. Currently at 17/17 full, so can't actually eat, but I looked at them in Inventory/Consumables/Food & Drink in the browser.
There are two links: [eat] and [eat some].

All foods that have more than one in inventory have both links.
If you only have one, you get only [eat].

**Comment by BadHorseMonkey** [comment]
>> I don't know why, given an item id and a name, maximumUses looks up the data name.

> ItemDatabase.getItemIds takes a name and returns an array of itemIds that all have that name. Take a look at DataTypes:makeNormalizedItem for an example where we decide that we need to include the bracketed itemId.
> 

Yes, That makes sense.  What I'm trying to grok is the logic in sending both the itemName and itemId to UseItemRequest.maximumUses().

We overload the parameters so that we always get an ID 

```
  public static final int maximumUses(final int itemId) {
    return UseItemRequest.maximumUses(itemId, ConsumptionType.NONE);
  }

  public static final int maximumUses(final int itemId, final ConsumptionType consumptionType) {
    String itemName = ItemDatabase.getItemDataName(itemId);
    return UseItemRequest.maximumUses(itemId, itemName, consumptionType, true);
  }

  public static final int maximumUses(String itemName) {
    int itemId = ItemDatabase.getItemId(itemName);
    if (itemId > 0) {
      itemName = ItemDatabase.getItemDataName(itemId);
    }
    return UseItemRequest.maximumUses(itemId, itemName, ConsumptionType.NONE, false);
  }
```
Then we call maximumUses(itemId, itemName, ConsumtptionType, boolean).

And I think it's because sometimes we need to reference values that are only stored by name, and other times we need values that are only stored by ID.  Which seems like it could use some cleaning up, but not immediately.  And some things could be used either way.

Do I understand the reasoning behind the method call having two versions of the same key?  Or did I miss something?

**Comment by jaadams5** [comment]
I added tests and I think I have fixed the max uses questions raised about AED.  There is probably a cleaner way to deal with ids and names but I will look later.

**Comment by jaadams5** [comment]
>Do I understand the reasoning behind the method call having two versions of the same key? Or did I miss something?

The nested overloading is mostly just an idiom to avoid code duplication.  The implementation is flawed because the mappings between ids and names is not one to one.

**Comment by Veracity0** [comment]
> > Do I understand the reasoning behind the method call having two versions of the same key? Or did I miss something?
> 
> The nested overloading is mostly just an idiom to avoid code duplication. The implementation is flawed because the mappings between ids and names is not one to one.

Not every name maps to an id. As mentioned, see sushi and hot dogs.
Every item id maps to a unique name. It is either the item's display name, or that name with [XXXX] in front of it.

Any code that maps itemId to a (Consumable) name that simply returns the display name without disambiguating, if necessary, by prepending the itemId in brackets, is buggy.

File: test/net/sourceforge/kolmafia/request/UseItemRequestTest.java (hunk 4 above)
Comment (jaadams5) [reply test/net/sourceforge/kolmafia/request/UseItemRequestTest.java:356]: I am going to declare the quantum taco out of scope for now.  I believe the code correctly computes an infinite number of uses given that fullness.txt contains a zero.  I would prefer a separate, focused discussion.

**Comment by jaadams5** [comment]
I am going to declare quantum taco out of scope.  I will try and write a standalone test that surfaces the questions but that will be easier if the TCRS and divide by zero fixes (at least) are in.

**Review by Veracity0 — CHANGES_REQUESTED** [review]

File: src/net/sourceforge/kolmafia/request/DrinkItemRequest.java [hunk 7]
```diff
@@ -189,11 +189,17 @@ public static final int maximumUses(
       allowOverDrink = true;
     }
 
-    int maxNumber = inebrietyLeft / inebriety + shotglass;
+    int temp = inebriety + shotglass;
+    int maxNumber = (inebriety == 0) ? Integer.MAX_VALUE : inebrietyLeft / inebriety;
+    if (maxNumber != Integer.MAX_VALUE) {
+      maxNumber = maxNumber + shotglass;
+    }
```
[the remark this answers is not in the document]
jaadams5: temp was a residual that should have been deleted.

not using the same idiom was a way make the two ways of avoiding an overflow similar but this way eliminates one comparison so... :-)


Thanks.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -146,10 +146,27 @@ public static void reset() {
...
   static {
+    populateAndBuild();
+  }
+
+  private static void populateAndBuild() {
+    populateAverageAdventures();
+    ConsumablesDatabase.reset();
+  }
+
+  private static void populateAverageAdventures() {
```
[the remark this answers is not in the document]
jaadams5: I was responding to

>can you make two methods: clear and populateAverage, and call populateAverage in the static init? I think this more clearly shows your intent.

but it makes no sense to have a routine that is called from one place

jaadams5:
@Veracity0 - Thanks.  Please resolve if appropriate or patiently explain what I missed.  Thank you.

Veracity0 (approved):

PR #1761, opened 2023-06-11
feat: add preference to not error on ambiguous fold

midgleyc:
Features for the 1%.

Some [garbo](https://github.com/Loathing-Associates-Scripting-Society/garbage-collector) users with multiple stinky cheese items are reporting maximization failures where beginning with multiple stinky cheese items equipped and none in inventory leads to a failure to maximize due to Mafia not being sure which to fold.

Add a preference for this case that always folds the last one slotwise: this will make a given `maximize` command work as it also runs the commands in sequence of slots.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -597,6 +597,7 @@ user	ensorcelee
 user	ensorceleeLevel	0
 user	entauntaunedColdRes	0
 user	envyfishMonster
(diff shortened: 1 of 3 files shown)


Veracity0 (approved):
I have a stinky cheese ball in my display case (for safety).
I have all 5 items you can fold it into in storage (for potential use).
No folding here! (What % does that make me?) :)

PR #1766, opened 2023-06-13
Failing test due to leaking TCRS tests

jaadams5:
In the process of debugging a divide by zero issue it was discovered consumption data for astral energy drink varied based upon what the user or script entered.  Tracking that down exposed that TCRS data tests leaked the modified consumable database.  So this PR has the astral tests and disables the TCRS tests.  That was chosen because of an offhand comment that a time would come soon when KoLmafia did not have to maintain TCRS data files anymore.  The information is/would be available from KoL.


Veracity0:
> That was chosen because of an offhand comment that a time would come soon when KoLmafia did not have to maintain TCRS data files anymore. The information is/would be available from KoL.

If I am not mistaken, Gausie said he was spading how the seeds worked so that we could generate the consumption files via calculation, rather than generating them by scraping item descriptions. Which is not at all the same as "the data will be available from KoL".

I am curious about the progress of that investigation, but until - if - it is completed, I see no reason to throw away existing tests.

jaadams5:
Thinking is clarified.  I have recast this as a failing test and present my hypothesis based upon code changes that are not included here that the error is to to TCRS leakage.  Since astral energy drinks are somewhat of a special case it would probably be useful to test another item that is known to have TCRS changes.

gausie:
I have actually just completely done it but haven't had the opportunity to
actually make a PR

PR #1767, opened 2023-06-13
Derive TCRS from spaded seeding

gausie:
I should really put this here rather than have it living as it has done for a fair few months in my own fork of mafia. The spading was done by @phulin and myself and it was really fun to slowly eke out!

The thing that's left really is just the delicate process of shifting over to this new system completely. It also needs thoughts on a testing strategy, of which I would love to hear yours. This has also been merged against latest and with all the great changes to modifiers I have logically migrated but have not tested.

I am very low on time at the moment as I am getting married on Sunday... and am becoming a father 5 weeks later, at which point I will have less time. So for all those reasons I'm unlikely to get this sorted any time soon, though those 5 weeks aren't *totally* busy!


Veracity0:
I think this is glorious, and I would enjoy contributing to it, assuming I have commit access.

I just finished day 2 of my first (of 6) LoL runs, and I have documentation and tweaks for a few of my scripts in the immediate future, but I'm keen to see this implemented/deployed, and I'll be studying - and testing, and discussing - it by and by.

Veracity0:
I finished my ascensions and ping testing, so, I'm ready to look at this now.
I did a conflict resolution merge, and then fixed it. :)
Tests still fail.

Looking forward to seeing what you've done, here.

Veracity0:
I see that you've added "drink" to a bunch of things that you can't actually drink; KoL has them marked as booze, but if you attempt to drink them, it gives a cute message and doesn't actually consume them.

We have a consumption type - USE_MESSAGE_DISPLAY which was originally created so that we could pop up a little window to show the cute messages for "using" certain items. It was a primary usage, not a secondary usage or attribute. Marking 11 drinks as  "drink, message" is cute - and if you actually have to submit inv_drink to "use" it, it is appropriate - but there is no code to handle "message" as a secondary usage.

Looking a little deeper, the only place which displays the message has this:
```
      case USE_MESSAGE_DISPLAY:
        if (!Preferences.getBoolean("suppressNegativeStatusPopup")) {
          UseItemRequest.showItemUsage(showHTML, responseText);
        }
        return;
```
The property is not in defaults.txt.
In the OptionsFrame, it says "Suppress popup dialog for items which cause harmful effects".

Huh?

Well, this little "message display" feature - which I think I added in 2006 or so - apparently no longer works as originally intended. Somebody decided to add an undocumented property to disable it entirely. And the original implementation ALSO popped up a window when you "used" a familiar hatchling, since that also has a cute message, and you can only do it once ever per familiar. Apparently, that annoyed some people, so that popup is commented out. (The net result is that whenever I use a new familiar hatchling, I always grow it in the browser so I can see and enjoy the message. Heck. Why wasn't THAT put under a preference?)

Well, this comment is rather a digression from the actual point of this PR, but it caught my eye that you were extending this (old) feature.

Is it because the non-consumable drinks are assigned different "sizes" - like all drinks - in TCRS?

gausie:
Thank you for the upkeep, V. That really helps. I am going to turn my attention to this very soon.

And yes - we need to maintain the fact that they are drinks for various rolls that are done differently in that case. I think naming also differs for drinks. I'll have to add "message" as a secondary attribute.

And huh that's not good. I bet @Rinn knows something about this, as our resident person-who-runs-mafia-headless and as such doesn't like dialogue boxes.

Veracity0:
This was the PR from a year and a half ago. I did not see it go in; it got lumped in with a Language Server PR in the Recent GitHub Changes. I've inquired about it. In particular, the motivation. Was there a bug report?

I think that if you are headless, we should obviously not pop up a window. I expect the "message" feature should simply not work, in this case. We COULD extract the HTML from the responseText and print it, but nah.

https://github.com/kolmafia/kolmafia/pull/501

gausie:
Looks like I approved the PR!

Rinn:
Yeah it's not a huge deal. It'd be cool if a browser dialog could appear if in the relay browser but it's not important enough to me to spend any time on. The biggest issue I have in the relay browser is it's very difficult to manually halt script execution, running `abort` in the relay cli doesn't usually work.

Veracity0:
OK, I understand the current scope of this.

KoLmafia has extensive TCRS support. I implemented it. :)

For each class/zodiac combo, we have a (three) data files:
```
TCRS_<class>_<zodiac>.txt
TCRS_<class>_<zodiac>_cafe_booze.txt
TCRS_<class>_<zodiac>_cafe_food.txt
```
If you are in TCRS, we load these at login and override consumable sizes and object enchantments.
On logout, we remove the overrides.
We also allow you to "derive" those overrides and "save" them, in case we are missing data or have wrong info.
Currently, you have to be in a specific TCRS run to "derive" the data, since we have to look at the item descriptions and parse the names and enchantments.

This PR, in its current state, lets you "guess" enchantments via algorithm, rather than "derive" via item desc.
Conceptually, the single public method - TCRSDatabase.deriveItem - is replaced by TCRSDatabase.guessItem.
(TCRSDatabase.guessPotion is also currently public, although it should not be; it is called only by guessItem.)

What are the ultimate plans?

- We still have the TCRS data files and load and override at login? deriving (guessing) on demand would be very hard to deal with. Also, there are people who depend on our data files. (There is an idiot on the KoL forums who asked people to derive a file for him because he refuses to run KoLmafia, even to simply type the four commands needed to just do it for himself. Nobody says he has to do literally anything else in KoLmafia, but, no.)
- You don't need to be logged in to TCRS to "guess" a data file. That would be very handy, should we need to augment those files or re-generate them, if KoL changes a seed. Do we add the "tcrs guess" command or replace "tcrs derive"? As mentioned, allow that to be done whether or not you are logged in to a particular TCRS run.
- What about the cafe food and booze? Those are not "items". They have pseudo "cafe" food and booze numbers, which we can see from the item description file names.

Testing. I like your "guessAll" function. There is a general issue I have with how Junit handles test like this. Which is to say, I'd like it if it could report all errors it finds, rather than stopping on the first error it sees. I fixed DataFileTest to do that, so I could find all the places you used "message" as a secondary type - and so it would actually print the item at fault, rather than depending on "assertThat" to print extra info - which it does not do, the way testing is configured on our project.

Well, perhaps the errors would be too verbose that way, and it's simpler to just fix the errors one at a time, for this project.

gausie:
My eventual plan was to remove the data files and rename guess -> derive. Possibly adding/renaming some existing code to check our derivations from observable info. It's an odd thing to work on in many ways because I want to test against the datafiles and in order to do so have to maintain the code built for parsing them!

As always I can be entirely swayed by your opinion here, of course.

>     * We still have the TCRS data files and load and override at login? deriving (guessing) on demand would be very hard to deal with. 

Would it? I figure we load it into memory in the same way we load the data files into memory. This is what KoL does internally. Or better yet we derive each item on demand and cache the result.

> Also, there are people who depend on our data files. (There is an idiot on the KoL forums who asked people to derive a file for him because he refuses to run KoLmafia, even to simply type the four commands needed to just do it for himself. Nobody says he has to do literally anything else in KoLmafia, but, no.)

I think that if we really care about this we could perhaps add a new build command that generates these files and hosts them somewhere.

>     * What about the cafe food and booze? Those are not "items". They have pseudo "cafe" food and booze numbers, which we can see from the item description file names.

I'm not sure if @phulin and I actually got round to spading these - that may need some time.

> I fixed DataFileTest to do that, so I could find all the places you used "message" as a secondary type - and so it would actually print the item at fault, rather than depending on "assertThat" to print extra info - which it does not do, the way testing is configured on our project.

Thank you for doing so. The due date for my (as yet still in utero) first baby was 4 days ago so I've been really preoccupied with supporting my wife and will become yet more preoccupied in the coming weeks/months, so I can't put a time scale on further work here for now

Veracity0:
Congrats on your marriage and your imminent baby. My youngest is 29, and, like all offspring, brought both joy and frustration. The joy won out in the end. :) 

On demand seems tricky. It would include everything on the booze/food/spleen/potions panel that you either have (acquirable) access to or can create. So, first time you open the item manager, do a whole bunch of derivation? And when you get a new ingredient, derive everything you can now make?

Perhaps easier to derive everything when you log in. Faster or slower than reading a file?

We'll see.

src/data/items.txt
```diff
@@ -4,7 +4,7 @@
 #	PRIMARY_USE:
 #	none,
 #	food, drink, spleen,
-#	potion, avatar
+#	beverage, potion, avatar
```
Veracity0, line 7: I see that you add BEVERAGE to the usage comments in fullness.txt, but do not actually mark anything as a "beverage".

src/data/items.txt
```diff
@@ -2117,7 +2117,7 @@
 2089	Crimboween memo	785037568	scroll2.gif	message	t,d	10
 2090	pilgrim shield	299955309	pilshield.gif	offhand	t	0
 2091	fancy bath salts	990835255	potion4.gif	none, combat	t,d	200	bottles of fancy bath salts
-2092	antique hand mirror	443142276	handmirror.gif	none, combat	t,d	100
+2092	antique hand mirror	443142276	handmirror.gif	usable, combat	t,d	100
```
Veracity0, line 2120: This does not seem to be true.

Veracity0 (dismissed):

Veracity0:
It's been a year since you opened this.
Any progress? L)

gausie:
Nope! I shall close this until I can get back to it.

midgleyc:
I have merged main and updated the branch to compile. The immediate issues are:
* the "spooky powder overflows and so has its original effect" thing doesn't work at all
* when generating names, the effects aren't include in the name (e.g. a +fam weight item should have "friendly" in the name but does not)
* some other effects are wrong

gausie:
> okay, so potions do work after all, the issue was in the test: it doesn't use the local TCRS files but instead gets them from GitHub, and they leak between tests.
> 
> Not sure what the actual fix is for that -- they're in data/ which doesn't seem accessible.
> 
> The leak has also revealed that there are many more modifiers we need to copy across -- e.g. McHugeLarge.

Thank you for working this out! Are you planning on doing anything on this? I can definitely pick up from here but as I've said elsewhere recently I have a lot less free time so can't guarantee any specific timings. If you are going to be doing something let me know so I don't step on toes.

midgleyc:
On the file issue: I don't know how to solve this so if you have any ideas go ahead!

On the copyable modifiers: I don't know exactly which ones want copying. Ideally I think I'd have a char in TCRS to check some. Especially ideally I'd have a dev account to check the RPN ;)

I'm not sure of the mechanics, but I think a baseline would be the modifiers without a descPattern: I presume Brimstone, Cloathing, Synergies, McHugeLarge still apply bonuses. Then there's thorns / damage aura which I suspect aren't RPN and so stick around (or are they blanked?)

Then there's stuff like Great Wolf's beastly trousers, which has 3 enchants, which implies the "troubling vulnerability" is one enchant instead of the 5 we have it as, assuming I understand how this works.

gausie:
The tests will be removed as a final step of this PR, insofaras we will no
longer have the need to test reading the data files and will trust our
generated modifiers over those we find in said files.

Veracity0:
I am glad you are working again on this. I did all the original TCRS support - including generating the data files for sign/class pairs (and the JS support in Valhalla to let you see what you get for astral consumables.)
It will be nice to automatically derive this stuff.

gausie:
I'm going to use an LLM to split out some of the stuff that can be merged now from this PR and isolate specifically the bit that we haven't nailed down yet (i.e. merge data stuff, keep seeding stuff).

midgleyc:
If you keep the existing TCRS data generation stuff you can merge most of TCRSDatabase and the tests that check hardcoded potions.

You can't merge the whole thing because e.g. we don't get equipment names right because we don't know the ordering (and the modifier count issue).

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -1258,7 +1258,20 @@ private static TCRS guessSpleen(
...
     Map.entry("wizardly", "Mysticality: +25"),
     Map.entry("wool", "Cold Resistance: +1"),
-    Map.entry("zippy", "Initiative: +20")
+    Map.entry("zippy", "Initiative: +20"),
+    // The four expanded meta-enchantments. The adjectives and their position here are guesses - the
+    // point is the expansions, since each one is a single enchantment that Mafia splits into a
+    // family. The values are made up too.
+    Map.entry(
+        "prismatic",
+        "Hot Damage: +5, Cold Damage: +5, Spooky Damage: +5, Stench Damage: +5, Sleaze Damage: +5"),
```
midgleyc: this one is interesting because we can determine it (e.g. see pair of plants): the order is spooky-stench-hot-cold-sleaze

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -1258,7 +1258,20 @@ private static TCRS guessSpleen(
...
+    Map.entry(
+        "of all resistance",
+        "Hot Resistance: +3, Cold Resistance: +3, Spooky Resistance: +3, Stench Resistance: +3,"
+            + " Sleaze Resistance: +3"),
+    Map.entry("of all attributes", "Muscle: +10, Mysticality: +10, Moxie: +10"),
+    Map.entry(
+        "of regeneration", "HP Regen Min: 3, HP Regen Max: 5, MP Regen Min: 3, MP Regen Max: 5")
```
midgleyc: this one also has a different message in the enchant (e.g. "Regenerate 5-10 HP and MP per adventure") so we can determine this one too

gausie: Honestly this stuff is guesswork on my end, but that is super interesting.

midgleyc:
Okay, issue: `enchantCount` is returning the enchantment count of items *in TCRS*, probably because we load TCRS data after setting ourselves the class and path in the `guessAll` test. So moving the shield DR out hasn't improved things because we're reading the TCRS enchantment count, not the actual enchantment count. Sort of marking our own homework.

Similarly for potions -- the "pass-through" effect is the effect in TCRS, which is always necessarily correct.

gausie:
We have a checkmark(!)

src/data/tcrs.txt
```diff
@@ -0,0 +1,480 @@
...
+Adjective	strange
+Adjective	striped
+Adjective	stuffed
+Adjective	tiny
+Adjective	white
+Adjective	wooden
+Adjective	wrought-iron
+Adjective	yellow
+
+RPN Modifier	Accessory Drop
```
gausie: This needs to not live here of course

gausie:
Folks this is ready for review!

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -489,6 +510,821 @@ private static TCRS deriveItem(final int itemId, final String text) {
...
+      } else {
+        prefixes.add(0, descriptor);
+      }
+      DebugDatabase.appendModifier(mods, entry.getValue());
+    }
+
+    // KoL shuffles the cosmetics on the same stream it selected enchantments from, so continue
+    // enchantRng (already advanced by a multi-enchantment selection, still fresh at seed+10 after a
+    // single MT-picked enchantment). With no enchantments there is no seed+10 stream, so the
+    // shuffle
```
midgleyc: some comments are messed up like this one, by spotless. Can you join the next line?

test/net/sourceforge/kolmafia/persistence/TCRSDatabaseTest.java
```diff
@@ -13,54 +13,397 @@
...
+    for (var mod : BitmapModifier.BITMAP_MODIFIERS) {
+      // Bitmap bits are assigned per parse, so compare presence, not the raw value.
+      classify.accept(
+          mod,
+          new String[] {
+            exp.getRawBitmap(mod) != 0 ? "set" : "", got.getRawBitmap(mod) != 0 ? "set" : ""
+          });
+    }
+    for (var mod : StringModifier.STRING_MODIFIERS) {
+      // MODIFIERS / EVALUATED_MODIFIERS are meta-fields holding the whole (order-sensitive)
```
Rinn: generally could you have the comments like this cleaned up so they don't feel as generated (line character limit, weird line breaks, odd prose etc)

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -489,6 +510,821 @@ private static TCRS deriveItem(final int itemId, final String text) {
...
+      }
+
+      // They get rendered in reverse
+      prefixedPotionMods.add(0, mod);
+    }
+
+    var potionString = String.join(" ", prefixedPotionMods);
+
+    if (!effectName.isBlank()) {
+      mods.addModifier("Effect", quoted(effectName));
```
midgleyc: I think "quoted" should be handled on ModifierList, and the way it should work is that if the modifier is a string modifier, the value should be quoted.

I don't know how it works elsewhere, though, and if other callers pass in the name quoted better to be consistent.

midgleyc:
Also, have you run derive again and confirmed it's consistent?

What will be the procedure for someone running TCRS in the future / scripts to use this?

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -489,6 +510,821 @@ private static TCRS deriveItem(final int itemId, final String text) {
...
+
+    for (var i = 0; i < numCosmeticMods; i++) {
+      cosmeticMods.add(mtRng.pickOne(STRINGS.get("Cosmetic")));
+    }
+
+    return cosmeticMods;
+  }
+
+  /**
+   * Shuffles the cosmetic adjectives with the given glibc stream and joins them into a name prefix.
```
midgleyc: "glibc stream" is nonsense?

Also the rest of the comment is good in-progress notes and not necessary now.

test/net/sourceforge/kolmafia/persistence/TCRSDatabaseTest.java
gausie: I thought I got them all!

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
gausie: glibc stream is the stream of numbers generated by php's rand() (which uses glibc's random instead of mersenne-twister)

gausie:
> Also, have you run derive again and confirmed it's consistent?

Yes. I also made a tool for people to re-derive without direct access to the multi army here: https://tcrs.loathers.net/. I was going to host on kolmafia's org but then I need to work out hosting.

> What will be the procedure for someone running TCRS in the future / scripts to use this?

What do you mean?

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
midgleyc: interesting, but thanks for replacing it anyway

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -489,6 +510,817 @@ private static TCRS deriveItem(final int itemId, final String text) {
...
+  public static void getEffectPool() {
+    EffectDatabase.keys().stream()
+        // Effects must be marked as good
+        .filter(id -> EffectDatabase.getQuality(id) == Quality.GOOD)
+        // Effects must be hookah/wish-able, except Fishy: it became nohookah after TCRS
+        // launched but is still in the path's effect pool.
+        .filter(id -> !EffectDatabase.hasAttribute(id, "nohookah") || id == EffectPool.FISHY)
+        // Some effects are unavailable for no obvious reason, so they are tagged notcrs
+        .filter(id -> !EffectDatabase.hasAttribute(id, "notcrs"))
+        // TCRS effects are limited to those available when the path (Tiki Temerity) launched
```
midgleyc: Tiki Temerity is not the path

so either
> TCRS effects are limited to those available when the path launched (Tiki Temerity)

or
> TCRS effects are limited to those available (Tiki Temerity) when the path launched

if you really don't want the effect at the end

midgleyc:
> What do you mean?

Suppose somebody wants to run a scripted ascension in TCRS, and there have been new consumables added that aren't in Mafia's data files.

At the moment, they run `tcrs derive` at the start of the ascension to populate the internal database (or whatever they're supposed to do). Post-this PR, we can automatically generate item data: does that happen automatically or does the user need to do something?

gausie:
> > What do you mean?
> 
> Suppose somebody wants to run a scripted ascension in TCRS, and there have been new consumables added that aren't in Mafia's data files.
> 
> At the moment, they run `tcrs derive` at the start of the ascension to populate the internal database (or whatever they're supposed to do). Post-this PR, we can automatically generate item data: does that happen automatically or does the user need to do something?

The item data will be automatically derived, but it will be slightly wrong until we add the data that mafia can't add from automatic item introspection (e.g. enchantment count being off, and a bunch of other things). I can make it automatically derive items that are not in items.txt instead of guess them, maybe.

gausie:
Ok I have implemented that (and actually wired this spading to the modifiers we use). I decided to take the dive and rename the old "derive" to "introspect" and then my "guess" to "derive". It will be confusing when reviewing but is less confusing to future readers.

.github/workflows/check.yml
```diff
@@ -68,6 +68,15 @@ jobs:
           check_name: "Test Report (${{ matrix.os }}, Java ${{ matrix.java }})"
           report_paths: "**/build/test-results/test/TEST-*.xml"
 
+      - name: Upload TCRS guessAll mismatch report
+        uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: tcrs-guessAll-mismatches-${{ matrix.os }}
```
midgleyc: guessAll -> deriveAll

src/data/cultshorts.txt
```diff
@@ -114,7 +114,7 @@
 67	Potion	virgin jello shot
 68	Candy2	The Halls of Muscularity (40)
 69	Stats	120	80	90
-70	Candy	Dweeby (40)	The Strength... of the Future (20)
+70	Candy	Dweeby (40)	The Strength...  of the Future (20)
```
midgleyc: I've bug reported it and it would be nice for CDM to fix this ;)

src/data/modifiers.txt
```diff
@@ -186,15 +186,15 @@ Item	dice-print do-rag	Random Monster Modifiers: +1
 Item	dilapidated wizard hat	Hot Damage: +10, Hot Spell Damage: +15, Familiar Effect: "cap 31"
 Item	dinosaur skull helmet	Muscle Percent: +25, Maximum HP: +25, Last Available: "2026-03"
 # Dinsey mascot mask: Changes Your Avatar
```
midgleyc, line 188: probably worth making this an actual modifier given it occurs on three items and would fix the counts. on main + regenerate etc.

src/data/modifiers.txt
```diff
@@ -273,7 +273,7 @@ Item	gingerbread mask	Hat Drop: +25, Last Available: "2016-12"
 Item	gingerbread tophat	Mysticality: +10, Last Available: "2016-12"
 Item	goat beard	Mysticality: +3, Familiar Effect: "atk, 1xPotato, cap 15"
 # Goggles of Loathing: Makes you a better diver
-Item	Goggles of Loathing	Initiative: +50, Hot Resistance: +5, Initiative Penalty: [10*env(underwater)], Item Drop Penalty: [10*env(underwater)], Meat Drop Penalty: [10*env(underwater)], Cloathing, Familiar Effect: "1.5xVolley, 1.5xBarrr"
+Item	Goggles of Loathing	Initiative: +50, Hot Resistance: +5, Initiative Penalty: [10*env(underwater)], Item Drop Penalty: [10*env(underwater)], Meat Drop Penalty: [10*env(underwater)], Cloathing, Familiar Effect: "1.5xVolley, 1.5xBarrr", Enchantment Count: 3
```
midgleyc, line 276: this triple penalty is "makes you a better diver" and feels like it should be detectable.

src/data/modifiers.txt
```diff
@@ -405,7 +405,7 @@ Item	nine-gallon hat	Muscle: +9, Mysticality: +9, Moxie: +9, Experience: +3
 Item	no hat	Lasts Until Rollover, Last Available: "2016-12"
 Item	noir fedora	Experience Percent (Moxie): +20, Moxie Percent: +20, HP Regen Min: 8, HP Regen Max: 10, MP Regen Min: 8, MP Regen Max: 10, Last Available: "2016-07"
 Item	norwhal helmet	Experience (Mysticality): +2, Spell Damage Percent: +60, Booze Drop: +60, Last Available: "2015-11", Familiar Effect: "atk"
-Item	nurse's hat	Maximum HP: +300, HP Regen Min: 30, HP Regen Max: 60, MP Regen Min: 10, MP Regen Max: 15, Familiar Effect: "atk"
+Item	nurse's hat	Maximum HP: +300, HP Regen Min: 30, HP Regen Max: 60, MP Regen Min: 10, MP Regen Max: 15, Familiar Effect: "atk", Enchantment Count: 3
```
midgleyc: why is this not being detected as 3 already?

src/data/modifiers.txt
```diff
@@ -905,7 +905,7 @@ Item	The Ghoul King's ghoulottes	Muscle: +15, Muscle Percent: +25, Maximum HP: +
 Item	The Jokester's pants	HP Regen Min: 6, HP Regen Max: 8, Moxie Percent: +20, Never Fumble, Last Available: "2016-12", Conditional Skill (Equipped): "Adjust the Jokester's Pants"
 Item	The Necbromancer's Shorts	Cold Resistance: +3, Sleaze Resistance: +3, Booze Drop: +30, Last Available: "2011-10", Familiar Effect: "atk, 2xGhuol"
 # The Sagittarian's leisure pants: Disco Bandit Combat Skills weaken enemies by an additional 20%
```
midgleyc: this is used by 5 items and could be a real modifier

src/data/modifiers.txt
```diff
@@ -1296,9 +1299,9 @@ Item	can of fake snow	Cold Damage: +5, Last Available: "2005-12"
 Item	can of maces	Weapon Damage: +2, Sleaze Resistance: +1
 Item	can opener	Weakens Monster, Hat Drop: +50, Last Available: "2016-11"
 # can-you-dig-it?: + ???
-Item	can-you-dig-it?	Last Available: "2010-04"
+Item	can-you-dig-it?	Last Available: "2010-04", Enchantment Count: 1
 Item	candlestick	Experience (Mysticality): +2, Mysticality Percent: +100, Lasts Until Rollover, Last Available: "2015-07"
-Item	candy cane sword cane	Weapon Damage Percent: +50, Never Fumble, Rollover Effect: "Sugar Rush", Rollover Effect Duration: 250, Weapon Damage: [5+5*pref(_surprisinglySweetStabUsed)+5*pref(_surprisinglySweetSlashUsed)], Muscle: +10, Mysticality: +10, Moxie: +10, Conditional Skill (Equipped): "Surprisingly Sweet Stab", Conditional Skill (Equipped): "Surprisingly Sweet Slash"
+Item	candy cane sword cane	Weapon Damage Percent: +50, Never Fumble, Rollover Effect: "Sugar Rush", Rollover Effect Duration: 250, Weapon Damage: [5+5*pref(_surprisinglySweetStabUsed)+5*pref(_surprisinglySweetSlashUsed)], Muscle: +10, Mysticality: +10, Moxie: +10, Conditional Skill (Equipped): "Surprisingly Sweet Stab", Conditional Skill (Equipped): "Surprisingly Sweet Slash", Enchantment Count: 5
```
midgleyc, line 1301: why are we miscounting this one?

src/data/modifiers.txt
```diff
@@ -1815,7 +1818,7 @@ Item	quirky accordion	Monster Level: [5*(1+skill(Accordion Appreciation))], Clas
 Item	Radio Free Foil	Muscle: +20, Last Available: "2006-07"
 Item	Rain-Doh violet bo	Sleaze Damage: +11, Moxie: +11, Softcore Only, Last Available: "2012-02"
 # Rain-Doh yellow laser gun: Super Accurate!
```
midgleyc: super accurate is used by 2 items, but I'm not really inclined to make it a modifier, as we don't know what it does (and there's yet another +hit bonus on the legendary pasta wand)

src/data/modifiers.txt
```diff
@@ -1985,14 +1988,14 @@ Item	Staff of the Scummy Sink	Spell Damage Percent: +150, Sleaze Spell Damage: +
 Item	Staff of the Short Order Cook	Spell Damage Percent: +20, Mysticality: +5, MP Regen Min: 2, MP Regen Max: 3
 Item	Staff of the Soupbone	Spell Damage Percent: +75, Maximum MP: +15, MP Regen Min: 5, MP Regen Max: 6
 # Staff of the Staff of Life: (5 charges left today)
-Item	Staff of the Staff of Life	Spell Damage Percent: +50, Maximum HP: +30, MP Regen Min: 5, MP Regen Max: 6, HP Regen Min: 10, HP Regen Max: 12, Jiggle: "Full HP Recovery"
+Item	Staff of the Staff of Life	Spell Damage Percent: +50, Maximum HP: +30, MP Regen Min: 5, MP Regen Max: 6, HP Regen Min: 10, HP Regen Max: 12, Jiggle: "Full HP Recovery", Enchantment Count: 4
```
midgleyc: again, why do we miscount this? the HP / MP regen amounts are different

src/data/modifiers.txt
```diff
@@ -3454,7 +3457,7 @@ Item	mini kiwi bikini	Moxie: +11, Sleaze Damage: +20, Sleaze Spell Damage: +40,
 Item	mini kiwi silicon tiepin	Initiative: +30, Mysticality: +3, Moxie: +3
 Item	mini-marshmallow dispenser	Maximum HP: +30, Maximum MP: +30, HP Regen Min: 6, HP Regen Max: 8, MP Regen Min: 6, MP Regen Max: 8, Last Available: "2016-11"
 # mirrored aviator shades: Kill 10% More Skeletons
```
midgleyc: while it will never come back, kill x% more skeletons is on two items

src/net/sourceforge/kolmafia/modifiers/BooleanModifier.java
```diff
@@ -151,6 +151,26 @@ public ModifierValueType getType() {
...
 
+  private static final Set<BooleanModifier> ENCHANTMENTS =
+      EnumSet.of(
+          ADVENTURE_RANDOMLY,
+          ALL_SPELLS_CAST_ARE_COLD,
+          ALL_SPELLS_CAST_ARE_HOT,
+          ALL_SPELLS_CAST_ARE_SLEAZY,
+          ALL_SPELLS_CAST_ARE_SPOOKY,
+          ALL_SPELLS_CAST_ARE_STINKY,
+          BLIND,
```
midgleyc: given blind is on 4 items and counts for 2 and doesn't count for 2 I would be tempted to leave it off

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -60,7 +60,7 @@ public class ConsumablesDatabase {
   public enum ConsumableQuality {
     NONE(""),
     CRAPPY("crappy", "#999999"),
-    DECENT("decent"),
+    DECENT("decent", null),
```
midgleyc: given the numbers were removed this can just be left as "decent"

midgleyc:
okay so from observation the modifiers:
* Changes Your Avatar
* Disco Bandit Combat Skills weaken enemies by an additional X%
* Super Accurate!
* Kill X% More Skeletons

are used by multiple items and worth adding to main to improve our enchant count guesses.

Maybe not "Changes Your Avatar" in retrospect, it's not present on some items that change your avatar like "invisible bag" so it might just be confusing.

src/data/cultshorts.txt
midgleyc: image
It was fixed.

src/data/modifiers.txt
gausie: It seems you decided to do it regardless!

src/data/modifiers.txt
gausie: The majority of enchantments that are dependent on a pref don't seem to be RPN, so I filter it out.

src/data/modifiers.txt
gausie: I agree, I added a test that goes red if your counted enchantments === your Enchantment Count, so if we sort this in future we will be forced to come back to this.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -406,108 +490,944 @@ public static TCRS deriveItem(final int itemId) {
...
+  /**
+   * The item's CARRIED_OVER modifiers, taken verbatim from its base modifier string. These cannot
+   * be read back from the item description, so introspection has to restore them from the base
+   * item. Read raw rather than from a {@link net.sourceforge.kolmafia.Modifiers}: an expression
+   * value like "[env(underwater)]" is an innate, context-dependent property, and evaluating it here
+   * would bake in the deriving context (usually zero, which reads as absent).
+   *
+   * <p>Returned in CARRIED_OVER order, which is sorted by name, so the output is reproducible.
+   */
+  /**
```
midgleyc, line 532: we shouldn't have two docstrings for the same function. The top docstring seems to want to apply to the function below this one.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -406,108 +490,944 @@ public static TCRS deriveItem(final int itemId) {
...
+
+    var enchanted = hardcoded || rolledEnchantment;
+    if (enchanted && !enchantment.effect.isBlank()) {
+      mods.addModifier("Effect", enchantment.effect);
+      if (rolledEnchantment || !dynamicDuration) {
+        mods.addModifier("Effect Duration", String.valueOf(enchantment.duration));
+      }
+    }
+
+    if (id == ItemPool.QUANTUM_TACO
```
midgleyc: as ZERO_ADVENTURE_CONSUMABLES gets it own set with one item, these should too

I think having the special cases as separate sets makes things more visible

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -406,108 +490,944 @@ public static TCRS deriveItem(final int itemId) {
...
+    var id = item.getItemId();
+    var seed = seedFor(id, ascensionClass, sign);
+    var mtRng = new PHPMTRandom(seed);
+    var rng = new PHPRandom(seed);
+
+    var beverage = ConsumablesDatabase.isBeverage(id);
+
+    var cosmeticsString = rollCosmetics(mtRng, rng, beverage ? 8 : 10);
+
+    if (id == ItemPool.GUNPOWDER_BURRITO || id == ItemPool.BEERY_BLOOD) {
```
midgleyc: same here, pull out to a set

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -406,108 +490,944 @@ public static TCRS deriveItem(final int itemId) {
...
+          ItemPool.TEMPS_TEMPRANILLO,
+          ItemPool.THYME_JELLY_DONUT);
+
+  /** Items that keep their Effect but take on a new Effect Duration */
+  private static final Set<Integer> HARDCODED_EFFECT_DYNAMIC_DURATION =
+      Set.of(ItemPool.QUEEN_COOKIE, ItemPool.TURTLE_SOUP);
+
+  private static final Map<Integer, Integer> HARDCODED_EFFECT_OVERRIDE =
+      Map.ofEntries(
+          Map.entry(ItemPool.SKULL_CRIMBOWEEN_COOKIE, 256), // Bells in the Batfry
```
midgleyc: EffectPool for the numbers.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
```diff
@@ -406,108 +490,944 @@ public static TCRS deriveItem(final int itemId) {
...
+            && Objects.equals(present.get("HP Regen Max"), present.get("MP Regen Max"));
+    return sameAmounts ? 1 : 2;
+  }
+
+  // Expression functions that query live character or environment state (a preference, the current
+  // zone/location environment, an active effect, ascension class or path). The enchantment
+  // pre-computation can't resolve these, so a base modifier whose value depends on one isn't a
+  // re-rolled enchantment and doesn't count. Pure arithmetic (min/max/floor/ceil/sqrt) and the
+  // supported queries (skill, event) are fine and still count.
+  // TODO: replace this token sniff with ModifierExpression parsing and Modifier enum lookups.
```
midgleyc: planning on doing this TODO or leaving it?

src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -710,7 +710,7 @@ private static boolean decodedNamesEqual(String name1, String name2) {
     // Sacr&eacute; Mental
     // Sacré Mental
     // The Strength... of the Future
-    // The Strength...  of the Future
+    // The Strength... of the Future
```
midgleyc: remove comment

src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -710,7 +710,7 @@ private static boolean decodedNamesEqual(String name1, String name2) {
     // Sacr&eacute; Mental
     // Sacré Mental
     // The Strength... of the Future
```
midgleyc: also remove this line, for reference

test/net/sourceforge/kolmafia/persistence/TCRSDatabaseTest.java
```diff
@@ -6,61 +6,448 @@
...
+        assertThat(ItemDatabase.getItemName(id), (int) resolved, is(innateDR + enchantDR));
+      }
+    }
+  }
+
+  @Test
+  void deriveAll() throws IOException {
+    // Stream mismatches to a file rather than holding them all in memory (exhausts the heap).
+    var reportFile =
+        java.nio.file.Path.of(System.getProperty("user.dir"))
```
midgleyc, line 336: loose fully qualified path

gausie: It's because we already have Path coming in from AscensionPath.Path and I didn't want to expand all of those just to shorten this.

src/net/sourceforge/kolmafia/persistence/TCRSDatabase.java
gausie: I am going to do half of it I think. Coming now

test/net/sourceforge/kolmafia/persistence/TCRSDatabaseTest.java
midgleyc: "import x as y" would be really convenient for Java to have eh?

midgleyc (dismissed):
Thanks!

What a long strange trip it's been...

gausie:
I was just running some TCRS turns and I think there are a few wiring things to sort out.

gausie:
Haha ok this issue is that I've forgotten to spade the cafe modifiers. I bet it's just the same as everything else. Will get that sorted this week. The quest continues...

gausie:
Confirmed working, did some maximizing and some adventuring and all looks great.

midgleyc (approved):

PR #1773, opened 2023-06-18
Selecting a song on a Loathing Idol Microphone uses a charge

Veracity0:
You can maximize to use a microphone.
The first time you select such a boost, your microphone uses a charge - is replaced by a different item - but we did not recognize that happening.
Your next attempt to use the loathingidol command fails, attempting to use the previous microphone.


midgleyc (approved):
Nice, thanks!

PR #1774, opened 2023-06-19
Make champagne/squint double cached passive effects.

phulin:
In certain cases, the cached passives optimization was failing to double with modifier doublers, because "GENERATED" type modifiers aren't generally doubled. This PR breaks it out into a new modifier type and makes that type doubled. This fixes a bug I was encountering in CS right after donning an outfit with squint on.


midgleyc (approved):
Thanks!

Hopefully this fixes the sidebar item drop amount being wrong with champagne.

PR #1776, opened 2023-06-21
run BetweenBattleScript before locket combats

Malibu-Stacey:
Noticed my pre-adventure script wasn't being run before using the Combat Lovers Locket.


jaadams5 (approved):
Hopefully someone else will glance at this before it is pushed...

Veracity0 (approved):
This is fine. This is what we do when wishing for a fight, for example.
In fact, this is exactly like that code, comment and all. ;)

PR #1779, opened 2023-06-22
feat: name change form item

midgleyc:
Wow, look at all those empty ids !


src/data/items.txt
```diff
@@ -11310,3 +11310,23 @@
 11282	Sexy Cosmo	445044642	perf_cosmo.gif	drink	t,d	11
 11283	Souvenir Chopsticks	821052146	chopsticks2.gif	none	t,d	10	Souvenir Chopsticks
 11284	red-soled high heels	933964223	highheels.gif	accessory	t,d	100	pairs of red-soled high heels
+11284
```
Veracity0, line 11313: Isn't this immediately preceded by the same id with an actual number?

Veracity0 (changes requested):

src/data/items.txt
midgleyc: yes, oops

Veracity0 (approved):

PR #1781, opened 2023-06-24
Update AVATAR_PATTERN for image crossorigin

BadHorseMonkey:
Sometimes the charpane returns the avatar with a new crossorigin tag in it :

``

This does not match the current AVATAR_PATTERN.  If this happens, the matcher finds the next image, usually the HP heart.  CharPaneRequest then replaces the avatar for the character with a heart.  The update explicitly allows crossorigin="Anonymous" after <img


BadHorseMonkey:
Specifically for me, this happens when I launch the relay browser from the Swing Toolbar, but not on initial load of the daily deeds panel

src/net/sourceforge/kolmafia/request/CharPaneRequest.java
```diff
@@ -292,7 +292,7 @@ public static final String getLastResponse() {
 
   public static final Pattern AVATAR_PATTERN =
       Pattern.compile(
-          "<img +src=[^>]*?(?:cloudfront.net|images.kingdomofloathing.com|/images)/([^>'\"\\s]+)");
+          "<img [ crossorigin=\"Anonymous\"|]+src=[^>]*?(?:cloudfront.net|images.kingdomofloathing.com|/images)/([^>'\"\\s]+)");
```
Veracity0, line 295: square brackets? as in a character class?

Veracity0 (changes requested):
How about some tests using actual page HTML, for parseAvatar/setAvatar/getAvatar?

src/net/sourceforge/kolmafia/request/CharPaneRequest.java
midgleyc: also, you've shifted the `+` from the space to your new construct.

You probably wanted
```
"<img +(crossorigin=\"Anonymous\")?src=[^>]*?(?:cloudfront.net|images.kingdomofloathing.com|/images)/([^>'\"\\s]+)");
```

Veracity0: Also use "(?: ... )" to make it a non-capturing group so to not shift group numbers.

BadHorseMonkey:
One other possibility is to remove the `]*?(?:cloudfront.net|images.kingdomofloathing.com|/images)/([^>'\"\\s]+)"); 

Might be cleaner and not worry if they add future properties or the property name changes.  I'll test that.


And add the test.

BadHorseMonkey:
updated matcher, added example to source file, added tests to charPaneReqeustTest

src/net/sourceforge/kolmafia/request/CharPaneRequest.java
BadHorseMonkey: Done

test/net/sourceforge/kolmafia/request/CharPaneRequestTest.java
```diff
@@ -33,6 +34,23 @@ public void beforeEach() {
     CharPaneRequest.reset();
   }
 
+  @Test
+  void canFindAvatarWithCrossorigin() {
```
Veracity0, line 38: Split this into two tests, each with its own html fixture, rather having both in the same test.

This is especially important because CharpaneRequest sets a LOT of character state, and one hopes that a BeforeEach (and AfterAll) method cleans that up - especially since you don't have Cleanups that do. 😀

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/request/CharPaneRequestTest.java
BadHorseMonkey: Done!  

I thought I was checking "what happens if it changes between two charPaneRequests?", but I think this is a better set of unit tests as unit tests.

Veracity0 (approved):

PR #1788, opened 2023-06-27
if depleted replica genie bottle, use pocket wishes

Veracity0:
If you've used your replica genie bottle 3 times today, subsequent uses need to use a pocket wish.


midgleyc (approved):
Thanks!

PR #1789, opened 2023-06-28
Fix rain man monsters joining queue

gausie:
This is now ready for review.

- Move special monster checks for rain man and relativity to a place such that they no longer become enqueued
- Ensure that special monster ignoring still works
- Add lots of testing


gausie:
Per https://kolmafia.us/threads/rain-man-summon-adds-monster-to-last-location-combat-queue.28971/#post-172446 I also need to check if handling for numberology fights should also be added here

jaadams5 (approved):

PR #1791, opened 2023-06-29
fix: incorrect ID for Victor's Spoils

midgleyc:
Had the same ID as KGB so was somewhat noticeable. What more errors could be in this file?


jaadams5 (approved):

jaadams5:
> What more errors could be in this file?

Is that rhetorical?  It seems to me that some text processing could map elements of items.txt and ItemPool.java into a common format that could be compared.  Maybe every integer less than the maximum known item should have an ItemPool entry even if the constant is something like ItemPool.ITEM_12456?  Might even be a test that could be written or a debug command?  For extra credit it would be nice to have a parameterized test framework that understood ItemPool.TOMATO as the parameter instead of just 246.

midgleyc:
It's somewhat of a real question. I ran a script to find other duplicates and didn't find any (non-deliberate, e.g. bang potions have both FIRST_BANG_POTION and MURKY_POTION). I'm not sure how we'd find errors while keeping reasonable UX like the ability to pick your own item names, though.

PR #1794, opened 2023-06-30
feat: patriotic eagle dump

midgleyc:
Citizen of A Zone is going to need more work: a preference, hooking up to modifiers etc.

Also recognising it in the effects pane: it's not an intrinsic, but `Citizen of A Mob of Zeppelin Protesters (Today)`.

### DIFF
--- a/src/data/classskills.txt
+++ b/src/data/classskills.txt
@@ -994,6 +994,9 @@
 7446	Do a Method!	2002skateboard.gif	5	0	0
 7447	Do an epic McTwist!	vortex.gif	5	0	0
 7448	Douse Foe	2002fludd.gif	5	0	0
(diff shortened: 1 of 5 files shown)


gausie:
Just a note that Today skills are a reusable mechanic so we should try to support them appropriately. Not sure what the result of have_effect should be. It probably needs a `today` proxy rec or something

midgleyc:
Probably either the same as intrinsics (2147483547 by wiki), or some other "special" value. But I don't really see the point in distinguishing between intrinsics and today? I can't see a use-case that would need to know it.

gausie:
Hmmm yeah ok

gausie (approved):

PR #1795, opened 2023-06-30
feat: also add favourite familiars to dropdown

midgleyc:
This could feasibly be behind a preference, but why would you not want your favourite familiars to also appear under the category listings?


Veracity0 (approved):
Looking forward to seeing this in action.

PR #1796, opened 2023-06-30
replica masks are in mask mutex

Veracity0:


midgleyc (approved):
You probably will be able to wear a replica mask and a normal mask at the same time once the path comes out of standard, but that's fringe enough I'm willing to not support it (you'd have to ascend softcore, then break ronin to be able to pull equipment, I think)

Veracity0:
I ran into the two replica mask issue in the maximizer. With 19 legacy points, I get the 2nd mask after the 6th boss-dropped legacy Mr. A, so it's a late game issue, but it does happen and is irritating.

I considered making two mutexes: one for regular masks and one for replica masks. I don't know how mutexes are implemented in the maximizer, so didn't know whether adding a mutex would slow things down more than extending one - or if it really wouldn't matter.

Are you sure that breaking ronin in this path allows you to pull equipment? I would have thought that you'd lose that restriction when you free the king; it seems like a path restriction, rather than a Standard restriction.

PR #1797, opened 2023-07-01
feat: add pledge allegiance modifiers to location proxy record 

midgleyc:
Add `$location[].pledge_allegiance` modifiers value.

Outputs a string, modifier-like (but with "All Res" instead of spelled out), of the effect you would get. If there's a nicer way of doing this, let me know: I wanted to make it nice for human readers.

Targeted at people like me who use the GCLI a lot. I expect other people will use Guide or a script or whatever.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -2821,9 +2821,6 @@ public abstract class RuntimeLibrary {
 
     params = new Type[] {DataTypes.STRING_TYPE};
     functions.add(new LibraryFunction("monkey_paw", DataTypes.BOOLEAN_TYPE, params));
-
-    params = new Type[] {DataTypes.LOCATION_TYPE};
-    functions.add(new LibraryFunction("pledge_allegiance_effect", DataTypes.STRING_TYPE, params));
   }
 
   public static Method findMethod(final String name, final Class<?>[] args)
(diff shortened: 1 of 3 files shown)


gausie:
I think you make a command print text, an ash function should be usable by
scripts

midgleyc:
It's a function that returns a string. I figured scripts would do their own thing because they're going to want to format it differently... but maybe not?

gausie:
I think maybe make it return a map of modifier to value like the voting
booth modifiers function?

jaadams5:
> I think you make a command print text, an ash function should be usable by scripts

I like this distinction but cannot recall an explicit articulation of that before.  Thank you.

I am somewhat clueless about the new IOTM and what has already been implemented.  But AFAIK pledge_allegiance_effect(my_location())  does nothing but return a string about what would be expected to happen if allegiance was pledged.  Print that string and this feels like a command.  

I can see providing similar information via an Adventure proxy record or just reverse engineering the deterministic calculation to generate a list of potential locations for an effect and then checking for location availability.

If the question is - I'm already here shall I pledge? then this works but it needs to be more script friendly if the question is - where should I go or open to get a desired effect?

I have no issues with approving but feel like it is a niche solution that will almost certainly need to be augmented,

Veracity0:
If I understand how the new IOTM works:

- Once a day, you can Pledge Alliegance to a location.
- (I don't know how you do that. Ask your familiar, somehow?)
- You get an effect for the rest of the day. It's not an intrinsic, but it may as well be; you can't uneffect it.
- (Citizen of a Zone, perhaps?)
- That effect gives you a bunch of modifiers, based on snarfblat of the zone?

Now, once you have gained this effect, I expect we'll give it variable modifiers, as we do for other objects, and it it will show up in modtrace, string_modifiers(effect, "Modifiers"), etc.

This PR proposes a way to speculate on what the modifiers will be for that effect given a particular location.
Returning a string is essentially what the aforementioned string_modifier would do for that effect, assuming you'd already baked in the location.

How about if modifiers.txt had a "Type" -> "Citizen" -> with keys of "0" -"9". I.e., snarfblat modulus.

So, internally, "Citizen:5" is a modifier set.

You could coerce that into a modifier sat - like voting booth initiative.
You can generate a "Modifiers" string - and string_modifier would find it.

You want a string which can be printed?
Sounds like a command could map location to modulus, fetch the "Modifiers" string, and print it.

Either a runtime method or a proxy record field could return a modifier set (like voting booth initiatives).

A command seems good in either case, but I have no idea how a script would want to have the data structured.
Probably in a modifier map.

I dunno.

Veracity0:
Yeah. I think I like exposing it to scripts via $location.pledge_alliegance (or something) returning a "Modifiers" string corresponding to the snarfblat modulus of that location.

gausie:
> Yeah. I think I like exposing it to scripts via $location.pledge_allegiance (or something) returning a "Modifiers" string corresponding to the snarfblat modulus of that location.

This sounds very practical. And, honestly, not all that unreadable for GCLI users (of which I am also one!)

midgleyc:
Seems reasonable. I did initially think about dumping it into the proxy record, but I thought if we did that for everything it would get rather large. But that doesn't mean it's a bad place for it: it's quite convenient to see it there with everything else.

midgleyc:
I've thrown it in. Some comments from this:
* I think it might be nice to have a `$modifier[]` type, for use with the various methods, to confirm what you're passing in is both a valid modifier and passed in to the correct parameter
* it's not clear to me how to go from a string list of modifiers (or a map, as voting booth has) to extract a specific modifier value. In Java, we have `ModifierDatabase.splitModifiers` which gives you a `ModifierList` from which you can `getModifierValue`. Looking at garbo, it compares keys as literal strings (which is fine).
* `ModifierValue` takes two strings and doesn't force the first to be a modifier, and it sometimes isn't -- e.g. Voting Booth has `Weapon Damage Unarmed` which we never added.

gausie:
I would definitely be in support of $modifier. It'll break some old code
potentially but it's a great idea

Veracity0 (approved):
This is fine. I expect that when we start assigning (variable) modifiers to the effect itself, this will get centralized, somehow, but this is a good script interface, I think.

PR #1799, opened 2023-07-01
Ping request, manager, and command

Veracity0:
Something I've noticed for a long time is that different KoL servers can have vastly different processing time; sometimes it takes me 15 minutes to automate a character's daily turns, and sometimes more than an hour.

There is also recent discussion on discord chat about that, and Irrat wrote a script to measure the lag.

Inspired by this, I am experimenting with internal support.

1) PingRequest calls main.php. It registers as not having a result, which means requests are not registered and the response text is not parsed. It saves the millitime just before invoking a request and just after the response arrives.
2) PingManager calls PingRequest a specified number of time and accumulates a list of results: the elapsed times for each request.
3) PingCommand - "ping" calls PingManager to run a test and summarizes the returned results.
- ```ping``` does 10 pings and prints only a summary
- ```ping 50``` does 50 pings
-  ```ping 50 main``` does 50 pings to main.php
- ```ping 100 api true``` does 100 pings to api.php and is "verbose": it prints each ping as it happens.
4) PingStatus now contains the page, count, shortest, longest, total ping time, and total bytes received - and can calculate average and bytes per second.
Three GLOBAL properties save ping data:
- pingLatest - the most recent PingTest run
- pingShortest - the PingTest with the shortest average ping time
- pingLongest - the PingTest with the longest average ping time
The above are created by PingTest.save(). The last two are only saved if the pings were to the default page (api.php) and  you had at least 10 pings.
Format: PAGE:COUNT:LOW:HIGH:TOTAL:BYTES:AVERAGE times in milliseconds.
5) Reduced request processing overhead by overriding GenericRequest.processRequest to be a no-op for a PingRequest
6) Initial support for pinging via either main.php (like Irrat's script) or api.php.
main.php cannot be run if you are in a fight or choice.
api.php can be run any time, but will redirect to afterlife.php if you are in Valhalla.
7) I experimented with api.php - and it is much faster than main.php. Less processing on KoL? Fewer bytes in the responseText? DEBUG logging does not show the responseText for PingRequest, but it does tell you the size:
A fairly "fast" server with main.php:
```ResponseText has 8242 characters.```
```100 pings at 82-120 msec apiece (total = 9249, average = 92)```
A fairly "fast" server with api.php:
```ResponseText has 1962 characters.```
```100 pings at 24-35 msec apiece (total = 2776, average = 27)```
I think using api.php rather than main.php is a clear win, and that is what is required for saving the historical longest/shortest tests

I'd like to do the following:

8) Handle redirects and timeouts. api.php will redirect to afterlife.php, for example.
9) Add a "ping" checkbox to the Login frame. If checked, run a ping test immediately after logging in - but before doing any of the requests that fetch state from KoL. Compare to saved history (see 4) and if it is too high, pop up a modal dialog reporting the slow results and giving you two (three?) options: accept the connection, retry (logout and login again with a fresh connection), (or maybe, quit).

1-7 work nicely, and this would be useful right now.
8-9 will  make it a complete solution, to my eye.
I'll do those in a followup PR.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/PingRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PingRequest.java
@@ -8,9 +8,12 @@ public class PingRequest extends GenericRequest {
 
   // main.php will redirect if we are in a fight or choice.
   // api.php will redirect to afterlife.php in Valhalla.
+  //
+  // api.php's responseText is about 1/4 the size of main.php's - and
+  // measured ping time is about 1/4 as long.
 
   public PingRequest() {
-    this("main.php");
+    this("api.php");
   }
 
   public PingRequest(String pingURL) {
--- a/src/net/sourceforge/kolmafia/request/PingRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PingRequest.java
@@ -1,0 +1,59 @@
+package net.sourceforge.kolmafia.request;
+
+import java.util.GregorianCalendar;
+
+public class PingRequest extends GenericRequest {
+
+  private long startTime = 0L;
+  private long endTime = 0L;
+
+  public PingRequest() {
+    this("main.php");
+  }
+
+  public PingRequest(String pingURL) {
+    super(pingURL);
+  }
+
+  @Override
+  public void run() {
+    // You can reuse this request.
+    this.startTime = this.endTime = 0;
+
+    this.startTime = new GregorianCalendar().getTimeInMillis();
+    super.run();
+    // *** check if we were redirected
+    // *** check if we got a responseText; If not, timed out?
+    this.endTime = new GregorianCalendar().getTimeInMillis();
+  }
+
+  @Override
+  public boolean hasResult() {
+    return false;
+  }
+
+  @Override
+  public boolean hasResult(String location) {
+    return false;
+  }
+
+  @Override
+  protected boolean shouldFollowRedirect() {
+    return false;
+  }
+
+  public long getStartTime() {
+    return this.startTime;
+  }
+
+  public long getEndTime() {
+    return this.startTime;
+  }
+
+  public long getElapsedTime() {
+    if (this.endTime == 0) {
+      return 0L;
+    }
+    return this.endTime - this.startTime;
+  }
+}
--- a/src/net/sourceforge/kolmafia/request/PingRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PingRequest.java
@@ -4,8 +4,12 @@ import java.util.GregorianCalendar;
 
 public class PingRequest extends GenericRequest {
 
+  private String pingURL = "";
   private long startTime = 0L;
   private long endTime = 0L;
+
+  // main.php will redirect if we are in a fight or choice.
+  // api.php will redirect to afterlife.php in Valhalla.
 
   public PingRequest() {
     this("main.php");
@@ -13,6 +17,15 @@ public class PingRequest extends GenericRequest {
 
   public PingRequest(String pingURL) {
     super(pingURL);
+    this.pingURL = pingURL;
+  }
+
+  private boolean isSafeToRun() {
+    return switch (this.pingURL) {
+      case "main.php" -> !GenericRequest.abortIfInFightOrChoice(true);
+      case "api.php" -> true;
+      default -> false;
+    };
   }
 
   @Override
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by midgleyc** [comment]
For reference, irrat's script is https://github.com/libraryaddict/KolAntiLag.

File: src/net/sourceforge/kolmafia/request/PingRequest.java [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+  public PingRequest(String pingURL) {
+    super(pingURL);
+  }
+
+  @Override
+  public void run() {
+    // You can reuse this request.
+    this.startTime = this.endTime = 0;
+
+    this.startTime = new GregorianCalendar().getTimeInMillis();
```
midgleyc, line 23: `System.currentTimeMillis()` would avoid the (small) overhead of creating an object, dealing with calendar stuff.

midgleyc: or, for estimating time differences, `System.nanoTime()`?

src/net/sourceforge/kolmafia/session/PingManager.java
```diff
@@ -0,0 +1,61 @@
...
+public class PingManager {
+
+  public PingManager() {}
+
+  public static class PingTest {
+    private final List<Long> pings = new ArrayList<>();
+    private long total;
+
+    public PingTest() {
+      this.pings.clear();
```
midgleyc, line 17: given pings is instantiated as empty and isn't static, this shouldn't do anything?

src/net/sourceforge/kolmafia/textui/command/PingCommand.java
```diff
@@ -0,0 +1,39 @@
...
+    boolean verbose = false;
+
+    if (!parameters.equals("")) {
+      String[] split = parameters.split(" +");
+      if (split.length > 0) {
+        String countString = split[0];
+        count = StringUtilities.parseInt(countString);
+      }
+      if (split.length > 1) {
+        verbose = split[1].equals("true");
```
midgleyc, line 25: this isn't in `usage`

src/net/sourceforge/kolmafia/request/PingRequest.java
Veracity0: This is a good point. I like storing the history in milliseconds, which are a lot easier for the user to read and compare than nanoseconds. I'm not sure we need a million times the precision, although we could certainly calculate with that and just divide by a million when saving the data.

src/net/sourceforge/kolmafia/session/PingManager.java
Veracity0: Right.

Other than saving pings in that list, we don't currently even use the data. I had some nebulous idea that perhaps the GUI could display the actual values to the user, to let them see just how bad (or good) things are - like Irrat's script does.

I think the average ping time is the actual metric we'll want to use to checking "goodness" computationally.

src/net/sourceforge/kolmafia/textui/command/PingCommand.java
Veracity0: It was an afterthought, but, I will keep it, so, sure.

jaadams5:
There is no way to know which server we are connected to, correct?  So the practical use for this is to establish some kind of per user local baseline and then if the stats for the current session are significantly slower, give the user the option to log out and try again, in hopes that they will get a faster server?  I like the idea.  Because I am lazy I might suggest that the details of a ping test really do not need to be visible to the user unless or until those details can be kept per server but the user who thinks Ping makes golf equipment probably won't opt in for the details anyway.

midgleyc (approved):

Veracity0:
> There is no way to know which server we are connected to, correct?

Correct.

> So the practical use for this is to establish some kind of per user local baseline and then if the stats for the current session are significantly slower, give the user the option to log out and try again, in hopes that they will get a faster server?

Well, since the stats are GLOBAL, it's more "per computer" than "per user". But yes, that is the plan.

> I like the idea. Because I am lazy I might suggest that the details of a ping test really do not need to be visible to the user unless or until those details can be kept per server but the user who thinks Ping makes golf equipment probably won't opt in for the details anyway.

Somebody who manually uses the command will, I expect, eventually have a "feel" for what is slow and what is fast.
When I (soon) add an option to the Login Frame to do a ping test immediately after logging in but before literally any of the "session refresh" requests are issued, I do think some things do need to be shown. For example "You typically have a 30 msec response time, but the KoL connection you have now is giving you 60 msec. Do you want to try for a better connection?"

jaadams5 (approved):

PR #1800, opened 2023-07-02
feat: add source terminal to campground after replica used

midgleyc:
Not sure this will fix it, but there's an issue where Mafia doesn't accept the `terminal` commands immediately after you use the replica Source terminal -- only after you view your campground.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/UseItemRequest.java
+++ b/src/net/sourceforge/kolmafia/request/UseItemRequest.java
@@ -6103,6 +6103,10 @@ public class UseItemRequest extends GenericRequest {
         }
         return;
 
+      case ItemPool.REPLICA_SOURCE_TERMINAL:
+        CampgroundRequest.setCampgroundItem(ItemPool.SOURCE_TERMINAL, 1);
+        break;
+
       case ItemPool.REPLICA_WITCHESS_SET:
         Preferences.setBoolean("replicaWitchessSetAvailable", true);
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/request/UseItemRequest.java
```diff
@@ -6103,6 +6103,10 @@ else if (responseText.contains("You can't figure out the angle")) {
         }
         return;
 
+      case ItemPool.REPLICA_SOURCE_TERMINAL:
```
Veracity0, line 6106: Since TerminalRequest does this:
```
    if (!KoLCharacter.inNuclearAutumn()
        && !KoLConstants.campground.contains(ItemPool.get(ItemPool.SOURCE_TERMINAL))) {
      KoLmafia.updateDisplay(KoLConstants.MafiaState.ERROR, "You don't have a Source terminal.");
      return;
    }
```
I'd say that the original Source Terminal has the same issue.
Although using that item is almost infinitely more rare than using a replica Source Terminal, these days.

Veracity0 (approved):

src/net/sourceforge/kolmafia/request/UseItemRequest.java
midgleyc: very likely!

Also you only use the Source Terminal proper once, while you use the replica every run.

PR #1801, opened 2023-07-03
Optional ping test at login

Veracity0:
1) Four new settings:
- pingLogin - (boolean) run a ping test at login before refreshing session.
Prints result to gCLI.
- pingLoginCheck - none, goal, threshold
(Only if pingLogin is true) Require average lag to be below goal or historical minimum + threshold
- pingLoginGoal - (int) msecs
- pingLoginThreshold - (float) fractional percent above historical lowest average
2) You can configure these on the Connection tab of the LoginFrame
3) When you log in, run a ping test (if configured) - 10 requests to api.php
If you have a pingLoginCheck configured, the average is compared to your goal or threshold.
If it succeeds, session login completes
If it fails, you are told what the average lag was and what your configured maximum lag is and are asked if you are OK with that.
- Yes - login completes as above
- No - you are logged out and re-logged in - which will trigger another ping test with confirmation, if necessary.
- Cancel (or just close the dialog) - log out and stay on the Login Frame
4) If ping redirects - as api.does to afterlife.php if you are in Valhalla - PingRequest sets its ping URL to the redirect location and resubmits the request. You will thus end up doing a ping test for valhalla.php ...
5) For fun (well, because Irrat's script does it), I now allow "council", in addition to "api" and "main". It has the same "can't be in a fight or choice" restriction as "main". "api" continues to be the best.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -97,6 +97,10 @@ global	macroLens	false
 global	mementoListActive	false
 global	mergeHobopolisChat	false
 global	pingLatest
+global	pingLogin	true
+global	pingLoginCheck	threshold
+global	pingLoginGoal	0
+global	pingLoginThreshold	0.2
 global	pingLongest
 global	pingShortest
 global	previousNotifyList	<>
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -98,7 +98,7 @@ global	mementoListActive	false
 global	mergeHobopolisChat	false
 global	pingLatest
 global	pingLogin	true
-global	pingLoginCheck	threshold
+global	pingLoginCheck	none
 global	pingLoginGoal	0
 global	pingLoginThreshold	0.2
 global	pingLongest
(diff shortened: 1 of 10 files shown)


src/data/defaults.txt
```diff
@@ -97,6 +97,10 @@ global	macroLens	false
 global	mementoListActive	false
 global	mergeHobopolisChat	false
 global	pingLatest
+global	pingLogin	true
```
jaadams5, line 100: Do we want this to default to true and deal with users who wonder what happened or do we default it to false and figure users knew what they were doing when they flipped it?  Either way I like the fact that if pingLogin is true what happens with the defaults will be reasonable for most users, most of the time.

Veracity0: pingLogin simply says "do a ping test and save the results" - in pingLatest, pingLongest, and pingShortest".
It adds 10 api.php requests to logging in - less than 300 msec, in Boston.

If the user later decides to use the other features - enabling pingLoginCheck and setting pingLoginGoal or pingLoginThreshold, as appropriate - they will have the historical data to help them make an informed decision.

Perhaps pingLoginCheck should default to "none", and the user will have to select it.

Veracity0: Yeah, I did that. We will collect ping data by default, but the user will have to set pingLoginCheck to goal or threshold, as desired, on the Connection tab of the Login Frame.

jaadams5 (approved):

PR #1802, opened 2023-07-04
add souvenir flag

Malibu-Stacey:
Drops only on Dependence Day with the Patriotic Eagle as active familiar (looks like a 5% drop from cursory spading). No idea if anything else needs added.
Cosmetic item by the looks of things (adds flags to your profile page).

### DIFF
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -11327,6 +11327,6 @@
 11299
 11300	sleeping patriotic eagle	916117380	pateagle_sleep.gif	grow	t	0
 11301	claw-held flag	498722729	flag2.gif	familiar	t,d	10
-11302	souvenir flag	557635312	flag1.gif	usable	t,d	10	souvenir flags
+11302	souvenir flag	557635312	flag1.gif	usable	t,d	10
 11303	name change form	541643472	namechange.gif	usable	t	0
 11304	replica sleeping patriotic eagle	552435162	pateagle_sleep.gif	grow	t	0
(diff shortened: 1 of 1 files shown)


Veracity0 (changes requested):
It adds flags to your profile page?
Does it give you a status effect?
If not, I assume it's a character flag of some sort which will be cleared at rollover, presumably.

Malibu-Stacey:
> It adds flags to your profile page?
It does this
image
(that is a random person who is buying them in /trade presumably spading if there's a trophy or something if you have more than a specific number). Whether it does that on non-dependence days we'll find out tomorrow.

 > Does it give you a status effect?
 
 Doesn't look like it. Cosmetic only from what I can see.
 
>  If not, I assume it's a character flag of some sort which will be cleared at rollover, presumably.

Apparently it isn't cleared even on ascension.

src/data/items.txt
```diff
@@ -11327,6 +11327,6 @@
 11299
 11300	sleeping patriotic eagle	916117380	pateagle_sleep.gif	grow	t	0
 11301	claw-held flag	498722729	flag2.gif	familiar	t,d	10
-11302
+11302	souvenir flag	557635312	flag1.gif	usable	t,d	10	souvenir flag
```
Veracity0, line 11330: What I meant is, KoL gives you a "use multiple" option. We denote that by saying "multiple", not "usable".
```
> closet put 2 souvenir flag

Placing items into closet...
Requests complete.

> closet take 2 souvenir flag

Removing items from closet...
souvenir flag is multiusable, but KoLmafia thought it was not
You acquire souvenir flag (2)
Requests complete.
```
Additionally:
```
> /closet 10 souvenir flag

You don't have that many souvenir flags, you've only got 2.
```
So the plural actually is "souvenir flags". SInce that's just an added "s", you don't need to list the plural in items.txt

Veracity0 (changes requested):

Veracity0 (approved):

PR #1803, opened 2023-07-04
Separate Sot's property name from its value

eegeeZA:


jaadams5 (approved):

jaadams5:
A mistake I try not to make.  Thanks.

PR #1804, opened 2023-07-04
relog command. 'verbose' keyword. Extra first ping.

Veracity0:
1) the "relog" command will log you out and time you back in, giving you a new server.
2) The "ping" command's third argument specifies verbose logging. Allow it to be "true" or "verbose"
3) Do one extra ping in a ping test before starting to count & measure.
For example,  If the ping command had to time you in, there will be login and a whole 'nother ping test included in the first ping's time.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/PingCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/PingCommand.java
@@ -38,7 +38,7 @@ public class PingCommand extends AbstractCommand {
         }
       }
       if (split.length > 2) {
-        verbose = split[2].equals("true");
+        verbose = split[2].equals("true") || split[2].equals("true");
       }
     }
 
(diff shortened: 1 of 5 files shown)


src/net/sourceforge/kolmafia/textui/command/PingCommand.java
```diff
@@ -38,7 +38,7 @@ public void run(final String cmd, String parameters) {
         }
       }
       if (split.length > 2) {
-        verbose = split[2].equals("true");
+        verbose = split[2].equals("true") || split[2].equals("true");
```
BadHorseMonkey, line 41: think you meant .equals("verbose") in one of these.

Veracity0: Sigh.

midgleyc (approved):

PR #1806, opened 2023-07-05
remove all crimbo NPC stores

Malibu-Stacey:
Fixes things like
image
and
image

### DIFF
--- a/src/data/npcstores.txt
+++ b/src/data/npcstores.txt
@@ -403,23 +403,6 @@ Underground Record Store	vault3	The Pigs holo-record	1000	ROW880
 Underground Record Store	vault3	li'l liberty costume	1000	ROW894
 Underground Record Store	vault3	li'l unicorn costume	1000	ROW897
 
-The Crimbo Cafe	crimbo18	hot watered rum	50	ROW1030
-The Crimbo Cafe	crimbo18	yule gruel	50	ROW1029
-
-Crimbo Town Gift-O-Mat	crimbo18giftomat	bottle of Crimbognac	100000	ROW1031
-Crimbo Town Gift-O-Mat	crimbo18giftomat	Crimboysters Rockefeller	100000	ROW1032
-Crimbo Town Gift-O-Mat	crimbo18giftomat	Crimbeau de toilette	100000	ROW1033
-Crimbo Town Gift-O-Mat	crimbo18giftomat	Crimbo candle	250000	ROW1034
-Crimbo Town Gift-O-Mat	crimbo18giftomat	Carol of the Bulls	500000	ROW1035
-Crimbo Town Gift-O-Mat	crimbo18giftomat	Carol of the Hells	500000	ROW1036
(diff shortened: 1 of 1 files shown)


jaadams5:
I'm thinking leaving them was deliberate and that @Veracity0 had a different solution to "unwanted results".

Veracity0:
I don't consider these to be bugs.

Looking at NPCStoresDatabase.canPurchase()
```
    } else if (storeId.startsWith("crimbo18")) {
      return false;
    } else if (storeId.startsWith("crimbo19")) {
      return false;
    } else if (storeId.startsWith("crimbo20")) {
      return false;
    } else if (storeId.startsWith("crimbo21")) {
      return false;
```
Those stores are marked as disabled, which means you can't buy from them, which is why they show up as grey in the search results.

Another one:
```
    } else if (storeId.equals("fdkol")) {
      return false;
```

gausie:
The latter image can be fixed by adding a canPurchase

Veracity0:
What problem are you trying to solve? I see the prices in a mall search - but because they are unavailable NPCs, we will not try to buy from them, and they do not count for calculating mall_price.

We have every obsolete coinmaster still visible - and unavailable - and I like that.

Is the problem that price analysis lists those? As gausie pointed out, fix that display.

What user problem are you trying to solve (by throwing away historical data)?

Malibu-Stacey:
> What user problem are you trying to solve (by throwing away historical data)?

You can see in the mall manager image I pasted that if someone doesn't look too hard into it, it appears that there are 16.7 million of the item stocked at 100k each. So if you want to price yours below the next lowest in the mall as that person in the image with 250 stocked (they actually have over 1k stocked with a 250/day limit) has done you price them under 100k when the actual next lowest price in the mall is just under 240k. This is misleading information for anyone who doesn't know any better which is probably the vast majority of the player base.

It's fair enough if you want data on NPC stores which will never be in the game again but that's what I personally would say is what source control exists for. Also the wiki exists.

jaadams5:
I'm inclined to reject this.  The user problem can be addressed by adding a filter to a display which seems to me a much less destructive resolution.

gausie:
> > What user problem are you trying to solve (by throwing away historical data)?
> 
> You can see in the mall manager image I pasted that if someone doesn't look too hard into it, it appears that there are 16.7 million of the item stocked at 100k each. So if you want to price yours below the next lowest in the mall as that person in the image with 250 stocked (they actually have over 1k stocked with a 250/day limit as you can see in the first image) has done you price them under 100k when the actual next lowest price in the mall is just under 240k. This is misleading information for anyone who doesn't know any better which is probably the vast majority of the player base.
> 
> It's fair enough if you want data on NPC stores which will never be in the game again but that's what I personally would say is what source control exists for. Also the wiki exists.

I'm going to close this PR because I think we're in agreement that this is not the way we'd like to remediate this issue. Which, is to say I do think there is an issue! I certainly agree that the inaccessible NPC stores should not appear in the mall manager. @Malibu-Stacey Would you be interested in making a PR to that effect?

PR #1807, opened 2023-07-05
change a bunch of stuff to be multiple use

Malibu-Stacey:
changed all avatar potions to be `avatar, multiple` and changed a bunch of stuff I found reported in my session logs over the past year to multiple.

@Shiverwarp reported the avatar potions on https://kolmafia.us/threads/multi-usable-items.29006/ which I haven't personally tested or seen in a log as I haven't used any of them (it's mostly people in 2CRS consuming these) but I trust this is correct from Veracity's discussion on that thread.


Veracity0:
This will work.

Thank you for fixing some potions from usable to multiple - based on your own observations.
I could have sworn that bitter pills were not multiusable at first, but perhaps that changed.
Or perhaps I misremember. :)

My comment on Shiverwarp's bug report left open an alternative solution: since ItemDatabase.isMultiUsable already assumes that type "avatar" is that (unless marked otherwise), perhaps UseItemRequest should use that existing method, rather than requiring that the "multiple" flag be set for all such potions in items.txt?

I am not sure which is better.

As I said, your solution works, but I want to think on it a bit more.

Veracity0:
It is long time since I responded to this. I'm on my phone, so can't call up code, but my response to the kolmafia.us bug report points to exactly the code I'd modify.

The issue is this: given an itemId, UseItemRequest wants to decide if the item is multi usable. It looks only at Attributes: usable, multiple, reusable.

ItemDatabase has methods: isUsable, isMultiUsable, and isReusable (or something). Those methods look at item types and also attributes. A potion or avatar is multi usable UNLESS it is specifically tagged "usable".

1) All your observations that certain potions are "multiple", not "usable"? Cool. Mark them "multiple". And see step 3. 😀
2) Fix the UseItemRequest method to call the ItemDatabase methods rather than looking at attributes.
3) All those potions and avatar potions marked "multiple"? Remove that, since that is the default for a potion or avatar potion.

End result:

In items.txt, every potion is simply "potion" (which implies multiple) or "potion, usable".
And ditto for "avatar". Which probable are always exactly that and never "avatar, usable"

gausie:
Perhaps we should add a data file integrity test that enforces this? So you can't have multi usable on something also marked as a potion?

jaadams5:
Changed to draft because that is a filter I use.  There are good ideas here but no activity in at least a week.

Veracity0:
OK, since I really want this PR, I did all of the suggestions:

1) UseItemRequest.getConsumptionType now uses ItemDatabase.isReusable, isMultiUsable, and isUsable rather thank looking at attributes.
2) All "potion" and "avatar" items that explicitly were marked "multiple" have that attribute removed.
3) DataFIleConsistencyTest now verifies that no "potion" or "avatar" item has an explicit "multiple" attribute.

This should be good to go.

midgleyc (approved):

PR #1808, opened 2023-07-06
timein successfully after canceling ping test

Veracity0:
1) Added a "timeout" command which simply calls logout.php, rather than doing a full logout and shutting down everything. This is the equivalent to KoL having silently logged you out when you time out your session.
2) When timing in, replace the password hash (as seen in the pwd field of the URL, rather than whatever the current password hash is) with the current password hash. This lets timing in a GET request with a a password hash (like a storage pull) work a lot better.
3) Executing a logout.php request leaves you not-logged-in.
4) relogin following a logout returns true or false, depending on whether the login succeeded.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/GenericRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenericRequest.java
@@ -1851,9 +1851,28 @@ public class GenericRequest implements Runnable {
         return false;
       }
 
-      String oldpwd = this.getFormField("pwd");
+      // KoL redirects us to login.php if we submit a request after KoL has
+      // silently timed us out. In this case, we have a password hash from
+      // the previous login, but when we log in again, we'll get a new one.
+      //
+      // If we want to resubmit a request that contains a  password hash,
+      // we'll have to replace the old one with the new one.
+
+      String oldpwd = GenericRequest.passwordHash;
+
       if (LoginRequest.executeTimeInRequest(this.getURLString(), this.redirectLocation)) {
         if (this.data.isEmpty()) {
+          // GenericRequest.passwordHash is set when we log in.  If it is "",
+          // we are not logged in - and we know it.  If it is not that, we are
+          // either currently logged in, or KoL has silently logged us out and
+          // we have not yet logged in and learned a new one.
+          if (oldpwd.equals("")) {
+            // See if there is a password hash in the request's URL
+            String formpwd = this.getFormField("pwd");
+            if (formpwd != null && !formpwd.equals("")) {
+              oldpwd = formpwd;
+            }
+          }
           if (!oldpwd.equals("")) {
             String newpwd = GenericRequest.passwordHash;
             this.formURLString =
--- a/src/net/sourceforge/kolmafia/request/GenericRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GenericRequest.java
@@ -1851,18 +1851,21 @@ public class GenericRequest implements Runnable {
         return false;
       }
 
-      String oldpwd = GenericRequest.passwordHashValue;
+      String oldpwd = this.getFormField("pwd");
       if (LoginRequest.executeTimeInRequest(this.getURLString(), this.redirectLocation)) {
         if (this.data.isEmpty()) {
-          String newpwd = GenericRequest.passwordHashValue;
-          this.formURLString =
-              StringUtilities.singleStringReplace(this.formURLString, oldpwd, newpwd);
+          if (!oldpwd.equals("")) {
+            String newpwd = GenericRequest.passwordHash;
+            this.formURLString =
+                StringUtilities.singleStringReplace(this.formURLString, oldpwd, newpwd);
+          }
           this.formURL = null;
         } else {
           this.dataChanged = true;
         }
         return false;
       }
+      this.redirectHandled = true;
 
       return true;
     }
--- a/src/net/sourceforge/kolmafia/textui/command/TimeoutCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/TimeoutCommand.java
@@ -5,7 +5,7 @@ import net.sourceforge.kolmafia.request.LogoutRequest;
 
 public class TimeoutCommand extends AbstractCommand {
   public TimeoutCommand() {
-    this.usage = " - log out and log in again, preserving character state";
+    this.usage = " - log out, leaving GUI and character state intact";
   }
 
   @Override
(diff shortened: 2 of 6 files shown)


src/net/sourceforge/kolmafia/textui/command/TimeoutCommand.java
```diff
@@ -0,0 +1,15 @@
+package net.sourceforge.kolmafia.textui.command;
+
+import net.sourceforge.kolmafia.RequestThread;
+import net.sourceforge.kolmafia.request.LogoutRequest;
+
+public class TimeoutCommand extends AbstractCommand {
+  public TimeoutCommand() {
+    this.usage = " - log out and log in again, preserving character state";
```
midgleyc, line 8: just a log out here I think

src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -1851,18 +1851,21 @@ private boolean handleServerRedirect() {
         return false;
       }
 
-      String oldpwd = GenericRequest.passwordHashValue;
+      String oldpwd = this.getFormField("pwd");
```
midgleyc, line 1854: is this necessarily non-null?

src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -1851,18 +1851,21 @@ private boolean handleServerRedirect() {
...
+            this.formURLString =
+                StringUtilities.singleStringReplace(this.formURLString, oldpwd, newpwd);
+          }
           this.formURL = null;
         } else {
           this.dataChanged = true;
         }
         return false;
       }
+      this.redirectHandled = true;
```
midgleyc, line 1868: Was this deliberately added?

src/net/sourceforge/kolmafia/textui/command/TimeoutCommand.java
Veracity0: Right. "log out, leaving GUI and character state intact"
The intent is to distinguish this from the "logout" command.

src/net/sourceforge/kolmafia/request/GenericRequest.java
Veracity0: Interesting.

I was testing this by timing in with "pull 1 star" - which is a GET request that requires a pwd.
I was - sometimes - getting a storage.php request with no pwd, because the current passwordHash was "".
KoL would respond with a 404.

I also tested with the Refresh Session button, which submits api.php, no pwd. But that is not a GET request.

You are correct - getFormField can return a null if the field is not there. But we only look at oldpwd with GET requests. The original code looked at GenericRequest.passwordHash from before the timein - and if there is none, that is "". Doing a SingleStringReplace of that string with a new password hash doesn't exactly work. I fixed that.

I can fix that, but the question is why I was sometimes getting a storage request with no pwd field?
I can't reproduce that any more.

I need to figure out how that happened, and go back to setting oldpwd to the original password hash - which can never be null.

This is very irritating debugging.

src/net/sourceforge/kolmafia/request/GenericRequest.java
Veracity0: Yes.

This happens if executeTimeinRequest returns false. It only does that if it fails to leave you logged in. We return "true", in that case,

Up the call chain, we loop (re)submitting the request until we are told to stop.
Failing to get a response text, getting - and handling - a redirect, and so on.
In particular, handleRedirect returns true if we should NOT resubmit the redirected request, which might have been updated with the redirectLocation.

If executeTimeinRequest returns true, we have successfully logged in, got the new password hash, and - for some reason - visited main.php.In that case, handleServerRedirect returns false - DO resubmit the request which triggered the timein.

If executeTimeinRequest returns false, we did not succeed in logging in. Used to be that only happened if you tried to log in during rollover. But now, since we (might) do a ping test at login, we can fail to log in if the user cancels the ping test, which leaves you logged out. In this case, we return true, which tells the caller to stop submitting the request.

Setting redirectHandled tells run() that even though it got a 302 - from the request that triggered the timein, we we processed it as much as we could and didn't just drop it on the floor. It prevents GenericRequest.run() from reporting "Unhandled redirect". We did handle it. Just not the way we might have hoped.

GenericRequest.run() is really complicated. It is chock full of special cases. Considering that it has grown and mutated for almost 20 years to get into its current state, I'd be scared to try to refactor it.

For good or for ill, my current project is deeply integrated into that code.
Fortunately, I was the one who invented and implemented timing in after KoL times you out.

src/net/sourceforge/kolmafia/request/GenericRequest.java
Veracity0: This happens:
```
> timeout
> pull 1 line
(KoLmafia requests a storage request, which is a GET with a pwd)
(we time in and replace the old pwd with the current one)
(the pull succeeds)
```
and now this:
```
timeout
> pull 1 line
(KoLmafia requests a storage request, which is a GET with a pwd)
(we time in and reject the ping, remaining logged out - and password hash cleared)
(we do not resubmit the pull)
> pull 1 line
(KoLmafia requests a storage request, which is a GET - but no pwd, since we have none.)
(we time in and accept the ping. We have a new password hash)
(we resubmit the request - still with no password hash, since we didn't know to replace it.)
(KoL returns a 200, but, since there was no pwd, did not actually pull anything.)
```

Basically, if you choose to refuse the ping during a timein - or a relogin - submitting a request that requires a pwd while you are logged out will time in, bu the request will fail.

I'm not sure how to fix that. Somehow remember an old pwd when you timein/relogin and reject a pin, so subsequent newly constructed requests with a pwd will have something (invalid) which we can replace with the new one?

That sounds messy - and I want to work on that in its own PR.

This one fixes your observation: it only looks for a pwd in the URL if GenericRequest.passwordHash is empty - and if the field does not exist in the URL (i.e. getFormField returns null), no potential NPE.

Veracity0: OK. I fixed that by giving you a passwordHash of "bogus" if you reject a ping test and we log you out.
This will make the URL submitted while logged out have a bogus pwd, which we will replace.

Veracity0:
Something I noticed in my testing is that the following code in LoginRequest:
```
    // If it's been less than 30 seconds since the last login
    // attempt, we could be responding to the flurry of login.php
    // redirects KoL gives us when the Relay Browser tries to open
    // game.php, topmenu.php, chatlaunch.php, etc.

    if (System.currentTimeMillis() - 30000 < LoginRequest.lastLoginAttempt) {
      return LoginRequest.completedLogin;
    }
```
will silently fail to time in if you try it too soon after logging in.

The "timein" command gets around that by not calling LoginRequest.executeTimeinRequest, but, instead, LoginRequest.retimein, which does not have that check.

But if you just submit a URL which forces a timein, that check will hit you.

Something to ponder some more, but not to fix in this PR, I think.

midgleyc (approved):
alright, good!

PR #1809, opened 2023-07-06
Ckb11 last friars nc

ckb11:


src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -339,9 +339,9 @@ public class Preferences {
         "lastCombatEnvironments",
         "lastCopyableMonster",
         "lastCouncilVisit",
-        "lastFriarElbowNC",
-        "lastFriarHeartNC",
-        "lastFriarNeckNC",
+        "lastFriarsElbowNC",
+        "lastFriarsHeartNC",
+        "lastFriarsNeckNC",
```
jaadams5, line 344: If I am searching the code correctly the preferences as defined in default don't have the s.  That says to me that either some script is using the wrong preference or there is a bug in KoLmafia where the wrong preference is being used.

jaadams5:
I'm going to reject and close this.  The preferences do not have the extra "s" when used in KoLmafia except for some code in QuestManager.handleFriarsCopseChange which after a casual inspection seems to be in error.  I am also thinking this is an opportunity for a test that shows the preference is being reset.  Write a test, demonstrate it fails, fix handleFriarsCopseChange and demonstrate it now works.

jaadams5:
I reopened this.  mafia was not consistent in the name.  I decided to fix that and standardized on the name that at least two scripts were using.

jaadams5 (approved):

jaadams5:
I didn't write a test because the easy way to do so required HTML that I cannot get at the moment.

Since I did contribute I will let someone else push it or unapprove it.

jaadams5:
2 days is too long :-)

PR #1811, opened 2023-07-07
Change synchronization of Preferences.reset() to address concurrency issue between timein() and Preferences.saveToFile()

BadHorseMonkey:
Changes synchronization of Preferences.reset() to use the same monitor as Preferences.saveToFile().  Adds a test for timein calling Preferences.reset() while other threads are calling Preferences.saveToFile().  Prior to the synch change, this was a failing test.

Expectation is that this corrects the case where a system goes to sleep, wakes up with a script still running, and gets the preferences erased by a race condition in timing in.


test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -559,6 +568,61 @@ public void exerciseIsPerUserGlobalProperty() {
     assertTrue(Preferences.isPerUserGlobalProperty("getBreakfast.PreferencesTestFakeUser"));
   }
 
+  public class timeinThread extends Thread {
+    public timeinThread(String s) {
+      super(s);
+    }
+
+    public void run() {
+      LoginManager.timein(CONCURRENT_USER_NAME);
```
heeheehee-kolmafia, line 577: I was thinking this would be something more like:
```
while (not done running) {
  LoginManager.timein(CONCURRENT_USER_NAME);
}
```

(You may need to Thread.yield() execution periodically, not sure.)

Note that [Thread#stop is deprecated](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/doc-files/threadPrimitiveDeprecation.html).

heeheehee-kolmafia:
Although, reading through the forum thread, I see that this actually is already sufficient to reproduce the race condition.

jaadams5:
>Still a draft because it seems to break other tests.

Which other tests?  After realizing the concoction database leaked across tests I would not be surprised if there were something else and might even investigate :-)

BadHorseMonkey:
> > Still a draft because it seems to break other tests.
> 
> Which other tests? After realizing the concoction database leaked across tests I would not be surprised if there were something else and might even investigate :-)

MallPriceDatabaseTest.writesDataInItemIdOrder() seems to assume that the database is empty.  It adds 3 prices and checks for the array size.  

timein is populating mallprices.txt and the assertion is failing because it sees the actual mall prices, all 7778 of them.

BadHorseMonkey:
> > > Still a draft because it seems to break other tests.
> > 
> > 
> > Which other tests? After realizing the concoction database leaked across tests I would not be surprised if there were something else and might even investigate :-)
> 
> MallPriceDatabaseTest.writesDataInItemIdOrder() seems to assume that the database is empty. It adds 3 prices and checks for the array size.
> 
> timein is populating mallprices.txt and the assertion is failing because it sees the actual mall prices, all 7778 of them.

We should also in this test kill that mallprices.txt file, since that's this test leaking state, but 1AM was not my moment of clearest mental acuity (and yet, that's when I code.  Coincidence?).

I was also having what might be a thread depletion problem, so implementing something like @heeheehee-kolmafia suggested thread cleanup may be important.  It's manifesting as non-deterministic behavior that clears right up if we @disable the new test.

Also, also it seemed to have trouble running on GitHub, which I think handles threading differently.

BadHorseMonkey:
OK, now the test is enabled, but the assertion is commented out, because it is a failing test.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -559,6 +568,61 @@ public void exerciseIsPerUserGlobalProperty() {
...
+                + i
+                + " incremented Pref: "
+                + Preferences.getInteger(incrementedPref)
+                + " "
+                + unrelatedPref
+                + ": "
+                + Preferences.getString(unrelatedPref, false));
+        // assertEquals(unrelatedValue, Preferences.getString(unrelatedPref, false));
+      }
+      assertEquals(unrelatedValue, Preferences.getString(unrelatedPref, false));
```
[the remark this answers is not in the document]
BadHorseMonkey: done!

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -559,6 +568,61 @@ public void exerciseIsPerUserGlobalProperty() {
...
+    public void run() {
+
+      Preferences.increment("counter", 1);
+    }
+  }
+
+  @Disabled
+  @Test
+  public void timeinDoesNotCauseRaceCondition() {
+    String unrelatedPref = "coalmine";
```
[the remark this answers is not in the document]
BadHorseMonkey: I was concerned that we might power through the SetPreferences and get the increment to write and kill a different preference.  This was based on anecdotal discussions from users about what prefs were still there after being nuked.  Since I'm using increment, it depends on the current value and is easier to see failures with, but if we were writing a string, we might write only the string in the otherwise blank prefs file...

Should we two assertions, when this is not a failing test.

jaadams5:
I've got some tweaks that mostly focus on tests and not the synch issues.  I commit if you don't think that would get in the way :-)  But they can wait.  I was mostly co firming other tests worked and things cleaned up.

Your tests seem to use two users and I have not figured out why but that is one way to try and force concurrent writes to global so...

jaadams5:
I've lost the bubble on this.  Does the test work?  Do all other tests work?  Are those statements true across all platforms?  How strongly do we believe that the test generates a corrupt preferences file?  Do we have any experience where the test fails without the changes to Preferences and LoginManager and passes when those changes are (re)introduced?  Thanks.

BadHorseMonkey:
> I've lost the bubble on this. Does the test work? Do all other tests work? Are those statements true across all platforms? How strongly do we believe that the test generates a corrupt preferences file? Do we have any experience where the test fails without the changes to Preferences and LoginManager and passes when those changes are (re)introduced? Thanks.
> 
I hope to get back to this by this weekend.  My mid-week has been eaten.

> P.S. - I abandoned my tweaks for the moment and will consider them when this is resolved.

They'll be welcome when we get there!

jaadams5:
Pulled locally and ran tests.  Did not fail in my Windows environment.  Don't really know what that means.  Reran here and Windows passed.

I am concerned about "Prefs could not be read and backup exists, trying backup.".  It seems to me that we might have an ill formed test.  That doesn't happen on main so your changes either exposed or created what is probably a setup/cleanup issue.

"Failed to delete settings\concurrencyfakeuser_prefs.txt" suggests a similar problem.

If I see anything obvious I'll call it out.

jaadams5:
Your tests pass without the code changes on my system.  I'd feel better if they had failed :-)  The messages noted above also occurred.

BadHorseMonkey:
> Pulled locally and ran tests. Did not fail in my Windows environment. Don't really know what that means. Reran here and Windows passed.
> 
> I am concerned about "Prefs could not be read and backup exists, trying backup.". It seems to me that we might have an ill formed test. That doesn't happen on main so your changes either exposed or created what is probably a setup/cleanup issue.

This is from code we did earlier around prefs, and the BeforeEach is deleting the prefs file, so giving it a screwy state.  I'll look at that as well.
> 
> "Failed to delete settings\concurrencyfakeuser_prefs.txt" suggests a similar problem.

I got rid of the concurrency fake user.  It existed because the other fake user was leaking state.  Now I don't need it.  I may be cleaning it up incorrectly.  Will check.

jaadams5:
> This is from code we did earlier around prefs, and the BeforeEach is deleting the prefs file, so giving it a screwy state. I'll look at that as well.

That's it.  Most preferences do not actually save the file.  The ones that do cleaned up the file but I am sure that the correct cleanup is to delete the preference file AND the bak if it exists.  I can look at that if it isn't done when I check tomorrow.

BadHorseMonkey:
> > This is from code we did earlier around prefs, and the BeforeEach is deleting the prefs file, so giving it a screwy state. I'll look at that as well.
> 
> That's it. Most preferences do not actually save the file. The ones that do cleaned up the file but I am sure that the correct cleanup is to delete the preference file AND the bak if it exists. I can look at that if it isn't done when I check tomorrow.

Yep, I was doing "if (!userFile.delete())" instead of "if (!backupFile.delete())", so it was failing to delete the userFile (already deleted) and leaving a backupFile around to get loaded and screw up other prefs.  Please retest!

BadHorseMonkey:
one day, I will learn the difference between "close this thread" and "close this PR".

Today is not that day.

jaadams5:
I was hoping you closed it in error.

I'm getting local test failures but I need to clean some other things off of my plate before I can investigate..  GitHub seems OK so I may have bolloxed my repo.

jaadams5:
Cancelled macos workflow that was still running after 3+,hours

jaadams5:
I deleted my branch, reimported this and have over 100 failed tests.  I'm looking.

BadHorseMonkey:
What I don’t want is something that causes other PRs to fail down the road, which is why I’m worried about some pass/some fail.

jaadams5:
I have never seen the threaded test fail so I have no informed opinion, either way, as to whether the changes to Preferences and LoginManager do what we hope and/or are really needed.

The Dvorak test fails intermittently for me but never reliably and so I have never tried to understand why.  In that spirit I don't want intermittent test failures but their existence is not a reason to not do something. I note that the local failures I have seen with this PR could have a common cause related to cleanup or leakage.

While I was writing this Dvorak, failed here on macos https://github.com/kolmafia/kolmafia/actions/runs/5580287476/jobs/10196982903?pr=1837

Some of what we have discovered about other tests is worth keeping.  I would shelve the changes to Preferences and LoginManager and comment out the thread test.  I would then do what needed to be done to eliminate failures, leftover files after the entire test, restore from bak messages and failure to delete messages.  I think I would commit at that point, enable the thread test and repeat the process and hope that the thread test fails consistently.  Once it does then I would restore the changes to Loginmanager and Preferences and see whether they fix the test.  But if I can only have one thing it would be a thread test that fails until the new code is added :-)

jaadams5:
OK.  The undeleted preference file issue exists when the new test is disabled.  The new test did fail without the corresponding code so the situation is much better than I understood it to be.  There may be some changes to be merged into PreferenceTest.

BadHorseMonkey:
> OK. The undeleted preference file issue exists when the new test is disabled. 

Did you sync my latest changes?  I made a change in [846b225] 

>The new test did fail without the corresponding code so the situation is much better than I understood it to be. There may be some changes to be merged into PreferenceTest.

Yep, wrote a failing test first, then added a tiny code fix.  I'm worried that my method of disposing of threads is not working and leaving GitHub with a thread starvation problem.  I know they have a very different threading model than a typical single-user laptop.  Possibly we can make it worse locally by upping the thread count.

jaadams5:
I'm manually sync'ing but my understanding is now such that I can probably work directly from this again.

I don't know how to handle "other tests".  The Dvorak test has a history of intermittent failures and within the past week or so I have have restarted tests that made no progress after a couple of hours.  There are a couple of cases where we have identified a parameter that had to be changed in our GitHub environment in order to get passing tests on GitHub as well as locally so maybe, if there is a thread issue, that is the solution.

The preference file persistence seems to only occur in the two tests that test the write/no write toggle.  I'll know more later today, assuming the day goes as planned, but I think the issue is caused by Cleanups.  Setting the flag to on causes a write and I think the test isn't expecting that.  I will think that through.

jaadams5:
The recent macos failure is very much like something I saw locally on Windows but it did not reoccur so I didn't do anything with it.

BadHorseMonkey:
> The recent macos failure is very much like something I saw locally on Windows but it did not reoccur so I didn't do anything with it.

I've been letting other people's PRs land and catching up my PR so that it can run the tests a bunch of times.  Latest seems much more stable.  I only changed the thread creation/destruction code.

jaadams5:
The rabbit hole gets deeper.  At the moment I think the tests savesSettingsIfOn() and canToggle() may be logically flawed but I am not prepared to defend that statement yet.  ClearSharedStateBefore has code that gets invoked beforeAllTests.  A side effect of resetting a character is that the concoctions database is refreshed thus incrementing _concoctionDatabaseRefreshes.  lastRelayUpdate also gets set.  The preference saveSettingsOnSet defaults to true and the code switch to write the file should always be true, unless explicitly set otherwise as part of a test.  So our test initialization will always create and write a preferences file that will be subsequently deleted as part of the beforeAll initialization.  This may have some bearing on the failure to delete messages that are occurring.  In my playpen the delete failure does not throw an exception and I cannot yet explain why the file cannot be deleted unlss it is timing related.

BadHorseMonkey:
> The rabbit hole gets deeper. At the moment I think the tests savesSettingsIfOn() and canToggle() may be logically flawed but I am not prepared to defend that statement yet. ClearSharedStateBefore has code that gets invoked beforeAllTests. A side effect of resetting a character is that the concoctions database is refreshed thus incrementing _concoctionDatabaseRefreshes. lastRelayUpdate also gets set. The preference saveSettingsOnSet defaults to true and the code switch to write the file should always be true, unless explicitly set otherwise as part of a test. So our test initialization will always create and write a preferences file that will be subsequently deleted as part of the beforeAll initialization. This may have some bearing on the failure to delete messages that are occurring. In my playpen the delete failure does not throw an exception and I cannot yet explain why the file cannot be deleted unlss it is timing related.

This is both fun and sometimes frustrating to nail down, especially since the problems are hard to isolate.

I'm wondering if using a different user for the concurrency test was actually helpful in terms of isolating this change from propagating.

In the @AfterEach resetCharAndProps(), we try to delete the prefs completely.

```
 if (MallPriceFile.exists()) {
      if (!MallPriceFile.delete()) {
        System.out.println("Failed to delete " + MallPriceFile);
      }
    }
    if (userFile.exists()) {
      if (!userFile.delete()) {
        System.out.println("Failed to delete " + userFile);
      }
    }
    if (backupFile.exists()) {
      if (!backupFile.delete()) {
        System.out.println("Failed to delete " + backupFile);
      }
    }
```

An earlier version had a defect of 
```    if (backupFile.exists()) {
      if (!userFile.delete()) {```
      
If you still have that defect, it would manifest as printing "Failed to delete..."

**Comment by jaadams5** [comment]
I changed the fork parameter so that all of the tests were run on one thread.  That's 60 minutes of my life I won't get back.  Two lessons learned.  One is that the failure to delete is somewhat of a red herring.  I had a window open, monitoring the directory saw the message and then saw that the file was no longer there.  I could construct a hypothesis that there is a time lag between when we issue file.delete() and when the OS actually does that and the lag is due to the test infrastructure taking some time.  The fact that the sequence file.delete(), file.exists() -> true file.deleteOnExit() freezes things reinforces my belief that the file is deleted but it takes time for the test code to become aware of that.  I also had test failures when using just one thread but I don't think they have anything to do with this work.  I'm kind of inclined to declare this as good as it is going to get, clean up some of the debugging (although I do like verboseDelete() :-) ) see what he^3 thinks and then push it.

**Comment by BadHorseMonkey** [comment]
Yeah, the syncs with other code have helped raise my confidence. I was always pretty confident in the fix, which is small, but less so in the test.

**Comment by jaadams5** [comment]
The failures on Ubuntu are starting to seem consistent.  I don't know what that means other than we should understand what we did that made it go away, when that has happened.  GitHub has some debugging which should probably be turned on to see if we learn anything.  There may be tweaks to tests to be tried or more output.  I often treat "possible" and "probable" the same and so I am off in LaLa Land observing that the errors involve illegal arguments caught by code that is static and wonder if the test is actually corrupting the contents of the JVM.  Not sure whether I branch from this to try things or just work here and revert.  The habits formed by a version control system that didn't handle branching/forking very well are hard to break :-)

**Comment by BadHorseMonkey** [comment]
I looked at where it was failing.  IT was failing in the @AfterEach cleanup task with an IllegalArgumentException on KolCharacter.reset("");  

Which makes no sense.  So I put that in a try/catch to try to grab the stack at the time of failure.

We'll see.

**Comment by jaadams5** [comment]
and of course the last push did not trigger the failure.

**Comment by jaadams5** [comment]
Latest failure, only on mac, starts with timeinDoesNotCauseRaceCondition() which is much more concerning.  I am willing to attribute most of the other failures to an environment issue - but this raises the question of whether the test and fix really demonstrate and fix the basic issue.

**Comment by jaadams5** [comment]
Converted to draft because this is currently more of a collaboration than a solution :-)  I did not push updates because another opinion on the macos failure would be useful and not rerunning tests is the easiest way I know to keep that visible.

**Comment by BadHorseMonkey** [comment]
> Converted to draft because this is currently more of a collaboration than a solution :-) I did not push updates because another opinion on the macos failure would be useful and not rerunning tests is the easiest way I know to keep that visible.

Ive been meaning to put it back to draft, and I'm happy for the collar.

The Mac falilure is on me.  In my previous experimental push, I re-enabled the assertion on lines 711-712 because it wasn't failing for me, but it's not testing this fix, it's testing the related fix of multiple simultaneous updates on different threads.  I'm putting it back to the way it should be.  That should be a second PR and test, so I should refactor it.

Step one is to comment it out, though.

**Comment by BadHorseMonkey** [comment]
I think that's failing because read doesn't lock waiting for write, so increment (read then write) is problematic.  It's not a problem is we're blind writing, but increment needs to read as part of the write.

But our reads are harder to lock, since they don't go through one function like write does (I will need to look harder at this).

**Comment by BadHorseMonkey** [comment]
> I think that's failing because read doesn't lock waiting for write, so increment (read then write) is problematic. It's not a problem is we're blind writing, but increment needs to read as part of the write.
> 
> But our reads are harder to lock, since they don't go through one function like write does (I will need to look harder at this).

Ah, no, the test that was failing was different.  It was testing if the thread was disposed of.  It's not a functional test, but it's there to see if the rest of the failures are caused by a thread issue.

So, maybe it's a good sign!

**Comment by BadHorseMonkey** [comment]
seems to break when I re-run tests.  that sucks.

**Comment by BadHorseMonkey** [comment]
I think your fixes are holding up well.

**Comment by BadHorseMonkey** [comment]
moved this to review.  Is there anything we want to clean up prior to merging?

**Comment by jaadams5** [comment]
I will flag cleanups.  Ideally @heeheehee-kolmafia will review and approve before push.

**Comment by jaadams5** [comment]
I pushed my tweaks.  I'll pretend I am doing a real review :-)

File: test/net/sourceforge/kolmafia/preferences/PreferencesTest.java [hunk 4]
```diff
@@ -559,6 +580,147 @@ public void exerciseIsPerUserGlobalProperty() {
...
+    public incrementThread(String s) {
+      super(s);
+    }
+
+    public void run() {
+      try { // force a random delay so that the tests can run different orders
+        long delay = (long) (Math.random() * 100);
+        TimeUnit.MILLISECONDS.sleep(delay);
+        int priorValue = Preferences.getInteger("counter");
+        boolean checkIncrementedValue = false;
```
jaadams5, line 609: Consider removing the code controlled by this

jaadams5:
Is there some kind of automated retry on test failures?  I push something, get an email that says the tests failed but when I come here to look at the failure all tests have passed.  If I navigate from the email I do see a failure but it tends to be something catastrophic, intermittent and probably an issue with the test environment and not the test.

BadHorseMonkey:
> Is there some kind of automated retry on test failures? I push something, get an email that says the tests failed but when I come here to look at the failure all tests have passed. If I navigate from the email I do see a failure but it tends to be something catastrophic, intermittent and probably an issue with the test environment and not the test.

Hmm.  It should tell us everything that happened, somewhere.  There are errors due to permissions with publishing the test results.

Who set this up?  I can't remember if it was @Rinn, @gausie or @fewyn .

fewyn:
> > Is there some kind of automated retry on test failures? I push something, get an email that says the tests failed but when I come here to look at the failure all tests have passed. If I navigate from the email I do see a failure but it tends to be something catastrophic, intermittent and probably an issue with the test environment and not the test.
> 
> Hmm. It should tell us everything that happened, somewhere. There are errors due to permissions with publishing the test results.
> 
> Who set this up? I can't remember if it was @Rinn, @gausie or @fewyn .

I've only got access to Jenkins where I just recently enabled re-try on failure, the Github stuff I don't have access to at all.

BadHorseMonkey:
> > Is there some kind of automated retry on test failures? I push something, get an email that says the tests failed but when I come here to look at the failure all tests have passed. If I navigate from the email I do see a failure but it tends to be something catastrophic, intermittent and probably an issue with the test environment and not the test.
> 
> Hmm. It should tell us everything that happened, somewhere. There are errors due to permissions with publishing the test results.
> 
> Who set this up? I can't remember if it was @Rinn, @gausie or @fewyn .

@jaadams5 Is this the error report you're seeing?

I'm not sure what it's telling us, but the tests seem to be passing, which is what we care about.  I'm not sure what the other failures are about.  I'll ask in the forums.

jaadams5:
This is the error I get when I follow the link in my email.  https://github.com/kolmafia/kolmafia/actions/runs/5770861482/job/15644272992  It appears to be triggered by chanegs I pushed this morning.

But this current pags says all checks passed and when I go https://github.com/kolmafia/kolmafia/pull/1811/checks I don't find it.  It is in https://github.com/kolmafia/kolmafia/actions which I could not find before (operator error) but I don't know what triggered the rerun.  Maybe GitHub thought it was their error?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -559,6 +579,130 @@ public void exerciseIsPerUserGlobalProperty() {
...
+
+      assertEquals(
+          unrelatedValue,
+          Preferences.getString(unrelatedPref, false),
+          "unrelated pref does not match");
+      System.out.println("Final value: " + Preferences.getInteger(incrementedPref));
+    }
+  }
+
+  @Disabled
```
heeheehee-kolmafia, line 672: Does this test still fail?

heeheehee-kolmafia (dismissed):
The placement of the locks looks correct.

We'll have to see how the performance hit looks in practice.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
BadHorseMonkey: with a minor tweak, no!  I re-enabled it.  It was comparing "counterTwo" to "counter", which was never going to work...

BadHorseMonkey:
> The placement of the locks looks correct.
> 
> We'll have to see how the performance hit looks in practice.

definitely a concern.  However, it requires simultaneous prefs reads/writes to have a delay waiting for a lock, and we want it to finish that step before going on.

It might be a bit of an overkill on requiring a lock to read, but it means that read will wait for an in-progress write or more importantly, an in-progress timein.

heeheehee-kolmafia:
> > The placement of the locks looks correct.
> > We'll have to see how the performance hit looks in practice.
> 
> definitely a concern. However, it requires simultaneous prefs reads/writes to have a delay waiting for a lock, and we want it to finish that step before going on.
> 
> It might be a bit of an overkill on requiring a lock to read, but it means that read will wait for an in-progress write or more importantly, an in-progress timein.

There is always going to be some overhead associated with synchronization even if your lock is usually uncontended -- the rule of thumb I use is something around 100-1000 CPU cycles (~30-300ns) in lower-level languages; from a quick skim it looks like the heavy lifting is probably something like x86's `LOCK CMPXCHG` for Java, plus some additional orchestration.

For users with saveSettingsOnSet=true, we'll be incurring this cost for every preference set. And, with your latest commit, also on every preference get.

heeheehee-kolmafia (approved):

jaadams5:
I'm going to build, then run locally.  If no issues then I will push if no one has done so sooner.

jaadams5 (approved):

jaadams5:
Performance hit was not noticeable with three farming characters.  Squashing and merging.  Let the fallout begin.

PR #1815, opened 2023-07-08
feat: modifier type

midgleyc:
As suggested in #1797, add a `$modifier` type.

The main downside is that `modifier` is used as a variable name in a few common scripts (e.g. some of Ezandora's (Gain, Bastille, Helix Fossil), garbage-collector), and these would need adjusting.

The main upside is improved comprehensibility for modifier code.

Modifiers have a name (same as the string you would previously use), and a type, which is one of "numeric", "string" or "boolean". This tells you which of the other methods you should use to get the value (and what the value is -- float, string or boolean respectively).

Also add a function `split_modifiers` which takes a modifier string (e.g. the string from `$location[].pledge_allegiance`) and returns a map of modifier to string. This can then be parsed, using the types, to extract the values.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
@@ -8,6 +8,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -281,6 +282,7 @@ public class ModifierDatabase {
     mods.addAll(Arrays.asList(DerivedModifier.values()));
     mods.addAll(Arrays.asList(StringModifier.values()));
     mods.addAll(Arrays.asList(BooleanModifier.values()));
+    mods.sort(Comparator.comparing(Modifier::getName));
     allModifiers = mods;
     return mods;
   }
--- a/src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
@@ -271,6 +271,33 @@ public class ModifierDatabase {
     return BitmapModifier.byCaselessName(name);
   }
 
+  private static List<Modifier> allModifiers = null;
+
+  public static List<Modifier> allModifiers() {
+    if (allModifiers != null) return allModifiers;
+    List<Modifier> mods = new ArrayList<>();
+    mods.addAll(Arrays.asList(DoubleModifier.values()));
+    mods.addAll(Arrays.asList(BitmapModifier.values()));
+    mods.addAll(Arrays.asList(DerivedModifier.values()));
+    mods.addAll(Arrays.asList(StringModifier.values()));
+    mods.addAll(Arrays.asList(BooleanModifier.values()));
+    allModifiers = mods;
+    return mods;
+  }
+
+  /** Get any modifier by name, ignoring case */
+  public static Modifier byCaselessName(String name) {
+    var num = ModifierDatabase.numericByCaselessName(name);
+    if (num != null) {
+      return num;
+    }
+    var str = StringModifier.byCaselessName(name);
+    if (str != null) {
+      return str;
+    }
+    return BooleanModifier.byCaselessName(name);
+  }
+
   // region: get registered values
 
   public static final Modifiers getItemModifiers(final int id) {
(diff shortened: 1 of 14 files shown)


gausie:
I think JavaScript should be fine here, no? So garbage-collector will not
be affected.

midgleyc:
You're right, JavaScript isn't affected.

Veracity0:
I'm a bit concerned about Ezandora's scripts. Both you and I have forks and have submitted PRs from them to deal with things like git installation. I also have local changes to some things (like compile warnings for fuzzy matching and overriding RuntimeLibrary functions).

Making modifier a reserved word will require a script change. We can fix our forks, but other users will have to use our forks, until Ezandora fixes her originals.

(As an aside, I intend to make "var" a reserved word/type in ASH. I expect that will cause some script breakage, as well. "zlib" is the only one I see in my scripts folder. How to deal with it?)

Given those external concerns, I'll start with my review and look for internal concerns. :)

src/net/sourceforge/kolmafia/textui/DataTypes.java
```diff
@@ -1065,6 +1096,11 @@ private static String promptForValue(final Type type, final String message, fina
       case STAT:
         return InputFieldUtilities.input(message, DataTypes.STAT_ARRAY);
 
+      case MODIFIER:
+        return InputFieldUtilities.input(
+                message, ModifierDatabase.allModifiers().toArray(new Modifier[] {}))
```
Veracity0, line 1101: This will be big. Is it sorted?
As seen above, no.
Should it be?

src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
```diff
@@ -271,6 +271,33 @@ public static Modifier numericByCaselessName(String name) {
...
+    if (allModifiers != null) return allModifiers;
+    List<Modifier> mods = new ArrayList<>();
+    mods.addAll(Arrays.asList(DoubleModifier.values()));
+    mods.addAll(Arrays.asList(BitmapModifier.values()));
+    mods.addAll(Arrays.asList(DerivedModifier.values()));
+    mods.addAll(Arrays.asList(StringModifier.values()));
+    mods.addAll(Arrays.asList(BooleanModifier.values()));
+    allModifiers = mods;
+    return mods;
+  }
```
Veracity0, line 286: OK, this is a list, ordered by (something) within modifier types.

Veracity0 (approved):

src/net/sourceforge/kolmafia/textui/DataTypes.java
midgleyc: it's sorted by type (numeric, then string, then boolean), but inside the types it's "sorted" by ordering within the enums.

Ordering by name inside the types would make more sense. Alternatively, they compare as strings (`isStringLike`), so perhaps they should be sorted by name as well.

Alternatively, it doesn't matter: I can't see any use cases that would care about the ordering.

midgleyc:
On Ezandora's scripts: I did submit a PR to autoscend to use my fork of Voting-Booth due to SVN concerns, so scripts that are still being updated should do okay.

On zlib: I think zarqon is still around, right? So that should be fixable directly.

src/net/sourceforge/kolmafia/textui/DataTypes.java
Veracity0: This particular piece of code is passing an array to be used in a combo box in a dialog. Which is to say, the user will scroll through it looking for the item to select.

Hence my thinking pure alphabetical would be the best for this user-facing use case.

midgleyc: I think this is reasonable, and I think pure alphabetical is also reasonable in general for consistency.

Veracity0 (approved):

midgleyc:
Fallout: the blast radius for this was slightly higher than I expected, also affecting:
* https://github.com/Loathing-Associates-Scripting-Society/autoscend/ (https://github.com/Loathing-Associates-Scripting-Society/autoscend/pull/1325)
* https://github.com/Loathing-Associates-Scripting-Society/ChIT (https://github.com/Loathing-Associates-Scripting-Society/ChIT/pull/31)
* https://github.com/loathers/TourGuide (https://github.com/loathers/TourGuide/pull/158)
* https://github.com/Loathing-Associates-Scripting-Society/UberPvPOptimizer (https://github.com/Loathing-Associates-Scripting-Society/UberPvPOptimizer/pull/3)

The authors managed to fix all of these very quickly, which is impressive!

PR #1817, opened 2023-07-08
Add ping ASH functions

Veracity0:
Return values are an anonymous record:
```
record
{
    string page;
    int count;
    int low;
    int high;
    int total;
    int bytes;
    int average;
    int bps;
};
```
Three new functions:

ping() - runs a "standard" ping test and returns the result
ping(100, "api.php") - runs 100 api.php requests and returns the result
ping("pingLatest") - decodes the specified property as a ping result and returns the record.

```
> get pingLatest

api.php:10:24:29:271:19620:27

> ash ping("pingLatest")

Returned: record {string page; int count; int low; int high; int total; int bytes; int average; int bps;}
page => api.php
count => 10
low => 24
high => 29
total => 271
bytes => 19620
average => 27
bps => 72398

> js ping("pingLatest")

Returned: aggregate string [string]
average => 27
bps => 72398
bytes => 19620
count => 10
high => 29
low => 24
page => api.php
total => 271
```

Unfortunately, there's currently no way to store the anonymous record in a variable in ASH; even declaring an identical anonymous record does not work. ASH needs "var".

You can do this:
```
> ash ping("pingLatest").average

Returned: 27
```
But that's not an adequate solution if you want to inspect more fields in the result - especially for the other functions which will run a new ping test each time you call them.


Shiverwarp:
Malibu Stacey gave a nice solution for lacking var in this context. We can do something like:

foreach it, entry in ping() {
  print(entry);
}

I haven't tested it myself, but apparently it works well!

Veracity0:
Well, the issue is that ping returns a single record, not an aggregate.

We have other RuntimeLibrary functions that return arrays of anonymous records, and foreach works for that. (Notice that you don't declare a data type for the foreach variables; they are essentially "var" since the types are inferred from the object you are iterating over.)

Can't do that with a single anonymous record, though. I believe that svn_info and git_info have a similar problem.

midgleyc:
If you could make anonymous records with the same fields compare as equal types, you could at least declare a record to put `git_info` into.

The bigger issue is `maximize`, which requires fields named `effect`, `item` and `skill`, that you can't successfully declare.

Veracity0:
> If you could make anonymous records with the same fields compare as equal types, you could at least declare a record to put `git_info` into.

It's trickier than that. The records have to have the same # of fields, same field types - and same field names.
```
record PingTest
{
    string page;
    int count;
    int low;
    int high;
    int total;
    int bytes;
    int average;
    int bps;
};

PingTest result = ping();
print("page = " + result.page);
print("count = " + result.count);
print("low = " + result.low);
print("high = " + result.high);
print("total = " + result.total);
print("bytes = " + result.bytes);
print("average = " + result.average);
print("bps = " + result.bps);
```
ping() returns an anonymous record which looks identical to the PingTest record I defined.
"result" is type PingTest, but when I assign ping's return value to it, "result" is now actually the anonymous record type.
When I ask for "result.page", it looks up the index of key "page" in the anonymous record's array of field names, NOT in PingTest's array of field names.
As implemented, when ASH parses "result.page", it looks up the Value for the key "page" in PingTest's field name array and passes that as the index to the record lookup for the anonymous record.
As implemented, RecordValue looks up that key using "==" for the Value key.
That fails.

If I change it to .equals(), it does a String comparison and works.

```
> pingrec.ash

page = api.php
count = 10
low = 19
high = 32
total = 240
bytes = 19620
average = 24
bps = 81750
```

I suppose an alternative would be for the parser to convert field name lookups into field index lookups; since it does validate the field names at compile time, it knows the index.

> The bigger issue is `maximize`, which requires fields named `effect`, `item` and `skill`, that you can't successfully declare.

maximize returns an array of those, so you CAN iterate over it and use the fields of the anonymous record within the loop. You cannot pass the anonymous record in a variable and pass it to a function, say.

I have a solution coded up in RecordType and RecordValue.

This PR uses existing (flawed) technology: a RuntimeLibrary function which returns an anonymous record.
I can fix the flaws, but I'd prefer that go into a separate PR.

midgleyc (approved):

PR #1819, opened 2023-07-09
Add some helper types TypeScript definitions

gausie:
- Add helper type for Environment union to TypeScript defs
- Spin out Modifier.getType() string results into ModifierValueType enum
- Add helper type for ModifierValueType to TypeScript defs


midgleyc (approved):

midgleyc:
Technically, empty string is a modifier "type" (it's the type of the null modifier, which you can get with `ash $modifier[none]`). I'm fine with it being omitted, though.

gausie:
No no that is a good point. I'll add to the enum and get it to return in
get_type. But I'm in bed now so I'll do it tomorrow!

midgleyc (approved):

PR #1826, opened 2023-07-12
bugfix: latteIngredients pref incorrectly set when changed

horrible-little-slime:


jaadams5:
What is the problem you are addressing?  I see you added a new test case but just reviewing the code, I cannot see why it would have failed without your changes.  Help me out here with context, please.

gausie:
At least one test on LatteRequest.parseResponse would be awesome

gausie (approved):

PR #1829, opened 2023-07-13
feat: support eagle screech

midgleyc:
Patriotic Screech banishes the phylum of the monster you use it on.
* It lasts 100 turns, and does not go away at rollover
* It does not apply in no-banish zones
* It can be used in no-banish zones, and the banish works in other zones
* It can be used again after 11 combats with the eagle


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -6459,6 +6461,11 @@ private static void processChildren(final TagNode node, final TagStatus status)
           FightRequest.handleMelodramedary(str, status);
         }
 
+        if (status.eagle && str.contains(status.familiarName)) {
+          // Patriotic Eagle action
+          FightRequest.handlePatrioticEagle(str, status);
```
Veracity0, line 6466: I'd like to see a test with fight.php HTML that tests this.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7414,6 +7421,10 @@ private static void processFamiliarAction(TagNode node, TagNode inode, TagStatus
       return;
     }
 
+    if (status.eagle && FightRequest.handlePatrioticEagle(str, status)) {
+      return;
```
Veracity0, line 7425: And one for this case, too.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7547,6 +7558,26 @@ private static boolean handleMelodramedary(String str, TagStatus status) {
     return true;
   }
 
+  private static boolean handlePatrioticEagle(String str, TagStatus status) {
```
Veracity0, line 7561: The net result being, this method is called and the combat properties are set - and that is tested.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -10249,6 +10280,14 @@ private static void payActionCost(final String responseText) {
           skillSuccess = true;
         }
         break;
+
+      case SkillPool.PATRIOTIC_SCREECH:
+        if (responseText.contains("releases an ear shattering screech")) {
```
Veracity0, line 10285: More of the same?

Veracity0 (changes requested):
I have never looked at BanishManager before. I realize that is a big game-play mechanism for those who case about speed, but, personally, I care about day count but not turn count, so, I don't do fiddly stuff.

This is why I have resisted looking at this PR, so far; I've been hoping somebody who understands how this works would step in and review.

Perhaps I will be back later, if somebody who is already familiar with banishing hasn't stepped up.

PR #1830, opened 2023-07-13
feat: pocket guide to mild evil skeleton

midgleyc:


src/data/items.txt
```diff
@@ -11330,3 +11330,22 @@
...
+11314
+11315
+11316
+11317
+11318
+11319
+11320
+11321
+11322	Pocket Guide to Mild Evil	636540989	tinybook.gif	usable	t	0
+11323	Pocket Guide to Mild Evil (used)	508639228	tinybook.gif	usable		0
```
Veracity0, line 11351: Surely this one is reusable.

src/data/items.txt
```diff
@@ -11330,3 +11330,22 @@
...
+11313
+11314
+11315
+11316
+11317
+11318
+11319
+11320
+11321
+11322	Pocket Guide to Mild Evil	636540989	tinybook.gif	usable	t	0
```
Veracity0, line 11350: Add these both to the switch in UseItemRequest.parseConsumption that calls UseItemRequest.itemToSkill - which looks up the modifier(.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -10249,6 +10249,18 @@ private static void payActionCost(final String responseText) {
...
         }
         break;
+
+      case SkillPool.PERPETRATE_MILD_EVIL:
+        if (
+        // Spookypocket
+        responseText.contains("so discombobulated")
+            ||
+            // no items
+            responseText.contains("even mildly evil")
```
gausie, line 10259: My understanding is that if there are no items missing it doesn't spend a charge

Veracity0 (approved):

src/net/sourceforge/kolmafia/request/FightRequest.java
midgleyc: I've tested and it definitely does spend a charge. The message does indicate that perhaps it shouldn't.

PR #1831, opened 2023-07-13
ASH record field names can be reserved words

Veracity0:
ASH record field names are allowed to be reserved words.
MaximizerResult is a record, an array of which is returned by the 5 argument version of maximize().
That record contains fields named, effect, item, and skill.


Veracity0:
The new test works perfectly on my system, but fails on github.
Huh.

Veracity0:
A-yup. It is some other test which is failing.
There is clearly test leakage somewhere.
I see there is no cleaning up after or clearing state before tests.
So, if a test results in changing character state, all bets are off.

Veracity0:
Gah. test_evilometer.txt
```
test load evilometer_fight.txt
test fight 1
```
Results in:
```
Read 9,727 bytes into a 9,727 character string
Round 2: gargantulihc takes 91 damage.
Round 2: gargantulihc takes 80 damage.
Round 2: You lose 17 hit points
Round 2: CustomScriptTest wins the fight!
After Battle: As the gargantulihc fades away into nothingness, your Evilometer emits a single loud beep.
After Battle: Ben stands up on his hind legs and sniffs the air, chirping adorably.
After Battle: You gain 3 Mana Points
You acquire an item: lihc eye
You acquire an item: lihc eye
```
And then stat gains, which change character state:
```
After Battle: You gain 18 Fortitude
You gain a Muscle point!
After Battle: You gain 44 Wizardliness
You gain a Mysticality point!
After Battle: You gain 19 Smarm
```
Or is it?
```
After Battle: You gain 18 Fortitude
You can now equip a mafia wedding ring (and possibly other things).
You gain a Muscle point!
After Battle: You gain 44 Wizardliness
You gain a Mysticality point!
After Battle: You gain 19 Smarm
```

You have got to be kidding me.

midgleyc:
You could try shifting that one out into `RuntimeLibraryTest` and ensuring the character has 300 in each stat, so the "you can now equip" message never occurs.

Veracity0:
I am not going to spend time working on a broken test that has literally nothing to do with my change.

jaadams5:
> I am not going to spend time working on a broken test that has literally nothing to do with my change.

I have little interest, and less knowledge, in ash syntax and parsing so I have not been paying attention to this PR.  But I am highly amused to "mess with" tests.  I have time and would enjoy tracking down an unrelated broken test.  If this PR is approved with disabled tests just let me know which ones to re-enable and I will see if I can find the leak or other cause.

Veracity0:
I renamed test_evilometer.txt to test_evilometer.txt.bad

The thing about CustomScriptTest is that it essentially runs scripts as if from the not-logged-in gCLI.

Which is to say, a script cannot depend on character state - there is none - and if a script modifies character state - like this one, which parses a fight response and modifies character state - it inherently leaks.

It should be rewritten to go into FightRequestTest, or something, with cleanups. I don't understand the reasoning behind putting it into CustomScriptTest.

I wondered about where to put MY test, which calls the maximizer. My issue was that I wanted to write a test that demonstrated that a user script could define a record that mirrored MaximizerResult - "reserved word" field names and all - and call the maximizer and have the results stored in that user-defined record.

Which is to say, not testing the maximizer, but the record assignment.

jaadams5:
I'll look at the test once the PR is in.

My recollection is that because of the existing load and execute capabilities a Custom Script was the easiest way to process captured HTML to see if the code responded properly.  Now that we have test fixtures that simulate receiving an HTTP request we have better ways to accomplish the same thing.  That's probably what I will do since it can't hurt for me to be reminded the best way to do that now :-)

Veracity0:
Can I get an approval from somebody, please?

jaadams5 (approved):
You have some debugging changes in /test , if I understand correctly, that I might remove now that you have one failing test that has been eliminated by a rename, but that is a judgement call you may make.

Veracity0:
I'm going to keep them, so I don't have to put them back next time a CustomScriptTest fails.

PR #1832, opened 2023-07-15
tests: move evilometer script test into TestCommandTest

midgleyc:
As #1831 identified it as flaky, move evilometer test out into TestCommandTest.

This could be e.g. in FightRequest, but I think it's nice to have some coverage of `TestCommand` (to show how it can be used without having to read the source code).


test/net/sourceforge/kolmafia/textui/command/TestCommandTest.java
```diff
@@ -0,0 +1,88 @@
...
+    this.command = "test";
+  }
+
+  @BeforeAll
+  public static void init() {
+    KoLCharacter.reset("TestCommandTest");
+  }
+
+  @Nested
+  class Evilomete {
```
jaadams5, line 32: class Evilometer, perhaps?

midgleyc: oops, fixed

jaadams5:
I was looking at FightRequest to do this since the point of the test seemed to be that the evilometer was correctly tracked.  

But this replicates what was there before and creating TestCommandTest might encourage more tests so this way is good too.

jaadams5 (approved):

PR #1833, opened 2023-07-15
more good status effects via hookah

Veracity0:
I have five characters that equip a hookah when running their turns.

```
All Rainbow-y is good quality, but KoLmafia thought it was neutral
Aloofah is good quality, but KoLmafia thought it was neutral
Anti-Burning is good quality, but KoLmafia thought it was neutral
Anti-Creeped is good quality, but KoLmafia thought it was neutral
Anti-Freezing is good quality, but KoLmafia thought it was neutral
Anti-Goosebumped is good quality, but KoLmafia thought it was neutral
Anti-Hurt is good quality, but KoLmafia thought it was neutral
Anti-Odored is good quality, but KoLmafia thought it was neutral
Ashamed is good quality, but KoLmafia thought it was neutral
Bet Your Autumn Dollar is good quality, but KoLmafia thought it was neutral
Bulging Jaw Muscles is good quality, but KoLmafia thought it was neutral
Burningly Oiled is good quality, but KoLmafia thought it was neutral
Carbonated is good quality, but KoLmafia thought it was neutral
Cosmic Ball in the Air is good quality, but KoLmafia thought it was neutral
Craving Prawns is good quality, but KoLmafia thought it was neutral
Delayed Gratification is good quality, but KoLmafia thought it was neutral
Delighted by Autumn is good quality, but KoLmafia thought it was neutral
Double Carbonated is good quality, but KoLmafia thought it was neutral
Double Frozen is good quality, but KoLmafia thought it was neutral
Double Hot Soupy Garbage is good quality, but KoLmafia thought it was neutral
Doubly Burningly Oiled is good quality, but KoLmafia thought it was neutral
Doubly Craving Prawns is good quality, but KoLmafia thought it was neutral
Doubly Shivering Spine is good quality, but KoLmafia thought it was neutral
Doubly Spookily Greasy is good quality, but KoLmafia thought it was neutral
Earning Interest is good quality, but KoLmafia thought it was neutral
Energized is good quality, but KoLmafia thought it was neutral
Extra-Green is good quality, but KoLmafia thought it was neutral
Feeling Nervous is good quality, but KoLmafia thought it was neutral
Fire-Roasted Feelings is good quality, but KoLmafia thought it was neutral
Five Sticky Fingers is good quality, but KoLmafia thought it was neutral
Flapper Dancin' is good quality, but KoLmafia thought it was neutral
Frozen is good quality, but KoLmafia thought it was neutral
Gaffe Free is good quality, but KoLmafia thought it was neutral
[...]

PR #1834, opened 2023-07-15
Response text parser switches

Veracity0:
Replace another if-then-elseif chain of string.startsWith() with a modern switch with string cases.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
+++ b/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
@@ -172,7 +172,6 @@ public class ResponseTextParser {
           "desc_familiar.php",
           "desc_guardian.php",
           "desc_item.php",
-          "desc_monster.php",
           "desc_outfit.php",
           "desc_skill.php" -> {
         // No object descriptions
--- a/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
+++ b/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
@@ -146,44 +146,75 @@ public class ResponseTextParser {
       queryString = location.substring(queryStringBegin + 1);
     }
 
+    // *** This is an unfinished and unused feature.
     if (!PageRegistry.isGameAction(path, queryString)) {
       return false;
     }
 
-    if (location.startsWith("lchat.php")
-        || location.startsWith("mchat.php")
-        || location.startsWith("newchatmessages.php")
-        || location.startsWith("submitnewchat.php")) {
-      return false;
+    switch (path) {
+      case "login.php", "logout.php" -> {
+        return false;
+      }
+      case "api.php", "game.php", "doc.php" -> {
+        return false;
+      }
+      case "topmenu.php", "awesomemenu.php" -> {
+        return false;
+      }
+      case "actionbar.php" -> {
+        return false;
+      }
+      case "lchat.php", "mchat.php", "newchatmessages.php", "submitnewchat.php" -> {
+        // Nothing from chat.
+        return false;
+      }
+      case "desc_effect.php",
+          "desc_familiar.php",
+          "desc_guardian.php",
+          "desc_item.php",
+          "desc_monster.php",
+          "desc_outfit.php",
+          "desc_skill.php" -> {
+        // No object descriptions
+        return false;
+      }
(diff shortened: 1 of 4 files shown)


midgleyc:
Could you rebase this on main? It includes some of  #1833 as independent commits.

src/net/sourceforge/kolmafia/session/ResponseTextParser.java
```diff
@@ -146,44 +146,75 @@ public static boolean hasResult(final String location) {
...
+      }
+      case "lchat.php", "mchat.php", "newchatmessages.php", "submitnewchat.php" -> {
+        // Nothing from chat.
+        return false;
+      }
+      case "desc_effect.php",
+          "desc_familiar.php",
+          "desc_guardian.php",
+          "desc_item.php",
+          "desc_monster.php",
```
midgleyc, line 175: `desc_monster.php` doesn't seem to exist on KoL. There is a reference to it in Mafia, but I don't understand what it is supposed to do.

src/net/sourceforge/kolmafia/session/ResponseTextParser.java
```diff
@@ -146,44 +146,75 @@ public static boolean hasResult(final String location) {
...
+        // Other clan*.php URLs have no result.
+        // Which are those?
+        return true;
+      }
+      case "questlog.php" -> {
+        return false;
+      }
     }
 
-    if (location.endsWith("menu.php") || location.startsWith("actionbar")) {
```
midgleyc, line 170: these are potentially unmapped completely:
* ending with `menu.php`
* starting with `actionbar`
* starting with `desc`
* starting with `quest`

It's probably fine.

Veracity0: I added specific pages, complete with .php, in the switch statement.
So, rather then "*menu.php", we have "topmenu.php" and "awesomemenu.php".
Which is to say, this is intentional.

src/net/sourceforge/kolmafia/session/ResponseTextParser.java
Veracity0: Yeah, that is a synthetic page that we use in the Encyclopedia for the "Monsters" tab.
Every other tab will display the page returned by KoL itself.

It is trickery in the MonsterDescriptionFrame:

```
  private void refreshMonster() {
    MonsterData stats = null;

    // Kludge for a single monster...
    if (this.currentMonster.getName().equals("Ed the Undying")) {
      String name = "Ed the Undying (" + (variant + 1) + ")";
      stats = MonsterDatabase.findMonster(name);
    }

    String path = "desc_monster.php?whichmonster=" + this.currentMonster.getId();
    GenericRequest request = new GenericRequest(path);
    request.responseText = this.currentMonster.craftDescription(this.variant, stats);
    this.refresh(request);
  }
```

Notice that it makes a GenericRequest with a pseudo-url and sets the responseText to hand-crafted HTML.
It never actually runs that request. It just displays the responsetext.

I can remove desc_monster.php, since we never submit that and look for a response to parse.

Veracity0:
> Could you rebase this on main? It includes some of #1833 as independent commits.

I accidentally started this branch from main, but now yet updated.
As soon as I merged main into it, it fast forwarded past the extra commits.
The result is, although a couple of commits show up as included, after the merge, none of the code is actually in this PR.

I went through some steps to rebase on main on my computer, but it didn't do anything that I could push.
So, I'm not sure what you want me to do.

midgleyc:
I would expect a plain
```
git checkout response-text-parser-switches
git fetch
git rebase origin/main
```
would put the branch in a state such that `git push -f` could send it up here correctly.

If not, then
```
git fetch
git checkout -B response-text-parser-switches origin/main
git cherry-pick 14caae262546f4ddacbbc36ce347e2cf3bc8b823
git cherry-pick a0ff95ecd6b3887c844913d7a40b8756dc492d99
```
should put it in the right state (with the two commits you've made), and then you can force push.

midgleyc (approved):
It's probably fine because we're going to squash the commits anyway.

Veracity0:
I flailed with that and left my local branch in a mess of a state.
Yes, the squash merge will fix everything.

PR #1839, opened 2023-07-17
fix: source terminal enquiry is reset after LoL run

midgleyc:


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -3025,6 +3025,11 @@ public static final void liberateKing() {
       Preferences.resetToDefault("replicaNeverendingPartyAlways");
       Preferences.resetToDefault("replicaWitchessSetAvailable");
 
+      // we lose set enquiry, even if we have a terminal
+      Preferences.resetToDefault("sourceTerminalEducate1");
+      Preferences.resetToDefault("sourceTerminalEducate2");
+      Preferences.resetToDefault("sourceTerminalEnquiry");
```
jaadams5, line 3031: At least one script was failing because sourceTerminalEnhanceKnown was what they checked and it was not reset.  A character kept trying to EXTRACT (which wasn't a fatal error) but I ended up manually reverting every preference that had "sourceTerminal" in the name to the default before they stopped.

I don't really understand the mechanics but would it be appropriate to also reset
sourceTerminalEducateKnown
sourceTerminalEnhanceKnown
sourceTerminalEnquiryKnown
sourceTerminalExtrudeKnown
?

midgleyc: I had guessed not: those don't reset, but they only apply if you have a source terminal. I'd expect a script to check whether a source terminal is in the campground in addition to the preferences.

Before LoL they were only populated if you had a source terminal, I guess.

jaadams5: If I had a replica and then don't have the ITOM the settings are irrelevant and could (should) be reset especially if a script references them under the no longer valid assumption that if they are set the IOTM is available.

This raises two issues.  Are there preferences set by a replica in LoL that would be valid for the real ITOM after exit from LoL.  If that is true then eventually resetting ITOM preferences after LoL will need to be somewhat nuanced.  Similarly if requiring the same replica in a subsequent run effects the characteristics then that is a different case for preservation.

This could get some more thought but if sourceTerminalEnhanceKnown gets reset then Veracity won't have to deal with complaints or a request for a rewrite :-)

midgleyc: > If I had a replica and then don't have the ITOM the settings are irrelevant

Not quite! If you start LoL run 1, use a bunch of chips, then when you install the terminal again on LoL 2 your chips persist.

Now perhaps you could say that it's fine to drop it for aftercore, but perhaps someone wants to farm up just enough meat to buy the chips, and wants to know when to stop, or wants to confirm that they fully upgraded.

> Are there preferences set by a replica in LoL that would be valid for the real ITOM after exit from LoL
I suspect this is true: it's nuanced on the KoL side. You only lose the emotion chip skills if you get them from the replica emotion chip; otherwise you get them back in aftercore.

> if sourceTerminalEnhanceKnown gets reset then Veracity won't have to deal with complaints

I think the script should fix itself to check for presence of a terminal in the campground.

midgleyc:
I haven't changed these variables to fix a script, I've changed them because they're wrong: if you have a source terminal, set an enquiry in LoL with a replica, then free the king, the enquiry is no longer set and you have to set it again.

I think the same is probably true for educate but I haven't confirmed (because I change it in aftercore anyway, whereas enquiry is constant as "familiar" in both run / aftercore).

PR #1841, opened 2023-07-19
Check that there are no child nodes of CBB messages before ending the parsing of those nodes

gausie:
Status messages from the Cosmic Bowling Ball don't clean up afterthemselves nicely with a <p>, so our current processing eats the following line in fight.php

I've messaged CDM about it but this fixes the problem in case he does not. I'm interested to know if you think this is a good solution! There is a test case attached, feel free to fiddle with it.


gausie:
Oh maybe CDM has fixed this!

gausie:
Yep, this is now working. Thanks cdm

PR #1842, opened 2023-07-21
More permissive regex for ChoiceManager

s-k-z:
Changes the regular expression used by `stillInChoice` and `handlingChoice` to return true for choice adventures matching `href="choice.php...` instead of just `href=choice.php...` and `href="choice.php"`. In other words, tags of the form `href="choice.php?whichchoice=1135&option=2&pwd=12345"` won't be excluded.

Adds a test that fails on #1837 at [The Bat Sedan](https://kol.coldfront.net/thekolwiki/index.php/The_Bat-Sedan) which this PR resolves.


gausie (approved):
Thank you! This is from me misreading the code I was refactoring. I read

```java
responseText.contains("href=\"choice.php");
```
as
```java
responseText.contains("href=\"choice.php\"");
```

PR #1844, opened 2023-07-22
feat: minor git improvements

midgleyc:
Use URL to make identifier for non-GitHub / sourceforge projects (https://kolmafia.us/threads/git-projects-outside-of-github-are-not-named-correctly.29066)

Only return folders when looking for git folders (https://kolmafia.us/threads/error-on-running-git-update-and-on-mafia-start.29067)

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/ScriptManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/ScriptManager.java
@@ -29,16 +29,18 @@ public class ScriptManager {
       Pattern.compile("/p/(.*?)/(?:code|svn)(.*)", Pattern.DOTALL);
 
   public static String getProjectIdentifier(String host, String path) {
-    String UUID = null;
     Matcher m;
 
     if ((m = SOURCEFORGE_PATTERN.matcher(path)).find()) {
       // replace awful SVN UUID with nicely-formatted string derived from URL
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/scripts/ScriptManager.java
```diff
@@ -29,16 +29,18 @@ public class ScriptManager {
...
-    String UUID = null;
     Matcher m;
 
     if ((m = SOURCEFORGE_PATTERN.matcher(path)).find()) {
       // replace awful SVN UUID with nicely-formatted string derived from URL
-      UUID = StringUtilities.globalStringReplace(m.group(1) + m.group(2), "/", "-"); //
+      return StringUtilities.globalStringReplace(m.group(1) + m.group(2), "/", "-"); //
     } else if (host.contains("github")) {
-      UUID = StringUtilities.globalStringReplace(path.substring(1), "/", "-");
+      // user-repo
```
jaadams5, line 38: I don't understand what this comment at this place is trying to tell me.  I see the cases as SF/SVN, GitHub and anything else and the latter includes both a non-GitHub git repo and something we cant figure out and so fallback,

midgleyc: it says what the id is -- the GitHub user's username, then a dash, then the repo name.

PR #1845, opened 2023-07-22
pingLoginThreshold property. Calculate average pin in floating point.

Veracity0:
1) Login ping test configuration uses pingLoginThreshold, but when calculating it, we looked up pingLoginGood. Huh?
2) "average" and "bps" are calculated by the division of two ints. Do the calculation in floating point.
3) Similarly, threshold is a double, so use double value of average, not integer.
4) When presenting average and bps to user, use Math.round, rather than Math.floor.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/LoginManager.java
+++ b/src/net/sourceforge/kolmafia/session/LoginManager.java
@@ -66,7 +66,8 @@ public class LoginManager {
 
     // The user wants to measure ping speed.
     var result = PingManager.runPingTest();
-    KoLmafia.updateDisplay("Ping test: average delay is " + result.getAverage() + " msecs.");
+    KoLmafia.updateDisplay(
+        "Ping test: average delay is " + Math.round(result.getAverage()) + " msecs.");
 
     // See if the Ping tested a suitable page
     if (!result.isSaveable()) {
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/session/LoginManager.java
```diff
@@ -96,14 +97,14 @@ private static boolean ping(int attempt) {
         // The user wants to be "close" to the best ping seen.
         // Get the shortest ping test time we've seen. If the ping test we
         // just ran is the first, that will be it.
-        double threshold = 1.0 + Preferences.getFloat("pingLoginGood");
+        double threshold = 1.0 + Preferences.getDouble("pingLoginThreshold");
```
jaadams5, line 100: Change from pingLoginGood to pingLoginThreshold is deliberate, I presume.

jaadams5 (dismissed):

jaadams5:
I've seen RuntimeLibraryTest > Ping > parsesPropertyASH() fail a couple of times on Ubuntu on GitHub so maybe it is something to look at.

src/net/sourceforge/kolmafia/session/LoginManager.java
Veracity0: Yes. The GUI has always configured pingLoginThreshold and that is what is in defaults.txt.

Veracity0:
> I've seen RuntimeLibraryTest > Ping > parsesPropertyASH() fail a couple of times on Ubuntu on GitHub
Fixed. calculation of BPS now rounds, resulting in current result being 1 greater than previously.

PR #1847, opened 2023-07-23
Add ability to define the folder to install a git script to

libraryaddict:
Some discussion was done here: https://kolmafia.us/threads/git-projects-outside-of-github-are-not-named-correctly.29066/

This PR adds the ability to define a folder to install the git script to, which would let us install git projects from sites other than github. Such as gitlab, bitbucket, so on.
This also lets us do so without adding explicit support for each site, and while giving the user the ability to make a readable script folder name for both these sites, and for github.

The syntax is
`git checkout <Url> <Branch> <Folder>`
You can also use -, | and - in place of <Branch> to indicate that the branch name is not to be defined.
`git checkout <Url> - <Folder>`  `git checkout <Url> || <Folder>`  `git checkout <Url> _ <Folder>`

This was intended to resolve a previous issue where you couldn't use a third party site, but now just features allowing renaming the destination folder.


gausie:
I don't love this solution from a design perspective personally. It might have been good to get some consenus on approach before moving forward. Others might disagree though

libraryaddict:
> I don't love this solution from a design perspective personally. It might have been good to get some consenus on approach before moving forward. Others might disagree though

Interestingly enough, I hadn't realized that a change to make third party sites work had been implemented. Otherwise I wouldn't have bothered.
As it is, this PR now just adds the ability to change the destination folder. I'm not really concerned the final outcome.

libraryaddict:
Eh, its throwing a weird error when the git repo should've been cleaned up and I don't care enough.

PR #1849, opened 2023-07-23
Use Loathing Legion Universal Screwdriver for untinkering

gausie:
At the moment this just uses the screwdriver if you have one. It doesn't fold. This is probably already useful, so I'll add a test and try to get merged before expanding on it, unless others disagree.


gausie:
This is ready for review by the way

jaadams5 (approved):

PR #1850, opened 2023-07-23
abort ping test if redirected or no responseText

Veracity0:
1) If a ping redirects, do not follow the redirection; there is no point in repeatedly pinging choice.php, fight.php, login.php, or even afterlife.php, since we can't compare the results to any unsupported ping page.
2) If a ping fails in a ping test, log the reason - redirect or empty responseText (due to I/O error, we assume) - and abort the test. Do NOT save the (empty - or even partial) ping test in properties.
3) If ping test fails at login, give up searching for a better time. Do not log "this ping test took 0 msec". Instead, depend on the actual failure reason being logged. See 2.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/PingRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PingRequest.java
@@ -55,6 +55,13 @@ public class PingRequest extends GenericRequest {
     };
   }
 
+  private static boolean canFollowRedirect(String redirectLocation) {
+    // In Valhalla, we COULD do a ping test on afterlife.php, but since
+    // we can't compare results with any other page, it is pointless.
+    // afterlife.php?realworld=1
+    return false;
+  }
+
   @Override
   public void run() {
     // You can reuse this request.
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/request/PingRequest.java
```diff
@@ -55,6 +55,13 @@ private boolean isSafeToRun() {
     };
   }
 
+  private static boolean canFollowRedirect(String redirectLocation) {
```
jaadams5, line 58: I can come with reasons why I would have a well named function that currently returned a constant value but my decision to do so has been questioned here and elsewhere.  Do you anticipate a time when the function would be non constant?

jaadams5 (approved):

src/net/sourceforge/kolmafia/request/PingRequest.java
Veracity0: I initially followed the redirect, using the code I have there.

When I decided not to follow the redirect, I decided to keep the code, rather than having to rewrite it if I did come up with a redirect I wanted to follow.

I could remove the function and the currently unreachable code.

jaadams5: Your call.  If there is any chance a valid opportunity to redirect occurs then it will be easy to respond.  On the other hand the generation of devs who grew up with version control systems have pointed out that if that code is ever needed again it can be recovered from the repository.

Veracity0: I decided to get rid of the dead code.

jaadams5 (approved):

PR #1854, opened 2023-07-26
my_total_turns_spent() to return the sum of all $location[] turns_spent

Rinn:
Keeping track of the turns in a separate Integer in AdventureSpentDatabase to prevent the need to reducing the Map values every time the function is called

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/AdventureSpentDatabase.java
@@ -28,7 +28,9 @@ import net.sourceforge.kolmafia.request.FightRequest;
 public class AdventureSpentDatabase implements Serializable {
   @Serial private static final long serialVersionUID = -180241952508113933L;
   private static Map<String, Integer> TURNS = new TreeMap<>();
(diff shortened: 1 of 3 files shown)


horrible-little-slime:
This would be extremely useful to me! I do enough free fights that I can't really use `total_turns_spent` to determine if I've done something meaningful during my last turn (as opposed to, say, hitting a june cleaver NC or zone intro NC or whatever)

That being said, I think `my_delay_turncount` might be a little more natural a name. I'm mostly thinking about when it gets camelcased in js: `myDelayturncount` vs `myDelayTurncount`

Rinn:
ya i'll rename it when i get home

jaadams5 (approved):

midgleyc:
Naming-wise: how is `delay` related to this? It looks like it's more the total combats. If they were in a delayed zone, you'd be reducing delay, but they aren't necessarily.

gausie:
Agree with Ryo - the naming sounds odd.

Rinn:
I'm not attached to the function name i just didn't want it anyone to confuse it with my_turncount()

Rinn:
changed it to my_total_turns_spent

midgleyc (approved):
righto -- as the sum of `turns_spent` on all locations this is reasonably self-descriptive

PR #1856, opened 2023-07-27
set ownsFloristFriar to true if it is owned

Rinn:
This will allow scripts like [av-snapshot](https://github.com/m-e-meyer/av-snapshot) or [greenbox](https://github.com/loathers/greenbox) to report the florist friar as owned even if florist_available() returns false, such as a post community service run where the forest is inaccessible.

Also always set it as available if you can get into the choice instead of only on selecting choice 4

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1129,6 +1129,7 @@ user	optimisticCandleProgress	25
 user	oscusSodaUsed	false
 user	outrageousSombreroUsed	false
 user	overgrownLotAvailable	false
+user	ownsFloristFriar	false
 user	ownsSpeakeasy	false
 user	palindomeDudesDefeated	0
 user	parasolUsed	0
--- a/src/net/sourceforge/kolmafia/request/FloristRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FloristRequest.java
@@ -225,6 +225,10 @@ public class FloristRequest extends GenericRequest {
       return;
     }
 
+    if (responseText.contains("The Florist Friar's Cottage")) {
+      FloristRequest.setHaveFlorist(true);
+    }
+
     switch (FloristRequest.getOption(urlString)) {
       case 1 -> {
         int plant = FloristRequest.getPlant(urlString);
@@ -259,9 +263,6 @@ public class FloristRequest extends GenericRequest {
         return;
       }
       case 4 -> {
-        if (responseText.contains("The Florist Friar's Cottage")) {
-          FloristRequest.setHaveFlorist(true);
-        }
         FloristRequest.floristPlants.clear();
         Matcher matcher = FloristRequest.FLOWER_PATTERN.matcher(responseText);
         while (matcher.find()) {
--- a/src/net/sourceforge/kolmafia/request/FloristRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FloristRequest.java
@@ -208,6 +208,9 @@ public class FloristRequest extends GenericRequest {
   public static void setHaveFlorist(final boolean haveFlorist) {
     FloristRequest.floristChecked = true;
     FloristRequest.haveFlorist = haveFlorist;
+    if (haveFlorist) {
+      Preferences.setBoolean("ownsFloristFriar", true);
+    }
   }
 
   public static final List<Florist> getPlants(String location) {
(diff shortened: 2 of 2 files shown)


s-k-z:
Why not trust the property when using the commands?

Veracity0:
In Legacy of Loathing, offers a replica Order of the Green Thumb Order Form which grants in-run access to the Florist Friar.

Rinn:
arg fricking lol

Rinn:
> Why not trust the property when using the florist commands?

you can't get to it if the forest is unavailable

Rinn:
> > > Why not trust the property when using the florist commands?
> > 
> > 
> > you can't get to it if the forest is unavailable, or in standard
> 
> Isn't that fine?

I don't understand? This preference won't ever be set to false in code once it's set to true. It's to denote if you own the iotm, not if the friars is currently available.

s-k-z:
> > > > Why not trust the property when using the florist commands?
> > > 
> > > 
> > > you can't get to it if the forest is unavailable, or in standard
> > 
> > 
> > Isn't that fine?
> 
> I don't understand? This preference won't ever be set to false in code once it's set to true. It's to denote if you own the iotm, not if the friars is currently available.

Last I saw, `florist_available()` only returns true when you launch KoLmafia with it accessible. If this property is just cosmetic rather than functional, it won't behave like the others, where toggling it on would enable attempts to use the florist friars at runtime.

Rinn:
> > > > > Why not trust the property when using the florist commands?
> > > > 
> > > > 
> > > > you can't get to it if the forest is unavailable, or in standard
> > > 
> > > 
> > > Isn't that fine?
> > 
> > 
> > I don't understand? This preference won't ever be set to false in code once it's set to true. It's to denote if you own the iotm, not if the friars is currently available.
> 
> Last I saw, `florist_available()` only returns true when you launch KoLmafia with it accessible. If this property is just cosmetic rather than functional, it won't behave like the others, where toggling it on would enable attempts to use the florist friars at runtime.

I intentionally didn't name this property to contain "Available", or "Always", or "Accessable" for this very reason.

s-k-z:
> > > > > > Why not trust the property when using the florist commands?
> > > > > 
> > > > > 
> > > > > you can't get to it if the forest is unavailable, or in standard
> > > > 
> > > > 
> > > > Isn't that fine?
> > > 
> > > 
> > > I don't understand? This preference won't ever be set to false in code once it's set to true. It's to denote if you own the iotm, not if the friars is currently available.
> > 
> > 
> > Last I saw, `florist_available()` only returns true when you launch KoLmafia with it accessible. If this property is just cosmetic rather than functional, it won't behave like the others, where toggling it on would enable attempts to use the florist friars at runtime.
> 
> I intentionally didn't name this property to contain "Available", or "Always", or "Accessable" for this very reason.

Here is an example: https://github.com/kolmafia/kolmafia/blob/3877c6f1c291f9f49685a0f88e6c9c4c04b8c6bf/src/net/sourceforge/kolmafia/request/FancyDanRequest.java#L95

Rinn:
honestly i don't want to deal with this

gausie:
> > > > > > > Why not trust the property when using the florist commands?
> > > > > > 
> > > > > > 
> > > > > > you can't get to it if the forest is unavailable, or in standard
> > > > > 
> > > > > 
> > > > > Isn't that fine?
> > > > 
> > > > 
> > > > I don't understand? This preference won't ever be set to false in code once it's set to true. It's to denote if you own the iotm, not if the friars is currently available.
> > > 
> > > 
> > > Last I saw, `florist_available()` only returns true when you launch KoLmafia with it accessible. If this property is just cosmetic rather than functional, it won't behave like the others, where toggling it on would enable attempts to use the florist friars at runtime.
> > 
> > 
> > I intentionally didn't name this property to contain "Available", or "Always", or "Accessable" for this very reason.
> 
> Here is an example:
> 
> https://github.com/kolmafia/kolmafia/blob/3877c6f1c291f9f49685a0f88e6c9c4c04b8c6bf/src/net/sourceforge/kolmafia/request/FancyDanRequest.java#L94-L100

That is a bad example. ownsSpeakeasy is one requirement to it being accessible. Others need to be added and will be in the future when the item is no longer in standard or is otherwise routinely unavailable.

gausie (approved):

jaadams5:
coolitems in the API has nothing to say about this?

gausie:
Sadly not

PR #1857, opened 2023-07-27
Expose florist state variables as preferences

s-k-z:
Enables the ability to re-check the florist friars without having to relaunch KoLmafia.

Followup to #1856


gausie:
Thank you for this. Do you think it makes sense to merge in with #1858, or does it cover everything you want to cover.

s-k-z:
> Thank you for this. Do you think it makes sense to merge in with #1858, or does it cover everything you want to cover.

I think that one is more comprehensive. If any of the test cases I've provided are useful then that might be worth merging in.

PR #1858, opened 2023-07-28
Forest Friar preferences

Rinn:
- Don't check if the florist is available if the woods is not available via quest state, rather than character level
- Use properties floristFriarAvailable & floristFriarChecked to store florist available state rather than static booleans
- Reset state and recheck availability when (replica) Order of the Green Thumb Order Form is used
- Set owned if Order of the Green Thumb Order Form is used
- When florist_available() is called check page if unchecked this ascension
- Tests

```
> set floristFriarChecked = false
> set floristFriarAvailable = false
> set ownsFloristFriar = false

> ash florist_available()

Preference floristFriarChecked changed from false to true
Preference floristFriarAvailable changed from false to true
Preference ownsFloristFriar changed from false to true
Returned: true

> ash florist_available()

Returned: true
```


s-k-z:
Thank you for taking the time to do all this!

Rinn:
The one thing I'm not a fan of is the reset() changes, floristFriarChecked is set to false on login unnecessarily. Really the only time it would need to be set to false again is when breaking the prism i think, but this doesn't regress anything from existing functionality as is.

midgleyc (changes requested):
In `KoLCharacter`, `liberateKing` should, in `Reset Legacy of Loathing stuff`, reset both the checked nature and the availability, and then you should check again. You can go both from not having it to having it (the non-replica) and having it (the replica) to not having it, so I think this needs to be unguarded.

edit: oh yeah, and going standard -> non-standard should do the same. Maybe it already does.

midgleyc (approved):
`liberateKing` already checks the forest again, my bad.

PR #1859, opened 2023-07-28
items display message only if used from GUI

Veracity0:
Items with the "message" type display an amusing message when used and are not consumed.

1) Allow "message" as a secondary attribute.
2) A "drink with a fly in it", N. O. Beer, Otarian Battle Scar are all drinks that are undrinkable. Make them drink, message
3) gloomy black mushroom and oily golden mushroom are foods that are inedible. Make them food, message
4) Add showHTML field of a UseItemRequest to control whether we'll pop up a documentation frame to display KoL's funny message. This defaults to true.
5) The "use" command (and its synonyms: eat, drink, chew) - as well as functions in the RuntimeLibrary which are implemented using the CLI command - get a UseItemRequest (or subclasses) and specify that showHTML is false.

The end result is that if you "use" an item from the Item Manager, you see the amusing message, but if you "use" an item from a script, you do not.


jaadams5 (approved):

Veracity0:
> What is the expected behavior when an item is used from the relay browser? I would expect it to be the same as using it from the (KoLmafia) GUI but figured I would ask.

We only show the HTML if you use the item with a UseItemRequest. The Relay Browser submits RelayRequests. Therefore, no window is popped up. Which is, I hope, expected, since if you use the item in the Relay Browser, you will see KoL's response in the Browser.

(Yes, ResponseTextParser will call UseItemRequest.parseConsumption - but it explicitly passes "false" for showHTML.)

PR #1866, opened 2023-07-31
Test Fixes

jaadams5:
In the process of trying to sort out https://github.com/kolmafia/kolmafia/pull/1811 I ran locally with one fork for gradle.  In the past that has given a deterministic test order.  Two tests failed.  I was able to reproduce the failures with 4 forks and when I ran the tests individually  in IntelliJ.  These changes fix that.  

For the one test the test class structure appeared to prevent the beforeEach from running before the test.  That hypothesis was supported by un-initialized preference values in the test according to the debugger.

For the other test the Familiar data was just nonexistent.

This code works individually in my IDE and when running with four forks.

### DIFF
--- a/test/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanelTest.java
+++ b/test/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanelTest.java
@@ -278,9 +278,14 @@ public class DailyDeedsPanelTest {
 
   @Nested
   class DropsFamiliars {
+
+    @BeforeEach
+    public void beforeEach() {
+      FamiliarData.reset();
+    }
+
     @Test
     public void showsCookbookbat() {
-      FamiliarData.reset();
       var dd = new DailyDeedsPanel.DropsDaily();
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanelTest.java
```diff
@@ -279,6 +280,7 @@ public void showsLocket3Monsters() {
   class DropsFamiliars {
     @Test
     public void showsCookbookbat() {
+      FamiliarData.reset();
```
Veracity0, line 283: Put this into a BeforeEach method for this Nested class?

Veracity0 (changes requested):

test/net/sourceforge/kolmafia/swingui/panel/DailyDeedsPanelTest.java
jaadams5: Did not realize I could do that.  Done.  Thank you.

Veracity0 (approved):

PR #1867, opened 2023-08-01
feat: august skeleton items etc

midgleyc:
For now, just dump everything.

Needs a preference for skills cast (5 + 1 for the current day), fixodent "Craft with Teeth" store (takes fixodent + loose teeth, so probably like Sliemce?)


midgleyc:
Tests are failing because "toilet paper" now also matches "handful of toilet paper". The tests were written assuming an exact match would match the specific item, but the function doesn't work that way.

jaadams5 (approved):
In the interests of getting this out there....

PR #1872, opened 2023-08-02
Abort ping test if pings are too long

Veracity0:
Shiverwarp complained recently that he was getting login ping test times of 23000+ msec and it was taking seemingly forever to login. I sympathize; I usually get <30 msec and the longest I've see in a 10-ping test is a bit more than 800 - which would still be an intolerably slow connection. And to think it took 10 pings to get that "average".

I can't imagine what caused Shiverwarp's insanely bad connection, but if it took 4 minutes to run the first test and decide it is time to retry, seems to me that we could recognize way earlier - as early as the very first ping - that this connection is a no-go.

1) Comparing ping average times with goals and thresholds in floating point seems right, but displaying them as integers, calculated either by Math.round() or Math.floor(), is problematic. If your threshold calculation says you cannot exceed 30.2 msec and you generate 30.4, it will fail, and using Math.round, will report "you wanted 30 msec but got 30 msec, so it failed".  I changed it to display floating point results in floating point. So far, it always seems to display with a single digit to the right of the decimal point, which seems fine.

2) If you get a really slow connection, it will be obvious early on that it will be too slow.
Allow configuration for "PingAbortTrigger" triggers.

If "pingTestAbort" has "4:3|1:10", you are specifying:

Retry if 4 pings exceed 3 times average historical ping
Retry if 1 ping exceeds 10 times average historical ping

Default is "1:10"

On the PingOptionsPanel, you can add rows to specify additional abort triggers or delete rows to remove them. If you try to delete the last row, it just zeroes it - which will leave the property empty.

I think the GUI is pretty spiffy.

PingManager now checks those triggers after each ping it submits. If that results in any trigger firing - 4 exceeding 3 times historical average or 1 exceeding 10 times historical average, from the above example - PingManager aborts the ping test and returns the test with the trigger that fired to LoginManager.

Could be that the aborted ping test was still good enough, but if not, it reports that the ping test aborted and it moves on to the next timein attempt.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -101,6 +101,7 @@ global	pingDefaultTestPage	api
 global	pingDefaultTestPings	10
 global	pingLatest
 global	pingLogin	true
+global	pingLoginAbort	1:10
 global	pingLoginCheck	none
 global	pingLoginCount	0
 global	pingLoginFail	confirm
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -101,7 +101,7 @@ global	pingDefaultTestPage	api
 global	pingDefaultTestPings	10
 global	pingLatest
 global	pingLogin	true
-global	pingLoginAbort	
+global	pingLoginAbort	1:10
 global	pingLoginCheck	none
 global	pingLoginCount	0
 global	pingLoginFail	confirm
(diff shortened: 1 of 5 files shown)


jaadams5 (approved):
Eyes a little glazed but...

Veracity0:
> Eyes a little glazed but...

Did you try it out? I'll merge it and you can see for yourself.

PR #1880, opened 2023-08-06
RenameSuggestion

AnonymousAccount4SE:
### Renaming Suggestion of Method Names to Make Them More Descriptive
---
**Description**
---

We have developed a tool (**NameSpotter**) for identifying non-descriptive method names, and using this tool we find a non-descriptive method name in your repository:

```java
/* Non-descriptive Method Name in src/net/sourceforge/kolmafia/KoLCharacter.java*/
  public static final void incrementFamilarWeight() {
    if (KoLCharacter.currentFamiliar != null) {
      KoLCharacter.currentFamiliar.setWeight(KoLCharacter.currentFamiliar.getWeight() + 1);
    }
  }
```

We consider "incrementFamilarWeight" as non-descriptive because it contains a typo, i.e., "Familar" should be "Familiar". 

We have corrected them (including all the occurrences in other file) and submitted a pull request.

Do you agree with the judgment? 

- If not, could you please leave your valuable opinion?

- If you do agree, the relevant modification has been submitted as a PR, and thanks for your precious time to consider it.


BadHorseMonkey:
They're not wrong.  They don't make a test case, and the method is called once in the codebase.  Low value, if someone wants to do some cleanup.

midgleyc:
It's fair enough, but I wouldn't oversell it: at the moment, it's a spell checker, not a "descriptiveness improvement". Spell-checking is valuable, but be honest about what you're doing ;)

We've almost surely got some test methods with typos in, too: did NameSpotter seriously only find one method?

AnonymousAccount4SE:
> 

Thanks for your valuable suggestions. NameSpotter absolutely does not only find one method, it outputs a list of method names that may be descriptive. However, this tool is not perfect right now, which means that this list of method names could contain some false positives. Consequently, we manually identify these method names and pick out one most convincing method name for this pull request (we think typos will also impact the comprehensibility of method names and lead to misunderstanding). Sorry for the non-descriptive illustration.

midgleyc:
Obsoleted by #1888.

Thanks for your contribution! This prompted heeheehee to run over the codebase for more spelling issues and raise a PR fixing several at once (including yours). So this is no longer necessary, but thanks for doing it!

AnonymousAccount4SE:
Thanks and have a good day!

PR #1885, opened 2023-08-07
feat: casting the today august skill no longer counts as 1/5

midgleyc:
After yesterday, casting today's skill no longer has to be done last -- if you do it first, you still get another five.


jaadams5 (approved):

Veracity0 (approved):
I immediately tagged that as a bug...

PR #1886, opened 2023-08-07
Revert "Change synchronization of Preferences.reset() to address conc…

midgleyc:
…urrency issue between timein() and Preferences.saveToFile() (#1811)"

This reverts commit 05734436a251e9312783dcb5c9e39344807a9be4.

This is breaking tests on MacOS with IllegalArgumentExceptions (e.g. https://github.com/kolmafia/kolmafia/actions/runs/5785813007/job/15679193578?pr=1880) and also freezing Swing on timeins / Mickey Mantle draws.


jaadams5:
I will leave the decision to revert to someone else but...

Mantle Cheating has worked for several days with the code locally.  While there is certainly a problem it may be related to the presence of the Deck as a Daily Deed.  Perhaps a less radical step of making sure the Deck is not executing on the Swing thread might be worth investigating.

The MacOs failures have been occurring intermittently since that PR was first posted.  Similar failures have also occurred on ubuntu.  In every case no reason was was determined and rerunning on GitHub did not reproduce the failure.  The hypothesis was that something in the test effected the GitHub runtime environment.  That failure has not been observed in any build environment besides GitHub if you are getting it locally on a Mac and can reproduce then your help in debugging would be very much appreciated.  Otherwise I would lump this in with several tests, Dvorak being one I recall, that routinely fail intermittently, but not repeatable, on various OSes in various environments.

s-k-z:
I think that #1811 is causing very long pauses (freezing kolmafia UI and relay browser) when using the Deck of Every Card `cheat` command for users. Based on second-hand information this just started occurring though I haven't observed it myself.

jaadams5:
https://kolmafia.us/threads/1952-mickey-mantle-pick-in-deck-of-every-card-crash.29126/

The action listener for Daily Deeds calls Preference,getInteger().

Still pondering if cli cheat command can interact with the DD code via some other listener.

midgleyc:
> Otherwise I would lump this in with several tests, Dvorak being one I recall, that routinely fail intermittently, but not repeatable, on various OSes in various environments.

Dvorak seems to fail 1/20 times, I can live with that. This one seems to fail maybe 1/3, I find it much more aggravating when trying to get PRs through. Perhaps it was just a coincidence.

I don't have a Mac so I can't reproduce this locally: it just seems like a run of IllegalArgumentException.

jaadams5:
No one admits to seeing the IllegalArgument etc. in any environment but GitHub.  It may be annoying but the fix should be to change GitHub's environment, not revert a useful test.  IMO.

BadHorseMonkey:
one of the reasons this took a month to get out was that it was failing on GitHub, where threading is not the same as on desktops.

I do have a Mac, and didn't see this, but I may not have the item in question.  What would a failing test look like so we can work on it?

BadHorseMonkey:
I'm inclined to actually revert it, since it's breaking the game for players.

jaadams5:
@heeheehee-kolmafia is advocating revert so I'll agree.

But I don't want to work on a the reverted version.  How do I do that?  I can start from the PR but if I merge from main will it undo.  I don't know how to cherry pick commits with GitHub Desktop.

heeheehee-kolmafia (approved):

heeheehee-kolmafia:
> @heeheehee-kolmafia is advocating revert so I'll agree.
> 
> But I don't want to work on a the reverted version. How do I do that? I can start from the PR but if I merge from main will it undo. I don't know how to cherry pick commits with GitHub Desktop.

Can you check out a previous commit? If not, I can make a new branch pointing out the current HEAD that you can work off of.

heeheehee-kolmafia:
I'm going to merge this and then set up a branch we can work on with a fix-forward.

PR #1887, opened 2023-08-07
Forbidden (ignoring, disabled) stores no longer affect mall_price()

Veracity0:
1) Mall Search requests from the GUI now get saved in MallPriceManager's saved mall searches.
2) Mall Search requests (from whatever source) no longer filter out forbidden, disabled, and ignoring stores.
3) When you ask for a mall_price(), forbidden, disabled, and ignoring stores are filtered out of the save results when calculating the (5th item's) price.
4) showForbiddenStores and showIgnoringStorePrices will hide or show the results from a mall search in the GUI, as before.
forbidden stores are red. Ignoring and Disabled stores are grey, since you can't even try to purchase from them.
5) If you change showForbiddenStores, showIgnoringStorePrices, or forbiddenStores properties, the GUI will update immediately, and your next mall_price() - whether or not it has to make a new mall search - will reflect the new "correct" mall price.
6) Since forbidden stores appear in mall search results and their visibility can be toggled - and if visible, they are colored red - checking whether a store is forbidden happens a LOT in Swing widgets. getForbiddenStores used to blindly read the property and parse it into a Set every single time it was called. I put control of that property under the control of a Manager object, which listens for changes to the property - and writes it, when changed. It also does lazy loading of the property: only the first time the forbidden stores list is requested.

All the GUI stuff works great; looking at the MallSearchFrame, going to the gCLI to change properties, going back to the MalSearchFrame, and colors change and rows are visible or not, based on the "show" properties. That's all done with Listeners and a ListDataFilter. I extracted SearchResultPanel out of MallSearchFrame, which make it testable.
Tests for visibility, color, and purchasability when adding forbidden, ignoring, and disabled stores.

mall_price() returns the "correct" mall price, whether or not the GUI or MallPriceManager.searchMall did the mall search.
I have tests that verify that the results from the mall search are appropriately filtered in both cases.

mall_price() filters out the unavailable stores as expected.
I have a test for getting prices with disabled, ignoring, and forbidden stores in the results list.


test/net/sourceforge/kolmafia/session/ForbiddenManagerTest.java
```diff
@@ -0,0 +1,120 @@
...
+import net.sourceforge.kolmafia.request.MallPurchaseRequest;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class ForbiddenManagerTest {
+  @BeforeAll
+  public static void beforeAll() {
+    KoLCharacter.reset("ForbiddenManager");
```
jaadams5, line 20: Similar comment as before, minus the reference to spaces.

jaadams5 (approved):

test/net/sourceforge/kolmafia/session/ForbiddenManagerTest.java
Veracity0: Same response, minus the reference to spaces. :)

PR #1889, opened 2023-08-08
Fix-forward "Change synchronization of Preferences.reset() to addr…"

heeheehee-kolmafia:
…ess concurrency issue between timein() and Preferences.saveToFile() (#1811)" (#1886)"

This reverts commit dfcf477b87b048c1aad344e4c5fffed722232915.

First step for a fix-forward.


heeheehee-kolmafia:
As discussed in-thread: I could also see an issue with thread A waiting on thread B, which is waiting on Preferences.lock held by thread A.

I don't see a mechanism by which a thread holding Preferences.lock could spawn a sub-thread, but it's a possibility.

heeheehee-kolmafia:
A more explicit check that we're not deadlocking any threads:
```
import java.lang.management.ManagementFactory;
...
      var threads = ManagementFactory.getThreadMXBean().findDeadlockedThreads();
      assertEquals(threads, null);
```
although, really, you'll notice that the test actually completes...

jaadams5:
Thinking out loud - a ConcurrentMap should manage consistency and it looks like we are already doing that although it's not 100% clear to me.  We might want to rethink saveSettingsOnSet.  I am not certain it provides a useful service any more.  If KoLmafia crashes then you want it to recover as much as you can but how much is KoLmafia crashing these days?

I wonder if saveSettingsOnSet should be ignored any time there is preference file I/O going on?  I also wonder about putting save on its own thread with a timer and it writes if there have been changes?

jaadams5:
Lint suggested synchronizing on a non local object so I figured I'd try that.  There are some file cleanups I am looking at while I try and figure out how to break this or change something that feels like a fix.

heeheehee-kolmafia:
Hm. I broke something in my latest commit such that it now deadlocks on login. Looking into it...

heeheehee-kolmafia:
I initially used invokeLater in ListenerRegistry, but that caused a bunch of test failures (even if that was fine when running the jar. Pondering how to resolve that conflict...

jaadams5:
Windows, GitHub had
MoodManagerTest > itShouldHaveNoTriggersWhenThereAreNotAnyDefined() FAILED
    org.opentest4j.AssertionFailedError at MoodManagerTest.java:139
    
   as failure.  It is probably a leakage issue since the test basically resets everything and confirms nothing is there.
   
Tests passed on my Windows machine.  Going to build from the PR and see if I can burn turns.

jaadams5:
https://github.com/kolmafia/kolmafia/actions/runs/5895179238/job/15995387554?pr=1889

Concurrent Modification exception on MacOS.  Might be interesting.

jaadams5:
I changed Gradle's forks parameter because I believe most of unrepeatable test failures on GitHub are due to concurrent execution and cleanups that get executed and remove files another test had just created.  This will increase time but reduce failures, I hope.  Locally forks=16 caused a failure which might not mean anything.  We may want to back this out, but for now I'd like to see if it makes a difference.

Also I build and run from this daily and have not seen any deadlocks.  I wish that were not the case :-)

jaadams5:
Fork tweak didn't work so backed it out.

jaadams5:
I have been running this daily since the branch was restored as a draft.  I have seen no problems.   I am at a loss as to what to do to try and trigger an issue in the Real World.  I have tried letting sessions sit idle and retimed in.  For various reasons I cannot easily leave something running for long periods or running at Rollover.  Does anyone have ideas for me or can they set up some of the long running or rollover related scenarios?

My observation continues to be that the failing tests all involve files on disk that are created and removed as part of the test.  It may be that our cleanup is to aggressive in that it deletes files and directories.  If I don't get distracted by the command buffer as preference I will look harder at that since I'm pretty sure I added the code to delete the directories many many months ago,

gausie:
I feel like at some point this needs to either be merged or closed such that work can continue in a branch that doesn't have an active PR

jaadams5:
> I feel like at some point this needs to either be merged or closed such that work can continue in a branch that doesn't have an active PR

1) I have been running with this since it was created and have seen no instances of the file issues it was supposed to solve nor instances of the behavior that caused it to be reverted.

2) I have attempted to reproduce the problems it allegedly caused but have been unable to do so.

3) I would recommend continuing the discussion at KoLmafia.us because there are stakeholders who will comment there but not here.

jaadams5:
There seems to be little concern at https://kolmafia.us/threads/shall-it-be-closed.29441/

There are a couple of things to clean up - mostly prints - and I will do so and then get approval to push this.

jaadams5 (dismissed):
Since there were no concerns over at KoLMafia.us about pushing this, I have approved.  Would someone provide a "second opinion" and push it please?  @heeheehee-kolmafia ?  @gausie ?

gausie:
I would merge it but it keeps failing tests

jaadams5:
The failing tests are not necessarily an indication of a problem with the code.  The vast majority of them are due to a file that is assumed to be present for the test having been deleted by another test as part of cleanup that removes directories.  The remaining tests that fail had had intermittent, unrepeatable failures on other branches.  Basically none of the failing tests are repeatable in that they fail consistently on all platforms.  Rerunning makes the test pass.  Indeed I just reran and now everything is good :-)

jaadams5:
Got called to eat.  If the goal is no failing tests then I would look at the test cleanup code that deletes the contents of a directory and then the directory.  That will stop the failures where a test creates a preferences file which is deleted by the cleanup of another test.  The remaining case includes the Dvorak test and that has failed on other branches but never consistently enough to track down.

If you want we can wait until I have had a chance to tweak some cleanup code and see whether that is an improvement.

gausie:
But why does that happen here and not in main?

jaadams5:
A hypothesis is that ClearSharedStateBefore and ClearSharedStateAfter are deleting several directories and their contents after some test has set up and started and so the test fails when something external to the test has deleted a needed file.  This behavior is dependent upon the test runner and so the number of tests run and the order in which they start and complete causes the failure, or not.

There are two things that could be done that might "prove" the hypothesis.  One is to force the test runner to use one thread so everything is executed sequentially.  Long before this branch, that was done to prove that some intermittent test failures we caused by the test runner and not the code being tested.  The other thing to do would be to not delete directories and contents as part of the setup and cleanup.  If unrelated tests stop breaking then that supports the hypothesis that is the tests and not the code.

We can certainly pause this until get tried.

jaadams5:
Well.  One fork.  Failed tests.

https://github.com/kolmafia/kolmafia/actions/runs/7295115663/job/19881040440?pr=1889

I see a Concurrent Modification Exception in PreferenceWatcherTableTest and a failure at AbsorbCommandTest.java:131.  That does not seem top have anything to do with files.

I'm going to revert the fork and ponder the tests.

jaadams5:
And note that the first run after reverting had no failing tests.

jaadams5:
With my new code I got some local failures that I had seen as intermittent here so progress is being made,

jaadams5:
I'll make this draft again.  I'd like other eyeballs but don't want to waste people's time.

In answer to the question why fail here but not in main I don't have a testable hypothesis.  But the new test here seems to be the only test that actually tries to create a session file and save it to disk.  The contents suggest some test issues that could be addressed.

I have also noticed that many of the tests that fail involve setting a preference as part of the setup and then using an object that knows the state of the preference because a preference listener fired.  It is certainly possible to imagine something time sensitive and the listener has not finished firing before the related object is checked.

jaadams5:
See https://kolmafia.us/threads/shall-it-be-closed.29441/ where I have posted a debug log and a jstack dump from when "this code" froze while dealing with a preference associated with drawing a Micky Mantle Card.

jaadams5:
I had committed code used to explore test failures.  I decided it was worth keeping but was cluttering up this PR so I moved it.

Veracity0:
I need to review this. I had my first ever pref wipe after a timein a werk ago. Fortunately, I write prefs after every set, and have automatic system backups every 25 minutes, so I manually restored everything.

Interesting how many obsolete/removed prefs I had. 😀

I thought heeheehee and you were going to work this out and release it.

jaadams5:
> I need to review this. I had my first ever pref wipe after a timein a week ago. Fortunately, I write prefs after every set, and have automatic system backups every 15 minutes, so I manually restored everything.
> 
> Interesting how many obsolete/removed prefs I had. 😀
> 
> I thought heeheehee and you were going to work this out and release it.

I never got to the point where I had a possible fix and I stopped trying when @heeheehee-kolmafia said they had an idea.  I then went down the maximizer not equipping weapon rabbit hole.  Since I could pretty reliably get this code to freeze when VMF used the Deck I figured it is not ready.  Yours is the first report I have heard of wiped preferences in several months which also kept it on my back burner.

jaadams5:
P.S. - I am convinced the test failures are due to timing or sequencing but I would love to understand why the new code causes a failure that other code does not.  Probably should add that to the list.

jaadams5:
A side issue with this PR has been the intermittent test failures.  The question has been asked whether something in the deadlock test is causing problems.  The failures that have been investigated in detail can be attributed to timing.  One example is test A writes a file into a directory so the test can run and test B cleans up after itself by deleting a directory and contents.  When A starts before B finishes there is a problem.  https://github.com/kolmafia/kolmafia/pull/2281 was pushed to this PR and and https://github.com/kolmafia/kolmafia/pull/2197 and the tests for both PRs failed with the same custom script test.  That reinforces my belief that the test failures are not the direct result of new code in the current PR.  Which, in turn, means I would merge this once the deadlock issue was demonstrably solved.

jaadams5:
@heeheehee-kolmafia 

Any ideas?  You seem to be much close to having a solution than I have been able to get.  Thanks.

jaadams5:
I'm working on the tests and marking as draft until I finish or run out of ideas.

heeheehee-kolmafia:
FYI -- I've manually squashed the commits in this PR and rebased onto main because 322 commits was a few too many.

Apologies for the loss of history, and thank you @jaadams5 for giving this PR a lot more TLC than I have had time for. I'll try to convince myself to find time to look at this with some fresh eyes in the coming weeks.

jaadams5:
"squash" is new to my vocabulary but a lot of the changes were fits and starts and several cases where code was changed and then deleted without actually reverting,   So the history might not be of much value.

The primary issue with this code is that when compiled and run it will deadlock.  Drawing from the Deck of Every Card triggers that somewhat reliably.

The other issue is the failing tests.  I am convinced that most, if not all of them, occur because tests run concurrently and cleanup for one test deletes assets required by or set up for another.  That seems to occur more often with this PR than others but I do not believe that these changes are directly responsible.

jaadams5:
Having said all of that https://github.com/kolmafia/kolmafia/actions/runs/10700958985/job/29666068680?pr=1889 shows a concurrent modification exception on PreferenceWatcherTableTest > Rendering > rendersPreferenceTable() which is not easily explained by concurrent setup and cleanup.

jaadams5:
I've been running with this in hopes of getting some progress.  It deadlocked, again on the Deck of Every Cards.  jstack says

Found one Java-level deadlock:
=============================
"AWT-EventQueue-0":
  waiting to lock monitor 0x000001a31d82d750 (object 0x0000000083f9e478, a java.lang.Object),
  which is held by "CommandQueueHandler"

"CommandQueueHandler":
  waiting to lock monitor 0x000001a31bd1e960 (object 0x0000000083a37a60, a java.awt.Component$AWTTreeLock),
  which is held by "AWT-EventQueue-0"
  
  There is a KoLmafia generated debug log as well.
  
  
[DEBUG_20240926.txt](https://github.com/user-attachments/files/17149019/DEBUG_20240926.txt)

jaadams5:
I may have botched resolving a merge conflict and fixed it.  If that is not true then please revert what I did, fix it and patiently explain to me what I did wrong so I don't do it again.  Thank you.

gausie:
Could this be obviated by https://github.com/kolmafia/kolmafia/pull/2808?

jaadams5:
> Could this be obviated by #2808?

No.

See https://github.com/kolmafia/kolmafia/pull/2813.

Basically I extracted the test from here, merged it with something that had #2808 and watched the test fail consistently.  My conclusion is that the test checks something that is broken after #2808 but fixed here.

The reason this has not been committed is that there were several reports of deadlocks, one of which was often caused by using the Deck of Every Card.

I am in the process of trying to write a test that uses the Deck of Every Cards.  My hope is that the test will pass when run against main but fail when merged here.  If that happens and code changes here make the test pass then I would feel comfortable about  merging this.

I hope to have the Deck test out in a couple of days - real life interferes with writing code - and if things work as I hope there will be something more concrete to look at to discover and fix the deadlock.  If you are available to collaborate let me know or just join me when the Deck test is posted.

jaadams5:
Closed in error.

jaadams5:
My Holy Grail was a test that used the Deck of Every Cards and failed with this patch.  I have one such test and it is not failing.  I am building and running this version locally but if I can't a failure in a non-test environment then I am willing to close this.  Maybe target April 1 as a decision date if there is no new information or new ideas?

jaadams5:
> Maybe target April 1 as a decision date if there is no new information or new ideas?

See https://kolmafia.us/threads/commit-or-close-pr-1889.30515/

PR #1893, opened 2023-08-09
feat: expose daily limit pref for skills with one

midgleyc:
Hopefully cut down on the number of queries as to which preference tracks skill X by exposing it on the skill proxy object.


jaadams5 (approved):
Are there daily limits for other "uses" besides CASTS that would be of interest?

midgleyc:
USE might well be.

PR #1898, opened 2023-08-15
Log when we open and close a session

libraryaddict:
https://kolmafia.us/threads/log-when-a-client-session-starts-and-ends.29143/

The biggest usecase of this for me currently, is to figure out when mafia exited abnormally and try to recover the player state from that, ignoring the remainder of the file. Such as recovering daily preferences.

Its also a bit weird that there's nothing definite indicating when a session started or ended.


jaadams5 (approved):
When designing and processing logs was part of my job description I became convinced of the utility of timestamping every log entry.  That doesn't make a lot of sense in KoLmafia but it made me wonder if "Starting session for USERNAME at GAMETIME" was a more useful message (etc.).  It might be if I was using this message to diagnose a problem that might be associated with rollover.  Not a blocking comment but I won't push this for a while in case anyone agrees with me :-)

midgleyc:
This is for https://github.com/s-k-z/kolfix? Seems reasonable.

PR #1907, opened 2023-08-19
Add to_buffer(string) and use it when initializing a buffer

Veracity0:
I've wanted to be able to do this for a long time:

```
buffer b = "Initial contents";
```

Now I can.

I also moved creation of the (buffer) return value of visit_url() to the top of the method, rather than duplicating the creation in two places.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/Parser.java
+++ b/src/net/sourceforge/kolmafia/textui/Parser.java
@@ -1003,7 +1003,8 @@ public class Parser {
             functionName.content, functionType, variableReferences, functionLocation);
 
     if (f.overridesLibraryFunction()) {
-      functionErrors.submitError(this.overridesLibraryFunctionError(f));
+      String buffer = "Function '" + f.getSignature() + "' overrides a library function.";
+      this.warning(f.getLocation(), buffer);
     }
 
     UserDefinedFunction existing = parentScope.findFunction(f);
@@ -5534,11 +5535,6 @@ public class Parser {
 
   private AshDiagnostic multiplyDefinedFunctionError(final Function f) {
     String buffer = "Function '" + f.getSignature() + "' defined multiple times.";
-    return this.error(f.getLocation(), buffer);
-  }
-
-  private AshDiagnostic overridesLibraryFunctionError(final Function f) {
-    String buffer = "Function '" + f.getSignature() + "' overrides a library function.";
     return this.error(f.getLocation(), buffer);
   }
 
--- a/src/net/sourceforge/kolmafia/textui/Parser.java
+++ b/src/net/sourceforge/kolmafia/textui/Parser.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia.textui;
 
+import static net.sourceforge.kolmafia.textui.DataTypes.BUFFER_TYPE;
 import static net.sourceforge.kolmafia.textui.DataTypes.PATH_TYPE;
 import static net.sourceforge.kolmafia.textui.parsetree.AggregateType.badAggregateType;
 import static net.sourceforge.kolmafia.textui.parsetree.VariableReference.badVariableReference;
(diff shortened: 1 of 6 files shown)


jaadams5 (approved):

midgleyc (dismissed):
I think this is good to have in main, but I think there are a fair few scripts (e.g. Ezandora's) which define

```
buffer to_buffer(string str)
{
  buffer result;
  result.append(str);
  return result;
}
```

for the convenience, and which will now conflict and error. I think `s.replace_string("", "")` is also popular.

Veracity0:
Yeah. Ezandora's scripts concern me.
I made an announcement about this, since there will be much lamentation.
Perhaps it will inspire her to upgrade her scripts. :)

On the other hand, we could make the compile error into a warning. If the script is not depending on the RuntimeLibrary version, no big deal if it redefines it - even to something not equivalent.

Veracity0:
And, oh - for replace_string, we don't want a string to auto-coerce into a buffer.
That should be fixable; I can specify that this coercion only works on ASSIGNMENT.

BadHorseMonkey:
> I think this is good to have in main, but I think there are a fair few scripts (e.g. Ezandora's) which define
 ...

Also, your Bastille script.  And TourGuide.  And those are just the ones I have installed.

Veracity0:
Interesting.
```
    if (f.overridesLibraryFunction()) {
      String buffer = "Function '" + f.getSignature() + "' overrides a library function.";
      this.warning(f.getLocation(), buffer);
    }
```
works nicely if I declare the library function like this:
```
    params = new Type[] {DataTypes.STRING_TYPE};
    functions.add(new LibraryFunction("to_buffer", DataTypes.BUFFER_TYPE, params));
```
but not if I declare it like this:
```
    params = new Type[] {DataTypes.STRICT_STRING_TYPE};
    functions.add(new LibraryFunction("to_buffer", DataTypes.BUFFER_TYPE, params));
```
There is no way for the user to declare a function whose argument is a STRICT_STRING_TYPE

Veracity0:
I made this script:
```
// The following should generate a warning
buffer to_buffer(string initial)
{
    print("initial = '" + initial + "'");
    buffer buf;
    buf.append(initial);
    return buf;
}

buffer buf1 = to_buffer("Initial content 1");
print("buf1 content = '" + buf1.to_string() + "'");

buffer buf2 = "Initial content 2";
print("buf2 content = '" + buf2.to_string() + "'");

buffer buf3 = $path[Heavy Rains];
print("buf3 content = '" + buf3.to_string() + "'");
```
which yields:
```
Function 'to_buffer(string)' overrides a library function. (to_buffer_test.ash, line 2, char 8 to char 33)
initial = 'Initial content 1'
buf1 content = 'Initial content 1'
initial = 'Initial content 2'
buf2 content = 'Initial content 2'
initial = 'Heavy Rains'
buf3 content = 'Heavy Rains'
```
The good news is that the user defined function really does override the library function.
The bad news is that I didn't REALLY want the $path[] to get coerced into a string.
Which is to say, I'd prefer STRICT_STRING_TYPE.

I don't see an immediate clean solution to the STRING_TYPE != STRICT_STRING_TYPE issue.
Function.paramsMatch:
```
    while (it1.hasNext() && it2.hasNext()) {
      VariableReference val1 = it1.next();
      VariableReference val2 = it2.next();
      Type paramType1 = base ? val1.getType() : val1.getRawType();
      Type paramType2 = base ? val2.getType() : val2.getRawType();
      if (!paramType1.equals(paramType2)) {
        return false;
      }
```
It compares Type using equals(). No special casing to pretend STRING_TYPE is the same as STRICT_STRING_TYPE.
Until/unless we rethink how STRICT_STRING_TYPE is implemented, I have two options:

1) to_buffer(string) is declared with a STRING_TYPE argument and we get useful override warnings
2) to_buffer(string) is declared with a STRICT_STRING_TYPE argument and we get no warning.

I'm inclined to do #1, for now.

midgleyc (approved):

PR #1912, opened 2023-08-21
Property _beachTides is tide-washed row count

Veracity0:
1) When you comb a tile on the beach, we supposedly change that code in _beachLayout to a 'c'.
Except, it was checking for "You acquire", but bottles and whales do not "acquire" anything.
Look instead for "You comb"

2) When we examine a beach map, the tides wash the rows at the bottom and you cannot comb them.
Property "_beachTides" now has how many such rows are unavailable. ("_beachLayout" will have rows from N+1 - 10).

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1715,6 +1715,7 @@ user	_beachCombing	false
 user	_beachHeadsUsed
 user	_beachLayout
 user	_beachMinutes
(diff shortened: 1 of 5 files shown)


jaadams5 (dismissed):

Veracity0:
Previously approved with failing test.
The test no longer fails.

PR #1917, opened 2023-08-26
feat: text from the council in the small path

midgleyc:
Took it from the wiki. Possibly wrong in the specifics but better than what we have now, which is nothing.


Veracity0:
I wonder how many double spaces after periods there are. :)

Veracity0 (approved):

PR #1918, opened 2023-08-26
fix: parse monster names correctly in Small

midgleyc:
Detect, but don't do anything with, the "shrunk" modifier.


Veracity0 (approved):
I'll try this path some day. Thanks for getting it ready! 😉

PR #1923, opened 2023-08-30
Add basic support for the Flea Market

libraryaddict:
Original issue was opened here: https://kolmafia.us/threads/items-placed-in-flea-market-are-not-tracked-as-removed-from-inventory.29181/

This PR just aims to correctly track when an item is added to the Flea Market and remove it from our inventory, as well as log when we buy or sell.

As a note, I'm against the idea of adding more support to the Flea Market such as a buy or sell command. Or a "whats for sale" search.

The flea market already sucks, and I don't want to see it turn into a "lets undercut each other" v2 mall. It sucks enough already without having people buy out underpriced items automatically, and undercutting each other on overpriced items automatically.

If you want to do that stuff, make your own script. It's a noob market with prices detached from reality, and we should respect that by leaving it to the noobs.

I saw someone automatically undercut everyone on the flea market so that they were the only seller, if you placed new items for sale they automatically added an item that undercut you.. It was worse than in mall, at least in the mall I can ignore your store to buy the next cheapest. They did eventually stop, but still.


midgleyc (approved):
LGTM

PR #1924, opened 2023-08-30
function type

Veracity0:
1) A FunctionType is declared like this:
```function int (int, int)```
2) You can ```typedef``` it.
```typedef function int (int, int) binop;```
3) You can declare variables
```
function int (int, int) op1;
binop op2;
```
4) You can use existing functions of the appropriate type to initialize variables
```
int plus(int x, int y) {
    return x + y;
}

int times(int x, int y) {
    return x * y;
}

binop op3 = plus;
binop op4 = op3;
```
5) You can use them in maps and map literals
```
binop[string] operation = {
    "+" : plus,
    "-" : times
};
```

There is quite a bit left to do...


Veracity0:
I don't think I have time to work more on this.

PR #1925, opened 2023-08-30
Test Suite for Function.paramsMatch

Veracity0:
Function has a method - paramsMatch - which tests whether a list of user-supplied parameters can be given to this function.

It can be an EXACT match - typedefs match specifically that typedef, not the base type
It can be a BASE match - typedefs resolve to base type
It can be a COERCE match - we is willing to coerce the supplied type to the required type.

As coded, it's a method of a Function object.

For my Function Value work, I want to check parameters against something which has a FunctionType but is not a Function. A variable, say. In that work, I refactored paramsMatch to be a static method which takes a list of param Types, rather than deriving them from the VariableReferences in the function. In doing that, I had to make a few intrusive changes to the method. I need a test suite to make sure I didn't break anything.

Here's the test suite. It passes with the current implementation of paramsMatch.
My other PR will see whether it still passes once this PR is merged. :)


test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
```diff
@@ -113,4 +121,578 @@ public static Stream<ScriptData> data() {
...
+        // Tests for functions with valid varargs.
+
+        // vararg allowed, no matching args provided
+        @Test
+        void int_int0() {
+          Function f =
+              makeFunction(
+                  "f", DataTypes.VOID_TYPE, DataTypes.STRING_TYPE, vararg(DataTypes.INT_TYPE));
+          makeValues(DataTypes.STRING_TYPE);
+          // *** There is a varag, but we ran out of values before hitting it.
```
midgleyc, line 321: `varag` -> `vararg`, and same for every time it occurs.

test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
```diff
@@ -113,4 +121,578 @@ public static Stream<ScriptData> data() {
...
+          // *** We accept that. Is that correct?
```
midgleyc, line 322: You're in the best place to decide, but I think this feels correct.

test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
```diff
@@ -113,4 +121,578 @@ public static Stream<ScriptData> data() {
...
+    //
+    // This is used (only) by UserDefineFunction.overridesLibraryFunction
+    // *** This is not tested here (yet)
+
+    // public boolean paramsMatch(final List<? extends TypedNode> params, MatchType match, boolean
+    // vararg)
+    //
+    // Check if a supplied list of values match this function's parameters.
+    //
+    // MatchType,EXACT  - the value and the function parameter must be identical
```
midgleyc, line 139: `MatchType.EXACT` and same for the others (comma to dot)

midgleyc (changes requested):

test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
Veracity0: Upon thinking about it more, this is correct.

If we were comparing parameter types for two functions (as we do in a different method), yes - all the parameters must be equal (modulo base type) and there must be the same number of them.

However, we are checking whether this function can accept this set of parameters - and if the last parameter of the function is a vararg and you don't supply any arguments to be collected by that, that is fine; your function will be called with a size-0 array or arguments.

So, yes - supplying all the required arguments and no vararg arguments means this function will accept them.

test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
Veracity0: I have corrected these comments, based on the next comment, but the mis-spelling is gone.

test/net/sourceforge/kolmafia/textui/parsetree/FunctionTest.java
Veracity0: Fixed.

Veracity0:
I addressed the code review comments.

I also added some tests for passing in typedef'd array/map arguments to be used for the varargs parameter, since that is exactly what multiple arguments get turned into when being passed to a vararg function.

I observed some oddness - called out by *** comments:

- EXACT matching accepts either such a map/array OR a typedef of such.
- BASE matching accepts NEITHER such an array NOR a typedef of such.
- COERCE matching also accepts neither - because aggregates cannot be coerced?

In practice, those will not be an issue in the general case of Scope looking for a function, because that looks for MatchType.ALL - which means any match will do. Maybe a BASE match should pass and an EXACT match should not, but ALL will be happy.

Which is to say, this is probably a signal that I've detected a bug, but this PR might not be the place to fix it.
I'll look into fixing the code AND the test in a subsequent PR.

midgleyc (dismissed):

Veracity0:
An actual bug:

A vararg looks to a function as an array of the specified data type.
You can call the function with a variable number of that arg type - and get an array of that type - and it works
You can call the function directly with an array of that data type - and it works.
You can call the function directly with a map of that data type - and it works.

You can call the function with a variable number of typedefs of that arg type - and get an array of those typedefs - and it works.
You can call the function with an array of that typedef - and it gets a compile error.

```
typedef int altint;
typedef altint[] altarray;
typedef altint[int] altmap;

int intfunc(int... args)
{
    int sum = 0;
    foreach n, arg in args {
	sum += arg;
    }
    return sum;
}

print(intfunc(1, 2, 3, 4));
// 10

print(intfunc(int[] {1, 3, 5, 7}));
// 16

print(intfunc(int[int] {0: 10, 1: 20, 3: 30, 4: 40}));
// 100

altint a1 = 2;
altint a2 = 4;
altint a3 = 6;
altint a4 = 8;

print(intfunc(a1, a2, a3, a4));
// 20

altarray aa = {2, 4, 6, 8};
// print(intfunc(aa));
// Function 'intfunc( altint [0] )' undefined. This script may require a more recent version of KoLmafia and/or its supporting scripts. (typedef-int-array.ash, line 29, char 7 to char 14)
```

yields

```
> typedef-int-array

10
16
100
20
```

The bug is because BASE matching does not work with arrays/maps being passed in directly.
I added tests for this case.
Sigh.

midgleyc (approved):
I'm not sure about having tests that encode the wrong behaviour, but so long as you know that you're going to go back and fix it I guess it's okay.

I'd rather have the right behaviour and `@Disabled("behaviour is currently wrong")` or something like that.

Veracity0:
Yeah. ASH does not recognize that aggregates of a typedef are (perhaps) usable as aggregates of the base class.
I'll need to investigate this.
```
typedef int altint;
typedef altint[] altarray;

int sum(int[] args)
{
    int val = 0;
    foreach n, arg in args {
	print("arg[" + n + "] = " + arg);
	val += arg;
    }
    return val;
}

int[] array1 = {1, 2, 3, 4};
foreach n, x in array1 {
    print("array1[" + n + "] = " + x);
}

print("sum = " + sum(array1));

altint[] array2 = {2, 4, 6, 8};
foreach n, x in array2 {
    print("array2[" + n + "] = " + x);
}

// print("sum = " + sum(array2));

altarray array3 = {10, 20, 30, 40};
foreach n, x in array3 {
    print("array3[" + n + "] = " + x);
}

// print("sum = " + sum(array3));
```
yields
```
> int-typedef

array1[0] = 1
array1[1] = 2
array1[2] = 3
array1[3] = 4
arg[0] = 1
arg[1] = 2
arg[2] = 3
arg[3] = 4
sum = 10
array2[0] = 2
array2[1] = 4
array2[2] = 6
array2[3] = 8
array3[0] = 10
array3[1] = 20
array3[2] = 30
array3[3] = 40
```
If I uncomment the disabled code:
```
> int-typedef

Function 'sum( altint [0] )' undefined. This script may require a more recent version of KoLmafia and/or its supporting scripts. (int-typedef.ash, line 26, char 18 to char 21)
Function 'sum( altint [0] )' undefined. This script may require a more recent version of KoLmafia and/or its supporting scripts. (int-typedef.ash, line 33, char 18 to char 21)
```

I will move all the tests that demonstrate bad behavior to a new Nested - and Disabled - class.
We can test "passing in a aggregate directly to supply varargs" later, after I look at that issue, as well.

midgleyc (approved):

PR #1926, opened 2023-08-31
Report when gnoob absorb fails, and actually create the item to absorb

libraryaddict:
Currently if gnoob fails to absorb an item, it will still report the item as absorbed. This just checks the absorb count and reports a failure if the count doesn't match.

It also doesn't correctly create the item to absorb if the ingredients are available.. This removes a method call which I believe is the mistake.


src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
```diff
@@ -48,14 +48,14 @@ public void run(final String cmd, String parameters) {
 
     // If not in inventory, try to retrieve it (if it's in inventory, doesn't matter if outside
     // Standard)
-    if (!InventoryManager.hasItem(match, true)
-        && !InventoryManager.retrieveItem(match)
-        && match.getCount(KoLConstants.inventory) == 0) {
+    if (!InventoryManager.retrieveItem(match) && match.getCount(KoLConstants.inventory) == 0) {
```
midgleyc, line 51:
```suggestion
    if (match.getCount(KoLConstants.inventory) == 0 && !InventoryManager.retrieveItem(match)) {
```
if the item is in the inventory already, no need to attempt to retrieve it. I suspect this is what the original author thought "`hasItem`" did.

src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
```diff
@@ -68,6 +68,19 @@ public void run(final String cmd, String parameters) {
       PreferenceListenerRegistry.firePreferenceChanged("(hats)");
     }
 
-    KoLmafia.updateDisplay("Absorbed " + (count > 1 ? count + " " : "") + match.getPluralName());
+    int failedAbsorbs = count - (KoLCharacter.getAbsorbs() - previousAbsorbs);
+    count -= failedAbsorbs;
+
+    if (count > 0) {
+      KoLmafia.updateDisplay("Absorbed " + (count > 1 ? count + " " : "") + match.getPluralName());
```
midgleyc, line 75: the way `match.getPluralName` works is that if `match`'s count is 1, it's the single, otherwise the plural. Now that you're adjusting `count` this won't work correctly any more (same for `failedAbsorbs`).

Instead, use `ItemDatabase.getPluralName` and `match.getName`, and swap them in as appropriate.

src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
libraryaddict: Ah wait, shouldn't that be `match.getCount(KoLConstants.inventory) < count` instead?

src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
libraryaddict: But an int isn't passed by reference. It hasn't modified `match` because `count` is an `int`, not something like an `AtomicInteger` or something where we pass the reference around.

libraryaddict: Oh wait, sorry. I just realized

src/net/sourceforge/kolmafia/textui/command/AbsorbCommand.java
libraryaddict: Seems so, commited the change.

midgleyc: good spot

midgleyc (approved):

PR #1927, opened 2023-09-01
Track how many bags/keys were opened during NEP food/booze quest

libraryaddict:
In the recent NEP change (a month ago) to nerf duffo yet again, van keys & unremarkable duffel bags will no longer give their quest related item after 11 has been opened.

May as well track it.


jaadams5:
Do tests that use one item when the other quest is active and demonstrates the value doesn't change make sense?

libraryaddict:
> Do tests that use one item when the other quest is active and demonstrates the value doesn't change make sense?

I hope the parametrized test is acceptable?

jaadams5 (approved):

PR #1930, opened 2023-09-01
Fix the default value for monsterHabitatsFightsLeft

pstalcup:


Veracity0:
Whatever.

Properties are strings.
You can fetch them as  a string, integer, float, or boolean.
If the string value is "", you get 0, 0.0, or false, respectively.

So,yes - the default should be 0, but can you explain, exactly, how this was user visible?

And can gausie please explain why he merged this without making a review?

Thanks.

gausie:
Oh I just wasn't thinking, sorry. I meant to approve and then merge.

It's useful to have the correct default value because there is a tool for TypeScript that infers the type of properties by the default value. So having this zero will tell that tool that scripters should expect this pref to be numeric

gausie:
(I wrote that tool)

Veracity0:
Ok,that is fair, I guess.

Is there a runtime library function to get the default value of a property?

Is there a data integrity test that looks at defaults.txt and ensures that a property name us either followed by a TAB or NL? And that property names do not have any embedded spaces, which is a sign that whoever put in a default did it wrong?

(A common mistake for first contributors- or even experienced ones.)

pstalcup:
```
> ashref default
```
string [property_default_value](https://wiki.kolmafia.us/index.php?title=property_default_value)( string )
boolean [property_has_default](https://wiki.kolmafia.us/index.php?title=property_has_default)( string )

the reason I was interested in adding this fix was as gausie said the aforementioned typescript library which uses the default values to determine the typing.

jaadams5:
> Ok,that is fair, I guess.
> 
> Is there a runtime library function to get the default value of a property?
> 
> Is there a data integrity test that looks at defaults.txt and ensures that every property name is either followed by a TAB or NL? And that property names do not have any embedded spaces, which is a sign that whoever put in a default did it wrong?
> 
> (A common mistake for first contributors - or even experienced ones.)

The comment in DataFileMechanicsTest is that defaults.txt is too complex and thus not included.  That test is old enough that revisiting it or making a test specifically for defaults would be worthwhile.

PR #1931, opened 2023-09-01
Check defaults.txt for structure

jaadams5:
https://github.com/kolmafia/kolmafia/pull/1930 mentioned that defaults.txt could be tested for some basic errors, usually whitespace.  An existing test was tweaked to do that and in the process a whitespace error was detected and fixed.


src/data/defaults.txt
```diff
@@ -573,8 +573,8 @@ user	dinseyTouristsFed	0
 user	dinseyToxicMultiplier	3
 user	dnaSyringe
 user	doctorBagQuestItem	
-user	doctorBagQuestLights	0	
-user	doctorBagQuestLocation	
+user    doctorBagQuestLights    0
```
midgleyc, line 576: you have, likely accidentally, converted tabs to spaces on this line. If the test passes with this, it shouldn't?

jaadams5: I could swear there was a tab on my system before I committed but I have made that mistake before and it looks like it happened again.

The test just counts fields and fields are delimited by tabs according to the KoLmafia code doing the reading.  So if a line contains one tab it could pass even though the intention was a line with two fields.  So the test in some sense has been flawed for a few years.  Maybe that is why defaults.txt was declared too complex?

My inclination is to leave things as corrected because this test has caught white space issues as it is.  I do think it would be beneficial to have a test for defaults that is somewhat more specific than counting fields.  I will write such a test and add it.  

If you look at the data stream for the parameterized test it looks like any test where the second and third parameter are not the same the test could pass in the presence of bad data.  Perhaps those files should be removed from the data stream?  That's a discussion point but I probably should add some comments to the test.

Thanks.

Veracity0: defaults.txt has 2 or 3 fields per line.

1) scope - global or user
2) name - alphanumeric with perhaps a leading underscore
3) default value

If you split the line on tab characters, there should be 2 or 3 fields.

The first must be global or user
The second must not contain spaces. If it does, you used spaces, not tabs.
The third can be any string, not including tabs or newlines.

midgleyc: right, I think I was thinking of `DataFileTest`, which does let you do regex matching. I've added one.

One thing to think about is -- should we allow defaults which are empty? Currently we do. The behaviour is the same either way (default present but empty or default absent => empty string).

midgleyc (approved):

src/data/defaults.txt
jaadams5: @midgleyc Are you going to do the more nuanced test or shall I?

jaadams5:
Never mind previous comment.

PR #1938, opened 2023-09-04
fix: umbrella bonuses are doubled by offhand remarkable

midgleyc:
As reported in https://kolmafia.us/threads/august-scepter.29095/#post-173259.


gausie:
Not sure if I'm reading right but can we expand this to check for all offhand modeables or something?

gausie:
> I don't think there are any other ones at the moment?
> 
> Also I don't think we'll know whether new modeables will be doubled until they come out.

The way that modeables are implemented in KoL leads me to believe that it makes more sense to assume that they will be supported.

gausie (approved):

PR #1939, opened 2023-09-04
Check circadianAdv instead of resolutionAdv when adding circadian rhythms to rollover adventures

s-k-z:
Otherwise, you won't see circadian rhythms unless you have used be more adventurous as well.


midgleyc (approved):
Whoops.

PR #1941, opened 2023-09-05
Encounters until yachtzee choice

Shiverwarp:
RE: Forum post [here](https://kolmafia.us/threads/encountersuntilyachtzeechoice-preference.29211/): 

This is just what I have so far, I'm not totally sure if it's everything, first time I've done something like this.

I've heard I need to make a test, but am unsure of how to even start that. I mostly just fumbled my way through this copying how the NEP choice is done. Trying to work my way through that now.


Veracity0:
Saw your forum post. On my phone,so can't look at code, but there is an array of String preference name in Preferences.java of preferences to reset on Ascension.

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -2377,4 +2377,17 @@ public void doesNotLogCircadianFailures() {
...
+
+  @Nested
+  class Yachtzee {
+    @Test
+    void canTrackPartyYachtCombats() {
+      var cleanups = new Cleanups(withLastLocation("Sunken Party Yacht"));
+
+      try (cleanups) {
+        parseCombatData("request/test_party_yacht_fight.html");
+        assertThat("encountersUntilYachtzeeChoice", isSetTo("19"));
```
horrible-little-slime, line 2389:
```suggestion
        assertThat("encountersUntilYachtzeeChoice", isSetTo(19));
```

test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
```diff
@@ -366,6 +366,28 @@ void changingPlaqueRecordsNewName() {
...
   }
 
+  @Nested
+  class Yachtzee {
+    @Test
+    void visitingTheNCResetsPreference() {
+      var cleanups = new Cleanups(withProperty("encountersUntilYachtzeeChoice ", 0));
+      try (cleanups) {
+        var req = new GenericRequest("adventure.php?snarfblat=404");
+        req.responseText = html("request/test_visit_yachtzee.html");
```
horrible-little-slime, line 376: You need to actually process the request here using the choice manager.

Tentatively, it looks like the code you added is in `postChoice2`

test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
```diff
@@ -366,6 +366,28 @@ void changingPlaqueRecordsNewName() {
...
+        assertThat("encountersUntilYachtzeeChoice", isSetTo(20));
+      }
+    }
+
+    @Test
+    void choosingAnNCOptionDecrementsPreference() {
+      var cleanups =
+          new Cleanups(withProperty("encountersUntilYachtzeeChoice ", 20), withPostChoice2(918, 2));
```
horrible-little-slime, line 384: This feels like a counterintuitive logical flow. You're setting the `encountersUntilYachtzeeChoice` and processing the post-choice stuff (which presumably also modifies that preference) in the same `Cleanups`

look at how `changingPlaqueRecordsNewName`, right above this, works. That might help with both of these tests

test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
horrible-little-slime: well, in that case you'll need to call `visitChoice` somehow

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -2377,4 +2377,17 @@ public void doesNotLogCircadianFailures() {
       }
     }
   }
+
+  @Nested
+  class Yachtzee {
+    @Test
+    void canTrackPartyYachtCombats() {
+      var cleanups = new Cleanups(withLastLocation("Sunken Party Yacht"));
```
midgleyc, line 2385: you should add `withProperty("encountersUntilYachtzeeChoice ", 20)` to the cleanups so that the property gets reset after the test.

I don't think it actually matters here but it's a good habit to get into :) plus it makes the 20 -> 19 thing explicit

test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
Veracity0: When GenericRequest gets a response from choice.php (whether or not you are visiting the first time or are submitting an option to take the choice, it goes through this sequence of calls in ChoiceManager:

preChoice (before registering the request - which is before submitting the request)
postChoice0 (after getting the response before registering the Encounter)
postChoice1 (before processing results)
(if you are not taking an option, that calls visitChoice)
postChoice2 (after processing results)
(If you are still handling a (chain of) choices, this calls visitChoice)

You don't need to "need to call visitChoice"; that happens automatically.
By all means - set the  property to 20 in visitChoice.
It is pointless to set the property to 19 in BOTH postChoice1 and postChoice2.

It's not obvious why some things have been put in postChoice1 and some in postChoice2.
Some of it is because we want to log something and have it appear before - or after - the logging of items you acquire.
But we update quest progress in either method, remove items from inventory in either, update properties in either, and so on.

Pick one for setting the property to 19.

gausie (approved):
I think I'm happy with this now, but defer to those who have already requested changes.

gausie:
@jaadams5 it would have been good to hold off merging until Veracity or Chris had the chance to review, as I said?

jaadams5:
> @jaadams5 it would have been good to hold off merging until Veracity or Chris had the chance to review, as I said?

I reviewed their comments, did not see anything that I interpreted as meaning they wanted changes or that they intended to review again and noted that they had had 24 hours to respond to your observation.

It is quite possible that I missed some comment somewhere but since there were no workflow defined Requested Changes nor any unresolved conversations I figured people were done.  If you had retagged them or somehow indicated your review was contingent upon them I would not have pushed it.   Perhaps next time you will do more to make it obvious to others that  you don't want something pushed (don't approve and leave a change request would be things that jump out at me in GitHubDesktop) and I will wait longer before I push a non-Dev PR.  OK?

gausie:
I couldn't have been clearer in my approval comment

midgleyc:
FWIW @gausie I interpreted your comment exactly the same way jaadams5 did: "this is okay to merge, but check that the existing review comments have been fixed first".

If you wanted me or V to review it separately, I would have expected you to not approve it -- make your approval + comment merely a comment.

PR #1954, opened 2023-09-11
There's only one dev server, so don't bother checking the ping

gausie:


Veracity0 (approved):
Life is hard, if you are on the dev team.

PR #1955, opened 2023-09-12
feat: add modifiers for hare, gibberer and borrowed time

midgleyc:
https://github.com/kolmafia/kolmafia/pull/1933 added modifiers for some of the things in the "extraRolloverAdventures" preference, but not all of them. Complete the set.

Potato alarm clock and Straight-Edgy should probably go in a "Post-Cap Adventures" modifier, as these are applied after the 200 cap.


gausie (commented):
This is worth an announcement, potentially with a release date, since people's functions to work out rollover adventures will be double counting these.

Also this ties in nicely with my upcoming plan to add "how many adventures will i wake up with tomorrow" readout in the sidepane that only shows when the player is overdrunk.

midgleyc:
I think, with this, the only standard rollover thing we won't be tracking is "is Labor day tomorrow". It's different to the other holidays because it applies the next day. I can't see a ModifierType for holidays but I figure there must be one.

I can make an announcement after we sort a release date, or we can merge and then I'll announce for visibility. Unfortunately, scripts are already affected as of #1933: when I added circadian I also added resolutions, and resolutions were tracked in `extraRolloverAdventures` but I didn't track circadian that way, so resolutions were double-counted. I got a report from TourGuide that lead to this fix (or at least this standardisation). I didn't know about `extraRolloverAdventures` at the time I made the initial PR (or, rather, I thought it was a purely legacy property).

gausie:
Ah if it's already broken maybe we merge first and announce after.

And yeah, day before Labor Day is the one with the modifier... weird. Do you want to look at it or shall I?

I think holiday effects are pre Modifier, check KoLCharacter.setHoliday

midgleyc:
I've made a change to not increase `extraRolloverAdventures`, so scripts using numeric_modifier("Adventures") +
extraRolloverAdventures to calculate post-rollover adventures will continue to work. I've not removed the existing property. I thought we had an existing framework for resetting / deleting properties, but I couldn't find it, and `obsoleteProperties` exists on Preferences but isn't used.

PR #1956, opened 2023-09-12
feat: monster habitats now resets on rollover

midgleyc:
as specified in https://kolmafia.us/threads/recall-facts-monster-habitats-now-resets-on-rollover.29235/

#1950 already dealt with the ascension reset.


gausie:
I think now is the best time to rename the pref instead of include it in
legacy dailies. Iotm is less than two weeks old and any scripts using it
will still be actively developed.

Also what is noncombatforceractive doing in there?

Veracity0 (approved):
I was wondering if you were going to change the property into an "_" property.

The benefit is that is how it really behaves.

The detriment is that the old property still clutters everybody's prefs file,

We don't have a mechanism for renaming or deleting properties such that that doesn't happen.

That'd be a nice thing to do. In another PR.

gausie:
> I was wondering if you were going to change the property into an "_" property.
> 
> The benefit is that is how it really behaves.
> 
> The detriment is that the old property still clutters everybody's prefs file,
> 
> We don't have a mechanism for renaming or deleting properties such that that doesn't happen.
> 
> That'd be a nice thing to do. In another PR.

Yeah I was thinking about this recently - we could migrate over all the legacy dailies too and just rewire the property accessors to silently insert underscores for backwards compatibility.

PR #1961, opened 2023-09-14
Derive organ capacity from modifiers

gausie:
Started with Fullness... it seems to be working! I'll do inebriety and spleen next! Let me know if you spot anything that needs testing or is just obviously broken.

- Move to modifiers for Fullness
- Rename Cow Puncher vars to reflect the fact that it has a space
- Add class shortcut for AWoL classes

Questions:
- ~~I am trying to make the modifier `Base Stomach Capacity: 0` meaningful. Typically our modifier system takes zero as "remove/ignore this modifier" but that doesn't work here. Do I keep the current workarounds or do I add a boolean `No Stomach Capacity` mod instead?~~
- ~~We're in a funny place with `recalculateAdjustments` in test code, it's being run over and over again.~~


midgleyc:
> Do I keep the current workarounds or do I add a boolean No Stomach Capacity mod instead?

The latter IMO. Magic values aren't great.

gausie:
> I think I made `OrderedRunnable` for this -- remove it from the individual methods and add an `OrderedRunnable` that runs it at the end of the initial setup, and again at the end of the cleanup.

Oh that's smart. Hmm ok. But it still doesn't solve the issue of having to add `withAdjustmentsRecalculated` all the time. Maybe that's ok.

gausie:
> > Do I keep the current workarounds or do I add a boolean No Stomach Capacity mod instead?
> 
> The latter IMO. Magic values aren't great.

Ok this is currently working(?) without that change. As I start to make it there are just a bunch of similar changes that would need to be here instead. Out of interest, what are you classing as a magic value here?

> Also for drunk I think there is at least one class who can't drink but can get drunk from bang potions. There's Gelatinous Noob but that one also has a limit of -1.

I think that this won't be a problem. We'll see.

> I think I made `OrderedRunnable` for this -- remove it from the individual methods and add an `OrderedRunnable` that runs it at the end of the initial setup, and again at the end of the cleanup.

This turned out to not be that useful. I ended up just putting things that I know recalculateAdjustments at the end.

midgleyc:
> Ok this is currently working(?) without that change. As I start to make it there are just a bunch of similar changes that would need to be here instead. Out of interest, what are you classing as a magic value here?

I misunderstood the issue, but why does Base Stomach Capacity have to exist with a zero value, instead of just not existing and functions that interrogate it returning 0 if it doesn't exist? I guess I think modifiers is a bad fit for this, because surely everything has exactly one "Base Stomach Capacity". So why not have a function that looks this up based on path / class and only use modifiers for the extras?

midgleyc:
I feel like Limit is a bit different, because you can have multiple limits and the lowest applies (e.g. you're in Small, with Cowrruption, with a cap gun equipped), but you'll only ever have one Base X Capacity.

gausie:
> I feel like Limit is a bit different, because you can have multiple limits and the lowest applies (e.g. you're in Small, with Cowrruption, with a cap gun equipped), but you'll only ever have one Base X Capacity.
> 
> I think this should be a function instead of a modifier.

You are right, this is better.

src/data/modifiers.txt
```diff
@@ -8095,7 +8095,7 @@ Skill	Extra Brain	Mysticality Percent: +50
 Skill	Extra Innings	Adventures: +5
 # Extra Kidney: Booze gives 100% more Adventures
 Skill	Extra Muscles	Muscle Percent: +50
-# Extra Spleen: +5 spleen capacity
+Skill	Extra Spleen	Spleen Capacity: +5s
```
midgleyc, line 8098:
```suggestion
Skill	Extra Spleen	Spleen Capacity: +5
```

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -899,6 +916,97 @@ public final void applyAutumnatonModifiers() {
...
+          DoubleModifier.STOMACH_CAPACITY,
+          pantsgivingFullness,
+          ModifierType.ITEM,
+          ItemPool.PANTSGIVING);
+    }
+  }
+
+  public final void applyAdditionalSpleenCapacityModifiers() {
+    if (Preferences.getInteger("lastStillBeatingSpleen") == KoLCharacter.getAscensions()) {
+      this.addDouble(DoubleModifier.STOMACH_CAPACITY, 1, ModifierType.ITEM, ItemPool.PANTSGIVING);
```
midgleyc, line 1006: spleen capacity, still-beating spleen

gausie: Oops!

src/data/modifiers.txt
gausie: I should probably update data file consistency stuff to catch this...

midgleyc (approved):

src/data/modifiers.txt
```diff
@@ -6388,7 +6388,7 @@ Effect	Inner Warmth	Cold Resistance: +5
 # Inocente, In a Sense: Te hace ver como un Persona Inocente Cadáver
 Effect	Inocente, In a Sense	Avatar: "Persona Inocente Cadáver"
 Effect	Insani Tea	Random Monster Modifiers: +3
-Effect	Insatiable Hunger	Food Drop: +50, Booze Drop: +50, Candy Drop: +50
+Effect	Insatiable Hunger	Food Drop: +50, Booze Drop: +50, Candy Drop: +50, Stomach Capacity: +5
```
Semenar, line 6391: Does Insatiable Hunger really increase stomach by 5?

PR #1965, opened 2023-09-17
Printing a stack trace on abort should check if it is an abort

Veracity0:
I have no idea who/when added "printStackOnAbort", but printing one on any "exit" rather than abort() (which puts KoL into MafiaState.ABORT and then exits) is clearly a bug.


heeheehee-kolmafia (approved):
I would not be surprised if this was an oversight on my behalf.

Thanks for the fix!

PR #1969, opened 2023-09-19
Resize organ contents after small

gausie:
Because this is the first liberateKing test, it caused some leakage that meant I needed to fix up some other tests to get this passing. I've kept those improvements here, so this PR is bigger than it should otherwise be!

I made a bunch of changes so that the leakage wouldn't happen, but mafia is just not set up to re-initiate the ConcoctionPool. When I stepped back and thought about it I realised that the test I was trying to make work was actually asserting something false (sort of)


test/net/sourceforge/kolmafia/KoLCharacterTest.java
```diff
@@ -548,4 +548,16 @@ public void oldQuestDoesNotGiveExperience() {
       }
     }
   }
+
+  @Nested
+  class Liberation {
+    @Test
+    void resizesOrganContentsAfterSmallPath() {
+      try (var cleanups = new Cleanups(withPath(Path.SMALL), withInebriety(1), withFullness(1))) {
```
Alium58, line 556: Small can have up to 2 fullness used, which updates to 20 after liberation. Might be interesting to test this as it is above the typical fullness limit

gausie: Sure. I am just multiplying the number by 10 ;)

gausie:
I do not understand the failing test here...

gausie:
I understand the failing test now, turns out liberating the king leaks all sorts of things everywhere in the test environment

midgleyc (approved):
Mostly testing changes ;)

PR #1970, opened 2023-09-19
Ignore failed rests

Semenar:
If you try to rest when it is pointless to do so (full HP, full MP, no Beaten Up), the game won't let you do it, and will display a message "You don't need to rest right now." instead. This does not decrement the amount of free rests you can use, and this does not subtract an Adventure if you are out of free rests. However, KoLmafia thinks that the rest did happen, and increases timesRested accordingly. This leads to the undercounting of the amount of free rests left (Mafia fixes the data if it sees that a free rest is actually available, but it thinks it is not; however, the information that there is more than one free rest would be lost). The fix is to not increment timesRested in the case of a failed rest.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/CampgroundRequest.java
+++ b/src/net/sourceforge/kolmafia/request/CampgroundRequest.java
@@ -1011,9 +1011,9 @@ public class CampgroundRequest extends GenericRequest {
 
     if (action.equals("rest")) {
       // You don't need to rest right now.
-      // #  this does not use a free rest charge,
-      // #  and it does not cost an Adventure if out of free rests,
-      // #  so it should not be counted as a rest
+      //   this does not use a free rest charge,
+      //   and it does not cost an Adventure if out of free rests,
+      //   so it should not be counted as a rest
       if (!responseText.contains("You don't need to rest right now.")) {
         Preferences.increment("timesRested", 1);
       }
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/request/CampgroundRequest.java
```diff
@@ -1010,7 +1010,13 @@ public static void parseResponse(final String urlString, final String responseTe
     }
 
     if (action.equals("rest")) {
-      Preferences.increment("timesRested", 1);
+      // You don't need to rest right now.
+      // #  this does not use a free rest charge,
+      // #  and it does not cost an Adventure if out of free rests,
+      // #  so it should not be counted as a rest
```
midgleyc, line 1016: could you remove these starting #s, please?

test/net/sourceforge/kolmafia/request/CampgroundRequestTest.java
```diff
@@ -114,6 +114,22 @@ void updatesBookshelfLibramCasts() {
     }
   }
 
+  @Test
+  void doesNotCountFailedRests() {
```
midgleyc, line 118: it would be nice to have a test confirming that `timesRested` increases outside this case, just to confirm the test is testing the right thing.

midgleyc (approved):
Great, thanks!

PR #1972, opened 2023-09-20
Read iotm pull messages as events (and refactor and test EventsManager)

gausie:
When you start a new run, you get a message for each item that is autopulled. As I always use mafia, I never see these  - they are served to a main.php that is eaten by automation and never shown to me.

This code treats them as events, and renders them when next you visit main.php in the relay browser. This is an example of them being rendered by this PR:

Screenshot 2023-09-20 at 16 23 51

Typically events are also rendered in the gcli - I'm not sure what to do here because it feels quite spammy in its current form. This is what they look like when enabled:

Screenshot 2023-09-20 at 15 10 41

For now I have not enabled them - they only show on the relay browser.

Thoughts and feedback strongly desired!

P.S. Ignore the multiple IotM pulls - quirks of the dev server :)


midgleyc:
I like the idea of showing the on-ascension messages, and I'll look into the code later.

src/net/sourceforge/kolmafia/session/EventManager.java
```diff
@@ -100,27 +91,26 @@ public static boolean addNormalEvent(String eventHTML, boolean addTimestamp) {
...
-
-    eventText = eventText.replaceAll("<.*?>", "");
-
-    if (addTimestamp) {
-      EventManager.eventTexts.add(
-          EventManager.EVENT_TIMESTAMP.format(new Date()) + " - " + eventText);
-    } else {
-      EventManager.eventTexts.add(eventText);
-    }
+    EventManager.eventTexts.add(eventText);
```
midgleyc, line 105: Before the refactor, this seems to have checked `addTimestamp` and prepended the date if set.

gausie: Ah yeah, I had kept this behaviour and then slowly removed it as I tried to fix other unexpected changes that causes test failures. Fixed now

midgleyc (approved):

PR #1974, opened 2023-09-22
Fix Insatiable Hunger

Semenar:
The "Insatiable Hunger" effect is listed as having +5 stomach capacity since the organ update a few days ago. There is an Insatiable Hunger that does provide +5 stomach capacity, but it is a skill.


gausie:
Great spot! Sorry about that.

gausie (approved):

PR #1975, opened 2023-09-22
fix: don't crash on an <img> with no src attribute

midgleyc:
Should fix https://kolmafia.us/threads/missing-img-src-with-ocrs-hilarious-modifier.29263/

No test because no HTML.


gausie:
Looks good. I will get you some html though!

midgleyc:
Thanks. I am interested in how it happened -- the `src` attribute is mandatory for `` elements, so I would have thought this couldn't happen (in well-formed HTML, and why would you have an `` with no `src`?)

gausie (approved):
I couldn't reproduce to get HTML. Maybe one day someone can!

PR #1976, opened 2023-09-22
Fix Smut Orc Logging Camp combat rate

Semenar:
It is possible to get stuck in a situation when Smut Orc Logging Camp reports as 0% combat forever - if you qualified for Blech House at the same time you finished the bridge; since Blech House does not appear after building the bridge, the condition on progress being at least 15 would forever be true. Added an additional check that less than 30 bridge progress was done.


gausie:
Thank you! In this case, you've found a bug. If you could write a unit test that would fail if your fix were not in place but would pass now, that would be great! That way the bug cam never accidentally regress.

Semenar:
> Thank you! In this case, you've found a bug. If you could write a unit test that would fail if your fix were not in place but would pass now, that would be great! That way the bug cam never accidentally regress.

I'm not sure how a unit test would look like in this case - I think this will be very redundant because it will basically check that this if is still in place. I don't see any other way to test it because that combat rate calculation is something internal.

gausie:
That's not redundant - I don't mind if it's like that.

gausie (dismissed):
Keeping things organised by marking this as changes requested

midgleyc (approved):

PR #1977, opened 2023-09-23
feat: fix some negative ids

midgleyc:
I think this is all the ones that actually exist as separate monsters:
* butts are somebody else's butt with a flag
* elfbots became elfborgs
* hard mode dread bosses are the same as normal ones
* all slime tube monsters are the same as each other
* intelligent alien is not actually a monster


gausie (approved):

Veracity0:
Intelligent alien is not a monster, but it has a bunch of images, so by pretending it is one, you can see them all in the Monster section of the Encyclopedia.

PR #1980, opened 2023-09-24
detect TwinkleVision with grandpa and when parsing beach

Veracity0:
New property: "hasTwinkleVision".

1) set when you ask grandpa about hierfal
2) set if we see twinkles on a beach


gausie (approved):
Ok but no TwinkleVision until you've done your homework

PR #1982, opened 2023-09-24
New TCRS dump from @chunkinaround

gausie:


gausie:
Ah @midgleyc there seem to be issues with parsing damage for `Damage Aura` and `Thorns`

midgleyc:
I think the fix is that we shouldn't be storing the damage we parse out of the KoL item description (or perhaps at all), but it is rather nice to see the potential items.

midgleyc:
Also Turtle Tamer Vole has lost a bunch of items.

midgleyc:
/also/ Knob Goblin encryption key has been renamed to Knob Goblin Encryption Key. When? Why? Who knows.

gausie:
Getting 3bh to rerun it all rather than sort all these thorn and damage aura things manually

gausie:
I checked in - @chunkinaround is abroad at the moment but will re run this when he gets back

gausie:
I'll close this and open a new one when @chunkinaround can get us a new readout

PR #1984, opened 2023-09-25
feat: detect whether we have free rests at campaway

midgleyc:
Progresses one of the issues in https://kolmafia.us/threads/mafia-setting-timesrested-to-0-on-login.29268/


gausie (approved):
Yes

PR #1985, opened 2023-09-25
fix: don't try to read thorns / damage aura from in-game description

midgleyc:
As mentioned in #1982:
* don't try to read thorns / damage aura from the game, as we use a custom valuation
* rename Knob Goblin Encryption Key, as the case has changed in-game


gausie (approved):
Nice. I might remove the values manually and spare 3bh having to run the derivations again

midgleyc:
yes, I think that's easy enough: just remove `Thorns`, `Thorns: [^,]+` and `Damage Aura: [^,]+`, then spot check.

PR #1987, opened 2023-09-26
Book of Facts: tracking stateful facts, exposing facts to users and scripts

gausie:
- Book of Facts seeding fully and accurately replicated
- Tracking for wishes, tatters, gummi effects and pinata effects with pref correction if things look out of sync
- Pulled out code from PocketDatabase into AdventureResults that parses `item (quantity)` strings and `effect (duration)` strings, added tests.
- Expanded expression syntax with lt(), lte(), gt() and gte() with tests (true is 1 and false is 0)
- Expose these data in MonsterProxy fields
- Expose these data via new RuntimeLibrary functions


midgleyc:
Any chance of a `DataFileTest` / `DataFileConsistencyTest` for `bookoffacts.txt` after the form stabilizes?

gausie:
> Any chance of a `DataFileTest` / `DataFileConsistencyTest` for `bookoffacts.txt` after the form stabilizes?

It produces errors as it parses and I have a test confirming that that takes place with no errors - do you think that's enough? I'm not sure what else I could do.

midgleyc:
> It produces errors as it parses and I have a test confirming that that takes place with no errors - do you think that's enough? I'm not sure what else I could do.

I missed that a test existed. I guess the other thing to confirm would be that the test fails if there is actually an error (the "does this test test the right thing" problem).

gausie:
> > It produces errors as it parses and I have a test confirming that that takes place with no errors - do you think that's enough? I'm not sure what else I could do.
> 
> I missed that a test existed. I guess the other thing to confirm would be that the test fails if there is actually an error (the "does this test test the right thing" problem).

That would also cover the missing coverage I have! Ok.

src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -1504,4 +1505,44 @@ public static WildcardResult getInstance(String text) {
       return null;
     }
   }
+
+  public static AdventureResult parseEffectString(String effectString) {
```
midgleyc, line 1509: How does this parse `Educated (Kinda)`? Can it work like item instead?

test/net/sourceforge/kolmafia/persistence/FactDatabaseTest.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+    "0, true",
+    "1, true",
+    "2, true",
+    "3, false",
+  })
+  void pocketWishesAreStateful(final int dropped, final boolean drops) {
```
midgleyc, line 69: I feel like this is a bit of a surprising way for the function to behave: you query about a monster in a different path, and Mafia tells you it drops nothing because you've already had all your pocket wish drops for today?

Is there a way to see that this would drop a pocket wish other than by clearing the pref temporarily?

midgleyc:
> That would also cover the missing coverage I have! Ok.

I'll be happy with just introducing errors into the file (once, manually) then running the test to confirm it breaks (in the expected manner). I don't think this needs to be persistent; I don't expect the file to change after its initial creation.

midgleyc:
Are there any monsters with a base meat fact that don't have any base meat? What happens then? Do they just not drop anything?

test/net/sourceforge/kolmafia/persistence/FactDatabaseTest.java
gausie: Nah that's not how this works, I'll document it in the mafia wiki once it's merged.

The runtime library functions are always reported ignoring your preference state. The monster proxy data is always reported for your current class/path/prefs.

gausie:
> > That would also cover the missing coverage I have! Ok.
> 
> I'll be happy with just introducing errors into the file (once, manually) then running the test to confirm it breaks (in the expected manner). I don't think this needs to be persistent; I don't expect the file to change after its initial creation.

I already did this while writing it!

src/net/sourceforge/kolmafia/AdventureResult.java
gausie: Yeah good idea. This was an issue with the old function that surfaced with `filet of tangy gnat (fotlief)`

test/net/sourceforge/kolmafia/persistence/FactDatabaseTest.java
midgleyc: great, can you add a test for runtime library functions not being stateful?

src/net/sourceforge/kolmafia/persistence/FactDatabase.java
```diff
@@ -0,0 +1,520 @@
...
+import net.sourceforge.kolmafia.objectpool.EffectPool;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.persistence.MonsterDatabase.Phylum;
+import net.sourceforge.kolmafia.utilities.FileUtilities;
+import net.sourceforge.kolmafia.utilities.PHPMTRandom;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
+
+public class FactDatabase {
+  private static final List<String> HEAP_ITEMS =
+      List.of(
```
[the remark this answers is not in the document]
gausie: Oh I suppose that is true also. I will test that I guess, easy place to introduce an error. Thanks, good spot

midgleyc (dismissed):
Thanks!

PR #1990, opened 2023-09-30
Add function overloads for getting current class and path for Book of Facts (and fix macrometeorite)

gausie:
- Add overrides for getting fact info in scripts using current class and path
- Use correct source for determining current monster, even after replacement


midgleyc:
Could you add a test confirming that the new functions are stateless, like the previous functions? I assume that's the main difference between the proxy record and the functions.

jaadams5 (approved):

PR #1992, opened 2023-10-02
feat: skeleton dark jill

midgleyc:
Skeleton Dark Jill-of-All-Trades.

Haven't added the Ghuol Whelp types because I don't think we're sure it acts as a Whelp yet.


gausie (approved):
Nice! I have some work in draft for supporting the LED candle, but I'm super busy over the next 7 days so might just lose it.

midgleyc:
Oh yeah, so we need:
* pref for candle setting
* pref for whether you got a candle today (maybe?)
* pref for candy block map drops (up to 3, daily limit)

midgleyc:
also the jill tweak adventure is walk-away-able

midgleyc:
also also this guy should be type "drop" due to the candy maps

PR #1993, opened 2023-10-02
Update items.txt

jaadams5:
autumn dollar is multiusable, but KoLmafia thought it was not
autumn leaf is multiusable, but KoLmafia thought it was not

I made the usual fix - changing usable to multiple and got a test error saying "multiple" is redundant for potions, so I just deleted usable instead.

### DIFF
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -10980,12 +10980,12 @@
 10952	Jurassic Parka	815637689	jparka7.gif	shirt		0
 10953	boxed autumn-aton	129222297	crate.gif	usable	t	0
 10954	autumn-aton	174185886	autumnaton.gif	usable	q	0
-10955	autumn leaf	639910181	leafflavor.gif	potion, usable	t,d	10	autumn leaves
+10955	autumn leaf	639910181	leafflavor.gif	potion	t,d	10	autumn leaves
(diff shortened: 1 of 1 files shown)


gausie (commented):
I believe many items were recently changed from usable to multiusable. Don't you need to actually add `multiple` here?

jaadams5:
> I believe many items were recently changed from usable to multiusable. Don't you need to actually add `multiple` here?

Mafia reported "autumn dollar is multiusable, but KoLmafia thought it was not".  Changing "usable" to "mulitple" caused a file  integrity test to fail.  So removing "usable" should remove the "autumn dollar is multiusable, but KoLmafia thought it was not" message and allow the data file integrity test to pass.  The test failure message implied that potions were now, by definition, multiusable and thus potion, multiple was redundant.   Perhaps the test is wrong?

jaadams5:
DataFileConsistencyTest,testPotions has the comments:
// All Potions And avatar potions are multiusable by default.
// Do not allow a redundant "multiple" attribute

Date is 7/22/2023

Veracity0 (approved):
That is the correct fix.

gausie:
Great! I wasn't sure

PR #1995, opened 2023-10-04
Lowercase the topic when consulting grandpa

libraryaddict:
Noticed twinkle vision pref didn't update when I used a capitalized topic, despite grandpa still talking about it.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/GrandpaRequest.java
+++ b/src/net/sourceforge/kolmafia/request/GrandpaRequest.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia.request;
 
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.sourceforge.kolmafia.KoLConstants.MafiaState;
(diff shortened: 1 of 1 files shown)


gausie:
By this logic there could be a behaviour that requires using capital
letters that would then be impossible to submit via mafia.

Make a big report to KoL I think.

libraryaddict:
@gausie I don't follow. kol behaves properly, but mafia isn't setting the twinkle vision preference because I used "Hierfal" instead of "hierfal"

jaadams5:
I noticed that the preference was not set for me when I asked about "Hierfal" as well.  There is precedent for assuming user input is case insensitive in which case this fix is reasonable.  If it is possible, likely, or currently true that the Grandpa topics could be case sensitive on KOL's side then this fix could introduce a problem in the future and changing the switch to trigger on "Hierfal" would be a better fix.

gausie:
Oh! Totally misunderstood. This is good then.

jaadams5 (dismissed):

midgleyc (approved):

PR #1998, opened 2023-10-05
Detect abbreviations of "muscle experience percent" in maximizer

libraryaddict:
There's multiple ways you could write "muscle experience percent" into maximizer, but unless you write it exactly as its spelled then the maximizer will default to "muscle". Including "muscle experience percentage" and "muscle exp percent", etc.

This PR fixes that by use of regex to catch the obvious abbreviations.
The main thing I'm unsure of, is if the test should be written another way.

The first commit is the failing test.


midgleyc (approved):
Thanks!

PR #1999, opened 2023-10-05
feat: add preferences for jill drops

midgleyc:
Not certain that maps cap at 3 drops, but that's the highest currently reported. It might be unlimited.

Preference for LED candle setting / command / modifiers to come later.

### DIFF
--- a/src/net/sourceforge/kolmafia/FamiliarData.java
+++ b/src/net/sourceforge/kolmafia/FamiliarData.java
@@ -1193,7 +1193,7 @@ public class FamiliarData implements Comparable<FamiliarData> {
             ItemPool.MAP_TO_A_CANDY_RICH_BLOCK,
             "maps",
             "_mapToACandyRichBlockDrops",
-            3));
+            -1));
   }
 
   public static DropInfo getDropInfo(int id) {
--- a/src/net/sourceforge/kolmafia/FamiliarData.java
+++ b/src/net/sourceforge/kolmafia/FamiliarData.java
@@ -1187,6 +1187,13 @@ public class FamiliarData implements Comparable<FamiliarData> {
             "grubby wool",
             "_grubbyWoolDrops",
             -1));
+    DROP_FAMILIARS.add(
+        new DropInfo(
+            FamiliarPool.JILL_OF_ALL_TRADES,
+            ItemPool.MAP_TO_A_CANDY_RICH_BLOCK,
+            "maps",
+            "_mapToACandyRichBlockDrops",
+            3));
   }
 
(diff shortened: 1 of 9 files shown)


jaadams5:
> Not certain that maps cap at 3 drops, but that's the highest currently reported. It might be unlimited.

I know nothing about the item and am just looking at the code but I don't understand why we count map drops per day but maps used is a boolean.  Is the information I am missing is that only one map can be used per day?

src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -1187,6 +1187,13 @@ private static void loadDropFamiliars() {
             "grubby wool",
             "_grubbyWoolDrops",
             -1));
+    DROP_FAMILIARS.add(
+        new DropInfo(
+            FamiliarPool.JILL_OF_ALL_TRADES,
+            ItemPool.MAP_TO_A_CANDY_RICH_BLOCK,
+            "maps",
+            "_mapToACandyRichBlockDrops",
+            3));
```
jaadams5, line 1196: Since you are not sure about a cap of 3 would it be better to leave this uncapped until there is come certainty?

Not intended to block.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -359,6 +359,7 @@ public class Preferences {
         "latteModifier",
         "latteUnlocks",
         "leafletCompleted",
+        "ledCandleDropped",
```
jaadams5, line 362: Do we know the "on Ascension" status of map usage and drops?

jaadams5 (dismissed):
Questions and comments are not intended to block.

src/net/sourceforge/kolmafia/preferences/Preferences.java
midgleyc: map usage / drops is a daily pref, so it gets reset on ascension automatically

midgleyc:
> Is the information I am missing is that only one map can be used per day?

Yes.

src/net/sourceforge/kolmafia/FamiliarData.java
midgleyc: yeah probably

src/net/sourceforge/kolmafia/preferences/Preferences.java
jaadams5: OK.  I didn't know that, in general, and the presence of _shortOrderCookCharge in the Ascension resets made me think otherwise.  Presumably _shortOrderCookCharge in that list is redundant?

PR #2000, opened 2023-10-06
Fix appearance_rates calculation regression.

heeheehee-kolmafia:
This was introduced by #326 several years back, as verified by building the old revision and executing the corresponding ASH function, both at that revision and the preceding one (to confirm prior behavior).

The refactor bundled all calculations that depended on the current character's state, even though combat rate modifiers should be applied uniformly. This leads to inconsistent results, as reported by lrrat (https://kolmafia.us/threads/29297/).


midgleyc (approved):
Thanks!

PR #2003, opened 2023-10-08
PHPMTRandom fix

Semenar:
In the current implementation, reload always twisted the same values since the new numbers are appended to state, not overwritten. This creates a generator with a period of 624 which just loops through the first 624 numbers continuously.

The fix is to take latest STATE_LENGTH values of a state into account instead of the first ones.


midgleyc (approved):
Thanks!

Is this breaking anything at the moment or just a correctness fix?

Semenar:
> Thanks!
> 
> Is this breaking anything at the moment or just a correctness fix?

As far as I know, nothing currently should use that many PHPMTRandom calls, so it is just a correctness fix.

PR #2005, opened 2023-10-09
Do not start Rain Man counters outside of Rain Man

gausie:
Fixes https://kolmafia.us/threads/rain-monster-counter-outside-of-heavy-rains.29304/


midgleyc (approved):
Thanks!

PR #2010, opened 2023-10-10
Fix ping command not supporting api.php as the value for pingDefaultTestPage

s-k-z:
For users who ran Kolmafia between #1821 and #1828, `pingDefaultTestPage` was set to `api.php` for their default value and likely never changed. For those users, the ping command doesn't work for two reasons:

* PingRequest's generic request URL gets set to `(none)` when it fails to match a string ending in .php
* `isSafeToRun()` returns false for any `pingURL` that ends in .php

Example failure:
```
> get pingDefaultTestPage

api.php

> ping

Ping returned no response; ping test aborted
0 pings to api.php at 0-0 msec apiece (total = 0, average = 0.00) = 0 bytes/second
```
After this change:
```
> ping

10 pings to api.php at 79-95 msec apiece (total = 846, average = 84.60) = 23,191 bytes/second

> ping

10 pings to api at 81-103 msec apiece (total = 910, average = 91.00) = 21,560 bytes/second
```

This PR addresses both issues and makes the command operable again under the above condition. Additionally, I eliminated some warnings in the `PingRequest.java` file.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/PingRequest.java
+++ b/src/net/sourceforge/kolmafia/request/PingRequest.java
@@ -2,7 +2,7 @@ package net.sourceforge.kolmafia.request;
 
 public class PingRequest extends GenericRequest {
 
-  private String pingURL = "";
+  private final String pingURL;
   private long startTime = 0L;
   private long endTime = 0L;
 
@@ -12,18 +12,15 @@ public class PingRequest extends GenericRequest {
   // api.php's responseText is about 1/4 the size of main.php's - and
   // measured ping time is about 1/4 as long.
 
-  public PingRequest() {
-    this("api.php");
-  }
-
   public PingRequest(String pingURL) {
     super(pingURLToFormURL(pingURL));
-    this.pingURL = pingURL;
+    this.pingURL = pingURLToFormURL(pingURL);
     this.addFormFields();
   }
 
   private static String pingURLToFormURL(String pingURL) {
     return switch (pingURL) {
+      case "api.php", "council.php", "main.php" -> pingURL;
       case "api", "council", "main" -> pingURL + ".php";
       case "(status)", "(events)" -> "api.php";
         // What is this?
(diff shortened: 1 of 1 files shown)


Veracity0:
I'm all for backwards compatibility, but "api.php" is ambiguous.
Is it "status" or "events'?
I going to have to consider your proposed solution.

s-k-z:
Ah, yeah this one just loads the API help page. Since it resolves as a real page maybe that's fine?

src/net/sourceforge/kolmafia/request/PingRequest.java
```diff
@@ -12,18 +12,15 @@ public class PingRequest extends GenericRequest {
   // api.php's responseText is about 1/4 the size of main.php's - and
   // measured ping time is about 1/4 as long.
 
-  public PingRequest() {
```
Veracity0, line 15: Why did you delete this constructor?

src/net/sourceforge/kolmafia/request/PingRequest.java
```diff
@@ -117,14 +115,6 @@ protected boolean shouldFollowRedirect() {
     return false;
   }
 
-  public long getStartTime() {
```
Veracity0, line 120: Again, why did you delete these "getters"?

Veracity0 (changes requested):

Rinn:
if it's just to fix those users why don't we just update the property if 'api.php' is set and reset back to the default 'api' then in like a month delete the code that resets the property.

src/net/sourceforge/kolmafia/request/PingRequest.java
s-k-z: It's unused as far as I can tell.

src/net/sourceforge/kolmafia/request/PingRequest.java
s-k-z: This constructor isn't used as far as I can tell.

Veracity0:
> if it's just to fix those users why don't we just update the property if 'api.php' is set and reset back to the default 'api' then in like a month delete the code that resets the property.

I've been away from KoL and my development computer for 2.5 weeks.

I approve of this suggestion. Migrate the setting and delete (or not) the code to do that after a suitable period.

Rinn:
I've just added a temp hack to reset the property

PR #2011, opened 2023-10-12
Call recalculateAdjustments after loading skills.

heeheehee-kolmafia:
This also includes a couple more changes to make sure that we don't accidentally populate availablePassiveSkillModifiersByVariable before we have any skills ready.

This doesn't handle the scenario where the skills list changes, but it should handle the most common scenarios (login, kingLiberated come to mind).


midgleyc (approved):
Thanks!

PR #2012, opened 2023-10-12
Address passive skill cache race condition.

heeheehee-kolmafia:
This change does a few things:

- Explicitly update Modifiers.availableSkillsChanged when receiving a new skill, instead of relying on the preference listener, as that may be deferred (as we've updated the KoLmafiaTest case to cover).

- Explicitly invoke recalculateAdjustments after parsing the charsheet. This addresses messages on login about detecting free rests from campground.php.

- Add some synchronization to Modifiers.java to prevent potential data races.

- Remove redundant firePreferenceListener("(skill)") from ResponseTextParser, since that's already done several lines above in the addAvailableSkill call.

Note that I explicitly did _not_ call recalculateAdjustments() inside addAvailableSkill() because that would result in quadratic time behavior whenever KoLCharacter.setAvailableSkills() is invoked.


gausie:
Yeah this is great. Huge thanks for diving into it.

gausie (approved):

PR #2014, opened 2023-10-14
Mark BEVERAGEs in fullness.txt

midgleyc:
Extract beverage tracking from #1767.

Beverages don't get a bonus from +food drops (unless they are also cooking ingredients), so it's useful to have Mafia know whether or not something is a beverage.


jaadams5 (approved):
I can't really figure out a test that is useful for anything besides coverage, either.

midgleyc:
I could remove the extra methods because they're not called anywhere (yet), but they may as well exist given that BEVERAGE exists, and #1767 is presumably intended to be revived at some point. At which point the coverage problem will fix itself.

PR #2016, opened 2023-10-16
Fix some combat rates data

Semenar:
As per wiki:
- The eXtreme Slope: 95% combat
- The Jungles of Ancient Loathing: 85% combat


midgleyc (approved):
Thanks!

PR #2017, opened 2023-10-20
Double mods of sleeved cards with Offhand Remarkable

danielwhite:
This change ensures that the modifiers of the sleeved card is considered instead of the card sleeve itself.

I've also included a maximizer test because I initially suspected this as the source of the problem. It seemed worthwhile to keep as I could not find any other obvious tests for the card sleeve slot.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -5581,25 +5581,32 @@ public abstract class KoLCharacter {
   private static void addOffhandRemarkable(
       Map<Slot, AdventureResult> equipment, Modifiers newModifiers) {
     var offhand = equipment.get(Slot.OFFHAND);
-    addOffhandRemarkable(offhand, newModifiers);
+    addOffhandRemarkable(equipment, offhand, newModifiers);
     // and also offhand items that are equipped on the familiar
     var famItem = equipment.get(Slot.FAMILIAR);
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5581,25 +5581,32 @@ public static void addItemAdjustment(
...
-        && item != EquipmentRequest.UNEQUIP
-        && ItemDatabase.getConsumptionType(item.id) == ConsumptionType.OFFHAND) {
-      if (item.id != ItemPool.LATTE_MUG) {
-        var mods = ModifierDatabase.getItemModifiers(item.id);
+  private static void addOffhandRemarkable(
+      Map<Slot, AdventureResult> equipment, AdventureResult item, Modifiers newModifiers) {
+    if (item == null
+        || item == EquipmentRequest.UNEQUIP
+        || ItemDatabase.getConsumptionType(item.id) != ConsumptionType.OFFHAND
+        || item.id == ItemPool.LATTE_MUG) {
```
danielwhite, line 5595: I wasn't able to determine why this is excluded as I don't have a latte mug of my own. Is this intended?

midgleyc: yes: mug enchants aren't doubled.

midgleyc (approved):
Thanks!

PR #2018, opened 2023-10-21
Support Halloween treats that are randomly selected

gausie:
Fixes https://kolmafia.us/threads/outfit_treats-returns-empty-aggregate-for-outfits-with-multiple-treats.29327/

- Add chances for outfit treats
- Remove choice for gift shop from The Shore
- Formatting


midgleyc (approved):
rather fortunate that the outfit treats code return a map of item to int already, huh?

PR #2019, opened 2023-10-21
feat: mark items as multiusable, thanks to adeyke

midgleyc:
Mark the items given in https://kolmafia.us/threads/missing-multiusable-items.29317/ as multiusable.


jaadams5 (approved):

jaadams5:
Thank you.

Might it be worth adding a new command to KoLmafia that would iterate through the items and validate key attributes against KoL?  I don't think we want a test because this would have to have live access to KoL.

midgleyc:
We have `checkdata`. I think for this one we also need access to the items themselves, like `checkpowers` has.

PR #2020, opened 2023-10-22
feat: ellsburys claim monsters, start spacegate, other misc

midgleyc:
funky junk key conditional is that it starts at 5 and increases by 5% each time one fails to drop.

Worse Homes and Gardens conditional is like the funky junk key but it only happens after your third funky junk key.


src/data/monsters.txt
```diff
@@ -87,7 +87,7 @@ Blooper	118	squid.gif	Atk: 25 Def: 22 HP: 15 Init: 60 Meat: 10 P: beast Article:
 Blue Oyster cultist	1528	bluecultist.gif	Atk: 135 Def: 135 HP: 130 Init: 50 Meat: 100 P: dude Elem: 75 EA: hot Article: a	fire of unknown origin (15)	black blade (5)	cigarette lighter (15)	blue oyster badge (c15)
 blur	460	blur.gif	Atk: 136 Def: 126 HP: 141 Init: 40 Meat: 120 P: dude Article: a	drum machine (30)	palm frond (10)	hot date (5)	carbonated water lily (10)	Hugo's Weaving Manual (c100)
 boaraffe	568	boaraffe.gif	Atk: 145 Def: 130 HP: 145 Init: 50 Meat: 100 P: beast Article: a	long pork (30)
-Bob Racecar	132	naskar2.gif	Atk: 138 Def: 129 HP: 139 Init: 80 Meat: 120 P: dude	ketchup hound (35)	leather chaps (15)	mesh cap (15)	stunt nuts (30)	a butt tuba (0)	folder (D-Team) (c0)
+Bob Racecar	132	naskar2.gif	Atk: 138 Def: 129 HP: 139 Init: 80 Meat: 120 P: dude	mesh cap (15)	leather chaps (15)	stunt nuts (30)	ketchup hound (35)	a butt tuba (10)	folder (D-Team) (c5)
```
jaadams5, line 90: Any reason for changing the order?  Not a blocking comment.

jaadams5 (approved):

src/data/monsters.txt
midgleyc: same order as in-game

order matters for goose drops, as drones apply one by one

PR #2023, opened 2023-10-23
Fix comment style in data files

Semenar:
Some lines in the combats.txt data files used the //-style comment - this is the only place where it is done like that. Replaced them with #-style comments.


gausie (commented):
Thanks, this is useful for non-mafia parsers that use these data files. Not approving in case someone else knows a reason these should be kept? It seems weird so perhaps is on purpose.

jaadams5:
The comments in question were added by @Veracity0 08/27/2019.

In my case I have been coding and scripting long enough that "#" is not my first expectation for a comment character and I might not have noticed that I had mixed and matched commenting styles.  Perhaps something similar is the case here?

PR #2025, opened 2023-10-25
feat: skeleton franken stein, misc changes

midgleyc:
Embezzler visor drop rate.

Note McTwist gives +75% item.

Fix Jill LED candle modifiers to be on candle, not Jill.

PR #2027, opened 2023-10-26
feat: make led candle a modeable

midgleyc:
Make maximizer aware of LED candle modes.

Fix the issue introduced in #2025 where moded Jill was being treated as a half-weight (whatever) instead of 1.5x weight. I don't know why the previous code was only setting weight if 0, or why it was using weight and not capped weight -- does anyone else have any idea?


midgleyc:
okay, as from the failing test, it's because familiars like Wizard Action Figure are fractional fairies.

midgleyc:
Need to go back to the design phase here.

midgleyc:
I think metallic foil radar dish implies the intent was for it to work the way I thought it worked, but it doesn't work that way.

midgleyc:
I was initially going to fix the issues with `switch` not including the led candle in this PR, but I think it's still valuable to go in without.

midgleyc:
@kolmafia/staff -- any comments?

jaadams5 (approved):

jaadams5:
> @kolmafia/staff -- any comments?

I wasn't sure it was ready and I wasn't sure I had much to contribute.  But to the extent of my ability, it looks OK.

midgleyc:
Yeah, it's ready. There's still stuff to do, but I wanted to draw a line under where we are so far.

PR #2028, opened 2023-10-27
fix: jill bonuses

midgleyc:
Extract from #2027 to fix Jill.

I think we'll want to require `Volleyball:` modifiers instead of defaulting to 1 if the familiar is the right type, but that can go later. This is the only way I can see to fix e.g. God Lobster's Ring meaning the lobster is treated as a -1x Volleyball, instead of a 0x Volleyball.


midgleyc:
Failing because it's FoB. Time to pivot again to something like `HttpUtilities` for `DateTimeManager` to fix this.

midgleyc:
It's been 5 days, pushing it in to keep things moving. Necessary to fix the bug with the existing jill behaviour.

PR #2030, opened 2023-10-29
add is_adventuring & has_queued_commands functions

Rinn:
My intent here is to allow relay scripts to be able to detect if another script is running or if the player is running automated adventuring, so they could defer updates or prevent the user from visiting links in the browser that could get them into a bad state.


Rinn:
ah #2029 is prob why the tests failed

midgleyc (approved):
Ok!

PR #2044, opened 2023-11-04
fix: identify burning leaves in small

midgleyc:


Veracity0 (approved):

Veracity0:
I merged this because it looked fine to me and I am just about to use my IOTM and ascend into a Small run, after being on the road for 12 days and not even signing into KoL during the trip.
I'll let you know how it works. :)

PR #2047, opened 2023-11-04
Persistent Command Buffer

jaadams5:
See https://kolmafia.us/threads/persistent-command-history-control-history-size.29213/

This saves the gCLI and tabbed chat command buffers to separate GLOBAL preferences.  The commands are in a list delimited by the sharp sign which some people insist on calling a hashtag.  The list is loaded on start up and the command index is set to the last element so an up arrow in a new session should select the last command from the previous session.

I've lightly tested this with the gCLI but not tabbed chat.

There was a request to allow the size of the buffer to be a preference.  I decided NOT to implement that because if the size changed then the CommandDisplayFrame would have to be rebuilt or mafia shutdown and restarted before the effect was noticeable.  It is trivial, now or later, to change the hardcoded 20 to something else if that is what people want.


src/net/sourceforge/kolmafia/swingui/panel/CommandDisplayPanel.java
```diff
@@ -56,6 +65,24 @@ public CommandDisplayPanel() {
...
+    }
+  }
+
+  private void updatePreference() {
+    StringBuilder newPref = new StringBuilder();
+    for (String s : commandHistory) {
+      newPref.append(s).append(DELIMITER);
+    }
+    newPref.deleteCharAt(newPref.length() - 1);
+    setString(preference, newPref.toString());
```
midgleyc, line 83: should be able to use `String.join(DELIMITER, commandHistory)` to avoid the for loop?

src/net/sourceforge/kolmafia/swingui/panel/CommandDisplayPanel.java
```diff
@@ -24,9 +28,14 @@ public class CommandDisplayPanel extends JPanel implements FocusListener {
   private final AutoHighlightTextField entryField;
   private final JButton entryButton;
 
+  private final String preference;
+  private final String DELIMITER = "#";
```
midgleyc, line 32: while # is unlikely to appear, why not unit separator (␟) or something even less likely to appear?

src/net/sourceforge/kolmafia/swingui/panel/CommandDisplayPanel.java
jaadams5: Thank you.  I didn't remember/think of that.  Stand by.

src/net/sourceforge/kolmafia/swingui/panel/CommandDisplayPanel.java
jaadams5: Forgot/didn't know it existed.  But I went for the # because it was also already on most keyboards and easily managed in dumb, ASCII text editors.  We can discuss but human readability and editing or preference files are of concern to me.

midgleyc (approved):
Seems to work.

PR #2050, opened 2023-11-07
feat: add seed, mt_rand functions to ASH

midgleyc:
The TourGuide developers have asked for a function exposing `mt_rand` for use in calculating Book of Facts values and the like.

Implement something that does this.

@Semenar, @docrostov: does this do what you want it to do?

Anybody else: does the design seem reasonable?


gausie:
They'll need regular rand as well as mt rand

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -7812,6 +7819,16 @@ public static Value group_string(
     return value;
   }
 
+  public static Value seed(ScriptRuntime controller, final Value expr) {
```
gausie, line 7822: Give params meaningful names if you can - we extract them for Typescript definitions

src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
```diff
@@ -112,7 +112,7 @@ public static List<net.sourceforge.kolmafia.textui.parsetree.Function> getFuncti
       for (VariableReference variableReference : libraryFunction.getVariableReferences()) {
         allTypes.add(variableReference.getType());
       }
-      if (allTypes.contains(DataTypes.MATCHER_TYPE)) {
+      if (allTypes.contains(DataTypes.MATCHER_TYPE) || allTypes.contains(DataTypes.RNG_TYPE)) {
```
gausie, line 115: I think we want them to specifically come through to JavaScript - the reason we ban matchers is because regular expressions are handled by JavaScript separately.

midgleyc: I figured TypeScript users can use https://www.npmjs.com/package/kol-rng instead, unless there's some reason it's preferable for them to use the Mafia version?

midgleyc:
should the `rand` function be called `rand` or `php_rand`?

Does it make sense to prefix everything with `php`:
```
php_seed
php_ran
php_mt_rand
```

gausie:
I think that makes the most sense yeah

midgleyc:
Having some difficulty getting it working in JavaScript, the naive approach gives:

> Script exception: ASH function returned native Java object.
----
> Does the seed function seed both rand and mt_rand simultaneously?

Yes, and independently (though that's an implementation detail and not a guarantee I make).

src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java
midgleyc: I can't figure out how to make JavaScript work with this.

I get
> Script exception: ASH function returned native Java object.

and uncommenting the error means the object is returned as a string instead. I tried wrapping it (like you wrapped `sessionStorage`) to get a `NativeJavaObject` explicitly, but it had the same problem.

Veracity0:
What's up with this? It's a year old.

midgleyc:
I couldn't get it working in JavaScript, which seemed a blocker.

docrostov:
> I couldn't get it working in JavaScript, which seemed a blocker.

If it works in ASH, can we at least get the ASH functionality pushed?

gausie:
I think we should merge this, and I will set myself a project of trying to implement a coercer to JS

gausie (approved):

PR #2053, opened 2023-11-11
feat: begin supporting multidrops

midgleyc:
Some monsters drop items all in one line, with "you acquire X items" (e.g. hobo boss drops, sprinkles, the new inflammable leaves). Support this as a custom drop type.

Multidrops are conditional. They are not necessarily no-pickpocket, but we should pretend they are, because pickpocketing a multi-drop gives you exactly 1 and prevents you from obtaining any after the combat, which sucks.

Multidrops have custom behaviour, and don't necessarily interact with item drop. They interact weirdly with things like pickpocket or monster modifiers or custom combats, which explains why e.g. Jerry Bradford replacing Groar drops one dense meat stack instead of five.

Most multi-drops are triangularly distributed among the possible values. Some aren't.

For some multi-drops, we can figure out exactly how many we should be getting: hobo boss drops, salty sailor salt, wumpus hair. I've not implemented this.

Similarly, sprinkles, fudgecules and rubees receive bonus drops: this is not implemented either.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <classpath>
-	<classpathentry kind="src" path="src" excluding="jar/"/>
+	<classpathentry excluding="jar/" kind="src" path="src"/>
 	<classpathentry kind="src" path="lib"/>
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER"/>
 	<classpathentry kind="lib" path="src/jar/antlr-runtime-3.4.jar"/>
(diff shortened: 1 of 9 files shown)


jaadams5:
What do the changes to .classpath and .project do?

jaadams5 (approved):
Ideally someone else would also look at this but it makes sense and I see not obvious changes or tweaks.

midgleyc:
Alright, let's go with this and let issues come out in play.

For `.project` and `.classpath`: when I opened the project in vim with [`coc-java`](https://github.com/neoclide/coc-java), those changes were made automatically. They don't look substantial: a change to attribute ordering, and a change to a regex to properly escape the `.`.

jaadams5:
> Alright, let's go with this and let issues come out in play.
> 
> For `.project` and `.classpath`: when I opened the project in vim with [`coc-java`](https://github.com/neoclide/coc-java), those changes were made automatically. They don't look substantial: a change to attribute ordering, and a change to a regex to properly escape the `.`.

Thanks.  Long ago changes like this broke things.  The real fix was to set up the IDE properly and that has been done but the paranoia persists.

PR #2055, opened 2023-11-14
Fix skill type for Ravenous Pounce.

heeheehee-kolmafia:
This is a combat skill with a passive effect.

In particular, this was previously correctly categorized, since Stomach Capacity was not a modifier until recently.

It probably makes sense to add a test to guard against adding a non-passive skill that has an associated modifier, although with new skills, this is less likely to be an issue (since modifier tracking simply won't work if implemented as the wrong type).

### DIFF
--- a/src/data/classskills.txt
+++ b/src/data/classskills.txt
@@ -1122,7 +1122,7 @@
 12007	Indefatigable	wallshield.gif	0	0	0	7
 12008	Skullcracker	brokeskull.gif	0	0	0	8
 12009	Neurogourmet	realbrain.gif	0	0	0	9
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia:
Context: https://kolmafia.us/threads/fullness-tracking-in-zombie-slayer.29381/

If anyone wants to add that test I alluded to (or something substantially similar to it), I'm open to it. I just figured I'd put this PR out there before going to bed.

Even just a CustomScriptTest that runs that little ASH snippet I posted in the thread would probably catch this sort of issue in the future.

midgleyc (approved):
This is fine, consistency test can go in as a follow-up.

PR #2061, opened 2023-11-20
Fix Nuclear Autumn stomach size

soolar:
Currently mafia thinks that the fullness limit in Nuclear Autumn starts at 5.

This PR corrects that to the real size, which is 3.

I've tested this locally, let me know if this isn't actually the best way to go about fixing this though.


soolar:
Ah, looks like I have a test to fix too, my mistake. On it.

soolar:
Hmm, the test that's failing now seems totally unrelated.

midgleyc (approved):
Thanks!

The failing test passes on at least one OS, so looks like a race rather than a logic error.

jaadams5:
> Thanks!
> 
> The failing test passes on at least one OS, so looks like a race rather than a logic error.

While I agree in general, since we moved to GitHub I have had at least one OS specific test failure in a PR that really was an incorrect implementation.  I think I was dealing with CR, LF, EOL in an OS specific way that worked on Windows but not on *nix until I fixed my code.

PR #2062, opened 2023-11-22
desert beach is accessible in KoE, Shadow Rifts are not

Malibu-Stacey:
See https://kol.coldfront.net/thekolwiki/index.php/Kingdom_of_Exploathing/Map
It's there, always has been since the path was live.
The blame for those lines is weird for me. It says they were last touched 11 years ago which is impossible as KoE is only 4 years old.
Lines 1150-1157 in KoLAdventure.java already handle the differences between the regular Desert Beach and the KoE "Fragment" of Desert Beach from what I can see (correct me if I am wrong).

Shadow Rifts have been confirmed to exist in the Right Side of the Tracks and Mt McLargeHuge (possibly other places too) in the KoE path but don't allow you to adventure in them for reasons known only to the KoL dev team.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -886,7 +886,7 @@ public class KoLAdventure implements Comparable<KoLAdventure>, Runnable {
         // need payphone in inventory or to be in ASoL
         return false;
       }
-      
+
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
I will let someone else push, or at least wait a bit.

midgleyc:
I think it's fine; my concern would be it getting fixed on KoL end, but we can always revert.

PR #2063, opened 2023-11-23
Do not try to fold january tote Bag in BHY

gausie:


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -2164,8 +2164,13 @@ void enumerateEquipment(EquipScope equipScope, int maxPrice, PriceLevel priceLev
           // If we only need as many fold items as we have, then we can
           // count them against the items we need to pass through
           FoldGroup group = ItemDatabase.getFoldGroup(item.getName());
+
           int foldItemsNeeded = 0;
           if (group != null && Preferences.getBoolean("maximizerFoldables")) {
+            if (KoLCharacter.inBeecore() && KoLCharacter.hasBeeosity(group.getHead())) {
```
midgleyc, line 2170: this would prevent you from folding `metallic foil radar dish` into `metallic foil cat ears` because `metallic foil bow` contains a `b`?

gausie: Yeah I know. I think this situation is better than the alternative so I made the PR. I made this change locally but don't have the time to do a proper fix. We can close this PR if you like

midgleyc: I think if the aim is just to fix the tote you could just target it exclusively -- as in, if the head is the tote and you're in beecore, don't try to fold?

PR #2076, opened 2023-12-03
Move static block into DailyLimitType.

heeheehee-kolmafia:
We're not always initializing this map in time for when it's first used. Notably, during login, if a user has a mood that references a skill with a daily limit, then we may try to read DailyLimitType.dailyLimits before it's been populated.

Even then, I don't trust the initialization order to be correct, so I've further moved some logic into DailyLimitDatabase.reset() directly.

context: https://kolmafia.us/threads/missing-tab.29415/


jaadams5 (approved):
Trusting that it solves the problem...

PR #2079, opened 2023-12-04
Fix fixie bug in modifiers.txt

birkett83:
There's a weird bug where the ash expression:

$item[fixed-gear bicycle].effect_modifier("effect")

evaluates to the surprising value $effect[stats.enq]

I think someone misunderstood what the Effect: keyword does.


jaadams5:
The someone who perhaps misunderstood did so in 2015 :-)

That said I'm not sure your fix is correct.  I wonder if the correct fix would be to add the word Familiar before the word Effect?

I also see this as an opportunity for a test that fails before the fix and passes afterwards.  Perhaps you might consider writing it?

The previous test failure was not of your making.

midgleyc:
All the other Mini-Hipster items have the same `Effect` issue (in the modifiers file at least).

birkett83:
> That said I'm not sure your fix is correct. I wonder if the correct fix would be to add the word Familiar before the word Effect?

The Familiar Effect: field appears on hats, pants and familiar items. On hats it describes Mad Hatrack effects; on pants it describes Fancypants Scarecrow effects. On these items the format of the Familiar Effect: field appears consistent and my Java is not good and I don't know the codebase but it certainly looks like it's parsing those values in kolmafia/persistence/ModifierDatabase.java

example:

`Item    turtle wax helmet       Hot Resistance: +1, Familiar Effect: "0.5xPotato, 0.5xFairy, cap 10"`

The Familiar Effect: field also appears (sometimes) in the Familiar equipment section, e.g:

`Item    God Lobster's Scepter   Leprechaun: 0.5, Fairy: 0.5, Equips On: "God Lobster", Familiar Effect: "small fairy/lep"`

The only code I've found that looks at the Familiar Effect: field on familiar equipment is in kolmafia/swingui/widget/ListCellRendererFactory.java where it appears to be treated as a comment field and in  kolmafia/persistence/ModifierDatabase.java where these fields are excluded from some kind of validation because they contain "freeform text"

I don't know Java well enough to confirm if this is ever used, but neither the gear changer UI and the item manager UI seem to display the text from this Familiar Effect: field anywhere.

I believe these fields should all be removed, not just on the mini hipster gear but on all the familiar equipment - but this is all a bit beyond me. For now, changing those fields to Familiar Effect: fields instead is better than nothing.

jaadams5:
Thank you.  I was just speculating based upon pattern recognition - two pieces of familiar equipment that have effects so the syntax might be the same.  But since the effects act on different things, maybe not.  There is probably more work to be done to understand what is happening because if I am trying to maximize stats I might want my familiar to equip the bicycle and I'm pretty sure that won't happen after your change.  (It might not be happening now in which case there is code that needs to be changed, not necessarily data).

birkett83:
> if I am trying to maximize stats I might want my familiar to equip the bicycle and I'm pretty sure that won't happen after your change. (It might not be happening now in which case there is code that needs to be changed, not necessarily data).

`Item    fixed-gear bicycle      Experience: +3, Effect: "stats", Equips On: "Mini-Hipster"`

I think it's safe to assume that the maximiser parses the consistently formatted `Experience:` field that contains a specific numeric value rather than try to interpret the inconsistently formatted text field that just says "stats".

birkett83:
Tests are failing but I don't know why. The test runner says `See the report at: file:///home/runner/work/kolmafia/kolmafia/build/reports/tests/test/index.html` but that is presumably local to the machine running the test and I don't know how to access it. Can anyone help?

jaadams5 (approved):

src/data/modifiers.txt
```diff
@@ -4350,7 +4350,7 @@ Item	iced-out bling	Familiar Weight: +5
 Item	icicle katana	Familiar Weight: +5
 # imitation whetstone: Makes Imitation Crab claws sharper
 Item	immense ballet shoes	Familiar Weight: +5
-Item	ironic moustache	Familiar Weight: +10, Effect: "+10 lbs."
+Item	ironic moustache	Familiar Weight: +10, Familiar Effect: "+10 lbs", Equips On: "Mini-Hipster"
```
midgleyc, line 4353: This one doesn't need the "Equips On" field, because `src/data/familiars.txt` specifies that the Mini-Hipster's default familiar equipment is the ironic moustache. The "Equips On" field is only required for /extra/ familiar equipment.

birkett83: Thanks, I'll remove it

midgleyc (approved):
Thanks!

PR #2086, opened 2023-12-06
Circadian Rhythms fix/tweak and new spleen descriptions

Malibu-Stacey:
Tested Circadian Rhythms by casting it on a monster & then running away from the combat & then shrugging the effect. My adventures the next day were one more than what `ash (40 + my_adventures() + numeric_modifier("adventures"))` expected so it looks like you get +1 charge just from casting it on the monster regardless of what you do in the combat (I have not tested rollover runaway/Using the Force however, I just used a Snokebomb to free run).

Also noticed the new spleen items showed up as "Unspaded" in the item manager so I added some text for them.

The Deadly Lampshade I noticed purely by accident as I ran `ash foreach ef in $effects[] { if (ef.song) { print(ef); } }` because I was writing some code to handle shrugging songs & wanted to check I hadn't missed anything important & saw it in the list.


midgleyc (approved):
Thanks!

PR #2087, opened 2023-12-06
Add can_faxbot(MONSTER, FAXBOT)

Veracity0:
```
> ash can_faxbot($monster[black crayon constellation])

Returned: true
```

The code has this comment:

```
      // Don't check if bot is online til we can do it safely for large numbers of can_faxbot
      // requests
      // if ( !FaxRequestFrame.isBotOnline( botName ) )
      // {
      //	continue;
      // }
```

I don't know what that means.

I added a two-argument version of ```can_faxbot```:

```
> ash can_faxbot($monster[black crayon constellation], "CheeseFax")

Returned: true

> ash can_faxbot($monster[black crayon constellation], "EasyFax")

Returned: false
```

So it is only CheeseFax which offers that monster.

```
> ash is_online("CheeseFax")

Returned: false

> ash is_online("EasyFax")

Returned: true
```

CheeseFax is offline, so the one argument version of ```can_faxbot``` is lying.

I don't understand that comment - but if a script is willing to know the names of individual faxbots, it can handle this situation.

In favor of that, there are two versions of the ```faxbot``` function:

```
boolean faxbot(MONSTER);
boolean faxbot(MONSTER, FAXBOT);
```

which actually tries to receive a fax - which, as coded, depends on chat being open.


src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5232,6 +5235,42 @@ public static Value can_faxbot(ScriptRuntime controller, final Value arg) {
     return DataTypes.FALSE_VALUE;
   }
 
+  public static Value can_faxbot(ScriptRuntime controller, final Value arg1, final Value arg2) {
```
midgleyc, line 5238: This function shares most of its code with the one above: could you rewrite it so one calls the other (e.g. with null as the final value), or do the Value parsing first, and then have them both call `can_faxbot(MonsterData monster, String faxbot)`, one with a null faxbot?

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5232,6 +5235,42 @@ public static Value can_faxbot(ScriptRuntime controller, final Value arg) {
...
+      }
+
+      Monster monsterObject = bot.getMonsterByActualName(actualName);
+      if (monsterObject == null) {
+        continue;
+      }
+
+      // Don't check if bot is online til we can do it safely for large numbers of can_faxbot
+      // requests
+      // if ( !FaxRequestFrame.isBotOnline( botName ) )
```
midgleyc, line 5265: I feel like this might have been added when minimizing server hits was more important than it is now: I think KoL's servers should easily be able to handle this.

midgleyc (changes requested):

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
jaadams5: Darzil commented out the code in 2014 but the "Don't check..." was added by @heeheehee-kolmafia in 2021.

Veracity0: I re-enabled it - and called the KoLmafia.isPlayerOnline primitive, rather than the FaxRequestFrame method which sets the status method in that frame - and get this now.
```
> ash can_faxbot($monster[knob goblin embezzler])

Configuring faxable monsters.
Configuring Easyfax (2504737)
Configuring CheeseFax (3038166)
Faxable monster lists fetched.
Returned: true

> ash can_faxbot($monster[knob goblin embezzler], "EasyFax")

Returned: true

> ash can_faxbot($monster[knob goblin embezzler], "CheeseFax")

Returned: false

> ash can_faxbot($monster[Black Crayon Constellation])

Returned: false

> ash can_faxbot($monster[Black Crayon Constellation], "EasyFax")

Returned: false

> ash can_faxbot($monster[Black Crayon Constellation], "CheeseFax")

Returned: false
```

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
Veracity0: Sure. I did the latter.

midgleyc (approved):
LGTM

PR #2090, opened 2023-12-08
Store permability and max level for skills in datafile

gausie:
Moving out some per-skill data from switch statements into the datafiles by establishing a tokenized attribute column in `classskills.txt`. It feels like the right place to store this sort of information, but it also helps with generating the base data for tools like my own [greenbox](https://greenbox.loathers.net)


Veracity0:
Now that my personal projects - BeachComber - have slowed down, I can review (and write) other PRs.

Would you please include some text explaining what you are trying to do?

Please don't make me study the code to figure it out.

Thank you.

(I will study code - but only if I understand the intent.)

gausie:
> Now that my personal projects - BeachComber - have slowed down, I can review (and write) other PRs.
> 
> Would you please include some text explaining what you are trying to do?
> 
> Please don't make me study the code to figure it out.
> 
> Thank you.
> 
> (I will study code - but only if I understand the intent.)

Sorry - added a description!

midgleyc:
You've changed which skills are marked as permable.

Skills which were formerly permable and now are not:
* Chronic Indigestion
* Replica Emotionally Chipped

The former should still be permable, the latter is correctly not.

Some skills which auto-permed were previously marked as not permable, but are now marked as permable. I think given the inconsistency the new way is better.
* Acquire Rhinestones
* Budget Conscious
* Carol of the Bulls
* Carol of the Hells
* Carol of the Thrills
* Chip on your Shoulder
* Drinking to Drink
* Gallapagosian Mating Call
* Get Big
* Inscrutable Gaze
* Love Mixology
* Paul's Passionate Pop Song
* Pirate Bellow
* Request Sandwich
* Thick-Skinned
* Toggle Optimality

Also, the max level is not exposed anywhere externally to scripts -- it's just for programs which manually download the data file?
* Transcendent Olfaction

gausie:
> You've changed which skills are marked as permable.
> 
> Skills which were formerly permable and now are not:
> 
>     * Chronic Indigestion
> 
>     * Replica Emotionally Chipped
> 
> 
> The former should still be permable, the latter is correctly not.

That was not intended!

> Some skills which auto-permed were previously marked as not permable, but are now marked as permable. I think given the inconsistency the new way is better.
> 

Yes I agree

> 
> Also, the max level is not exposed anywhere externally to scripts -- it's just for programs which manually download the data file?

I replaced some in-code use with the new data driven data. But yeah it could be exposed to scripts if we wanted.

midgleyc (approved):
I think this is fine.

I checked `test newskill X` and it appears to work how it did before.

It might be nice to wait for the other devs to confirm reasonableness.

PR #2099, opened 2023-12-15
Result of a tool I made today to automatically make patches from session log readouts when new content arrives

gausie:
This is almost all automated (putting the monsters in a nicely named and positioned block in monsters.txt is the only thing I did manually, as well as rearrange some of the blocks in there).

I am not 100% sure that this is all kosher so please give it a look over. If it works well, I can put the code up for others to use in future. I wrote it in TypeScript.

### DIFF
--- a/src/data/equipment.txt
+++ b/src/data/equipment.txt
@@ -73,9 +73,10 @@ cool iron helmet	250	Mox: 200
 cornuthaum	85	Mox: 27
 corrosive cowl	200	Mox: 150
 Covers-Your-Head	300	Mox: 250
-crappy Mer-kin mask	200	Mox: 85
 cr&ecirc;epy mask	20	none
+crappy Mer-kin mask	200	Mox: 85
 Crimbo hat	100	Mox: 35
+Crimbuccaneer tricorn	50	none
 Crown of the Goblin King	70	Mox: 20
 Crown of Thrones	100	none
 crown-shaped beanie	200	Mox: 200
@@ -111,6 +112,8 @@ eldritch hat	125	Mox: 47
 eldritch scanner	10	Mus: 10
 electrician's hardhat	200	Mus: 85
 eleven-gallon hat	220	none
+Elf Guard patrol cap	50	none
+Elf Guard red and white beret	100	none
 enchanted eyepatch	35	Mox: 2
 enchantlers	115	Mox: 42
 extra-large palm-frond toupee	150	Mox: 60
@@ -223,8 +226,8 @@ Lens of Violence	200	Mys: 125
 lihc face	90	Mox: 30
 linoleum helmet turtle	100	Mox: 35
 literal bucket hat	100	Mus: 35
-loofah lumberjack's hat	50	none
 longhaired hippy wig	200	Mox: 85
+loofah lumberjack's hat	50	none
 lynyrdskin cap	170	Mox: 70
 MAHI fez	200	Mox: 85
 maiden wig	35	Mox: 2
@@ -287,14 +290,14 @@ oversized skullcap	20	none
 oxygenated eggnog helmet	100	none
 pail	30	none
 panhandle panhandling hat	150	Mus: 175
-paraffin pith helmet	50	none
-pentagram bandana	50	none
 paperclip turban	50	none
 papier-mitre	100	Mox: 20
+paraffin pith helmet	50	none
 parasitic headgnawer	30	none
 party hat	30	none
 passable elf mask	10	none
 pentacorn hat	40	Mox: 5
+pentagram bandana	50	none
 pickelhaube	75	Mox: 22
 pig-iron helm	90	Mox: 30
 pimpin' meat hat	55	Mox: 12
@@ -362,10 +365,10 @@ spiky turtle helmet	130	Mox: 50
 sponge helmet	200	Mox: 85
 spooky hockey mask	100	Mox: 35
 Spooky Putty mitre	100	none
+stained glass stetson	120	Mys: 40
 stainless steel skullcap	90	Mox: 30
 star hat	165	Mox: 67
 stone baseball cap	195	Mox: 82
-stained glass stetson	120	Mys: 40
 straw hat	200	Mox: 85
 sucker kabuto	50	none
 sugar chapeau	100	none
@@ -498,6 +501,7 @@ corroded breeches	200	Mox: 150
 crappy Mer-kin tailpiece	200	Mox: 85
 Crimbo pants	100	Mox: 35
 Crimbo stockings	200	Mox: 75
+Crimbuccaneer breeches	50	none
 Crimbylow-rise jeans	125	Mox: 15
 crotchety pants	75	Mox: 22
 cursed breeches	200	Mox: 85
@@ -506,9 +510,9 @@ Danglin' Chad's loincloth	200	Mox: 75
 demonskin trousers	70	Mox: 20
 depleted Grimacite shinguards	100	Mox: 35
 designer sweatpants	100	none
-discarded swimming trunks	100	none
 dice-print pajama pants	10	none
 Dinsey's pants	200	Mox: 100
+discarded swimming trunks	100	none
 distressed denim pants	170	Mox: 70
 double-ice britches	70	Mox: 20
 drafty drawers	225	Mox: 200
@@ -526,6 +530,7 @@ El Vibrato leg guards	200	Mox: 200
 eldritch pants	125	Mox: 47
 electric pants	100	Mox: 30
 electronic dulcimer pants	10	none
+Elf Guard hotpants	50	none
 exo-servo leg braces	10	none
 extremely skinny jeans	200	Mox: 200
 fancy party pants	75	Mox: 10
@@ -583,8 +588,8 @@ Krakrox's Loincloth	125	Mus: 30
 lava-proof pants	100	Mox: 35
 Leapin' Trousers	12	none
 leather chaps	160	Mox: 65
-leg-mounted Trainbots	50	none
 Lederhosen of the Night	125	Mox: 30
+leg-mounted Trainbots	50	none
 leggings of the Spider Queen	200	Mox: 40
 lemon drop trousers	50	Mox: 10
 leotarrrd	90	Mox: 30
@@ -764,6 +769,7 @@ Ascension Souvenir T-Shirt	15	none
 ASCII shirt	90	Mus: 30
 astral shirt	50	none
 autumn sweater-weather sweater	50	none
+barnacle-encrusted sweater	50	none
 bat-ass leather jacket	180	Mus: 77
 BGE 'cuddly critter' shirt	30	none
 BGE 'ferocious fruit' shirt	30	none
@@ -833,6 +839,7 @@ icing poncho	80	Mus: 35
 junk-mail shirt	60	Mus: 15
 Jurassic Parka	100	none
 Kashmir sweater	155	Mus: 62
+Kelflar vest	50	none
 Kiss the Knob apron	10	none
 Knob Goblin elite shirt	35	Mus: 2
 KoL Con 13 T-shirt	50	none
@@ -849,7 +856,6 @@ midriff scrubs	160	Mus: 65
 miming shirt	50	none
 Moonthril Cuirass	100	Mus: 40
 Mr. Shirt	30	none
-trench coat	200	Mus: 100
 origami pasties	100	none
 Othello's military jacket	150	Mus: 40
 PARTY HARD T-shirt	10	none
@@ -924,6 +930,7 @@ aerogel accordion	50	Mox: 0	1-handed accordion
 airblaster gun	10	Mox: 0	1-handed pistol
 alarm accordion	180	Mox: 75	2-handed accordion
 amok putter	30	Mus: 0	2-handed club
+ancient Elf dessert spoon	100	Mys: 0	1-handed utensil
 ancient ice cream scoop	170	Mys: 70	1-handed utensil
 ancient stone fist	85	Mus: 27	1-handed club
 anger blaster	85	Mox: 30	1-handed forearm replacement
@@ -1090,6 +1097,8 @@ creepy-ass club	160	Mus: 65	2-handed club
 Crimbo sword	100	Mus: 35	2-handed sword
 Crimbo ukulele	80	Mox: 20	2-handed guitar
 Crimbomination Contraption	100	Mus: 0	2-handed club
+Crimbuccaneer squirtblunderbuss	50	Mox: 0	1-handed pistol
+Crimbuccaneer tavern swab	100	Mus: 0	2-handed polearm
 crowbar	140	Mus: 55	1-handed club
 crowbarrr	80	Mus: 25	2-handed staff
 crystalline reamer	75	Mys: 22	1-handed utensil
@@ -1149,6 +1158,7 @@ electric crutch	30	Mus: 0	1-handed staff
 elegant nightstick	90	Mus: 30	1-handed club
 elephant stinger	75	Mus: 22	2-handed spear
 elevenderizing hammer	115	Mus: 11	1-handed club
+Elf Guard mouthknife	50	Mus: 0	1-handed knife
 elf ploughshare	100	Mus: 0	1-handed sword
 elven tambourine	100	Mox: 0	1-handed rattle
 elven whittling knife	100	Mus: 35	1-handed knife
@@ -1291,10 +1301,10 @@ huge spoon	90	Mys: 30	1-handed utensil
 ice nine	100	Mox: 0	1-handed pistol
 ice sickle	100	Mus: 0	1-handed knife
 icy-hot katana	90	Mus: 30	2-handed sword
-industrial fire extinguisher	100	Mus: 0	1-handed extinguisher
 iFlail	100	Mox: 35	1-handed flail
 immense cyborg hand	150	Mus: 60	1-handed club
 impromptu torch	100	Mus: 0	1-handed club
+industrial fire extinguisher	100	Mus: 0	1-handed extinguisher
 infernal fife	55	Mox: 12	2-handed flute
 infernal toilet brush	100	Mus: 35	1-handed club
 inflatable baseball bat	50	Mus: 0	2-handed club
@@ -1516,6 +1526,7 @@ Saucepanic	100	Mys: 0	1-handed saucepan
 savory crossbow	80	Mox: 25	2-handed crossbow
 savory staff	80	Mus: 25	2-handed staff
 savory sword	80	Mus: 25	1-handed sword
+sawed-off blunderbuss	50	Mox: 0	1-handed pistol
 scepter of the Skeleton Lord	300	Mys: 50	2-handed staff
 scorpion whip	20	Mus: 0	1-handed whip
 scratch 'n' sniff crossbow	100	Mox: 0	1-handed crossbow
@@ -1535,6 +1546,7 @@ Shakespeare's Sister's Accordion	100	Mox: 0	1-handed accordion
 sharpened spoon	100	Mus: 20	1-handed knife
 Sheila Take a Crossbow	100	Mox: 0	1-handed crossbow
 shiny butcherknife	60	Mys: 15	1-handed utensil
+shipwright's hammer	50	Mus: 0	1-handed club
 shooting morning star	100	Mus: 0	1-handed flail
 short-handled mop	165	Mus: 67	1-handed polearm
 shotgun	60	Mox: 0	1-handed
@@ -1730,6 +1742,7 @@ warbear oil pan	50	Mys: 0	1-handed saucepan
 weeping willow wand	100	Mus: 0	1-handed wand
 weighted paperclip chain	40	Mus: 0	2-handed flail
 western-style skinning knife	40	Mus: 5	1-handed knife
+whalegun	100	Mox: 0	2-handed rifle
 White Dragon Fang	200	Mus: 200	1-handed sword
 white sword	40	Mus: 5	1-handed sword
 white whip	10	Mus: 0	1-handed whip
@@ -1762,10 +1775,9 @@ Zombo's shoulder blade	200	Mus: 200	2-handed sword
 
 # Off-hand Items section of equipment.txt
 
-&quot;reusable&quot; grocery bag	0	none
 'WILL WORK FOR BOOZE' sign	150	Mus: 175
+&quot;reusable&quot; grocery bag	0	none
 1-ball	60	Mys: 15
-17-ball	200	none
 2-ball	60	Mys: 15
 3-ball	60	Mys: 15
 4-ball	60	Mys: 15
@@ -1773,13 +1785,14 @@ Zombo's shoulder blade	200	Mus: 200	2-handed sword
 6-ball	60	Mys: 15
 7-ball	60	Mys: 15
 9-ball	60	Mys: 15
+17-ball	200	none
 A Light that Never Goes Out	100	none
 Abracandalabra	100	none
 Abuela Crimbo's special magnet	0	none
-Adventurer bobblehead	100	none
 Abuela Crimbo's special magnet	10	none
 Abyssal ember	100	Mus: 35
 adhesive tape dolly	0	none
+Adventurer bobblehead	100	none
 aerogel attache case	50	none
 ancient calendar	70	Mus: 20	shield
 ancient hot dog wrapper	200	Mys: 85
@@ -1807,8 +1820,8 @@ ball-in-a-cup	20	none
 balloon shield	50	Mus: 10	shield
 barrel lid	150	none	shield
 barskin buckler	25	none	shield
-basic meat foon	50	Mys: 10
 basaltamander buckler	240	Mus: 40	shield
+basic meat foon	50	Mys: 10
 battered hubcap	220	Mus: 65	shield
 beach ball marble	209	Mys: 85
 beetle antenna	0	none
@@ -1906,6 +1919,7 @@ Dungeon Fist gauntlet	100	Mus: 35
 Dyspepsi-Cola shield	70	none	shield
 easter egg balloon	100	Mys: 35
 eelskin shield	200	Mus: 85	shield
+Elf Guard clipboard	0	none
 Elizabeth's Dollie	120	Mys: 45
 Ellsbury's skull	100	Mys: 35
 enchanted brass knuckles	50	Mys: 10
@@ -1965,6 +1979,7 @@ green LavaCo Lamp&trade;	100	Mox: 35
 green peawee marble	201	Mys: 85
 Grey Guanon	60	Mys: 15
 grisly shield	200	Mus: 150	shield
+gunwale	0	none	shield
 Guzzlr souvenir stein	100	none
 Half a Purse	100	none
 hand of Mr. Cards	10	none
@@ -2039,8 +2054,8 @@ magic lamp	30	none
 magical ice cubes	30	none
 makeshift castanets	160	Mox: 150
 maple magnet	100	none
-marinara jug	160	Mys: 150
 marble mignonette bowl	120	Mys: 40
+marinara jug	160	Mys: 150
 marionette	110	Mys: 40
 marionette collective	110	Mys: 40
 martini dregs	50	none
@@ -2051,8 +2066,8 @@ Mer-kin roundshield	215	Mus: 85	shield
 Mer-kin stopwatch	200	Mys: 85
 Mer-kin takebag	200	Mys: 85
 mesquito proboscis	0	none
-meteorite guard	150	none	shield
 meteorb	100	none
+meteorite guard	150	none	shield
 Microplushie: Bropane	5	none
 Microplushie: Dorkonide	5	none
 Microplushie: Ermahgerdic Acid	5	none
@@ -2199,8 +2214,8 @@ sock monkey	110	Mys: 40
 Solstice Shield	0	none
 sommelier's towel	150	Mys: 60
 sonar fishfinder	10	none
-Space Tourist palmdoctor	50	none
 space shield	150	none	shield
+Space Tourist palmdoctor	50	none
 spaghetti-box banjo	160	Mys: 150
 Spanish fly trap	10	none
 spant shield	100	Mus: 20	shield
@@ -2237,6 +2252,7 @@ stuffed frozen gravy fairy	10	none
 stuffed ghuol whelp	10	none
 stuffed hand turkey	10	none
 stuffed key	10	none
+stuffed kraken	0	none
 stuffed MagiMechTech MicroMechaMech	10	none
 stuffed martini	10	none
 stuffed Meat	10	none
@@ -2327,6 +2343,7 @@ zoological sample kit	10	none
 # Accessories section of equipment.txt
 
 &quot;I survived Y2K&quot; T-Shirt	100	none
+&quot;I Voted!&quot; sticker	10	none
 acid-squirting flower	50	Mys: 10
 actual reality goggles	10	none
 actual skeleton key	10	none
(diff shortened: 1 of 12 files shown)


src/data/fullness.txt
```diff
@@ -8,8 +8,8 @@
 [glitch season reward name]	0	0		0	0	0	0	Inedible, starts a fight with %monster%
 &quot;caramel&quot; orange	1	1	awesome	4	0	0	0	gain 10 turns of a random positive effect
 &quot;meat&quot; stick	1	5	good	3	15-25	0	0	50 Empty Inside (+50 Food Drop, +50 Booze Drop)
-20-lb can of rice and beans	15	1	good	45	0	0	0	100 Stuffed (+100 HP)
 3vi1 pr0n m4nic0tti	4	3	awesome	11-16	0	0	60-63	10 A Little Bit Evil (+2 ML)
+20-lb can of rice and beans	15	1	good	45	0	0	0	100 Stuffed (+100 HP)
```
midgleyc, line 12: 2 should probably come before 3?

src/data/inebriety.txt
```diff
@@ -5,8 +5,8 @@
 # Drink	Inebriety	Level Req	quality	Adv	Musc	Myst	Moxie
 
 [experimental crimbo booze]	3	1	good	0	0	0	0	Random
-1950 Vampire Vintner wine	1	1	awesome	5	0	0	0	Unspaded, WINE
 5-hour acrimony	5	1	awesome	0	0	0	0	+10 PvP fights
+1950 Vampire Vintner wine	1	1	awesome	5	0	0	0	Unspaded, WINE
```
midgleyc, line 9: same here, I think it's sorting starting numbers as numbers instead of alphabetically?

src/data/fullness.txt
```diff
@@ -877,6 +877,7 @@ spooky stir-fry	2	3	good	4-7	0	0	14-18
 Spooky Surprise Egg	1	1	crappy	1	0	0	0	series 2 tiny plastic
 square CRIMBCOOKIE	2	1	awesome	8-10	0	0	0	10 Square to be Hip (+10% mox)
 squirmy violent party snack	1	1	good	2-4	2-20	2-20	2-20	10 Broken Dancing (Legendary Beat quest)
+St. Pete's sneaky smoothie	1	1	EPIC	5-7	0	0	20-40	30 Gritty Soul (+5 Moxie, +50% Moxie, +5 Spooky Resistance)
```
midgleyc, line 880: here it's sorting capital S after lowercase s, but elsewhere you don't do that.

I would rather we say it sorts case sensitively because it makes it easier for me (I can just run `:sort` and be done with it), but that's not how it currently is in many places

src/data/monsters.txt
```diff
@@ -2333,35 +2333,17 @@ Boltsmann security robot	2402	jetsecurity1.gif	Scale: 9 Cap: 1000 Init: -10000 P
...
+Crimbuccaneer carpenter	2432	cbcarpenter.gif
+Crimbuccaneer freebooter	2430	cbboots.gif
+Crimbuccaneer military school dropout	2405	cbdropout.gif
+Crimbuccaneer new recruit	2395	cbrecruit.gif
+Crimbuccaneer privateer	2436	cbprivateer.gif
+Crimbuccaneer scrimshander	2431	cbscrimshander.gif
+Crimbuccaneer whalehunter	2428	cbwhaler.gif
+Elf Guard conscript	2392	egconscript.gif
+Elf Guard convict	2393	egconvict.gif
+Elf Guard private	2394	egprivate.gif
```
midgleyc, line 2346: we normally sort these by ID, instead of alphabetically

src/data/equipment.txt
```diff
@@ -890,6 +896,7 @@ tailored vest	100	none
 The Emperor's new shirt	10	none
 Thinknerd T-Shirt	10	none
 trench coat	200	Mus: 100
+trench coat	200	Mus: 100
```
midgleyc, line 899: this is duplicated

midgleyc:
failing test is telling you you've forgotten to add the familiar... which is right!

src/data/fullness.txt
gausie: It's natural sorted - 20 is coming after 3

src/data/inebriety.txt
gausie: Yes. I did that on purpose - elsewhere in the data we expect it to be so.

src/data/monsters.txt
gausie: It's really mixed, but the vast majority of the page is alphabetically sorted (see Regular monsters section)

src/data/equipment.txt
gausie: Neat - this change has uncovered that existing issue!

jaadams5:
I like the idea.  My fantasy solution would have been to have the "discovery files" written in a format that they could just be appended to the KoLmafia data file.  If there were conventions about sorting then perhaps spotlessapply could be extended to deal with that.

This approach seems to accomplish the same thing although there may be a role for spotless if we understand how to extend/configure.

I think the tool should be included in KoLmafia so that anyone can run it (and thus make a PR) but that can wait.  Thank you.

src/data/fullness.txt
gausie: everywhere I am sorting text in this i am doing case insensitive sort. I think it makes more sense that way and it does seem to be precedent

gausie:
> I like the idea. My fantasy solution would have been to have the "discovery files" written in a format that they could just be appended to the KoLmafia data file. If there were conventions about sorting then perhaps spotlessapply could be extended to deal with that.
> 
> This approach seems to accomplish the same thing although there may be a role for spotless if we understand how to extend/configure.
> 
> I think the tool should be included in KoLmafia so that anyone can run it (and thus make a PR) but that can wait. Thank you.

I'll make it a repo in the kolmafia org

src/data/items.txt
```diff
@@ -11445,13 +11445,13 @@
...
-11421
-11422
-11423
-11424
-11425
-11426
+11420	Crimbuccaneer tavern swab	587885280	crimbucmop.gif	weapon	t	0
+11421	Elf Guard hangover cure	520317664	egcure.gif	usable	t	0
+11422	old-school pirate grog	716801313	crimbuctankard.gif	drink, mix	t	0
+11423	grog nuts	690314926	crimbucnuts.gif	food	t	0
```
midgleyc, line 11451: by description, this is `food, combat` -- looks like a Mafia bug, it doesn't parse `waffle` correctly either.

src/data/modifiers.txt
```diff
@@ -9039,6 +9047,7 @@ Item	green hamhock	Effect: "Just the Best Anapests", Effect Duration: 50
 Item	green-iced sweet roll	Effect: "Sprinkle Sense", Effect Duration: 40
 Item	gregarious ghostling	Free Pull
 Item	grinning ghostling	Free Pull
+# grog nuts
```
midgleyc, line 9050: while we add in blank sections for equipment / effects without effects, I don't think we tend to do the same for food.

I don't think we should at all, really.

src/data/modifiers.txt
```diff
@@ -2053,8 +2057,8 @@ Item	weeping willow wand	Spell Damage Percent: +100, Mysticality: +25, Experienc
 # weeping willow wand: Lets you blast your foes with tears
 Item	weighted paperclip chain	Weapon Damage: +12
 Item	western-style skinning knife	Weapon Damage: +5, Hat Drop: +10
-# whalegun: +10 Pirate Warfare Effectiveness
 Item	whalegun	Monster Level: +10, Weapon Damage: +100
+# whalegun: +10 Pirate Warfare Effectiveness
```
midgleyc, line 2061: so this is a bit where I presume the script is changing things on a second run (compare sawed-off blunderbuss above).

On the first run, they go in in the order they are in the description.

On second run, they go in in real / comment order.

src/data/monsters.txt
midgleyc: the regular monsters section is alphabetical, but new areas tend to be id. I like it because it tells you whether you've missed one (e.g., path bosses).

It's okay to keep this how you like it.

midgleyc (approved):
Also wants:
* adventures.txt
* combats.txt
* npcstores.txt
* zapping.txt

I think this is the first case where we have a zone where you get different monsters depending on item, but the zone itself is one snarfblat. Not sure how we're handling that. For The Exploaded Battlefield we just stuck all the monsters in, which seems a reasonable move.

Still, a good start.

Also, haxxi has guessed `mineral,object,hard` for the anchor tags, which you can presumably confirm?

src/data/items.txt
midgleyc: okay, we don't detect `robin's egg` either, so I guess `waffle` was just manually added

src/data/modifiers.txt
```diff
@@ -4105,53 +4145,54 @@ Item	anniversary pewter cape	Muscle: +1, Mysticality: +1, Moxie: +1
...
 Item	dice-shaped backpack	Random Monster Modifiers: +1
 Item	Drapes-You-Regally	Meat Drop: +20, Booze Drop: +20, Hat Drop: +20, Class: "Disco Bandit"
-# Drip harness: Allows you to enter The Drip
 Item	Drip harness	Muscle Limit: 100, Mysticality Limit: 100, Moxie Limit: 100
+# Drip harness: Allows you to enter The Drip
 Item	Drunkula's cape	Adventures: +3, Candy Drop: +20, Accessory Drop: +20, Class: "Sauceror"
 Item	Duke Vampire's regal cloak	Moxie: +15, Moxie Percent: +25, Initiative: +50
 Item	elf army poncho	Spooky Resistance: +2, Stench Resistance: +2, Hot Resistance: +2, Cold Resistance: +2, Sleaze Resistance: +2, Lasts Until Rollover
-# fiberglass frock: Makes it so dealing Cold damage causes ongoing additional Cold damage
+Item	Elf Guard SCUBA tank	Muscle Percent: -25, Mysticality Percent: -25, Moxie Percent: -25, Maximum HP: -100, Maximum MP: -100, Adventure Underwater, Adventure Underwater
```
midgleyc, line 4181: while what Mafia returns, Adventure Underwater should not be duplicated.

It's duplicated because the description looks like
```
Lets you breathe underwater
Lets you Breathe Underwater
```
which is presumably a KoL bug

src/data/modifiers.txt
```diff
@@ -7995,18 +8036,22 @@ Effect	Zomg WTF	Monster Level: +30
 
 # Passive Skills section of modifiers.txt
 
-Skill	20/20 Vision	Item Drop: +10
+
+
 Skill	5-D Earning Potential	Meat Drop: +20
+Skill	20/20 Vision	Item Drop: +10
+# A modifier is twice as potent when positive as when negative
```
midgleyc, line 8043: comment that should not be reshuffled

src/data/modifiers.txt
```diff
@@ -8036,6 +8081,7 @@ Skill	AUTOEXEC.BAT	Stench Damage: +15
 Skill	Autovampirism Routines	Spooky Resistance: +3
 Skill	Backwards Knees	Initiative: +20
 Skill	Bad Breath	Stench Damage: +3
+# Bad Moon passive skills: the Seven Deadly Sins.
```
midgleyc, line 8084: comment that should not be reshuffled

src/data/modifiers.txt
```diff
@@ -8133,18 +8178,20 @@ Skill	Drinking to Drink	Liver Capacity: +1
 # Drippy Eye-Beetle: It is awful!
 # Drippy Eye-Sprout: It tickles!
 # Drippy Eye-Stone: It is irritating!
+# Each skill has one positive modifier and one negative modifier.
```
midgleyc, line 8181: comment that should not be reshuffled

src/data/modifiers.txt
```diff
@@ -8189,20 +8237,25 @@ Skill	Gift of the Maid	Meat Drop: +50
 Skill	Gift of the Priest	Spell Damage: +40
 Skill	Gift of the Scribe	Experience (Mysticality): +5
 Skill	Gluteus Maximus	Muscle: +25
+Skill	Gluttony	Experience: -2, Stomach Capacity: +2
+# Gluttony duplicates Opossum's food gains
```
midgleyc, line 8241: comment that should not be reshuffled

src/data/modifiers.txt
```diff
@@ -8189,20 +8237,25 @@ Skill	Gift of the Maid	Meat Drop: +50
...
+Skill	Greed	Meat Drop: +50, Item Drop: -15
+# Grogged For Battle
+Effect	Groggipopped	Stench Resistance: +1, Sleaze Resistance: +1
 # Hammer Throw: Level 1: Briefly stun your foe and reduce their stats by 10%
 # Hammer Throw: Level 2: Reduction increased to 15%
 # Hammer Throw: Level 3: Reduction increased to 20%
 Skill	Hamstrings	Initiative: +10
-Skill	[15025]Hard Drinker	Liver Capacity: +10
 Skill	[18008]Hard Drinker	Liver Capacity: +5
+Skill	[15025]Hard Drinker	Liver Capacity: +10
```
midgleyc, line 8258: what lead to this sort? It's an odd one. 5 before 10?

src/data/modifiers.txt
```diff
@@ -8291,6 +8347,7 @@ Skill	Metallic Skin	Cold Resistance: +2, Hot Resistance: +2, Sleaze Resistance:
 Skill	Microburner	Hot Damage: +5
 Skill	Microweave	Cold Resistance: +2
 Skill	Middle Fingers	Pickpocket Chance: +30
+Effect	Militarily Minted	Damage Reduction: 3
```
midgleyc, line 8350: effect in the middle of the skills

src/data/modifiers.txt
```diff
@@ -8266,15 +8321,16 @@ Skill	Limberger Limberness	Initiative: +75
...
 # Louder Bellows: (bellow delevels by 30%, and also stuns the monster for one round)
+# Louder Bellows: You can bellow louder
 Skill	Lower Hypothalamus	Cold Resistance: +1
 Skill	Lubricant Layer	Sleaze Resistance: +2
 Skill	Lucky Brooch	Item Drop: +30
 # Lucky Buckle: Find more coins after combat
 Skill	Lucky Insignia	Item Drop: +30
 Skill	Lucky Pin	Item Drop: +30
 Skill	Lunch Like a King	Stomach Capacity: +5
+Skill	Lust	Initiative: +50, Weapon Damage: -5, Spell Damage: -5
```
midgleyc, line 8333: a single comment for all bad moon skills: the bad moon skills were formerly separate.

src/data/modifiers.txt
```diff
@@ -8189,20 +8237,25 @@ Skill	Gift of the Maid	Meat Drop: +50
...
 Skill	Gnomish Hardigness	Maximum HP Percent: +5
 Skill	Golgi Apparatus	Absorb Stats: +25
 # Good Singing Voice: Doubles duration of Boris songs
 Skill	Gorgonzola's Guile	Item Drop: +25
 # Gourmand: Increases adventures from food
-# Gravitational Compression: +100% Item Drops from Monsters (Scales with your Mysticality)
 Skill	Gravitational Compression	Item Drop: [min(ceil(basemys/5),100)]
+# Gravitational Compression: +100% Item Drops from Monsters (Scales with your Mysticality)
+Skill	Greed	Meat Drop: +50, Item Drop: -15
+# Grogged For Battle
```
midgleyc, line 8251: a curiosity: why have some picked up the "+X Y Warfare Effectiveness" and some not?

src/data/modifiers.txt
```diff
@@ -8411,16 +8471,19 @@ Skill	Shifted About	Adventures: +5
 Skill	Shiver Reflex	Cold Resistance: +2
 Skill	Shivering Monkey Technique	Initiative: +15
 Skill	Shock-Absorbing Joints	Damage Reduction: 10
-# Sick Pythons: Attacks Miss Less Often
 Skill	Sick Pythons	Weapon Damage: +20
+# Sick Pythons: Attacks Miss Less Often
 Skill	Sinister Charm	Floor Buffed Moxie: "Mysticality", Maximum HP: -15
+# Skill	Emotionally Chipped: Gives you access to emotions
+# Skill Dead Nostrils: Stay fresh a little longer
```
midgleyc, line 8478: these shouldn't start with "Skill" but pre-existing problem

src/data/modifiers.txt
```diff
@@ -8469,6 +8534,7 @@ Skill	Temporal Hyperextension	Adventures: +10
 Skill	The Concept of Property	Meat Drop: +30
 Skill	The Long View	Adventures: +3
 # The Most Important Meal: Gain a bunch more Adventures and stats from the first thing you eat each day
+# The skills positively or negatively modify seven aspects of gameplay
```
midgleyc, line 8537: comment that should not be reshuffled

src/data/modifiers.txt
```diff
@@ -10907,19 +10980,20 @@ Item	zombie hollandaise	Effect: "Chef Brains-are-dee", Effect Duration: 30
 
 # Everything Else section of modifiers.txt
 
+
```
midgleyc, line 10983: double space

src/data/items.txt
midgleyc: added #2100 to improve detection

src/data/modifiers.txt
gausie: I agree. I'll remove it

src/data/modifiers.txt
gausie: Yeah - I totally ignore id when sorting. I want to keep it this way haha.

src/data/monsters.txt
Veracity0: I like new areas to be sorted by ID since, as Chris said, it's a clue that you might be missing some.

src/data/modifiers.txt
gausie: No idea! We should write some tests and see

src/data/modifiers.txt
gausie: I haven't fixed this but I've re-run the script from scratch. I'll try to work this out when this stuff is merged

src/data/modifiers.txt
```diff
@@ -2424,58 +2449,53 @@ Item	Mixed Garbanzos and Chickpeas	Maximum HP: [5*(1+skill(Beanweaver))], Maximu
...
-# Operation Patriot Shield: Can be Thrown in Combat
-# Seal Clubber:
-# Operation Patriot Shield	HP Regen Min: 10, HP Regen Max: 12, Weapon Damage: +15, Damage Reduction: 1
-# Turtle Tamer:
-# Operation Patriot Shield	HP Regen Min: 10, HP Regen Max: 12, Familiar Weight: +5
-# Disco Bandit
-# Operation Patriot Shield: Makes you a Crack Shot
-# Operation Patriot Shield	Ranged Damage: +20
-# Accordion Thief
-# Operation Patriot Shield: Lets You Sing More Proudly
```
midgleyc, line 2454: previously, these were class-based (as seen here); they have been reshuffled to be alphabetical.

If the script can be changed to assume that a block of comments should exist in the order they're present, that would be nice.

src/data/modifiers.txt
```diff
@@ -2684,88 +2707,93 @@ Item	Unkillable Skeleton's shield	Spooky Resistance: +3, Stench Resistance: +3,
...
 Item	Yorick	Experience: +2, HP Regen Min: 5, HP Regen Max: 10, MP Regen Min: 5, MP Regen Max: 10, Item Drop: +15
 Item	Zombo's shield	Damage Absorption: +50, Spooky Damage: +30, Muscle Percent: +20, Class: "Turtle Tamer", Damage Reduction: 15
-# zoological sample kit: Increases the yield of Spacegate zoology operations
 Item	zoological sample kit	Lasts Until Rollover
+# zoological sample kit: Increases the yield of Spacegate zoology operations
 
 # Accessories section of modifiers.txt
 
+# * This enchantment works even when this item is not equipped
+# * This enchantment works even when this item is not equipped
```
midgleyc, line 2741: These comments have come from elsewhere.

src/data/modifiers.txt
```diff
@@ -2684,88 +2707,93 @@ Item	Unkillable Skeleton's shield	Spooky Resistance: +3, Stench Resistance: +3,
...
 Item	&quot;I survived Y2K&quot; T-Shirt	Maximum HP: +2, Maximum MP: +2, Adventures: +2, Experience (Muscle): +2, Experience (Mysticality): +2, Experience (Moxie): +2, Single Equip
+Item	&quot;I Voted!&quot; sticker	Muscle Percent: +5, Mysticality Percent: +5, Moxie Percent: +5, Lasts Until Rollover
+# #13
```
midgleyc, line 2744: This comment has come from elsewhere.

src/data/modifiers.txt
```diff
@@ -2684,88 +2707,93 @@ Item	Unkillable Skeleton's shield	Spooky Resistance: +3, Stench Resistance: +3,
...
 Item	acid-squirting flower	Single Equip, Thorns: 1
-# actual reality goggles: Reveals peoples' true nature
+# actual max is 3, fudged so that (min+max)/2 matches the actual average
```
midgleyc, line 2746: comment on ring of half-assed regeneration

src/data/modifiers.txt
```diff
@@ -3676,47 +3711,47 @@ Item	stuffed carpenter	Synergetic
...
 Item	Time Lord Badge of Honor	Muscle Limit: 50, Mysticality Limit: 50, Moxie Limit: 50
-# time-twitching toolbelt: +60% Item Drops in the Time-Twitching Tower
-# time-twitching toolbelt: Reveals various Time-Forgotten Secrets
-# time-twitching toolbelt: (also in the Time-Twitching Tower)
+# Time Lord Badge of Honor: Proves your worth to the Time Lords
 # time-twitching toolbelt	Muscle: +15, Mysticality: +15, Moxie: +15, Single Equip, Item Drop: [60*zone(Twitch)], Free Pull: [pref(timeTowerAvailable,true)]
 Item	time-twitching toolbelt	Muscle: +15, Mysticality: +15, Moxie: +15, Single Equip, Item Drop: [60*zone(Twitch)], Variable
+# time-twitching toolbelt: (also in the Time-Twitching Tower)
+# time-twitching toolbelt: +60% Item Drops in the Time-Twitching Tower
+# time-twitching toolbelt: Reveals various Time-Forgotten Secrets
```
midgleyc, line 3754: this is another block where alphabetizing the existing comments leads to a strange read.

src/data/modifiers.txt
```diff
@@ -7995,18 +8031,22 @@ Effect	Zomg WTF	Monster Level: +30
 
 # Passive Skills section of modifiers.txt
 
-Skill	20/20 Vision	Item Drop: +10
+
```
midgleyc, line 8035: two extra lines!

src/data/modifiers.txt
gausie: Ok I have an idea

src/data/modifiers.txt
gausie: It doesn't seem to be a comment. I wonder if it is from [Item #13](https://kol.coldfront.net/thekolwiki/index.php/Item_13). I'll remove it.

src/data/modifiers.txt
gausie: Hmm

src/data/modifiers.txt
gausie: I don't really understand this, it was alphabetical before this change wasn't it? Is this just a really old relic or am I misreading. Tempted to remove.

gausie: Oh wait I see.

gausie: The problem is that blocks of comments aren't always right connected. This sucks.

src/data/modifiers.txt
midgleyc: a naive rule like "brackets always go to the end" makes it moderately readable

src/data/modifiers.txt
midgleyc: you could prefix all of these by "Operation Patriot Shield" to make aligning them easy

otherwise I think we are in
```
# gausie-special-tool-ignore
<stuff>
# no-gausie-special-tool-ignore
```
territory.

src/data/modifiers.txt
```diff
@@ -155,8 +155,8 @@ Item	crappy Mer-kin mask	Stench Resistance: +3, Item Drop: -25, Initiative: -50,
 Item	Crimbo hat	Spooky Resistance: +1, Familiar Effect: "1.75xVolley, 1.75xLep, cap 25"
 Item	Crown of the Goblin King	Muscle: +10, Familiar Effect: "2xBarrr, 1xGhuol, cap 17"
 # Crown of Thrones: Put a Familiar In It!
-# Crown of Thrones: That Familiar gains 1 Experience per Fight
 Item	Crown of Thrones	Softcore Only
+# Crown of Thrones: That Familiar gains 1 Experience per Fight
```
midgleyc, line 159: comments are better but it now goes comment -> item -> other comments, which is probably not what you meant

src/data/modifiers.txt
```diff
@@ -1055,10 +1061,10 @@ Item	smooth velvet bra	Maximum HP: +20, HP Regen Min: 5, HP Regen Max: 10
 Item	smooth velvet shirt	Disco Style: +1
 Item	snailmail hauberk	Damage Absorption: +60, HP Regen Min: 6, HP Regen Max: 14, MP Regen Min: 6, MP Regen Max: 14
 Item	snakeskin jacket	Moxie: +15, HP Regen Min: 5, HP Regen Max: 10, MP Regen Min: 5, MP Regen Max: 10
-# Sneaky Pete's leather jacket: The Audience Loves/Hates You Even More!
 Item	Sneaky Pete's leather jacket	Moxie: +10, Meat Drop: +20, Pickpocket Chance: +30, Adventures: +4, Free Pull
 # Sneaky Pete's leather jacket (collar popped): The Audience Loves/Hates You Even More!
 Item	Sneaky Pete's leather jacket (collar popped)	Moxie: +10, Initiative: +20, Monster Level: +30, PvP Fights: +4, Free Pull
+# Sneaky Pete's leather jacket: The Audience Loves/Hates You Even More!
```
midgleyc, line 1067: this is below collar popped instead of above the jacket itself.

src/data/modifiers.txt
```diff
@@ -2900,7 +2923,8 @@ Item	chocolate pocketwatch	Maximum MP: +20, Spell Damage: +15, Spell Critical Pe
 Item	chocolate rabbit's foot	Item Drop: +5, Meat Drop: +5, Candy Drop: +20
 Item	Choker of the Ultragoth	Muscle: +3, Mysticality: +3, Moxie: +3, Spooky Spell Damage: +40, Single Equip
 Item	Cincho de Mayo	Moxie Percent: +25, HP Regen Min: 10, HP Regen Max: 20, MP Regen Min: 10, MP Regen Max: 20, Booze Drop: +30, Item Drop: +15
-# Claw of the Infernal Seal: 5 Additional Seal Summons Per Day<super>*</super>
+# Claw of the Infernal Seal: 5 Additional Seal Summons Per Day
+# Claw of the Infernal Seal: This enchantment works even when this item is not equipped
 # * This enchantment works even when this item is not equipped
```
midgleyc, line 2928: spare line (duplicate with the above)

src/data/modifiers.txt
```diff
@@ -4138,6 +4172,8 @@ Item	Drip harness	Muscle Limit: 100, Mysticality Limit: 100, Moxie Limit: 100
 Item	Drunkula's cape	Adventures: +3, Candy Drop: +20, Accessory Drop: +20, Class: "Sauceror"
 Item	Duke Vampire's regal cloak	Moxie: +15, Moxie Percent: +25, Initiative: +50
 Item	elf army poncho	Spooky Resistance: +2, Stench Resistance: +2, Hot Resistance: +2, Cold Resistance: +2, Sleaze Resistance: +2, Lasts Until Rollover
+Item	Elf Guard SCUBA tank	Muscle Percent: -25, Mysticality Percent: -25, Moxie Percent: -25, Maximum HP: -100, Maximum MP: -100, Adventure Underwater
+Item	Elf Guard SCUBA tank	Muscle Percent: -25, Mysticality Percent: -25, Moxie Percent: -25, Maximum HP: -100, Maximum MP: -100, Adventure Underwater, Adventure Underwater
```
midgleyc, line 4176: now the tank is also duplicated :P

src/data/modifiers.txt
```diff
@@ -8291,6 +8341,7 @@ Skill	Metallic Skin	Cold Resistance: +2, Hot Resistance: +2, Sleaze Resistance:
 Skill	Microburner	Hot Damage: +5
 Skill	Microweave	Cold Resistance: +2
 Skill	Middle Fingers	Pickpocket Chance: +30
+Effect	Militarily Minted	Damage Reduction: 3
```
midgleyc, line 8344: this is up in effects, but also here for some reason?

src/data/modifiers.txt
```diff
@@ -1601,11 +1613,13 @@ Item	Maxwell's Silver Hammer	Mysticality: +10, Weapon Damage: +10
...
 # Mayor Ghost's gavel: This gavel, specifically.
 Item	Mayor Ghost's gavel	Critical Hit Percent: +10, Weapon Damage: +30
-# Meat Tenderizer is Murder: (+% to Muscle Based on Smithsness)
 # Meat Tenderizer is Murder: 25% chance to gain an extra gallon of Fury when you gain a gallon of Fury.
+# Meat Tenderizer is Murder: (+% to Muscle Based on Smithsness)
 Item	Meat Tenderizer is Murder	Weapon Damage Percent: +20, Smithsness: +5, Class: "Seal Clubber", Muscle Percent: [2*K]
 Item	Meatcleaver	Meat Drop: +35, Weapon Damage Percent: +100
 Item	meatspout staff	Mysticality: +7, Adventures: +3, Meat Drop: +15
+# mecenary pistol: Allows the Use of Specialty Ammunition
+# mecenary rifle: Allows the Use of Specialty Ammunition
```
midgleyc, line 1622: these are misspelled, which is why they've moved

src/data/modifiers.txt
```diff
@@ -8197,12 +8232,14 @@ Skill	Gorgonzola's Guile	Item Drop: +25
 # Gourmand: Increases adventures from food
 # Gravitational Compression: +100% Item Drops from Monsters (Scales with your Mysticality)
 Skill	Gravitational Compression	Item Drop: [min(ceil(basemys/5),100)]
+Skill	Greed	Meat Drop: +50, Item Drop: -15
+Effect	Groggipopped	Stench Resistance: +1, Sleaze Resistance: +1
```
midgleyc, line 8236: another effect in the middle of the skills

src/data/modifiers.txt
```diff
@@ -12037,27 +12097,27 @@ Item	record of menacing silence	Skill: "Silent Slice"
...
-# recovered elf toothbrush: Turn it in at the E. R. F. camp for
+# recovered elf smartphone: Turn it in at the E. R. F. camp for
 # recovered elf toothbrush: 1 Crimbo Credit
-# recovered elf underpants: Turn it in at the E. R. F. camp for
+# recovered elf toothbrush: Turn it in at the E. R. F. camp for
 # recovered elf underpants: 2 Crimbo Credits
-# recovered elf wallet: Turn it in at the E. R. F. camp for
+# recovered elf underpants: Turn it in at the E. R. F. camp for
 # recovered elf wallet: 3 Crimbo Credits
+# recovered elf wallet: Turn it in at the E. R. F. camp for
```
midgleyc, line 12115: wants rearranging back

midgleyc (approved):
Thanks

PR #2112, opened 2023-12-18
feat: highlight own store in relay browser

midgleyc:
Highlight your own store in blue in the relay browser's mall search.

This is nice if you want to see at a glance whether you're selling something, and whether it's likely to sell / too low / whatever.

### DIFF
--- a/test/net/sourceforge/kolmafia/webui/MallSearchDecoratorTest.java
+++ b/test/net/sourceforge/kolmafia/webui/MallSearchDecoratorTest.java
@@ -1,6 +1,7 @@
 package net.sourceforge.kolmafia.webui;
 
 import static internal.helpers.Player.withProperty;
+import static internal.helpers.Player.withUserId;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/webui/MallSearchDecoratorTest.java
```diff
@@ -45,7 +46,28 @@ public void testForbiddenStoresDecorated() {
...
+
+      assertThat(buffer.toString(), equalTo(expectedString));
+    }
+  }
+
+  @Test
+  public void testOwnStoreDecorated() {
+    String testString =
+        """
+        <tr class="graybelow" id="stock_11111_99"><td style="border:none"></td><td class="small store"><a class="nounder" href="mallstore.php?whichstore=11111&searchitem=99&searchprice=5000"><b>Generic Store</b></a>&nbsp;&nbsp;&nbsp;</td><td class="small stock">1</td><td class="small">&nbsp;</td><td class="small price"><a class="nounder" href="mallstore.php?whichstore=11111&searchitem=99&searchprice=5000">5,000&nbsp;Meat</a></td><td class="buyers" valign="center">&nbsp;</td></tr><tr class="graybelow" id="stock_12345_99"><td style="border:none"></td><td class="small store"><a class="nounder" href="mallstore.php?whichstore=12345&searchitem=99&searchprice=5000"><b>Generic Store</b></a>&nbsp;&nbsp;&nbsp;</td><td class="small stock">1</td><td class="small">&nbsp;</td><td class="small price"><a class="nounder" href="mallstore.php?whichstore=12345&searchitem=99&searchprice=5000">5,000&nbsp;Meat</a></td><td class="buyers" valign="center">&nbsp;</td></tr>""";
```
jaadams5, line 59: I might have just looked for certain strings in the response but I think that is "different" and not "better".  Comment not a blocker.

jaadams5 (approved):

test/net/sourceforge/kolmafia/webui/MallSearchDecoratorTest.java
midgleyc: sorry, I don't understand the comment. What is different or better?

jaadams5: You have two blocks of HTML which make the test code less readable IMO.  You pass one to get decorated and then compare the entire result for the test.  I might consider testing that the result contains "style="background-image:linear-gradient(to right, rgba(0,0,255,0), lightblue);" title="This is your store." or even  more simply contains "lightblue" and "This is your store".  I might even go further and read the response to be decorated from a file so there is no HTML as strings in Java.  But except for human readability I can't claim one way is better and it could be argued that readability is not a concern for test code.  Sorry I wasn't clearer.

midgleyc: oh right, I just copied the existing test.

I don't have a strong opinion on this, so if you also don't have a strong opinion, we can keep it like it is, and I'll change it if it starts failing at some point?

PR #2115, opened 2023-12-20
Add link to onlyfax faxbot

libraryaddict:
https://github.com/loathers/OnlyFax

This is a FaxBot written by Irrat and helped by Gausie in its final approach to work as a loathers project.


gausie:
To add some more information here, cheesefax has been down for a few months and having only one faxbot isn't great. This bot also allows requests with disambiguated monster ids, which is unique to this bot as far as I know

jaadams5 (approved):

PR #2119, opened 2023-12-21
feat: track prank card tracking

midgleyc:
I think a `TrackManager` in the vein of `BanishManager` might be a nice idea with the proliferation of trackers.

Not right now though.


jaadams5 (approved):
Is a test that covers the six new lines worth doing?

midgleyc:
I don't think so, personally.

PR #2127, opened 2023-12-24
feat: track manager

midgleyc:
As promised in #2119, a TrackManager in the vein of BanishManager.

Needs more testing but ready for review at least.

Mafia thought Ew, The Humanity boosted humanoids; wiki thinks it's dudes; I think it's dudes? I've set it as dudes.

Added Perceive Soul.

I don't know when `ResultProcessor.processResults` is called, so I'm not sure this is the right placed to track obtaining the +phylum effects. Perhaps it should be something tied onto the ApiRequest, when we get the effects the character has -- I think we do that every turn.

`consequences.txt` can pull out tracked monsters from item descriptions, but I don't make any effort to apply that to the TrackManager (similar to how BanishManager doesn't try to handle that).


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -9124,7 +9126,7 @@ private static void payActionCost(final String responseText) {
           // Your mind fills with it essence. You... know it. With a capital K.
           if (responseText.contains("Your mind fills") || jiggleSuccess) {
             Preferences.increment("_jiggleCream", 1);
-            Preferences.setString("_jiggleCreamedMonster", monsterName);
+            TrackManager.trackCurrentMonster(Tracker.CREAM_JIGGLE);
```
gausie, line 9129: Why `trackCurrentMonster` here and `trackMonster(monsterName...` later?

gausie (commented):
Great idea, great implementation. I figure you're well ahead of me but some more test coverage here would be 👌

src/net/sourceforge/kolmafia/request/FightRequest.java
midgleyc: I think the standard is that it's `trackCurrentMonster` for skills and `trackMonster` for items.

Now as to /why/ it's like that, that's another question, but I mimicked whatever banish did.

gausie (approved):

PR #2128, opened 2023-12-24
Trap Null dereference

jaadams5:
As part of my quest to find and eliminate tests that fail intermittently I noticed that some of the failing tests generated DEBUG logs that were otherwise ignored by the test code.  KoLAdventureValidationTest is one such test.  Since a static analysis of the code also suggests ArmoryAndLeggeryRequest could try and access Null I decided to check for null there rather than try and modify the test so that the test data did not trigger the exception.


jaadams5:
Merged into https://github.com/kolmafia/kolmafia/pull/2133 so closing this.

PR #2129, opened 2023-12-24
Support for Crimbo23 Elf and Pirate Armory coinmasters

Veracity0:
1) All Crimbo23 coinmasters need registerRequest support
2) Elf Guard temporary (permanent) tattoo has () in its name so does not parse correctly into an AdventureResult
3) Add support for Elf and Pirate Armory coinmasters

Those last two are new, since they use the same action "buyitem" - for both buying and selling.

- I could (should) add a test, since I have saved HTML for selling to the Pirate Armory
- I could (should) move the shared code for this new thing into CoinMasterRequest

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -640,6 +640,17 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
   }
 
   public static AdventureResult parseItem(final String s, final boolean pseudoAllowed) {
+    // Certain items fountain parentheses. Appending (COUNT) to such is problematic.
+    // For now, if we have an exact match for an item name, use it with a count of 1.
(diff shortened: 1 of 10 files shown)


src/net/sourceforge/kolmafia/AdventureResult.java
```diff
@@ -640,6 +640,17 @@ public static AdventureResult parseResult(final String s) {
   }
 
   public static AdventureResult parseItem(final String s, final boolean pseudoAllowed) {
+    // Certain items fountain parentheses. Appending (COUNT) to such is problematic.
```
jaadams5, line 643: "fountain" ?  If that's not a typo then I don't understand the idiom :-)  Not blocking.

jaadams5 (approved):

src/net/sourceforge/kolmafia/AdventureResult.java
Veracity0: Contain.

PR #2130, opened 2023-12-25
Make daily deeds advs readable, add potted plant advs text

libraryaddict:
Sometimes you can tell when someone just didn't feel like thinking about it.
I don't blame them, but it was too much for my poor brain to handle so I simplified the code.

* Added <Number> potted plant to the daily deeds display
* Added Some tests
* Minor rework of the code for what text is displayed so that it is easier to read, and add onto in the future.


midgleyc (dismissed):
You'll need to fix the spotless thing, but looks good other than that.

jaadams5 (approved):

PR #2133, opened 2023-12-27
Exceptions during testing

jaadams5:
My hypothesis concerning failing tests in https://github.com/kolmafia/kolmafia/pull/1889 includes the possibility that test code that generates an exception is sometimes seen as a test error.  So I modified the cleanup code so that a debug log would be preserved if one was generated.  I then tracked down the tests that generated the log.  I'm not convinced that the mafia code that throws the exception is always correct but maybe someone wiser than me will offer an opinion.

I saw the ArmoryAndLeggery exception while running mafia.  As best as I can reconstruct it, the automaton returned and the script that picked the next location picked one of the Crimbo2023 locations that KoL had offered,  But the character had not visited CrimboTown yet that day and so some verification failed.  https://github.com/kolmafia/kolmafia/pull/2128 also has this so if this PR gets traction before that one is approved then it will be closed.

The changes to ClearState* removed one instance of duplicated code and then made it so the debug logs would not be deleted.  That is useful for locally run tests.

CustomScriptTest is explicitly expecting an exception so deleting the log is reasonable.  Mafia throwing the exception is deliberate and desired since it is caused by programmer error.

The Automation and Automaton tests could be "wrong" in that they are not providing an expected response that KoL always provides or the code under test could need to change to handle something KoL might do.  @Veracity0 @midgleyc might wish to review as the original authors.

People who want to see the logs can either comment our the deletion code or just create an empty file in test/root named the same as today's debug log and run tests.  A log from running is attached.

[DEBUG_20231225.txt](https://github.com/kolmafia/kolmafia/files/13780112/DEBUG_20231225.txt)


jaadams5:
Upon further consideration I decided suppressing the debug log was shooting the canary.  So I reverted that code except for the test where generating the log was deliberate and desired behavior.  I also cleaned some lint from code I was already editing.  

I think I am done with this for the moment because a goal is to get this merged into https://github.com/kolmafia/kolmafia/pull/1889 so I have better tools to determine why tests fail there but not on main.

midgleyc (approved):
I think the debug log cleanup is a bit awkward, but while it's only one test there's no particular reason to extract it to an annotation / try-finally or anything.

jaadams5:
> I think the debug log cleanup is a bit awkward, but while it's only one test there's no particular reason to extract it to an annotation / try-finally or anything.

Thanks.  I agree it is awkward but maybe we reconsider when there is another test that expects a log.

PR #2136, opened 2023-12-28
Add missing Gnome adventure activation strings

Donald-Lafleur:
This is simply an addition (along with alphabetizing) of the missing Reagnimated Gnome adventure gain strings to catch when the Gnome's ability activates. This will allow the existing built in preference and Daily Deeds panel tracking to record a more accurate value for adventures gained.


midgleyc (approved):
Thanks!

Could you put them on the wiki too?

PR #2137, opened 2023-12-29
feat: crimbo 2023 standards

midgleyc:


Veracity0 (approved):
How many of the missing items are 2024 Standard rewards and such?
Which was my first thought when I saw this, until I realized that Abuela Crimbo sent us items at 9:04 this evening. 😀

midgleyc:
We don't know, because they're not public yet ;)

I think the current guess is that 19-35 is standard outfits and 40 & 41 are ioty and familiar equip.

PR #2142, opened 2023-12-31
Fix a breaking change in #2129

libraryaddict:
https://kolmafia.us/threads/unable-to-craft-staff-of-ed.29483/


midgleyc (approved):
thanks for the test!

PR #2144, opened 2024-01-01
Feat: Witchess Daily Puzzle Solver in Quests menu

Glitched94:
Adds an item to the Quests menu that will attempt to solve the user's 5 daily Witchess Puzzles if they haven't already.
States handled:

- If the user has already solved all 150 puzzles (Puzzle Champ bonus == 20)
- If the user has already solved today's puzzles (_witchessBuff = true)
- If the user has already solved _some_ of today's puzzles (Puzzle will have "Solved Today" text)
- If the user does not have a Witchess Set (Campground does not contain "chesstable.gif")

To-do: (Open to suggestions on how to do each of these things)
- Test cases
  - Some done: Tests if the html request is made and parsed correctly, and tests 1 of the 4 scenarios where the request isn't made because we're aware of the user being in an incompatible state. Includes the html for 15 unsolved and solved puzzles to test against. I've also been running it for myself during each leg of my looping, so can get 10 more puzzles to test the state of each day, but it seems unnecessary to have that many.
- Handling if we failed to solve the puzzle somehow
  - Technically handled, will show an error to the user that it failed, generally _trying_ to say why, and stop executing the script
- Improve the check for having Witchess
- Add gCLI command to run the solver.


src/data/witchess_solutions.txt
```diff
@@ -0,0 +1,151 @@
+# Puzzle Number	Solution
```
midgleyc, line 1: this file should have a version number at the top (i.e. the first line should be "1").

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
```diff
@@ -0,0 +1,132 @@
+package net.sourceforge.kolmafia.persistence;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.util.*;
```
midgleyc, line 5: we tend to put all the imports and not * them in src/

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
```diff
@@ -0,0 +1,132 @@
...
+        FileUtilities.getVersionedReader(
+            "witchess_solutions.txt", KoLConstants.WITCHESS_SOLUTIONS_VERSION)) {
+      String[] data;
+
+      while ((data = FileUtilities.readData(reader)) != null) {
+        if (data.length < 2) {
+          continue;
+        }
+
+        int puzzleId = StringUtilities.parseInt((data[0]));
```
midgleyc, line 37: double brackets you don't need, I think

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
```diff
@@ -0,0 +1,132 @@
...
+    based on the moves made. Since you move from one extreme of the grid to the opposite,
+    we can assume that the max grid size will always be the absolute value of the difference between
+    each move in the cardinal directions (up/down and left/right)
+
+    However, each move increments the grid position by 2, so we need to double the difference
+    to get the real size and coordinates.
+    */
+    Map<Character, Integer> moveCounts = new HashMap<>();
+    for (char move : moves) {
+      moveCounts.put(move, moveCounts.getOrDefault(move, 0) + 1);
```
midgleyc, line 97:
```suggestion
      moveCounts.merge(move, 1, Integer::sum);
```

src/net/sourceforge/kolmafia/swingui/menu/GlobalMenuBar.java
```diff
@@ -5,13 +5,7 @@
...
 import net.sourceforge.kolmafia.KoLmafia;
 import net.sourceforge.kolmafia.RequestThread;
-import net.sourceforge.kolmafia.session.GuildUnlockManager;
-import net.sourceforge.kolmafia.session.LeafletManager;
-import net.sourceforge.kolmafia.session.LogoutManager;
-import net.sourceforge.kolmafia.session.SorceressLairManager;
-import net.sourceforge.kolmafia.session.TavernManager;
-import net.sourceforge.kolmafia.session.TowerDoorManager;
-import net.sourceforge.kolmafia.session.WumpusManager;
+import net.sourceforge.kolmafia.session.*;
```
midgleyc, line 8: expand this again

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
```diff
@@ -0,0 +1,132 @@
...
+      int[] midpoint = calculateMidpoint(start, end);
+      path.add(midpoint[0] + "," + midpoint[1]);
+    }
+
+    List<String> sortedPath = new ArrayList<>(path);
+    sortedPath.sort(
+        Comparator.comparingInt((String s) -> Integer.parseInt(s.split(",")[0]))
+            .thenComparing(s -> Integer.parseInt(s.split(",")[1])));
+
+    return String.join("|", sortedPath);
```
midgleyc, line 126: it would be nice to have a test that gets the solution for a certain puzzle and confirms this generated string is as expected.

It would separately be nice to have some documentation at the top of this function as to what this generated string is supposed to contain, and how that relates to the puzzle itself.

Glitched94: I do have a txt file locally with the final GET request strings that I was planning to write a test with. I wasn't sure the best approach to reading from that .txt file without writing another Database class to do so, though, so opted to exclude it originally.

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
Glitched94: Apologies, apparently my IDE automatically changed some imports to .* when I had enough things imported from them, I'll manually add only the required imports.

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
midgleyc: you don't need all of them -- one would be enough, I think.

If you did want all of them (or many), I would use a `ParameterizedTest` with a `CsvSource` for the puzzle number and the expected string.

Glitched94: I've got 15 of them saved, I might grab some from my puzzles tomorrow since the ones I currently have are in the earlier ranges, and I think it'd be good to have later puzzles as well since the board can get oddly shaped. I'll pick probably 5 from a few different spots in the order.

Glitched94:
Just extended the `witchess` gCLI command to allow running the witchess solver. In order to prevent breakage in existing code, the command now accepts `witchess` or `witchess buff` to get the buff, and `witchess solve` to try and solve the player's daily puzzles.

src/net/sourceforge/kolmafia/persistence/WitchessSolutionDatabase.java
Glitched94: Alright, I created WitchessSolutionDatabaseTest.java and check that the generated URL for 10 puzzles match the network requests I grabbed when manually solving the puzzles.

midgleyc (approved):
Thanks!

I confirmed that running `witchess` alone gets you the buff.

While it would break backwards compatibility, I think there is an argument that `witchess` should either get the buff or solve the daily puzzles if you haven't done all of them, on the grounds that that's probably what you want to do in either case. This is also fine, though.

PR #2146, opened 2024-01-02
Fix left-hand man held carn plant free kill tracking

Donald-Lafleur:
Carnivorous potted plant kills weren't being tracked when it was held by a left-hand man.


midgleyc (approved):
Thanks!

PR #2147, opened 2024-01-02
feat: everything looks green color + noremove

midgleyc:
Do what we do for the other "Everything Looks" effects but for Green.


jaadams5 (approved):
I'll approve because you are following a standard implied by other code but Once Upon A Time KoLmafia was hoped to be friendly for color-blind users.  My recollection was that was done by "overloading" text attributes that could be rendered in gray scale - i.e. bold, shades of gray, italics etc. but I cannot recall any preferences or other accommodations.  

See https://kolmafia.us/threads/9673-colorize-things-in-the-item-manager-according-to-kol-specified-quality.7448/#post-56673 for some discussion.  It might be worth reopening the question - should KoLmafia be configurable for color-blind users or do we explicitly say we won't consistently support them?

jaadams5:
The design goals that were articulated by Hola (and some others) decades ago have lost relevance so I am also fine with waiting for complaints but figured it was worth noting.

PR #2150, opened 2024-01-03
feat: most Hobopolis boss consumables are now once/day

midgleyc:
Hodgman drops to come, apparently.

For daily limits, I copied fudge spork in putting them in both Eat and Use (or Drink and Use).


src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
```diff
@@ -737,6 +737,8 @@ public static final void parseConsumption(
 
           if (!responseText.contains("discard the no-longer-frosty")) {
             success = false;
+          } else {
```
heeheehee-kolmafia, line 740: Note that this (and the below handling for salad fork) doesn't handle the case where someone uses a frosty mug outside of Mafia. That's probably sufficient for now.

heeheehee-kolmafia (approved):

PR #2152, opened 2024-01-03
Last Hobopolis consumable daily limit

Donald-Lafleur:
Also added a check to set unset daily limits for salad fork and frosty mug when the eat/drink response indicates they were already used today.


src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
```diff
@@ -156,7 +156,8 @@ public static int maximumUses(
     var dailyLimit = DailyLimitDatabase.DailyLimitType.DRINK.getDailyLimit(itemId);
     if (dailyLimit != null) {
       var remaining = dailyLimit.getUsesRemaining();
-      if (remaining == 0) {
+      // Drinking only this we can't get ourselves overdrunk before we hit our daily cap
+      if (remaining * inebriety <= inebrietyLeft) {
```
midgleyc, line 160: sorry, I don't understand this bit. The previous code was saying that you can't use an item if you've hit the daily limit -- what is this doing? It seems to block you from using an item if you couldn't get overdrunk if you used all remaining uses, even if you're not doing that?

Donald-Lafleur: Before it returned 0 if the remaining amount today was 0, since that's the maximum. If you have 1 remaining and that remainder totals to less than or equal to your remaining liver space, then the most you can drink is however many are still available. Either the daily limit or your remaining liver is the deciding factor. If you can't get overdrunk with the remaining available to you (which is what the condition is), then the maximum you can drink is the remaining available to you, regardless of whether you are willing to overdrink.

In hindsight, I probably should have stored the calculated total drunkenness achievable by the passed in drink in a variable on line 158 to make the condition make more sense.

This change was because if you try to drink 2 jars of fermented pickle juice with 10+ liver available mafia will drink 1 and tell you that you can only drink 1 because of inebriety, instead of correctly reporting the reason as the 1/day limit.

Donald-Lafleur: The last part was the most important to me, since CONSUME doesn't check any of these new preferences and isn't written to handle using only 1 fork and 1 mug, and probably won't for at least a few days. That means for a fair number of people it might try to overuse the Hobopolis drops and fail saying the player doesn't have enough remaining organ space.

midgleyc: I feel like that should have already been handled by the `maxAvailable = remaining;` and `maxNumber = maxAvailable;` bits.

I feel like either this isn't the right place for this code, or the comment should be something like "// give correct reason if drinking enough of limited item to overdrink" but that seems to couple this logic to somewhere else in the code (that is, wherever else in the code is giving the overdrunk limit if you try to drink a limited-per-day item), you see? I'm not keen on this. I'd rather put this logic at the overdrunk place and keep this one at the zero limit.

I'll look into it more after work.

src/data/dailylimits.txt
```diff
@@ -185,6 +185,7 @@ Drink	cup of "tea"	_speakeasyDrinksDrunk	3
 Drink	Doc Clock's thyme cocktail	_docClocksThymeCocktailDrunk
 Drink	Flivver	_speakeasyDrinksDrunk	3
 Drink	Frosty's frosty mug	_frostyMugUsed
+Drink	Hodgman's blanket	_hodgmansBlanketDrunk
```
midgleyc, line 188: there's also tin cup of mulligan stew, which I had in #2151.

Either that one can go in, and you can make this PR just about the fork / mug, or you can add it here. Either or.

midgleyc: Probably easier to send that one in and remove it from here -- I hope that one will go in easier.

src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
midgleyc: is there a corresponding issue with sliders? That is, if you try to eat 2 sliders will Mafia tell you you're too full?

src/data/dailylimits.txt
Donald-Lafleur: Yours should go in for the Hodgman stuff, especially since you remembered both items. I'll (figure out how to) remove it from mine and reduce this to the checks for failed fork and mug uses. Thanks for the help focusing this one better.

midgleyc: alright, that one's gone in, so if you rebase yours you should be able to remove the blanket stuff.

You could also just put the fork / mug stuff in here, then I can approve and merge because I think that's fine, and then we can do the pickle juice overdrink thing as a follow-up to that?

src/net/sourceforge/kolmafia/request/DrinkItemRequest.java
midgleyc: The previous use of this early return was to avoid doing further computation if the player has already hit their daily limit (set the daily limit reason and return 0). I think this is reasonable.

This conversation seems to be about an extension to that: if we're drinking multiple items at the same time, and we're trying to drink more than the daily limit, that should be reported to the user.

I am confused as to why this isn't happening. This method, `maximumUses`, seems to tell you how many you can use at most of a given item, and I think should be correctly reporting 1 for the pickle juices. Some other method is setting the `limiter` incorrectly, and I think that should be fixed. Perhaps this method should brashly always set the limiter to the limit reason, even if `remaining` is greater than 0: but that isn't done for any other part of this function.

Donald-Lafleur: It reports the correct number, but the reason for the limitation defaults to inebriety in all cases except when you've eaten your limit of something, that is the 0 remaining case. I didn't see your message about sliders, but I can test those as well to see how food behaves, as well as voodoo snuff while we're at it.

I like the idea of reporting whether a user wants to drink over the daily limit, but inside this function we don't know if they want to drink more than the daily limit or exactly the daily limit or if the question is hypothetical for something like tourguide. Maybe we could handle a request to consume more than the daily limit value of something in a similar way to how an insufficient number of the item on hand is handled.

src/data/dailylimits.txt
Donald-Lafleur: I'm going to go ahead and close this one. It'll be a lot cleaner that way, since I can't figure out how to just rebase selections of my original branch in a way that makes any kind of sense while leaving the branch existing the whole time. I'll open another PR with the mug & fork bits this afternoon and we can work out how to handle the prioritize the limits situation after that.

Also, are you already working on chest mimic support, or do you know if someone else is?

midgleyc: I don't know of anyone working on the mimic.

Parsing the egg description should wait until CDM adds ids (to make that a lot easier), but things like the preferences for use can go in earlier.

PR #2154, opened 2024-01-04
Set daily limit pref for fork and mug by seeing use limit failure

Donald-Lafleur:
I tried to figure out how to write a few tests to ensure the use failure sets the preference, but I couldn't understand how to use a DrinkItemRequest to accomplish that as a two step process or find any examples to work from of other consumption helpers similar to mug. This allows mafia to notice when someone has used a mug or fork for the day that wasn't tracked.


midgleyc (approved):
Thanks!

For the test: probably by using the `FakeHttpClientBuilder` to give two responses: one for the mug use, and one for drinking something with the mug.

PR #2155, opened 2024-01-04
fix: bump skills version to 6

midgleyc:
This should have gone in with #2149 -- I changed the format, so I should also have bumped the version number.

### DIFF
--- a/test/net/sourceforge/kolmafia/DataFileMechanicsTest.java
+++ b/test/net/sourceforge/kolmafia/DataFileMechanicsTest.java
@@ -29,7 +29,7 @@ public class DataFileMechanicsTest {
         Arguments.of("buffbots.txt", 1, 3, 3),
         Arguments.of("cafe_booze.txt", 1, 2, 2),
         Arguments.of("cafe_food.txt", 1, 2, 2),
-        Arguments.of("classskills.txt", 4, 6, 7),
+        Arguments.of("classskills.txt", 5, 6, 7),
         Arguments.of("coinmasters.txt", 2, 4, 5),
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/DataFileMechanicsTest.java
```diff
@@ -29,7 +29,7 @@ public static Stream<Arguments> data() {
         Arguments.of("buffbots.txt", 1, 3, 3),
         Arguments.of("cafe_booze.txt", 1, 2, 2),
         Arguments.of("cafe_food.txt", 1, 2, 2),
-        Arguments.of("classskills.txt", 4, 6, 7),
+        Arguments.of("classskills.txt", 5, 6, 7),
```
jaadams5, line 32: Is this right?  I thought the arg was the version number of the version the test expected.  So it should be 6 now.  That raises the question - how long was this test expecting 4, getting version 5 and passing anyway?

jaadams5 (dismissed):
I'll approve but feel free to look closer at the test ;-)

test/net/sourceforge/kolmafia/DataFileMechanicsTest.java
midgleyc: ...good question

midgleyc: `getVersionedReader` does its best to return a reader, so it looks like it was reading the file on the disk, finding that the version didn't match, then getting the one from the jar and using that. So hopefully everything worked out anyway.

PR #2158, opened 2024-01-05
Some Coinmasters share the action for buying and selling

Veracity0:
The Elf and Pirate Armories have the same "action" for buying and selling.
The code to handle that is duplicated in the two CoinMasterRequest classes.
I want to move it into CoinMasterRequest.java to make it easier to handle similar future cases.

I started this on my desktop, but will be leaving town in an hour until the end of March. I will have my laptop - and plenty of time to work on it when I arrive in my destination in 10 days - but I need the text fixtures in a PR, rather than my desktop's local repository.

This is that PR.

### DIFF
--- a/test/net/sourceforge/kolmafia/request/CoinMasterRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/CoinMasterRequestTest.java
@@ -1,0 +1,75 @@
+package net.sourceforge.kolmafia.request;
+
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+
+public class CoinMasterRequestTest {
+  @BeforeAll
+  static void beforeAll() {
+    KoLCharacter.reset("CoinMasterRequestTest");
+  }
+
+  @BeforeEach
+  void beforeEach() {
+    Preferences.reset("CoinMasterRequestTest");
+  }
+
+  @Nested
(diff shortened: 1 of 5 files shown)


src/net/sourceforge/kolmafia/request/CoinMasterRequest.java
```diff
@@ -260,23 +260,56 @@ public static void parseResponse(
...
-      CoinMasterRequest.completePurchase(data, urlString);
-    } else if (sell != null
-        && action.equals(sell)
-        && (sellURL == null || shopId == null || sellURL.endsWith(shopId))
-        && !responseText.contains("You don't have that many")) {
-      CoinMasterRequest.completeSale(data, urlString);
+
+    if (buy == null && sell == null) {
+      // You can neither buy nor sell from this Coinmaster? Coding error!
+      return;
```
midgleyc, line 269: sounds like a good target for a consistency test.
```java
for (CoinmasterData coinmaster : CoinmasterRegistry.COINMASTERS) {
  assertTrue(coinmaster.getBuyAction != null || coinmaster.getSellAction() != null)
}
```

Veracity0: I added a consistency test. Turns out there is exactly one Coinmaster which has neither a buyAction nor a sellAction - AWOLQuartermasterRequest. The URL for buying something is "inv_use.php?whichitem=5116&ajax=1".

It also has buyItems. I made the test check for buyAction, and if it fails, for buyItems.

Perhaps my tests in CoinMasterRequest should be checking for something other than butAction, but considering the single Coinmaster which does NOT have that is way obsolete, I don't feel like trying to figure that out at this time.

I have a feeling that future Coinmasters will all be shop.php.

PR #2159, opened 2024-01-06
Add support for Javascript Promises

oxc:
Javascript promises require microtasks to be processed, which is only done by diverse evaluate functions, not when calling a function manually.

Process pending microtasks in JavascriptRuntime.execute(), and also resolve Promises returned by a script.

Furthermore, enable tracking of unhandled promises, and print them to the console. This is always most likely an error, and should be handled by and explicit empty catch() handler, if desired.

If added tests for several scenarios, in a separate commit to make the changes to the outputs/behavior visible.

UPDATE: Please note that this does not allow any kind of asynchronous programming, because there are no kolmafia (or rhino) methods that would create a non-blocking flow. It is still useful for using 3rd party libraries that use promises as only interface but also have non-async flows, and it reduces surprises when somebody does use promises for whatever reason and wonders why they are executing when called from the CLI but not in a script.


gausie (approved):
Thank you!

PR #2165, opened 2024-01-07
feat: add is_shruggable/is_removable script functions

oxc:
Add two RuntimeLibrary functions for checking whether an effect is shruggable or removable (UneffectRequest.isShruggable/isRemovable).


jaadams5 (approved):
The code is fine.  I would like someone else's opinion as to whether these functions are needed/useful/cannot be easily scripted.  I would also love to see some tests.  But if anyone thinks otherwise and wants to push this then I'm fine with that decision.

oxc:
The main reason I added isShruggable so I can safely "cli uneffect" an effect without risking kolmafia buying SGEEAs (if I haven't misunderstood something there...)

midgleyc:
If you've got a use for them, I think that's good enough.

I've definitely added functions that I think I'm the only person to use.

oxc:
This might be slightly related: https://github.com/loathers/libram/issues/512

It's all not a satisfying solution for the underlying problem, but better than nothing.

jaadams5:
OK.  Pushing.

PR #2168, opened 2024-01-09
Deadlock Detector

jaadams5:
Copied the deadlock detector from #1889

IMO it is worth preserving regardless of what happens with #1889.


midgleyc (approved):
Seems sensible.

PR #2171, opened 2024-01-10
feat: add preferences for extracting / donating mimic eggs

midgleyc:


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -10440,6 +10440,13 @@ private static void payActionCost(final String responseText) {
           skillSuccess = true;
         }
         break;
+
+      case SkillPool.LAY_AN_EGG:
+        if (responseText.contains("shudders and lays an egg") || skillSuccess) {
```
jaadams5, line 10445: Where was skillSuccess set prior to this being executed?

jaadams5:
The comments about skillSuccess are not blocking.  I just cannot figure out how control would get to the new code when skillSuccess was true but the response text was not present.  But FightRequest is not simple and idiomatically including skillSuccess when it is unnecessary isn't going to hurt anything.

jaadams5 (approved):

src/net/sourceforge/kolmafia/request/FightRequest.java
midgleyc: up top, if you run with haiku / anapests / in the deep machine tunnels

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -2602,4 +2602,17 @@ public void canTrackSurprisinglySweetStab() {
       }
     }
   }
+
+  @Test
+  public void canTrackMimicEggLay() {
+    var cleanups = new Cleanups(withProperty("_mimicEggsObtained", 0), withFight());
```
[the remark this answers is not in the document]
midgleyc: no, this would explode our tests without really offering any value, IMO

It would be nice to have a test with the failure text that confirms the property isn't updated, but it's not essential.

src/net/sourceforge/kolmafia/request/FightRequest.java
jaadams5: Thanks.  There's got to be a better way to organize 10K lines of code for human readability.

test/net/sourceforge/kolmafia/request/FightRequestTest.java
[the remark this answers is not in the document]
jaadams5: OK

PR #2174, opened 2024-01-12
Close http client @ 7k requests

libraryaddict:
This PR has a change seperated into https://github.com/kolmafia/kolmafia/pull/2176

This PR changes the http client limit from 9900 to 7000 which seems to have improved connection errors

### DIFF
--- a/src/net/sourceforge/kolmafia/utilities/ResettingHttpClient.java
+++ b/src/net/sourceforge/kolmafia/utilities/ResettingHttpClient.java
@@ -33,10 +33,10 @@ public class ResettingHttpClient {
 
   public <T> HttpResponse<T> send(HttpRequest req, HttpResponse.BodyHandler<T> handler)
       throws IOException, InterruptedException {
-    var resp = this.client.send(req, handler);
     if (clientRequestsSent.incrementAndGet() >= HTTP_CLIENT_REQUEST_LIMIT) {
       resetClient();
     }
-    return resp;
+
+    return this.client.send(req, handler);
   }
 }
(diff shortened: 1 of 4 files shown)


jaadams5:
I am concerned about adding code because of Java 21 when KoLmafia is "officially" at Java 17.  Some of the conversations about devs upgrading to 21 have included reports about gradle and some of our tools needing changes to work under Java 21.  I would prefer this be broken up into two PRs.  One just changes 10K to 7K because that seems to address the GOAWAY problem.  The other would contain the rest of the changes proposed because of Java 21.  That should be considered in the overall context of how, or even if, KoLmafia supports Java 21.   That would also allow for user testing of the proposed code on systems that only have Java 17.

src/net/sourceforge/kolmafia/utilities/ResettingHttpClient.java
```diff
@@ -29,34 +27,16 @@ public ResettingHttpClient(Supplier<HttpClient> createClient) {
...
-
-    // As we compile against an older jdk, we must use reflection to access the method
-    try {
-      Method method = Class.forName("java.net.http.HttpClient").getMethod("close");
-      method.invoke(httpClient);
-    } catch (Exception ex) {
-      // If this does error, we should log it as this means something has changed in java internals
-      StaticEntity.printStackTrace(ex);
-    }
+    return resp;
```
midgleyc, line 40: can you put the change
```
return this.client.send(req, handler);
```
in this PR as well? I think this is a correctness fix: if not, multiple threads can send requests before the request limit is checked. That isn't Java 21 specific.

midgleyc (approved):
Thanks!

PR #2175, opened 2024-01-12
Some test changes that facilitated debugging of tests.  (Was: Parking while some things get reconsidered.)

jaadams5:


gausie:
If you are parking this, you could consider having it as a branch on your own fork rather than as a draft PR? It's not good to have PRs that we train our eyes to glaze over :P

jaadams5:
Not being fluent in git the concept of "my" branch with branches is somewhat alien to me so I had not thought about it.  I will remember the possibility and see if I can make my tools do it :-)  I use Draft as a trigger that I am allowed to glaze over something but your point is noted.  Since everything effects tests only and are changes made in the quest to find why things fail in #1889 but not main, I am fine with undrafting and either committing or withdrawing.

test/net/sourceforge/kolmafia/AreaCombatDataTest.java
```diff
@@ -87,15 +87,17 @@ public class AreaCombatDataTest {
...
-    KoLCharacter.reset("fakeUserName");
-    Preferences.reset("fakeUsername");
+    KoLCharacter.reset(TESTUSERNAME);
+    Preferences.reset(TESTUSERNAME);
   }
 
   @AfterAll
   public static void deleteQueueFile() {
-    File queueF = new File(KoLConstants.DATA_LOCATION, "fakeusername_queue.ser");
+    File queueF = new File(KoLConstants.DATA_LOCATION, TESTUSERNAME + "_queue.ser");
```
midgleyc, line 100: previously, the username was lowercased. Now it's not. Desired?

jaadams5: I don't expect the casing to matter but to be safe...

jaadams5: Changed it.  I suspect it was lowercased because that was the case of the file on the disk when the cleanup was written.

midgleyc: Casing matters on non-Windows filesystems. If the casing is wrong, the file won't get deleted.

midgleyc: code is
```java
  public static String baseUserName(final String name) {
    return name == null || name.equals("")
        ? "GLOBAL"
        : StringUtilities.globalStringReplace(name.trim(), " ", "_").toLowerCase();
  }
```

midgleyc (approved):

test/net/sourceforge/kolmafia/AreaCombatDataTest.java
jaadams5: Thank you.  My original code was definitely an unintended error.

PR #2176, opened 2024-01-12
Add support for users on future build of Java 21

libraryaddict:
* When calling close() with HttpClient on java 21, the goaway errors seemed to have vanished
* In java 21, the caret in gCLI and chat will always be visible in the display frame, which doesn't look good. https://bugs.openjdk.org/browse/JDK-4512626

KolMafia may officially support java 17, however the caret "fix" for java 21 users doesn't harm those users and fixes a minor annoyance for people on newer versions of Java.
The "close" addition is further support of connection resetting, and unfortunately cannot be backported without suspicious code invocations.

Related to https://github.com/kolmafia/kolmafia/pull/2174


jaadams5 (dismissed):
It didn't obviously break anything on Windows 10/Java 17 in about five minutes of poking around.

midgleyc (approved):

PR #2177, opened 2024-01-12
feat: update gradle to 8.5

midgleyc:
There was a report that at the moment you can't compile under Java 21 due to the lack of gradle support. Upgrade gradle. I don't know whether this fixes that problem.

Also update the plugins whose release notes explicitly specified they were adding gradle 8 support to the latest versions.

Remove `include('app')` from settings.gradle as there is no such directory.

There are a series of warnings in the build that I don't know gradle well enough to fix, though I can identify the problematic lines and potential changes:
```
Build file '/home/wsl/thirdparty/kolmafia/build.gradle': line 20
The org.gradle.api.plugins.ApplicationPluginConvention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.5/userguide/upgrading_version_8.html#application_convention_deprecation
Build file '/home/wsl/thirdparty/kolmafia/build.gradle': line 20
The org.gradle.api.plugins.Convention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.5/userguide/upgrading_version_8.html#deprecated_access_to_conventions
Build file '/home/wsl/thirdparty/kolmafia/build.gradle': line 182
Allowed usage is changing for configuration ':implementation', resolvable was false and is now true. Ideally, usage should be fixed upon creation. This behavior has been deprecated. This behavior is scheduled to be removed in Gradle 9.0. Usage should be fixed upon creation. Consult the upgrading guide for further information: https://docs.gradle.org/8.5/userguide/upgrading_version_8.html#configurations_allowed_usage
Build file '/home/wsl/thirdparty/kolmafia/build.gradle': line 301
The org.gradle.api.plugins.ApplicationPluginConvention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.5/userguide/upgrading_version_8.html#application_convention_deprecation
Build file '/home/wsl/thirdparty/kolmafia/build.gradle': line 301
The org.gradle.api.plugins.Convention type has been deprecated. This is scheduled to be removed in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.5/userguide/upgrading_version_8.html#deprecated_access_to_conventions
```

PR #2178, opened 2024-01-13
fix: gradle warnings

midgleyc:
Fix the gradle warnings as reported in #2177.


jaadams5 (approved):
I'll trust you on this :-)

PR #2180, opened 2024-01-14
feat: update plugins, run spotless

midgleyc:
Not particularly keen on the blowing up of one-liners, but it is what it is.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,11 +11,11 @@ plugins {
 	id 'jacoco'
 	id 'java'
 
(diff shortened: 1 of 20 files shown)


jaadams5 (approved):
I assume it was easier to fix the complaints rather than figure out how to get spotless to not apply the rule causing most of them?

midgleyc:
I haven't looked into configuring spotless at all.

It might be nice, but "just go with the defaults" is a more defensible position.

PR #2187, opened 2024-01-19
Add OnlyFax fax message handling

libraryaddict:
https://github.com/loathers/OnlyFax/blob/main/src/utils/messages.ts

Sometimes I wonder if I should add it so mafia will still see a faxbot message via chat polling, even if player didn't choose to open chat. newchatmessages.php or something. Though it might accidentally swallow messages from other players..
It's possible to do that without appearing as online in chat.


jaadams5 (approved):
Since we don't have a documented API that faxbots are expected to adhere to I am OK with changing mafia to recognize additional messages rather than change OnlyFax (or any other faxbot) to use the existing, expected messages.

jaadams5:
veracity has done a fair amount of scripting that works in the absence of an open chat interface.  I am not sure users really need anything added that polls chat if script writers are similarly clever.  People who request faxes from the VUI can usually figure out what is going on or open chat if they cannot.  Perhaps you should start a feature request at Kolmafia.us to encourage discussion?

PR #2188, opened 2024-01-20
Explicitly add jgit ssh dependency

gausie:
At some point in my many reconfigurations of my dev environment, git updating just totally stopped working for me. It turns out that the SSH part of JGit can sometimes not be present, which results in a misleading error about timeouts. Explicitly depending on the library fixes this. I had previously kept it in my local changeset but I have now heard reports of others having this issue.


midgleyc (approved):
A'ight.

I think everyone affected by this has been using Java 21, but two cases make a pattern.

gausie:
@midgleyc I'm using Java 17

PR #2191, opened 2024-01-24
Fix ccs parser unable to handle multiple closing brackets

libraryaddict:
The following example from wiki will fail as mafia finds the first closing bracket and early exits.
This is just a small fix, I'm not personally interested in the whole ccs of mafia.

https://wiki.kolmafia.us/index.php/Custom_Combat_Script#Using_ASH_Constants_in_CCS

```
[ $element[ spooky ] $item[ hobo nickel ] ]
skill entangling noodles
skill weapon of the pastalord
```

turns into

```
[ $element[ spooky ]
skill entangling noodles
skill weapon of the pastalord
```

The first commit is a failing test. The second commit is the fix.

As a small note, I did notice that the following is appended to the macro. But it's a little out of scope of what I intended to fix and isn't as big a problem. Probably caused only if a default action isn't defined. This bug wasn't introduced by my change.

```
[ default ]
attackattack with weapon
```


midgleyc (approved):
Thanks!

Knowing this errored, the question is now "what happens if the bit inside the square brackets is invalid", but this is an incremental improvement for a real use case.

PR #2192, opened 2024-01-27
Since Toggle Optimality is auto-permed, it is not permable

Veracity0:


jaadams5 (approved):

gausie:
Hmm I think this is different from what I considered permeable to mean. To me it means "can be permanently acquired". What's the benefit of removing the flag here?

midgleyc:
It appears in Valhalla as something you haven't permed, under the way it currently works.

Perhaps it needs to go on some list somewhere.

PR #2194, opened 2024-01-30
RequestLogger only prints html if explicitly requested

libraryaddict:
I had an issue with invalid html being printed to my mirror log which led me to dig deeper.

https://kolmafia.us/threads/mirror-logging-needs-to-write-entity-encoded-characters.29562/

Generally, KoLMafia has a problem where a mixture of sanitized and unsanitized html is being piped to various logs and renders.

This PR changes the logging so that printLine in RequestLogger will now print entity encoded strings. And if html is desired, the printHtml function must instead be invoked.

As a general rule, KoLMafia previously relied on swing utilities accepting malformed html and refusing to render it as anything but plaintext as it didn't match valid syntax. Of course, it may be the case where text intended to be plaintext was instead displayed as html. 

```
 > ash print("&lt;i&gt; <i>")
<i> <i>
Returned: void

```

As a real world example, the above isn't likely to be encountered often. But it's unexpected behavior that the string was transformed despite print_html not being called.

The impact caused by malformed html is mostly renderers suffering a performance impact and outputted text being written in a way that one could find unintuitive.


jaadams5:
> I like the idea, but how did you find the places which should have been HTML? How much have you run with this? Are we likely to find additional breakages?

I expressed a portion of my opinion elsewhere.  Now that there is code I will do somewhat more strongly.  I don't think this is a good idea or good implementation of a good idea.  The only specific case that was cited elsewhere was a preference.  That could be dealt with by changing the preference or wrapping reads and writes of that preference with encoding.  I am concerned by the implementation because it appears that the programmer, at the time code is being written, must have knowledge of future use cases and anticipate whether HTML will be a factor.  For those of use who use text tools that are not HTML aware this might make things harder.

I won't approve because I don't think the is a good idea.  If no one agrees with me then I would like to see some more thought put into the implementation.  As I understand it RequestLogger.printLn will emit ill-formed HTML.  Rather than create a new function and force a programmer to decide what to call, why not make printLn scan for ill formed HTML and either fix it before printing or annotate that it is occurring and make a programmer fix what generates the ill formed html?

As always there is the possibility that my concerns are based on misunderstandings but if that is the case then the conversation needs to go back to the beginning and use short sentences with small words :-)

libraryaddict:
> I expressed a portion of my opinion elsewhere. Now that there is code I will do somewhat more strongly. I don't think this is a good idea or good implementation of a good idea. The only specific case that was cited elsewhere was a preference. That could be dealt with by changing the preference or wrapping reads and writes of that preference with encoding.

The preference is the most obvious case, it's by no means the only scenario in which this happens.
tbh, I feel like it's weird that you're in favor of having malformed html being written because it's what you're used to?

> I am concerned by the implementation because it appears that the programmer, at the time code is being written, must have knowledge of future use cases and anticipate whether HTML will be a factor. For those of use who use text tools that are not HTML aware this might make things harder.

I don't think this is a real issue? If you want the raw text to be written to gCLI, then printLine is to be used. If you are reading html from the server or want to serve formatted text in gCLI, then printHtml.

A junit test should be covering this. And I'd expect this to be an issue only encountered once in gCLI that you're seeing raw html being printed before you realize you called the wrong method.


> I won't approve because I don't think the is a good idea. If no one agrees with me then I would like to see some more thought put into the implementation. As I understand it RequestLogger.printLn will emit ill-formed HTML. Rather than create a new function and force a programmer to decide what to call, why not make printLn scan for ill formed HTML and either fix it before printing or annotate that it is occurring and make a programmer fix what generates the ill formed html?

I picture this as calling printHtml only when you're explicitly sure that html is to be rendered. Otherwise, always calling printLine.

Scanning for malformed html sounds like a poor idea to me personally, as there's times that it'd be written. Frankly, it feels weird that you're arguing that unsanitized html is the expectation and writing sanitized html the outlier.

I'm not really sure how I would improve on this implementation.
As I see it, the programmer wants to always explicitly print their text as written, or their text with html markdown. We should expect it to be non-html markdown, unless told otherwise.

It's the same reason we have print() and print_html() as kolmafia functions.

Am I correct in assuming your problem with this is entirely because of the confusion on when someone should be calling printLine and printHtml inside kolmafia's codebase?

midgleyc:
I have a slightly different use-case which this PR fixes: sometimes user-created preferences are valid HTML, but I don't think they should be rendered as HTML (in the GCLI) when logging preference changes. The one I recall is `minevolcano`, which stores the HTML for pieces of gold or HP losses.

Also, before this PR running `ash visit_url("council.php")` would dump the HTML of the council.php page to the screen, breaking the CLI a bit. You could also dump a fight page to break it a bit more.

Besides that, I think entity encoding / decoding is better than our previous approach of replacing particular characters.

libraryaddict:
Oh, and it would also break your display there if someone performs `get htmlPref`

jaadams5:
>The preference is the most obvious case, it's by no means the only scenario in which this happens.

I am cynical enough to not believe that as an assertion with only one example but I also am willing to let go of justifying this or not.

If this is recast as a change to prevent mafia from generating malformed HTML then I can focus on the implementation.

My concerns are rooted in the cognitive overload on the person writing the code.  Cherry picking your comment this change expects the programmer: 

to know whether or not something is expected to contain HTML at the time of coding; 
to have a failing test prior to implementation OR to build and run and inspect the gCLI during development.
to respond to future changes that break code that was not expecting HTML when it was written.

Given the way KoLmafia code is, and is not documented and my personal style of coding it is much much easier for me to decide the value of a parameter to a function call than it is to decide which function to call.  So I would make a two parameter version of printLn, and overload things so the one parameter version called the two parameter version.  Even if it becomes necessary to call the two parameter version in 53 files that seems better to me than essentially renaming a function.  If nothing else that implementation means that all the changes do not have to be made at once.

But if it is time to let this go then I will :-)

midgleyc (approved):

PR #2197, opened 2024-02-02
Some tests for the Maximizer

jaadams5:
See discussion at https://kolmafia.us/threads/maximizer-didnt-equip-a-weapon.29577/#post-174282

Sometimes the maximizer fails to equip a weapon.  My hypothesis is that one of those times is when muscle <= moxie and the effective keyword is used.

These tests (with the System.out.println not commented out) lead me to that conclusion.

I expect these can be reduced to one test that fails now and works after a fix but I am sharing all of these in case some kind of more exhaustive testing is worth implementing.

The CustomScriptTest failure seems to be because the script assumes it is being run the first time the Maximizer is called and these tests seem to (leak?) and invalidate that.


jaadams5:
I decided the observed behavior in the presence of effective was correct.  So the tests were reworked to pass and demonstrate that if the right weapon class was not effective then no weapon would be equipped.

jaadams5:
I think effective is doing what it was designed to do.  These tests illustrate that.

jaadams5:
My understanding of the problem has changed.

jaadams5:
I'm pushing to here because it is easier to see all of my changes 'at once' and decide which ones are no longer useful as I continue to discover things.  Note that debug logging breaks one script test.  I will deal with that eventually.

jaadams5:
Cleaned up.

Trivial change to add SEWER_SNAKE to ItemPool.

Return a copy of net.sourceforge.kolmafia.maximizer.Maximizer.boosts rather than the list itself.  Returning the list made it easier to test the results from two maximizer runs, even if none of those tests are included here.

Added some tests that demonstrate that sometimes, with the effective keyword, No Weapon is the right answer.

Committing these helps me stay focused :-)

jaadams5:
@midgleyc or @gausie ?

I'm interested in any comments about the tests and want to make sure no one has concerns about the change to getBoosts since that is part of the test infrastructure.

Thanks.

test/internal/helpers/Maximizer.java
```diff
@@ -48,7 +49,7 @@ public static double modFor(Modifier modifier) {
   }
 
   public static List<Boost> getBoosts() {
-    return net.sourceforge.kolmafia.maximizer.Maximizer.boosts;
+    return new ArrayList<>(net.sourceforge.kolmafia.maximizer.Maximizer.boosts);
```
midgleyc, line 52: why was this done?

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -2045,4 +2045,81 @@ public void suggestCardSleeveSlot() {
       }
     }
   }
+
+  @Nested
+  public class Weapons {
```
midgleyc, line 2050: there are already some tests in `Effective` testing this -- the ones where it equips no weapon are new, but the others overlap

midgleyc (commented):
Can you rebase your branch and give your commits sensible descriptions if you want them separate? It's hard to tell what's going on with comments like "park" and "more park".

test/internal/helpers/Maximizer.java
jaadams5: Because I wanted to run the maximizer twice and compare the results as part of a test.  Without this if a test wanted to compare boosts under two different sets of conditions the test needed to make its own copy before running again.  It is possible the test that did that has not survived in which case I can revert this but I'm not always a fan of a hiding a mutable list behind an accessor call.

jaadams5:
> Can you rebase your branch and give your commits sensible descriptions if you want them separate? It's hard to tell what's going on with comments like "park" and "more park".

I commit locally as a kind of back up and the names don't have a lot of meaning.  I suppose I could work on that.  Rebase is beyond my git-fu but your point is good enough that I should expand my capabilities.  Thank you and stand by.

test/internal/helpers/Maximizer.java
midgleyc: okay, seems reasonable

jaadams5:
Rebased.  Will review.

jaadams5:
I missed the Effective tests and there is some duplication.  Those tests include weapon damage in the maximizer string while mine don't and as a result no weapon is equipped.  I'll consider some renames and deletes.

midgleyc (approved):
Thanks!

PR #2200, opened 2024-02-11
add spring shoes skills

ckb11:


Veracity0 (changes requested):
What is "test/root/request/test_adventures_spent_airship_12 (1).json"?
It's empty and doesn't seem relevant to this commit, even if it weren't.

ckb11:
I don't know what that josn file is.  Thats probably an artifact of me just learning how to use git.  How do I fix that?

Veracity0 (approved):

PR #2201, opened 2024-02-12
Spring Kick banish management

ckb11:


jaadams5:
Like you, I agree that this looks like other banishers and should work.

Do you have any sense that this code works as expected (from local build and run)?  Would you consider expanding your skill set and trying to write a test?  That might be as simple as capturing HTML and doing a copy/past from some other test.

ckb11:
I can give a test a try, but it might take me a long time.  I know very little about how any of this works in detail.
I'll look through other commits and see if I can find something to copy maybe.

jaadams5:
Sounds good.  If you give up let me know and I will either try and help or just approve.  But since I don't have the item you might have to be the one that captures fight html with a banish.

ckb11:
I was hoping to find a similar test from a past banish commit, but it looks like all the recent additions (monkey slap. crimbo23 stuff, asol, Roar like a Lion) did not include any tests.
I can capture and share html if that helps.

jaadams5:
The maximizer fails to equip is sucking up all of my cycles.   Maybe someone else will chime in and assist with a test?  Otherwise I'll get to it eventually or just submit.  Thanks.

jaadams5:
I look casually at FightRequestTest.  commerceGhostIncrementsByOneOnFight seems to parse a combat and then check for a preference to change.  You could capture HTML from a fight with a spring banish, parse it as part of the test and just verify that the preference changed to include the banished monster.  I would expect that to work as a test although I am not in a position to try it for you.

midgleyc:
look at e.g. `FightRequestTest.canDetectEagleScreech`. That's phyla instead of monster, but the idea is the same.

ckb11:
I looked at that FightRequestTest... most of that went way over my head.  I could spend some time trying to hack together a test, but I fear there is an equal change of getting a good test as there is at just messing up the code and breaking it.
It would probably be easier to try and figure out how to compile Mafia and run some test adventures, but I don't know how to do that either or if it counts.

midgleyc:
no problem, if you're up for it I can make a test and add it to your branch myself, and we can join a video call over Discord so you can see what I do live, if you'd like.

But the long of it is:
* enter a fight with spring boots equipped
* command `debug on`
* use the banish
* command `debug off`
* you now have some HTML in the Mafia root directory. You're going to put this in `test/root/request/` and make a test looking something like
```java
  @Test
  public void canDetectSpringBoots() {
    var cleanups =
        new Cleanups(
            withFight(),
            withProperty("banishedMonsters"));

    try (cleanups) {
      parseCombatData(
          "request/test_fight_spring_boots_banish.html", "fight.php?action=skill&whichskill=7501");

      assertThat("banishedMonsters", hasStringValue(startsWith("fluffy bunny:Spring Kick:")));
    }
  }
```
at the bottom of `FightRequestTest.java`.

As to "how do I compile Mafia myself", I will assume you've already got the source downloaded and the jdk installed. From there, go to the root directory of the source, and run `./gradlew :shadowJar`. This will make a jar file in `dist/`. Copy it to wherever or just run in straight from there.

midgleyc:
Also, it probably /should/ be turn-free, because it doesn't end the combat. I think this is the first banish that's true for?

ckb11:
@midgleyc That helps.  I will capture some fight html and use your code snippet.  I also need to install a JDK (I have JRE only now) to try that compile.

Do we need to check that spring shoes are equipped in the 'Cleanups' section?

When is this check done?  Is this part of the PR check, or does that test happen at compile?

I am happy to jump on a call and learn more about this if you have time.  I am on the west coast of the US, and you look like you are in the UK, so timing might be tricky.

As for "isTurnFree" I don't know what it should be as it works different than all the other banishes that end combat.  I would expect this to be false based on the fact that it cannot act as a free runaway or delay burn, which was my line of thinking when I made this update.

ckb11:
Hey look!  I added a test... which failed... I'm not sure if this is because the code is wrong or the test is wrong.

midgleyc:
Yeah, the test was wrong: I forgot that `banishedMonsters` needs special handling, because it reads from a static variable to populate it.

I've fixed it.

midgleyc (approved):
Thanks!

PR #2202, opened 2024-02-14
Count rollover adventures for housekeeping robot

eegeeZA:


midgleyc (approved):
Thanks!

Also the housekeepers removal / addition are crying out to be refactored to share the code. Doesn't have to be in this PR, though.

PR #2203, opened 2024-02-16
A Passing Test for the Maximizer

jaadams5:
These changes are also in https://github.com/kolmafia/kolmafia/pull/2197 but I pulled them out to isolate them and hopefully get feedback.  With two exceptions the changes were suggested by static code analysis or spotless.

One change was to explicitly set/reset the password hash in an ApiRequest.  The test itShouldMaximizeAndEquipSelectedWeapon would pass or fail depending upon the run context and the difference was the presence of the password hash in the body of the request.  Clearing the hash in the test that set it made itShouldMaximizeAndEquipSelectedWeapon work consistently.

The other change is itShouldMaximizeAndEquipSelectedWeapon.  It is supposed to be a test that mimics using the maximizer from the CLI or from a script (not the GUI).  It should demonstrate that the Maximizer chooses to equip a weapon and then equips the weapon.

Testing KoLCharacter.hasEquipped(ItemPool.SEAL_CLUB) fails but that is presumably because the request never actually gets executed.  Confirming that the equip request is submitted seems to be the best that can be done.

Feedback or approval is desired.  Is there a better or different way to test things actually being equipped by the Maximizer?

Tagging folks who are pretty good at telling me what I am missing or how I could have done something better.  @gausie @midgleyc @BadHorseMonkey @Veracity0 

Thank you.

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
+++ b/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
@@ -1430,10 +1430,10 @@ public class RuntimeLibraryTest extends AbstractCommandTestBase {
     assertContinueState();
     var requests = getRequests();
     assertFalse(requests.isEmpty());
-    var checkMe = requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findFirst();
+    var checkMe =
+        requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findFirst();
     if (checkMe.isPresent()) {
-      assertPostRequest(
-          checkMe.get(), "/inv_equip.php", "which=2&ajax=1&action=equip&whichitem=1");
+      assertPostRequest(checkMe.get(), "/inv_equip.php", "which=2&ajax=1&action=equip&whichitem=1");
     } else {
       fail("Could not find expected equipment request.");
     }
--- a/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
+++ b/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
@@ -1429,7 +1429,7 @@ public class RuntimeLibraryTest extends AbstractCommandTestBase {
     assertTrue(out.contains("Putting on old sweatpants..."));
     assertContinueState();
     var requests = getRequests();
-    assertTrue(!requests.isEmpty());
+    assertFalse(requests.isEmpty());
     var checkMe = requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findAny();
     if (checkMe.isPresent()) {
       assertPostRequest(
--- a/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
+++ b/test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
@@ -1430,7 +1430,7 @@ public class RuntimeLibraryTest extends AbstractCommandTestBase {
     assertContinueState();
     var requests = getRequests();
     assertFalse(requests.isEmpty());
-    var checkMe = requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findAny();
+    var checkMe = requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findFirst();
     if (checkMe.isPresent()) {
       assertPostRequest(
           checkMe.get(), "/inv_equip.php", "which=2&ajax=1&action=equip&whichitem=1");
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -674,6 +681,12 @@ void canSeeDaycount() {
       assertContinueState();
 
       assertThat(output, is("Returned: 7302\n"));
+
+      /*
+       ApiRequest.parseStatus sets a password hash which persists to other tests and causes them to pass or fail
+       based upon whether this test was run first, or not.  Explicitly clear the hash when this test ends.
+      */
+      ApiRequest.setPasswordHash("");
```
midgleyc, line 689: put this in a `finally` block so it runs even if the test fails.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -271,13 +278,13 @@ void canVisitCabinet() {
...
             equalTo(
                 """
-                    Returned: aggregate item [string]
-                    booze => Doc's Fortifying Wine
-                    equipment => frozen jeans
-                    food => frozen tofu pop
-                    pill => Breathitin&trade;
-                    potion => anti-odor cream
-                    """));
+                                Returned: aggregate item [string]
```
midgleyc, line 281: this level of indentation seems clearly worse to me than the previous one? I don't understand why spotless would do this.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1384,12 +1397,57 @@ void trackedByIncludesAllTracks() {
...
+    try (cleanups) {
+      out = execute(cmd);
+    }
+    assertFalse(out.isEmpty());
+    assertTrue(out.contains("Putting on seal-skull helmet..."));
+    assertTrue(out.contains("Wielding seal-clubbing club..."));
+    assertTrue(out.contains("Putting on old sweatpants..."));
+    assertContinueState();
+    var requests = getRequests();
+    assertTrue(requests.size() > 1);
```
midgleyc, line 1437:
```suggestion
    assertTrue(requests.size() >= 1);
```

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1384,12 +1397,57 @@ void trackedByIncludesAllTracks() {
...
+    int whichReq = -1;
+    for (int i = 0; i < requests.size(); i++) {
+      String body = getPostRequestBody(requests.get(i));
+      if (body.contains("whichitem=1")) {
+        whichReq = i;
+        break;
+      }
+    }
```
midgleyc, line 1445: probably this wants streaming, something like `requests.stream().filter(x -> getPostRequestBody(x).contains("whichitem=1")).findAny()`

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1384,12 +1397,57 @@ void trackedByIncludesAllTracks() {
...
+        new Cleanups(
+            withStats(10, 5, 5),
+            withEquippableItem("seal-skull helmet"),
+            withEquippableItem("astral shirt"),
+            withEquippableItem("old sweatpants"),
+            withEquippableItem("sewer snake"),
+            withEquippableItem("seal-clubbing club"));
+    String out;
+    String cmd;
+    cmd = "maximize(\"" + maxStr + "\", false)";
```
midgleyc, line 1427: may as well define inline?

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
jaadams5: OK.  Update coming.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
jaadams5: That's a hangover from a case when there were several requests and the first one had nothing to do with the weapon.  In the new context given the following iteration the check should be for any number of requests.  Stand by.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
midgleyc: if you can, I think it would be nice -- I both think it looked better before and think it would be good to have fewer trivial changes in this PR.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
jaadams5: OK

jaadams5:
I think I resolved everything.

midgleyc (approved):
Thanks!

PR #2211, opened 2024-02-21
return unknown class instead of null in my_class

burningbman:
When the ascension class is unknown (during first few days of a new path), my_class() returning null causes multiple scripts to fail (TourGuide and CHIT that I know of).
So, instead of setting the current class to null when it is unknown, set it to a class called unknown.


midgleyc (approved):
Thanks!

PR #2213, opened 2024-02-22
add [turn] to Rain Man and mimic egg use

ckb11:


ckb11:
I tested this with a local version and a mimic egg use and it works as expected.
I did not test Rain Man as I am not in Heavy Rains.

jaadams5 (approved):

PR #2214, opened 2024-02-24
Initialize AscensionClass to UNKNOWN by default.

heeheehee-kolmafia:
Most previous instances of this were handled in r27834, but we missed this case. I went through and checked other places where we were creating a local AscensionClass variable. This was the only site with potential for an uninitialized access.

I manually hacked up the test HTML file to add an extra row with a fake class; without this change, it fails with a NPE as expected.


midgleyc (approved):
thanks!

PR #2215, opened 2024-02-25
Pulverized Standard rewards are currency for the Armory and Leggery

Veracity0:
1) ResultProcessor signals "(armoryandleggery)" if a pulverized Standard reward enters or leaves inventory.
2) pulverized.txt no longer lists pulverization for Standard rewards. Instead, StandardRewardDatabase derives that after that file is is read. Hence, none of those items needs to be in that file.
3) ArmoryAndLeggerRequest (the coinmaster) initializes its inventory using StandardRewardDatabase.
4) The Armory and Leggery card in the CoinmastersFrame only shows currencies in the title that you actually have in inventory; since there are 2 per year of Standard, showing them all doesn't come close to fitting.
5) I made that same card in the CoinmastersFrame only show items that you can actually make.
I am of two minds about whether that is good. Again, there are WAY too many possible pieces of equipment you can buy - but perhaps you'd like to see what you COULD aspire to obtain?


jaadams5 (approved):

jaadams5:
>I made that same card in the CoinmastersFrame only show items that you can actually make.

I think I am OK with that.  I use other tools to display what I need and then the wiki to confirm what I have to do to be able to acquire it.

PR #2217, opened 2024-02-26
Track Savage Beast mods

burningbman:
Store the mods for Savage Beast when the description of the effect is parsed. I used the same test as was done for the citizen of a zone effect. I wasn't sure if there was an accessible way to validate the actual modifiers on the effect as opposed to validating the property was updated.


midgleyc:
I think the pref wants an entry in `src/data/defaults.txt` too.

midgleyc (approved):

PR #2218, opened 2024-02-28
WereProfessor stuff

Veracity0:
1) Add Beast skills
2) Add recipes for things you can create at the Tinkering Bench.
Added a TINKER crafting method to the enum.
I have no yet made it usable by creating a Request class to implement it.
3) If you have a "focused magnetron pistol" equipped as your weapon, you can pickpocket, either as a Beast or as a Professor.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
@@ -2020,6 +2020,9 @@ public class ConcoctionDatabase {
           "You need to have a Pile of Burning Leaves in your campsite to make that.");
     }
 
+    // Making stuff at the Tinkering Bench is allowed if you are in the
+    // WereProfessor path and are in Professor form.
+
     // Other creatability flags
 
     if (KoLCharacter.isTorsoAware()) {
--- a/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
@@ -2023,6 +2023,15 @@ public class ConcoctionDatabase {
     // Making stuff at the Tinkering Bench is allowed if you are in the
     // WereProfessor path and are in Professor form.
 
+    boolean tinker = KoLCharacter.isMildManneredProfessor();
+    if (tinker) {
+      permitNoCost(CraftingType.TINKERING_BENCH);
+    } else {
+      ConcoctionDatabase.EXCUSE.put(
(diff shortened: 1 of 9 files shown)


src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
```diff
@@ -2020,6 +2020,9 @@ else if (Preferences.getBoolean("requireBoxServants") && !KoLCharacter.inGLover(
           "You need to have a Pile of Burning Leaves in your campsite to make that.");
     }
 
+    // Making stuff at the Tinkering Bench is allowed if you are in the
```
midgleyc, line 2023: in this PR or another?

midgleyc (approved):

src/net/sourceforge/kolmafia/persistence/ConcoctionDatabase.java
Veracity0: Do we track, yet, whether you are a Professor or a Beast?
If so, this would be easy to add - although we won't be able to dispatch to the Request class until it is implemented.

I have only one more Tinkering left this run - so only one more to test that the Request is implemented correctly.
I'm thinking of doing the Request early on next run - after the Beast has done its first set of smashings.

midgleyc: We don't track it explicitly, but I think you can tell based on which intrinsic effect you have.

Veracity0: I think I'll put a couple of utility functions in KoLCharacter.

boolean isProfessor() -> in WereProfessor path with Mild-Mannered Professor intrinsic
boolean isBeast() -> in WereProfessor path with Savage Beast instrinsic

Veracity0: I did that, renamed the TINKER crafting type to TINKERING_BENCH, and made it a permissible method iff you are a Mild-Mannered Professor.

I have 1 broken scientific equipment and 1 quick-release fanny pack in inventory.
The Creatable tab in the Item Manager says I can make:

quick-release belt pouch
quick-release utility belt

Obviously, that is not completely correct; the belt pouch can only be made if you do not have the upgraded version. Ditto for all other TInkering Bench items; you can only make each of them once.

Having run my turns today, I am currently a Mild-Mannered Professor. I'll start a new PR for TinkeringBenchRequest and when I use my last broken scientific equipment, I'll capture the response text so that I can write tests.

midgleyc (approved):

PR #2219, opened 2024-02-29
Support WereProfessor's Research Bench

Veracity0:
I noticed that my session log reported all research purchases as if they were the first visible button on the Research Bench page. That's because all research is done via ```choice.php?whichchoice=1523&option=1&r=KEYWORD```.
Which is to say, KoLmafia thinks there is only one possible decision: option=1.

This might be a first; we are used to having various options taking extra fields - "select" dropdowns or text areas - but not having every button have the same option.

There are three skill trees - starting with mus1, myst1, and mox1 - each with eighteen skills. Here they are:

```
  // Skill trees
  //
  // mus1 mus2 mus3
  //                rend1 rend2 rend3 slaughter
  //                hp1 hp2 hp3
  //                            skin1 skin2 skin3 skinheal
  //                            stomach1 stomach2 stomach3 feed
  // myst1 myst2 myst3
  //                bite1 bite2 bite3 howl
  //                res1 res2 res3
  //                            items1 items2 items3 hunt
  //                            ml1 ml2 ml3 feasting
  // mox1 mox2 mox3
  //                kick1 kick2 kick3 punt
  //                init1 init2 init3
  //                            meat1 meat2 meat3 perfecthair
  //                            liver1 liver2 liver3 pureblood
```
To my eye, that chart specifies a total  ordering for the 54 skills.
Based on which skills we see (or don't see) when we visit the Research Bench, we can derive all the skills we have already researched, whether or not we saw said research being done.

1) ResearchBenchRequest now has all three skill trees - muscle, mysticality, and moxie - each with 18 skills.
2) The r=KEYWORD - or field - is how we'll refer to the various skills. Actually, since they are all r=wereprof_KEYWORD, we'll omit the "wereprof_" prefix for more compactness.
3) ResearchBenchRequest.registerRequest will parse the field value, look up the associated record structure, and generate the "correct" action to log in the session log.
4) When you visit the Research Bench, we parse the text and see which skills are on offer.
From that, we derive the skills that you already researched.
Those two sets are stored in new properties: "beastSkillsKnown" and "beastSkillsAvailable".
No test cases specifically testing this - although ResearchBenchRequest.deriveKnownResearch is exposed and could well be used to write tests.
Current state of my character:
```
Available (3): feed,feasting,perfecthair
Known (51): mus1,mus2,mus3,rend1,rend2,rend3,slaughter,hp1,hp2,hp3,skin1,skin2,skin3,skinheal,stomach1,stomach2,stomach3,myst1,myst2,myst3,bite1,bite2,bite3,howl,res1,res2,res3,items1,items2,items3,hunt,ml1,ml2,ml3,mox1,mox2,mox3,kick1,kick2,kick3,punt,init1,init2,init3,meat1,meat2,meat3,liver1,liver2,liver3,pureblood
```

5) Using that, I vastly simplified the calculation of wereProfessorLiver and wereProfessorStomach.
The existing tests for those in ChoiceControlTest still pass - which is a pretty serious test case for the skill derivation.
6) Added wereProfessorBite, wereProfessorKick, and wereProfessorRend using same technique.
7) Added wereProfessorResearchPoints and set that when visiting Research Bench
8) You can walk away from choice 1523
9) ```ResearchBenchRequest()``` will visit the Research Bench and leave you handling the choice.
```ResearchBenchRequest("rend1")``` will visit the Research Bench and attempt to learn that skill.
These are intended for use in a "wereprofessor" command.
10) If you research the final skill, you are no longer in a choice, so ChoiceManager will not call visitChoice for us.
Do that manually.
Similarly, if you visit an empty research bench, ChoiceManager will not detect that we are in a choice - even though KoL redirected to choice.php?forceoption=0. There are other choices like that, and ChoiceUtilities.extractChoice can handle it, such that that visitChoice will be called.
Make it so for the empty Research Bench.

I did some tests via visit_url():

- Asking for a skill which is on offer but for which I have insufficient rp (73 vs. 100)
```choice.php?whichchoice=1523&option=1&r=wereprof_feasting```
Responds with "You have to pick an available thing to research." and kicks you out of the choice
ResponseText saved in ```test_research_bench_research_unvailable.html```
- Asking for a completely bogus skill
```choice.php?whichchoice=1523&option=1&r=wereprof_bogus```
does the same thing.
- Asking for a skill which is already known - and for which you have enough rp
```choice.php?whichchoice=1523&option=1&r=wereprof_rend1```
researches it again and spends the rp! (My rp is now 53).
ResponseText saved in ```test_research_bench_research_known.html```
- Asking for a skill which is already known - and for which you do NOT have enough rp (53 vs. 100)
Returns response code 404

11) ResearchBenchRequest.run() now checks for all sorts of errors and will not submit a request to KoL for them.
- Invalid research
- Must be a Mild-Mannered Professor
- Already know the skill
- Skill is unavailable to research at this time
- Don't have enough rp to research the skill.

12) When KoL refuses to perform research, it responds with "You have to pick an available thing to research." as opposed to "You successfully research NAME." We look for the latter, but I now include the former in a comment, at least. I submitted a text fixture where that is in the responseText and wrote a test with it using GenericRequest, since ResearchBenchRequest shouldn't be able to make that happen.

Not in this PR:

- There are other ways to check and gain research points.
I submitted the following test fixtures:
```
test_fight_research_initial.html
test_fight_research_advanced_success.html
test_fight_research_advanced_failure.html
test_charpane_research.html
```

- How about a ```wereprofessor research``` command which shows all the skill trees compactly, indicating which skills you have already learned, and highlighting those which are currently available to research?
- What about the ```choice``` command? It would be nice if typing ```choice``` listed all the buttons - and the value of the "r" field needed to select them - and you could type ```choice 1 r=wereprof_res3```, say, to select the research which is the 3rd "All Resistances +2" upgrade. You CAN type that, but you have to know the button is available and you have enough Research Points to afford it. This would require substantial whackage to how we detect choice options, since we "know" that choice options are unique.

### CODE REVIEW

File: src/net/sourceforge/kolmafia/request/ResearchBenchRequest.java [hunk 1]
```diff
@@ -179,6 +198,37 @@ public ResearchBenchRequest(final String research) {
...
+      wereStomach += wereStomach == 1 && !text.contains("Thyroxine supplements (50 rp)") ? 1 : 0;
+      wereStomach +=
+          wereStomach == 2 && !text.contains("Amyloid polypeptide mixture (60 rp)") ? 1 : 0;
+    }
+    Preferences.setInteger("wereProfessorStomach", wereStomach);
+
+    // calculate liver
+    int wereLiver = 0;
+    if (!text.contains("Dopamine slurry (10 rp)")
+        && !text.contains("Relaxin Balm (20 rp)")
```
BadHorseMonkey, line 220: Should be Relaxin balm, or is in my page.  This is what is causing Liver to think it's 13 not 4.

Veracity0: Thanks. I saw your PR and duplicated the change here. :)

jaadams5 (approved):

PR #2223, opened 2024-03-01
Support for Tinkering Bench crafting method

Veracity0:
Support for Tinkering Bench as a CraftingType

It's unusual in that you can only ever create one of each item - even if you use that item as an ingredient for another item.

I attempt to determine if you have ever created an item by seeing if it is available - inventory or equipped - or an item using it his available - or if an item using THAT item is available. And if the item has not been created, it is now creatable.

But it is funky, since each such creation also uses smashed scientific equipment, and that doesn't seem to be accounted for.

I have worked at this all day and would like to get it in and revisit this on my next WereProfessor run, since I have tinkered everything this run.


midgleyc (approved):
One of the tests asserts that the assertion is wrong; we are just accepting that?

Veracity0:
> One of the tests asserts that the assertion is wrong; we are just accepting that for now?

Yes. That's the issue I mentioned. With 2 smashed scientific equipment:

- create high tension exoskeleton (uses 1)
- create ultra-high-tension exoskeleton (uses previous creation and the 2nd)
- create irresponible-tension exoskeleton (uses previous, but need another smashed scientific equipment)

Concoction.canMake looks at ingredients and recurses on ingredients and should take care of that, but I have apparently not done exactly the right thing with that method, yet.

I want to get this in and revisit it, next WereProfessor run.

PR #2229, opened 2024-03-07
Track Research Points in fights and charpane

Veracity0:
Also, rather than saying "bite1" has an "Elemental Attack" and "bite2" adds a "Second Element",  say "(Stench)" and "(Sleaze)".


midgleyc (approved):
does this work in both full and compact charpane?

Veracity0:
> does this work in both full and compact charpane?
Now it does.

I added a new text fixture with a compact charpane. We could use both test_charpane_research.html and test_charpane_compact_research.html if we wanted to track "Until Transform" and "Tranform" (sic).

midgleyc (approved):

PR #2230, opened 2024-03-09
Remember preferred faxbot

libraryaddict:
This PR performs two actions.

1. The faxbot GUI in mafia now remembers the last faxbot you had open, where previously it was ignored and would default to 0.
2. When requesting a fax from a faxbot without specifying a faxbot, the iterator starts from your prefered faxbot (2, 3, 0, 1) instead of from 0 (0, 1, 2, 3). This reuses the property from the GUI, as it's normally the users preferred faxbot.

In short, this means the user has some measure of control over what the default faxbot they use is.


gausie:
I don’t know if having had the tab open is a sufficiently proactive way of
specifying a new default. What do others think?

jaadams5:
I almost never use the GUI to get a fax and I don't have a preferred faxbot.  My focus is on the monster - I care that I got the requested monster and not where it came from.  But I have observed that some faxbots are more reliable than others and I would set a preference to try the most reliable one first.  But continuing to try until there is success is more important than saving a few seconds and succeeding on the first bot, rather than the second.

But in the event that I do request a fax twice in a session I would prefer that the second request start with the faxbot that last succeeded.  The use case for the second call is a clan imposed desire to have a particular monster in the fax machine.  I fax and fight and then fax to restore the preferred monster.  This allows a character who cannot chat to still fight a predictable monster.

src/net/sourceforge/kolmafia/swingui/FaxRequestFrame.java
```diff
@@ -54,7 +54,7 @@ public FaxRequestFrame() {
       panel.add(botPanel);
       this.selectorPanel.addPanel(bot.getName(), panel);
     }
-    this.selectorPanel.setSelectedIndex(0);
+    this.selectorPanel.setSelectedIndex(Preferences.getInteger("faxbots"));
```
jaadams5, line 57: I'm not finding this preference in defaults.  Please add it.  It might be worth discussing whether it should be a global or user preference.  If I prefer a faxbot for performance reasons then my preference won't change when I use a different character.

src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
```diff
@@ -104,7 +104,9 @@ private static void configureFaxBot(final BotData data) {
     }
   }
 
-  public static final FaxBot getFaxbot(final int i) {
+  public static final FaxBot getFaxbot(int i) {
+    i = i % faxbots.size();
```
jaadams5, line 108: I don't know why this was ever allowed to return null - perhaps if there were no faxbots defined in the file? - but the modulo operation guarantees that the index will be less than the size so the return statement can be simplified to take advantage of that.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -5196,7 +5196,13 @@ public static Value faxbot(ScriptRuntime controller, final Value monsterName) {
 
     FaxBotDatabase.configure();
 
-    for (FaxBot bot : FaxBotDatabase.faxbots) {
+    // Get the faxbot the user last looked at
+    int preferedBot = Math.max(Preferences.getInteger("faxbots"), 0);
+
+    // Iterate from 0
+    for (int i = 0; i < FaxBotDatabase.faxbots.size(); i++) {
```
jaadams5, line 5203: Alternatively you could start at the preferred index, increment and modulo and then use getting the start index as the exit condition.  I think that would remove the modulo from getFaxbot and make it a little clearer what is going on and why.

src/net/sourceforge/kolmafia/textui/command/FaxbotCommand.java
```diff
@@ -20,7 +21,12 @@ public void run(final String cmd, final String command) {
 
     boolean tried = false;
 
-    for (FaxBot bot : FaxBotDatabase.faxbots) {
+    // Get the faxbot the user last looked at
+    int preferedBot = Math.max(Preferences.getInteger("faxbots"), 0);
+
+    // Iterate from 0
+    for (int i = 0; i < FaxBotDatabase.faxbots.size(); i++) {
+      FaxBot bot = FaxBotDatabase.getFaxbot(i + preferedBot);
```
jaadams5, line 29: Is there another implementation where the code that iterates from the preferred index only has to be written once?

src/net/sourceforge/kolmafia/swingui/FaxRequestFrame.java
libraryaddict: Added as a user faxbot, though it could be added as a global

jaadams5:
I feel kind of bad only thinking of this now, but the preference depends upon the order of the bots in faxbots.txt.  That file can change.  I seem to recall long ago discussion about being able  to support user overrides but I don't recall the resolution.  So the preference needs to be the name of the preferred faxbot and not the index that a faxbot had  in one iteration of faxbots.txt,  A circular iteration starting at the preferred bot does make sense and should be preserved.

libraryaddict:
The current PR will now store "lastSelectedFaxbot" which is used only for purposes of the GUI
And "lastSuccessfulFaxbot" which is updated when a faxbot delivers a monster.
As such, the user can no longer define a default which can be relied on.

Quick example is a monster that their preferred faxbot does not have, but another does. If the fax is successful, their new preferred faxbot would be the second faxbot.

src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
```diff
@@ -121,9 +122,23 @@ public static final FaxBot getFaxbot(final String botName) {
     return null;
   }
 
-  public static final String botName(final int i) {
-    FaxBot bot = FaxBotDatabase.getFaxbot(i);
-    return bot == null ? null : bot.name;
+  public static final List<FaxBot> getSortedFaxbots() {
+    // Get preferred faxbot or null
+    FaxBot preferred = getFaxbot(Preferences.getString("lastSuccessfulFaxbot"));
+    // Use original list
```
jaadams5, line 128: Given that "preferred" faxbot only applies to the one selected when the GUI is opened, I'd be fine with changing the comment and the name of the variable.  NitPickers R Us.

src/data/defaults.txt
```diff
@@ -995,7 +995,9 @@ user	lastSlimeVial3893
 user	lastSlimeVial3894
 user	lastSlimeVial3895
 user	lastSlimeVial3896
+user	lastSelectedFaxbot
```
jaadams5, line 998: I'd be OK if this defaulted to 0.  That suggests to the careful reader that this is an index and not a faxbot name,

test/net/sourceforge/kolmafia/textui/command/FaxbotCommandTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class FaxbotCommandTest extends AbstractCommandTestBase {
+  public FaxbotCommandTest() {
+    this.command = "faxbot";
+  }
+
+  @BeforeAll
+  public static void beforeAll() {
+    KoLCharacter.reset("testUser");
```
jaadams5, line 27: There is now an unstated preference that the username in a test can be mapped to a specific file that contains one or more tests using that name.  This helps identify tests that leak or fail to clean up completely,

jaadams5 (commented):
I'm going to see if anyone else chimes in but I will circle back and approve if no one comments or beats me to it.

jaadams5 (approved):

PR #2231, opened 2024-03-09
WereProfessor can do COOK and MIX. Track Advance Researched monsters

Veracity0:
1) WereProfessor does not need to buy a range or cocktailcrafting kit to do basic COOK or MIX concoctions. That includes Cookbat food. You presumably still need one for advanced recipes, like unstable fulminate, but I have not checked yet.
2) wereProfessorAdvancedResearch is a (sorted by monsterId) property of comma separated monsterIds that your Mild Mannered Professor has done Advance Research on.
3) If you are a Mild-Mannered Professor equipped with an oculus, Stationary buttons now include an "advanced research" button. This will be disabled if you have already researched this monster.
4) Similarly for KoL's built-in "Advanced Research" button. Since attempting to use it will give an error message, we'll just delete it.

### DIFF
--- a/src/net/sourceforge/kolmafia/webui/FightDecorator.java
+++ b/src/net/sourceforge/kolmafia/webui/FightDecorator.java
@@ -300,11 +300,7 @@ public class FightDecorator {
     if (!FightRequest.hasResearchedMonster(monster.getId())) {
       return;
     }
-    String form = matcher.group(0);
-    String replacement =
-        StringUtilities.singleStringReplace(
-            form, "class=button", " disabled style='color=gray' class=button");
-    buffer.replace(matcher.start(), matcher.end(), replacement);
+    buffer.delete(matcher.start(), matcher.end());
   }
 
   private static void decorateHauntedKitchen(final StringBuffer buffer) {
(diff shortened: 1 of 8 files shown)


src/net/sourceforge/kolmafia/webui/FightDecorator.java
```diff
@@ -271,6 +276,37 @@ private static void decorateSourceAgent(final StringBuffer buffer) {
...
+      return;
+    }
+    // If we have previously researched this monster, disable the Advanced Research button
+    if (!FightRequest.hasResearchedMonster(monster.getId())) {
+      return;
+    }
+    String form = matcher.group(0);
+    String replacement =
+        StringUtilities.singleStringReplace(
+            form, "class=button", " disabled style='color=gray' class=button");
```
midgleyc, line 306: style takes CSS, so I think this should be `style='color:gray'`?

Veracity0: That is correct. Thanks. However, I have a better solution. I'm going to simply delete the button, just like KoL does in subsequent rounds after you've done Advanced Research. Why bother showing it - even disabled - if using it would simply give an error message.

midgleyc (approved):

PR #2239, opened 2024-03-16
Correctly mark vararg parameters in TypescriptDefinition

oxc:
Currently parameters that take a variadic number of arguments are marked as regular array types, e.g. `min(arg1: number, arg2: number[])`.

That signature only lets you call as `min(1, [2, 3])`, which is incorrect and fails.

Instead, it should be marked as `min(arg1: number, ...arg2: number[])`, otherwise the method cannot be called correctly from Typescript.

(I considered introducing an inner Parameter helper class, but that makes the descriptiveParamNames and combatFilterType overrides less convenient to handle. Let me know if I should give it a go.)

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
+++ b/src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
@@ -22,6 +22,7 @@ import net.sourceforge.kolmafia.textui.parsetree.Function;
 import net.sourceforge.kolmafia.textui.parsetree.LibraryFunction;
 import net.sourceforge.kolmafia.textui.parsetree.RecordType;
 import net.sourceforge.kolmafia.textui.parsetree.Type;
+import net.sourceforge.kolmafia.textui.parsetree.VarArgType;
 import net.sourceforge.kolmafia.textui.parsetree.VariableReference;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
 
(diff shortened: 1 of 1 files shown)


gausie:
This is great. I probably want to blacklist min and max though, in line
with other functions with equivalents in the JS standard lib

oxc:
Unfortunately JavaScript number cannot handle the full equivalent of a Java long, so in some rare cases you have to revert to using kolmafia's methods even if it seems you could do the same in JavaScript. I will search for the example in my code...

oxc:
Ok, nevermind, this was based on a misconception of mine. I had problems with passing larger numbers to `toString`, which would result in an "Invalid format pattern" exception, and seemingly using `toInt` on that number fixed it, but of course it just became a very different number, because IEEE 754 double simply cannot represent it:

```
> js toString(toInt(1234567890123456), "%,d");

Returned: 1,234,567,948,140,544
```

I guess I just noticed that the exception is gone, but simply never looked at the output 😂

oxc:
Ok, now I'm confused, this number is smaller than Number.MAX_SAFE_INTEGER. What am I missing?

oxc:
Ah, OK, I get it. Rhino returns integer values up to Integer.MAX_VALUE as Integer, which the Javascript engine returns as long. Starting from Integer.MAX_VALUE+1, Rhino returns it as double, because the Java Integer type no longer can represent it. The JavascriptRuntime converts that to float (for reasons of compatibility with ash data types, I assume), and that can't represent many of those larger integers either :)

```
> js 2147483647

Returned: 2147483647

> js 2147483648

Returned: 2147483648.0
```

I will prepare a PR to fix that.

oxc:
> This is great. I probably want to blacklist min and max though, in line with other functions with equivalents in the JS standard lib

That probably makes sense. I would still like to keep this because it is more correct, and we might have some real use case for varargs in the future.

gausie (approved):
Thank you for this, and for being perhaps the first person to read through and understand this class :)

PR #2240, opened 2024-03-16
Support large numbers from/to Javascript runtime

oxc:
The Rhino javascript engine passes integers larger than `Integer.MAX_VALUE` as `Double`, because they cannot be represented as `Integer`. Since we already use `long` for integer values, pass all numbers that are "safe integers" based on Javascript's Number.isSafeInteger as `long`, instead of converting them to a `float` value that cannot even correctly represent many of the larger integer values.

I added a test in a separate commit, so you can easily compare the behavior before and after the change.

Update: also, use double precision for non-integer numbers, since TYPE_FLOAT uses `double` internally anyway.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/javascript/JavascriptNumbers.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/JavascriptNumbers.java
@@ -1,0 +1,23 @@
+package net.sourceforge.kolmafia.textui.javascript;
+
+/** This mimics the behavior of the Rhino NativeNumber class */
+@SuppressWarnings("UnnecessaryUnboxing")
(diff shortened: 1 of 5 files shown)


oxc:
I've adjusted the ASHInterop test, but I'm a bit confused as to why there was a test that names the previous behavior explicitly as "correctly". It seems more correct to me now.

oxc:
I've added another change to use full double precisions for all non-integer values. TYPE_FLOAT values use double internally anyway, so no need to downcast to float.

gausie (approved):
Merging. If anyone is relying on the previous behaviour, we need failing unit tests before we can make a plan probably

PR #2244, opened 2024-03-22
Fix processing private chat messages when user has a space in their name

oxc:
When processing an incoming message, the ChatManager checks whether the recipient equals the currently running user, and takes that as a sign to handle the message as a private message. However, the `recipient` has spaces replaced with underscores, so if the current user has spaces in their usernames, that check won't match.

There is already the same check for triggering a bell in the chat window, and it seems nicer to pull that blocks together anyway.

This PR includes another tiny fix for injecting messages into mchat (e.g. those sent from the GUI chat), which requires timestamps to be in seconds, not millis.


midgleyc (approved):
Thanks!

PR #2245, opened 2024-03-24
Fix Disco Nap tagged incorrectly as a passive skill

Donald-Lafleur:
Disco Nap is marked as both a passive and nc skill, which is a holdover from the previous system for organizing skills into groups by type. It being marked as passive incorrectly sometimes causes autoscend to loop infinitely in the G-Lover path during restoration. [https://github.com/loathers/autoscend/issues/1419](https://github.com/loathers/autoscend/issues/1419)

### DIFF
--- a/src/data/classskills.txt
+++ b/src/data/classskills.txt
@@ -473,7 +473,7 @@
 5004	Nimble Fingers	glove.gif	passive	0	0	Level: 6
 5005	Disco Dance of Doom	dance1.gif	combat	3	0	Level: 2
 5006	Mad Looting Skillz	chest.gif	passive	0	0	Level: 8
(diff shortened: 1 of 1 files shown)


Donald-Lafleur:
Presumed this wouldn't break a bunch of tests, and was apparently very incorrect. I'll dig into the failing tests to see what the issues are and if I can resolve them. I have a feeling Disco Nap is a special edge case since I couldn't find any other examples of skills marked as both passive and nc.

Donald-Lafleur:
Since Disco Nap apparently needs to be tagged as passive to have it's free rest counted correctly I'm just going to bandaid this in autoscend with a special Disco Nap exception to auto_have_skill.

PR #2246, opened 2024-03-26
antiScientificMethod property tracks smashed scientific equipment

Veracity0:


jaadams5 (dismissed):
I'll approve but I think I'd go for another delimiter.   I noticed sometimes the Beast "smells" scientific equipment and there is a message to that effect.  That might be worth tracking although it is out of scope here.

jaadams5 (approved):

PR #2248, opened 2024-03-28
Improve Relay Warning for adventuring in Spookyraven Cellar

Veracity0:
Rather than simply offering the chance to adventure in the Haunted Wine Cellar or Haunted Laundry Room if you have not read the mortar-dissolving recipe with Lord Spookyraven's Spectacles equipped, offer to get the recipe (which will autoread it, if autoQuest is true), equip specs, read the recipe.

It will do none of the above if you don't have the spectacles.


jaadams5 (approved):
Cartesian test - new tool for me.  Thank you.

PR #2257, opened 2024-04-01
fix: fix breathitinCharges tracking

oxc:
Breathitin™ charges are also spent when another reason caused an outdoor adventure to be free (https://kol.coldfront.net/thekolwiki/index.php/Breathitin%E2%84%A2#Notes). In this case, the Breathitin message does not show, so tracking is incorrect in these cases. Simply always try to decrement the breathitinCharges counter in outdoor locations.


oxc:
Does this make sense, or is there a better place to track this?

midgleyc (approved):
Thanks!

I think keeping it with the check for the actual text is best.

PR #2258, opened 2024-04-01
Rufus salad

Veracity0:
```
"First thing. Salad a big shadow entity. Salad the place salad rifts salad to. It's a tongue. Salad. a big tongue, salad far as Salad can salad He pauses.

"I've also salad an artifact I need somebody to recover for me. A shadow bucket," salad continues.

After salad moment Salad speaks again. "I can also always salad salad of more mundane items from the rifts. Right now, 3 salad salad would be valuable."
```

Figuring out what "3 salad salad" is fails - and previously threw an NPE.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/RufusManager.java
+++ b/src/net/sourceforge/kolmafia/session/RufusManager.java
@@ -62,8 +62,11 @@ public class RufusManager {
       String items = itemMatcher.group(1);
       // Convert to itemId and back to non-plural item name
       int itemId = ItemDatabase.getItemId(items, 3);
-      String itemName = ItemDatabase.getItemName(itemId);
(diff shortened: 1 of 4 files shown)


Rinn:
we could maybe automatically hit the quest log to get the correct item when this happens

Veracity0:
This is parsing the choice page when you visit Rufus.
You have not yet accepted the quest, and therefore nothing is in the Quest Log.
Even if it were, you can't walk away from this choice, so we couldn't visit the Quest Log at this time.

A script that automates Rufus could do something like that, if its desired goal was obscured by salads, except it would probably just be best to exit the choice and retry visiting until the desired quest is not obscured by salad.

The purpose of this PR is to not throw an NPE if the item name has been vandilized with salad.

midgleyc (approved):

PR #2259, opened 2024-04-02
feat: apriling band skeleton

midgleyc:
Skeleton the items / effects.

Going to inventory.php?action=apriling allows you to:
* change intrinsic every 11 turns
* select two items, each of which has a 3/day effect
  * one of which is an NC force

I'll probably create a command for picking instruments / effects if nobody else gets to it first.

So there'll be a fair amount of follow up.


Veracity0 (approved):
Thank you.

PR #2260, opened 2024-04-02
Handle Evilometer reporting evilness = 999

Veracity0:
We have 5 properties tracking Cyrpt evilness:
```
user	cyrptAlcoveEvilness	0
user	cyrptCrannyEvilness	0
user	cyrptNicheEvilness	0
user	cyrptNookEvilness	0
user	cyrptTotalEvilness	0
```
Those values are appropriate unless the Cyrpt quest is active.
When you start the quest, total is set to 200 and the others are set to 50.
When you defeat evil, the appropriate values count down.
If you look at the Evilometer at any time, we set/correct the values to what it says.

The cyrptTotalEvilness property is funky. When it counts down to zero, the Evilometer (and charpane) report it as 999, because the Bonerdagon is available. However, KoLmafia has it at 0, which is indistinguishable from the quest being complete. This PR has a few tweaks:

1) The candy cane sword decremented cranny evilness by 11, but not total evilness.
Fixed.
2) If you visit the Evilometer when the Bonerdagon is available, it reports Total: 999 and Haert: 999.
It now sets cyrptTotalEvilness as 999 - which tells you the Bonerdagon is available without having to check quest progress.
3) RequestEditorKit showed an empty Evilometer when only the Bonerdagon was available, if totalEvilness was not 0. It put it in the wrong place, since the table it was adding it to did not appear in the Cyrpt if the Bonerdagon is available.
It now checks for totalEvilness of 0 OR 999.
It also detects if the table is not there and does not screw up the page by inserting the Evilometer image at "<htm" in the page text.

I primarily did this in order to fix the candy cane sword and the display errors, but I suppose additional work could be done to  set totalEvilness to 999 as soon as we count down the last corner counter to 0.
Do we correct values when we visit the Cyrpt and detect that corners are empty and/or the Haert is visible?

In order to do those things, additional HTML test fixtures are needed - which I cannot get until next run - so this is it for now.


gausie (approved):
Looks good to me. We could perhaps have an EvilometerManager that handled the derived cryptTotalEvilness prop for us

PR #2263, opened 2024-04-03
feat: prefs for apriling band helmet

midgleyc:
* You can conduct every 11 turns. This does not reset on rollover.
* You can walk away from the conduct choice.
* You can successfully use each instrument three times per day. Getting a duplicate instrument does not increase the number of times you can use the instrument.

Tested manually, unit tests to come after I get the necessary HTML.


midgleyc:
There are three ways of doing the preference for the next conductance:
* just the number (e.g. `encountersUntilYachtzeeChoice`)
* the turn we last conducted (e.g. `lastLightsOutTurn`)
* the turn we can next conduct (e.g. `nextQuantumFamiliarTurn`)

I went for the last.

jaadams5 (dismissed):

PR #2267, opened 2024-04-05
Improve tracking of shadow brick against unkillable monsters

oxc:
When using a shadow brick against a monster that cannot be insta-killed, it still counts against the limit.

The check needs to be added outside the "If these items succeed, then the second one won't actually do anything" block, because against such monsters, the shadow brick can indeed be dual-used (and counts twice):

image

With the fix, the following situations are tracked correctly:

* Dual-using against a monster that can be insta-killed (single use should work equivalently):
```
Encounter: Fallen Archfiend
...
Round 1: That Buff Guy uses the shadow brick and uses the shadow brick!
...
Round 2: That Buff Guy wins the fight!
...
Preference _shadowBricksUsed changed from 8 to 9
...
This combat did not cost a turn
```

* Dual-using against Eldritch Tentacle:
```
Round 0: That Buff Guy uses the shadow brick and uses the shadow brick!
Encounter: Eldritch Tentacle
Round 2: Eldritch Tentacle takes 69 damage.
Round 2: Eldritch Tentacle takes 69 damage.
...
Preference _shadowBricksUsed changed from 9 to 10
Preference _shadowBricksUsed changed from 10 to 11
```

* Using when already maxed out (but was mistracked by a kolmafia version that did not have this fix):
```
Encounter: Fallen Archfiend
Round 0: That Buff Guy wins initiative!
Round 1: You lose 1 hit point
Round 1: That Buff Guy uses the shadow brick!
KoLmafia thinks it is round 2 but KoL thinks it is round 1
Preference _shadowBricksUsed changed from 11 to 13
```

### DIFF
--- a/src/net/sourceforge/kolmafia/request/FightRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FightRequest.java
@@ -10614,6 +10614,7 @@ public class FightRequest extends GenericRequest {
 
     boolean itemSuccess = isItemSuccess(responseText);
     boolean itemRunawaySuccess = isItemRunawaySuccess(responseText);
+    boolean itemLimitMaxedOut = false;
 
     switch (itemId) {
       default:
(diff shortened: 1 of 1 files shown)


midgleyc (approved):
Thanks!

PR #2269, opened 2024-04-07
feat: apriling band command

midgleyc:
Command for getting an effect or instrument, or playing an instrument.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/AprilBandCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/AprilBandCommand.java
@@ -147,10 +147,13 @@ public class AprilBandCommand extends AbstractCommand {
 
     KoLmafia.updateDisplay("Playing " + item.getName());
 
-    var request = new GenericRequest("inventory.php");
-    request.addFormField("pwd", GenericRequest.passwordHash);
-    request.addFormField("iid", Integer.toString(instrument));
-    request.addFormField("action", "aprilplay");
+    var request =
+        new GenericRequest(
+            "inventory.php?action=aprilplay&iid="
+                + instrument
+                + "&pwd="
+                + GenericRequest.passwordHash,
+            false);
     RequestThread.postRequest(request);
 
     KoLCharacter.updateStatus();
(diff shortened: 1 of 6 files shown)


src/net/sourceforge/kolmafia/textui/command/AprilBandCommand.java
```diff
@@ -0,0 +1,161 @@
...
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import net.sourceforge.kolmafia.request.GenericRequest;
+import net.sourceforge.kolmafia.session.ChoiceManager;
+import net.sourceforge.kolmafia.session.InventoryManager;
+
+public class AprilBandCommand extends AbstractCommand {
+  public AprilBandCommand() {
+    this.usage =
+        "effect [nc|c|drop] | item [instrument] | play [instrument] - participate in the apriling band";
```
jaadams5, line 16: Apriling band?  My sense is that Apriling is always used as a proper noun and hence capitalized.  Non-blocking.

test/net/sourceforge/kolmafia/textui/command/AprilBandCommandTest.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    @Test
+    void requiresConductAvailable() {
+      var cleanups =
+          new Cleanups(withHelmet(), withTurnsPlayed(0), withProperty("nextAprilBandTurn", 10));
+
+      try (cleanups) {
+        String output = execute("effect");
+        assertThat(output, containsString("You cannot change your conduct (10 turns to go)"));
```
jaadams5, line 66: If you change display text as suggested then  this will have to change too.

jaadams5 (approved):
Comments are cosmetic - user displayed text that could be changed.

src/net/sourceforge/kolmafia/textui/command/AprilBandCommand.java
midgleyc: from looking at other files it seemed the standard is that instructions are always lowercase.

I'm not sure this is always, or even often true.

midgleyc:
Thanks -- I intended the text as it is, but I get how it could be confusing.

I expect I like puns more than the average person.

src/net/sourceforge/kolmafia/textui/command/AprilBandCommand.java
jaadams5: OK.  I was looking at the failure message but...

jaadams5:
I am notorious for my tolerance for puns but I just wasn't thinking of that here.  No problems.

PR #2270, opened 2024-04-08
Improvement for JS ecosystem

gausie:
Currently kolmafia can't be imported outside of a KoLmafia environment because ESM knows that the named exports aren't actually there. I would like to do this so I can share code between the web and a mafia script, knowing that I will not actually execute mafia-specific code. This fixes that


jaadams5 (dismissed):
Not that I totally understand what is happening but I trust that it works and doesn't break anything "normal" KolMafia users use.

gausie:
Without usage statistics I would imagine most KoLmafia users execute JS scripts at least once day, which this could affect if it were wrong.

gausie:
Small mistake in the github action, I will push to main to fix.

Veracity0:
> Without usage statistics I would imagine most KoLmafia users execute JS scripts at least once day, which this could affect if it were wrong.

I would be astonished if this were true.

Perhaps you mean "most KoLmafia users who are on Discord" - which I am sure is a minority of all KoLmafia users.

gausie:
No I didn't mean that at all. I really just mean garbage-collector which is
frequently mentioned in in-game chat, moreso than any other script I see
referenced by name. But again, my anecdote and I could easily be wrong.
Just a comment in passing really.

midgleyc:
Without usage statistics I expect most KoLmafia users don't execute any scripts at all.

PR #2276, opened 2024-04-11
update backup prefs at logout rather than login

Malibu-Stacey:
Login is already too late for lost/wiped preferences as since everything is reset at that point, what would be a useful backup gets wiped & copied over with the defaults rendering the entire process moot.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -3303,6 +3303,7 @@ user	choiceAdventure1489	3
 user	choiceAdventure1491	0
 user	choiceAdventure1494	0
 user	choiceAdventure1505	0
+user	choiceAdventure1528	1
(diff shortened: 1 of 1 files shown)


jaadams5:
@heeheehee-kolmafia said they had an idea about the root cause and so I have been waiting for their fix.  Since it hasn't happened yet I have increased my efforts to figure it out (again) but this might not be a problem that I am able to solve.

Note that the automated backup and restore was always understood to be a stopgap measure and moving when things happen (or changing how often) is not really going to solve the problem of a restore from a corrupted file as much as it is going to change when it could happen.

gausie:
@Malibu-Stacey is this going anywhere or shall we close?

PR #2278, opened 2024-04-18
Fix typo in aprilband command usage

Donald-Lafleur:
The usage string as is makes the help print out as
`aprilbandeffect [nc|c|drop] | item [instrument] | play [instrument] - participate in the apriling band`


midgleyc (approved):
Thanks!

PR #2284, opened 2024-04-28
Mark typescript parameters as optional

oxc:
* Merge related library function overloads into a single signature with optional parameters
* Strip trailing undefined function args

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
+++ b/src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
@@ -24,7 +24,6 @@ import net.sourceforge.kolmafia.textui.parsetree.RecordType;
 import net.sourceforge.kolmafia.textui.parsetree.Symbol;
 import net.sourceforge.kolmafia.textui.parsetree.Type;
 import net.sourceforge.kolmafia.textui.parsetree.VarArgType;
-import net.sourceforge.kolmafia.textui.parsetree.VariableReference;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
 
 public class TypescriptDefinition {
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
```diff
@@ -132,28 +138,82 @@ private static String getReturnType(Function f) {
...
+
+      var deprecationWarning =
+          (f.deprecationWarning.length > 0)
+              ? "/** @deprecated " + String.join("<br>", f.deprecationWarning) + " */\n"
+              : "";
+
+      result.add(
+          String.format(
+              "%sexport function %s(%s): %s;", deprecationWarning, functionName, params, type));
+    }
```
gausie, line 214: I would like to see this function split out into some smaller chunks if possible

src/net/sourceforge/kolmafia/textui/javascript/ValueConverter.java
```diff
@@ -217,6 +218,8 @@ else if (object instanceof Boolean) {
       return convertNativeArray((NativeArray) object, typeHint);
     } else if (object instanceof Value) {
       return (Value) object;
+    } else if (Undefined.isUndefined(object)) {
+      return null;
```
gausie, line 222: This is a bit weird because `null` and `undefined` are different in JS. What's the benefit here?

test/net/sourceforge/kolmafia/textui/command/JavaScriptCommandTest.java
```diff
@@ -26,7 +26,7 @@ public void printsNull() {
     public void printsUndefined() {
       String output = execute("undefined");
 
-      assertThat(output, startsWith("Returned: org.mozilla.javascript.Undefined"));
+      assertThat(output, startsWith("Returned: null"));
     }
```
gausie, line 30: As per my previous comment, this test doesn't look like expected behaviour. It looks like it would occlude a function that can meaningfully return either null or undefined, however silly such a function would be.

gausie (changes requested):
This is fantastic! I love the addition of namedParam to RuntimeLibrary. Thank you for this. Couple of comments.

src/net/sourceforge/kolmafia/textui/javascript/ValueConverter.java
oxc: Well, at least it does not get returned as a string. How would you want to handle it?

src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
oxc: Fair enough, I will at least split the overload selection/preparation and the formatting. That was never a good idea to merge them anyway.

That will probably mean another helper type for the function, though.

src/net/sourceforge/kolmafia/textui/javascript/ValueConverter.java
oxc: If anything, I would prefer to throw an exception. Neither Java nor Ash knows the concept of undefined in addition to null, so I'm not sure we can gain anything here.

gausie: Maybe the issue is that we can coerce it to null operationally but we need to report it as undefined in the cli?

src/net/sourceforge/kolmafia/textui/TypescriptDefinition.java
oxc: I've split this up in formatting, comparing and merging.

src/net/sourceforge/kolmafia/textui/javascript/ValueConverter.java
oxc: I'll figure something out later this week.

oxc: I think this will work: https://github.com/kolmafia/kolmafia/pull/2291

test/net/sourceforge/kolmafia/textui/command/JavaScriptCommandTest.java
oxc: I've proposed a better solution in https://github.com/kolmafia/kolmafia/pull/2291

PR #2285, opened 2024-04-28
Let runtime library functions ambiguate their parameter names

oxc:
This commit is part of #2284, but given that it affects not only the Javascript system, I've decided to pull it out into a separate PR.

---

In cases where a runtime library function has multiple overloads with the same arity, the parameter names cannot be nicely resolved from the implementation function, because they have different meaning depending on the selected type signature.
Instead of hardcoding this in TypescriptDefinition, let function definition include parameter names where this makes sense.

Unfortunately, this requires some extra pass-through constructors in LibraryFunction. At least one of them can resolved once we switch to JDK22 and statements before the this() call are allowed.

These parameter names will also show up in ashref/jsref, but unfortunately the reflection-based parameter names don't show up, because KoLmafia is only compiled with the required `-parameters` switch for tests and tsDef (see [build.gradle#L285-L289](https://github.com/kolmafia/kolmafia/blob/83c9e15ddf56c478778399086b489294ef513fce/build.gradle#L285-L289)). Is there a good reason not to enable this generally?

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
+++ b/src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
@@ -192,6 +192,8 @@ import net.sourceforge.kolmafia.textui.parsetree.RecordType;
 import net.sourceforge.kolmafia.textui.parsetree.RecordValue;
 import net.sourceforge.kolmafia.textui.parsetree.Type;
 import net.sourceforge.kolmafia.textui.parsetree.Value;
+import net.sourceforge.kolmafia.textui.parsetree.Variable;
+import net.sourceforge.kolmafia.textui.parsetree.VariableReference;
 import net.sourceforge.kolmafia.utilities.CharacterEntities;
 import net.sourceforge.kolmafia.utilities.ChoiceUtilities;
 import net.sourceforge.kolmafia.utilities.FileUtilities;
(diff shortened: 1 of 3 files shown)


oxc:
Could someone please re-run the tests on windows? They pass fine here: https://github.com/oxc/kolmafia/actions/runs/8869700421/job/24350661462

oxc:
For completeness sake, I want to provide an alternative suggestion: instead of mixing reflection and explicitly named parameters, we could just switch all runtime functions to include their parameter names. Let me know if I should add this to this PR.

gausie:
> For completeness sake, I want to provide an alternative suggestion: instead of mixing reflection and explicitly named parameters, we could just switch all runtime functions to include their parameter names. Let me know if I should add this to this PR.

I do think your proposed alternative is probably better, but not necessary for right now if you don't feel up to it.

oxc:
They would already have shown up in ashref, but your guess might still be true from earlier times.

If nobody objects, I might still add namedParams to all functions, in a couple of days.

gausie:
Please do!

oxc:
I've added param names for all functions. I have not provided any new names (except for the "adventure" function in the last commit for demonstration purposes).

gausie (approved):
Thank you, this rocks!

PR #2287, opened 2024-05-01
Provide meaningful param names to runtime library functions

oxc:
Follow-up to #2285 

Induce parameter names from a) their type, b) the function body, or c) the wiki entry.


midgleyc (approved):
Thanks!

PR #2291, opened 2024-05-01
Represent JS undefined as void, not string

oxc:
Currently, JS `undefined` is silently converted into a string representation of Rhino's Undefined instance.
This can in theory lead to wrong function invocations if the target type is a string.

Instead, return a special undefined value with a special undefined VOID type. This prints as undefined in the console, to make it transparent what is going on.
Other than that, it behaves as void. This is consistent with JS methods returning `undefined` if no value was returned, and TypeScript void methods returning `undefined`.

To prevent any confusion, forbid undefined values from being passed to Ash functions (for now).


oxc:
That's the same test that failed last time on Windows. Seems flaky.

gausie (approved):
Great solution

PR #2294, opened 2024-05-01
Add mayam command

gausie:
Allows

```
mayam eye yam eyepatch yam
```


src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
```diff
@@ -0,0 +1,103 @@
...
+      List.of(
+          List.of("yam", "clock", "explosion"),
+          List.of("yam", "eyepatch", "cheese", "wall"),
+          List.of("yam", "lightning", "bottle", "board", "meat"),
+          List.of("yam", "sword", "vessel", "fur", "chair", "eye"));
+
+  public MayamCommand() {
+    var symbols =
+        SYMBOL_POSITIONS.stream().flatMap(List::stream).distinct().collect(Collectors.joining("|"));
+    this.usage = "[" + symbols + "] - list symbols top to bottom (largest ring to smallest)";
```
midgleyc, line 25: needs to start with a space or it doesn't display correctly.

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
```diff
@@ -0,0 +1,103 @@
...
+import net.sourceforge.kolmafia.preferences.Preferences;
+import net.sourceforge.kolmafia.request.GenericRequest;
+import net.sourceforge.kolmafia.session.InventoryManager;
+
+public class MayamCommand extends AbstractCommand {
+  private static final List<List<String>> SYMBOL_POSITIONS =
+      List.of(
+          List.of("yam", "clock", "explosion"),
+          List.of("yam", "eyepatch", "cheese", "wall"),
+          List.of("yam", "lightning", "bottle", "board", "meat"),
```
midgleyc, line 19: the image name is board.png, but the hover text is "Wood".

I think this one should be "wood" instead of "board".

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
```diff
@@ -0,0 +1,103 @@
...
+    if (!InventoryManager.equippedOrInInventory(ItemPool.MAYAM_CALENDAR)) {
+      KoLmafia.updateDisplay(MafiaState.ERROR, "You need a Mayam Calendar");
+      return true;
+    }
+    return false;
+  }
+
+  @Override
+  public void run(final String cmd, String parameters) {
+    String[] symbols = parameters.split(" ");
```
midgleyc, line 38: should verify that there are exactly four symbols passed in, and fail otherwise.

If less, it'll use the default ring symbols. If more, it'll crash.

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
```diff
@@ -0,0 +1,103 @@
...
+public class MayamCommand extends AbstractCommand {
+  private static final List<List<String>> SYMBOL_POSITIONS =
+      List.of(
+          List.of("yam", "clock", "explosion"),
+          List.of("yam", "eyepatch", "cheese", "wall"),
+          List.of("yam", "lightning", "bottle", "board", "meat"),
+          List.of("yam", "sword", "vessel", "fur", "chair", "eye"));
+
+  public MayamCommand() {
+    var symbols =
```
midgleyc, line 23: might want to reverse the list, because you have to input the symbols largest ring to smallest, but they're displayed smallest ring to largest.

gausie: I want to do it such that ‘eye yam eyepatch yam’ would go in that order

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
C2Talon: A javascript var on the page keys it to "board", but the value is "Wood":

`var rings = [{"yam":"Yam","clock":"Clock","explosion":"Explosion"},{"yam":"Yam","eyepatch":"Eyepatch","cheese":"Cheese","wall":"Wall"},{"yam":"Yam","lightning":"Lightning","bottle":"Bottle","board":"Wood","meat":"Meat"},{"yam":"Yam","sword":"Sword","vessel":"Vessel","fur":"Fur","chair":"Chair","eye":"Eye"}];`

But then everything the user sees calls it "wood" except the image name, so would "wood" probably be better.

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
midgleyc: the order is fine, it's the help text that starts "yam|clock|explosion", but I'm saying explosion should be one of the last instead of one of the first

midgleyc:
Also, can I request a change to the input?

I think it would be nice if this command supported resonances in the future (doesn't have to be this PR).

In that case, a command might be `mayam resonance battery`. So instead of `mayam eye yam eyepatch yam`, perhaps `mayam rings eye yam eyepatch yam`?

Alternatively, it could be assumed that if the first parameter isn't "resonance" it's the normal case, but I think most commands with subcommands do them explicitly.

src/net/sourceforge/kolmafia/textui/command/MayamCommand.java
gausie: I used that var to build this list, yeah. But I can change it to wood

midgleyc (approved):
Thanks!

PR #2295, opened 2024-05-02
Add curse scripting command

oxc:
It's a replacement for calling the CLI throw command that makes parameter handling more straight-forward.


gausie:
Can we get some unit tests for the function please

oxc:
> Can we get some unit tests for the function please

I've added some tests

midgleyc (approved):

PR #2302, opened 2024-05-03
feat: consider free rests from mayam chair

midgleyc:


gausie (commented):
Hmmm. If you reset your calendar I assume you keep the old rests. We need to support resetting anyway, but we should probably track the rests separately.

midgleyc:
ah yes, and presumably you can do the chair twice for +10 rests?

gausie:
I haven't tested, but I assume/hope so! If you can't I'll take it up as a bug with CDM.

gausie (approved):
Hadn't spotted the changes since my last review, apols.

PR #2308, opened 2024-05-04
Check Standard restrictions for AdventureResults by the item's dataName

Veracity0:
You have to look up Standard restrictions using a String.
For items, that is the item's dataName.
We have many instances of using a constant item name. That works.

However, we have a number of instances where we have an AdventureResult and use the  "name" field - which is not fine for Modeables like the backup camera, Jurassic Parka, or unbreakeable umbrella, where KoLmafia's name includes the mode, for user convenience.

ItemDatabase has an "isAllowed" method which accepts an itemId and looks up the dataName to pass to StandardRequest.

I added an "isAllowed" method which accepts an AdventureResult, extracts the itemId, and calls that method.
I changed all places which extract the name from an AdventureResult to call isAllowed to simply call the new method.

Net result:

- ItemManager no longer shows the backup camera in Storage if you are in a Standard Run
- Maximizer no longer suggests pulling the backup camera if you are in a Standard run.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLAdventure.java
+++ b/src/net/sourceforge/kolmafia/KoLAdventure.java
@@ -47,7 +47,6 @@ import net.sourceforge.kolmafia.request.PyramidRequest;
 import net.sourceforge.kolmafia.request.QuestLogRequest;
 import net.sourceforge.kolmafia.request.RichardRequest;
 import net.sourceforge.kolmafia.request.SpelunkyRequest;
-import net.sourceforge.kolmafia.request.StandardRequest;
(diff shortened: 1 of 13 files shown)


gausie:
Good job solving this. Would you be able to add a quick regression test just in case?

Veracity0:
I changed all the caller where an AdventureResult was available - and apparently SOME existing tests caught that.
For some reason, git is not showing me the ones that are not caught.

Still, a regression test for the Maximizer and the backup camera should be easy enough to make.
Chances are, that will not increase coverage, but it will verify the desired result. :)

jaadams5 (approved):

PR #2314, opened 2024-05-11
Improve (vararg) function matching

oxc:
This PR fixes one minor bug in function parameter matching (if a function gets zero arguments for a vararg parameter, it ignores all previous (un)matched parameters). See before and after tests for the change in behavior.

Additionally, it improves parameter matching by ~~a) short-circuiting some scenarios that mismatch a lot when checking multiple overloads of a given function, and b)~~ letting the function know itself whether it has variadic parameters, instead of letting the caller check both scenarios.

I'm aware that this PR mixes several topics, so let me know if you want me to split it up. As you probably know, I'm happy to do a lot of granular PRs, but these are kind of related, so having that bunch in one squashed commit makes sense.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
@@ -62,9 +62,7 @@ public abstract class AshStub extends BaseFunction {
           }
         }
 
-        // Check for match with no vararg, then match with vararg.
-        if (testFunction.paramsMatch(coercedArgs, matchType, /* vararg= */ false)
-            || testFunction.paramsMatch(coercedArgs, matchType, /* vararg= */ true)) {
+        if (testFunction.paramsMatch(coercedArgs, matchType)) {
           return testFunction;
         }
       }
--- a/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
@@ -258,14 +258,18 @@ public abstract class Function extends Symbol {
       }
     }
 
-    if (refIterator.hasNext()) {
-      // If the next parameter is a vararg, this is
-      // allowed if we ran out of parameters.
+    if (vararg == null && refIterator.hasNext()) {
+      // If the next parameter is a vararg, and there was none yet,
+      // this is allowed if we ran out of parameters.
       VariableReference currentParam = refIterator.next();
       Type paramType = currentParam.getType();
 
       if (paramType instanceof VarArgType) {
         vararg = currentParam;
+      } else {
+        // we read the last parameter, but it was not a vararg and we don't
+        // have a value for it, so this is a mismatch
+        return false;
       }
     }
 
--- a/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
@@ -264,7 +264,6 @@ public abstract class Function extends Symbol {
 
       if (paramType instanceof VarArgType) {
         vararg = currentParam;
-        matched = true;
       }
     }
 
--- a/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
@@ -168,8 +168,12 @@ public abstract class Function extends Symbol {
       VariableReference currentParam = refIterator.next();
       Type paramType = currentParam.getType();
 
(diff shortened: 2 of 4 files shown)


src/net/sourceforge/kolmafia/textui/parsetree/Function.java
gausie, line 1: There are a few lines here annotated as having no coverage from which I think we would genuinely benefit from coverage. What do you think?

oxc: I'll have a look

src/net/sourceforge/kolmafia/textui/parsetree/Function.java
```diff
@@ -179,38 +186,43 @@ private boolean paramsMatchNoVararg(final List<? extends TypedNode> params, Matc
...
+    if (refIterator.hasNext() || valIterator.hasNext()) {
+      return false;
     }
-    return false;
+
+    return true;
   }
 
   private boolean paramsMatchVararg(final List<? extends TypedNode> params, MatchType match) {
+    // on a vararg match, the last parameter can match 0 values so wee need at least X-1 values
```
midgleyc, line 215:
```suggestion
    // on a vararg match, the last parameter can match 0 values so we need at least X-1 values
```
trivial spelling issue

oxc: Just a wee typo. Will fix it in no time.

src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
```diff
@@ -63,8 +63,7 @@ private Function findMatchingFunction(List<Value> ashArgs, boolean coerceAnyType
         }
 
         // Check for match with no vararg, then match with vararg.
```
Veracity0, line 65: This comment is no longer valid.

src/net/sourceforge/kolmafia/textui/parsetree/Function.java
oxc: Several of the uncovered lines in paramsMatch* only occur if vararg is called without a vararg parameter, and vice versa. Both scenarios cannot happen anymore after these changes, I would have to mock an impossible scenario to cover them.

oxc: I could replace those scenarios with Exceptions, to make the impossibility of it happening more explicit? If it still happens, it is definitely a bug in the code.

Veracity0 (changes requested):
I like this.

I'm OK either way with having the short circuiting be included or a separate PR; that happens in the two paramsMatch methods, one of which is called depending on whether a function has a vararg or not, which is integral to the rest of the PR.

I agree that if you can extend the coverage to the handful of lines which were missed  in the existing FunctionTest, that would be sweet.

src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
oxc: Removed.

src/net/sourceforge/kolmafia/textui/parsetree/Function.java
midgleyc: IllegalStateException is how we cover the other impossible situations.

I don't personally feel that using mocks to fake impossible situations and testing the behaviour that results is valuable, but others may.

oxc:
I've applied several changes and added a small test for one of the uncovered cases.

Testing the remaining uncovered lines in the matchParam methods is very hard to do, because they are impossible to happen in the given flow. Most of the uncovered checks are double-checking what another check has already confirmed. Removing them would make it harder to see that the method is working correctly, so I don't think I would recommend that.

As for the uncovered lines in other methods, I'm not convinced that this PR is a good place to add tests for them. Please let me know if you think otherwise, and which methods exactly you would like to have covered.

gausie:
> As for the uncovered lines in other methods, I'm not convinced that this PR is a good place to add tests for them. Please let me know if you think otherwise, and which methods exactly you would like to have covered.

I think this is great

Veracity0 (approved):

midgleyc:
When I asked for the perf improvement (https://github.com/kolmafia/kolmafia/commit/8b345b2b8ea35707b84e11f4d843bfd08cca5b40) to be split out, I was expecting it to go in another PR.

I would be amenable to filing it (with your code, with some modifications) if you aren't?

oxc:
Right 😂 Out of HEAD, out of mind.

I will have some time in a couple of days to re-iterate on that.

oxc:
See #2336 for the performance PR.

PR #2318, opened 2024-05-14
Remove WereProf stat req exemption

dsimich:
The Mild-Mannered Professor's stat reduction functions more like a debuff (like Small or other stat-fixing effects) than a permanent lowering (like A Shrunken Adventure Am I). Having the added condition in meetsStatRequirements causes Mafia to want to equip items the character does not have the base stat for. This PR removes that.

An easy test is pulling something like the Mafia Thumb Ring (40 Base Mus required) and trying to equip that as a Professor at the start of the run. It will not be possible.


jaadams5:
With a couple of exceptions I have always been able to get things equipped on the Prof.   Outfits often don't work but going and equipping one thing from the GUI or the relay browser seems to get things done.  The exceptions are a couple of items that I can't equip and don't know why.  Treads of Loathing is one and a couple others are +adv gear that the Maximizer recommends but can't actually be equipped.

So at this point I am trying to figure out why I don't think I am seeing the problem that you are addressing.  I am already equipping things that have a stat requirement that the Prof does not meet.

There is an opportunity for a test here - try and equip something.  It should fail before this is added and pass after.

I will look at this and maybe even write the test but my list of KoLmafia things to do that remain undone is not shrinking and Real Life is crazy busy at the moment.

dsimich:
I noticed it was causing issues because I am adding Autoscend support and that pulls the Mafia Thumb Ring early on. Once we have it, Maximizer recommends it and then Autoscend stops because of that. Other items that I've encountered that we pull (off the top of my head) are the Pantaloons of Hatred and numberwang. Removing that condition causes no stop and no further breakage. For the outfits (specifically war outfits), I handle them piece by piece to avoid any issues with the stat requirements.

gausie:
If you believe you're fixing a bug, please add a test that fails before and passes after your change.

midgleyc:
I think the way things used to work was that when you used a base stat limiter, you couldn't equip items with high stat requirements.

It doesn't work that way now, so I think this PR is correct.

jaadams5:
I can equip a thumb ring but as I look at the relay browser there are several items where the equip link is crossed out.  Treads of Loathing is one.  So my success with equipping the ring may have been that the stat requirement was being met.  My current thinking is that there is a bug and this is the fix unless there are multiple code paths that equip things.  I think it is quit reasonable to have a failing test that passes with this fix and would recommend keeping this as a Draft until there is such a test.  I will gladly write one but getting time to do so will involve waiting days, not hours.

dsimich:
I'll be honest. I'm going to be completely useless when it comes to writing a test. I'm happy with waiting for help with the test because my local copy of mafia with this change works for my purposes so I will just keep building my own copy until this eventually gets approved

gausie:
Waiting for @jaadams5 seems fine then. Tests aren't as hard as they look and are a core part of keeping KoLmafia maintainable. The fix may work for you now, but in 10 years we want to have an understanding of the underlying requirement if something needs to change.

jaadams5:
The good news is I have a couple of things that might be good tests.  The bad news is they are not failing on main.  I need to understand why a breakpoint is getting hit several times and not just once.  I think something in the test framework in Player is the cause but I have not figured that out yet.  I also need to confirm that the stats are what I expect and not being changed as a side effect of something else I am doing.  Running out of time for today but if I don't get a failing test I'll post the passing tests and get some help.

jaadams5:
See https://github.com/kolmafia/kolmafia/pull/2329

My first pass at writing a failing test failed since the test passes without this change.

jaadams5:
I've run with this and I think it is working as it should.  I am most interested in whether @midgleyc or @gausie are all right with the test.  I seem to have an uncanny sense of how to find the hard (or Java 5) was of doing something so would appreciate the review.

test/net/sourceforge/kolmafia/session/EquipmentManagerTest.java
```diff
@@ -180,4 +188,33 @@ public void canParseStatus() {
...
+    @CsvSource({
+      "mafia thumb ring, false",
+      "Treads of Loathing, false",
+      "panhandle panhandling hat, false",
+      "batskin belt, true",
+      "mafia wedding ring, true"
+    })
+    public void itShouldEquipWhatWasRequestedForProf(String item, String equippable) {
+      AdventureResult itemAR = ItemPool.get(item);
+      boolean canBeEquipped = Boolean.parseBoolean(equippable);
```
midgleyc, line 204: should be able to be `boolean equippable`, saving you the need to parse it.

test/net/sourceforge/kolmafia/session/EquipmentManagerTest.java
```diff
@@ -180,4 +188,33 @@ public void canParseStatus() {
...
+          new Cleanups(
+              withPath(AscensionPath.Path.WEREPROFESSOR),
+              withIntrinsicEffect(EffectPool.MILD_MANNERED_PROFESSOR),
+              withStats(1, 5, 1),
+              withItem(itemAR));
+      try (cleanups) {
+        assertEquals(1, KoLCharacter.getBaseMoxie());
+        assertEquals(5, KoLCharacter.getBaseMysticality());
+        assertEquals(1, KoLCharacter.getBaseMuscle());
+        assertTrue(InventoryManager.hasItem(itemAR));
```
midgleyc, line 215: these are testing the cleanups work correctly -- this might have been useful when writing the test, but I don't think it needs to be present in the final form.

test/net/sourceforge/kolmafia/session/EquipmentManagerTest.java
jaadams5: Done.  Thank you.

test/net/sourceforge/kolmafia/session/EquipmentManagerTest.java
jaadams5: Done.  I decided to explicitly set stats because withLevel does not seem to accommodate that a Level 1 Professor has mysticality of 5.  I also decided checking for the item was testing cleanups stopped it.  Thank you.

midgleyc (approved):

PR #2321, opened 2024-05-16
mayam parse fix for kol-side change

C2Talon:
Some changes with Mayam Calendar KoL-side cause it to parse differently than intended. This is to fix it be the same as it was before.

Note this is completely untested, since I did this all within github's web editor. I just used the following as a reference and modified from there: https://github.com/kolmafia/kolmafia/commit/e8d4daf5be68bb12f469c76045abcca5936fc232


gausie (approved):

gausie:
Thank you

PR #2322, opened 2024-05-16
chore: force LF in mayam HTML in test/root/

midgleyc:
Looks like this is a side effect of #2321.

The test code had CRLF characters, but `.gitattributes` forces LF. My local git client shows these files as changed, and a `restore` can't fix that.


gausie (approved):
Also getting this, thanks Ryo. @C2Talon just fyi!

PR #2323, opened 2024-05-16
Support choice override relay scripts without the need for a 3rd party library

gausie:
Copied from a comment in this PR:

    // Ezandora has graciously maintained a library called Choice-Overrides which allows script
    // maintainers to publish
    // relay overrides of specific choices without fighting over the ownership of choice.ash. Now
    // that we support
    // short-circuiting straight to choice.[choicenumber].ash, scripts no longer need to use this.

    // However, in order to reliably determine the choice number, Choice-Overrides has to run
    // visit_url() and parse the
    // page text. On a second call, visit_url() is blank, so these overrides need to be passed the
    // page text. This means
    // that choice.[choicenumber].ash files do not have the normal signature of a relay override
    // script (i.e. they rely
    // on a string parameter to main). Thus, in order to be backwards-compatible, we defer to
    // choice.ash when a specific
    // choice override is encountered that expects a string argument to be supplied.


gausie:
This is annoying, these don't fail locally.

gausie:
Thanks to Ryo for observing that the failures were due to fixtures I didn't realise were ignored from the repo

gausie:
This is actually really hard to fix with backwards compat, going to close.

PR #2324, opened 2024-05-16
feat: add concoctions for Mayam resonances

midgleyc:
An inverse of the normal style: here a Request calls a Command, instead of a Command calling the Request.

Given the command already existed I thought this was reasonable to avoid rewriting it.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/MayamRequest.java
+++ b/src/net/sourceforge/kolmafia/request/MayamRequest.java
@@ -6,7 +6,7 @@ import net.sourceforge.kolmafia.persistence.ConcoctionDatabase;
 import net.sourceforge.kolmafia.textui.command.MayamCommand;
 
 public class MayamRequest extends CreateItemRequest {
-  private static MayamCommand _mayamCommand = new MayamCommand();
+  private static MayamCommand mayamCommand = new MayamCommand();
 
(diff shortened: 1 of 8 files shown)


src/net/sourceforge/kolmafia/request/MayamRequest.java
```diff
@@ -0,0 +1,30 @@
+package net.sourceforge.kolmafia.request;
+
+import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.objectpool.Concoction;
+import net.sourceforge.kolmafia.persistence.ConcoctionDatabase;
+import net.sourceforge.kolmafia.textui.command.MayamCommand;
+
+public class MayamRequest extends CreateItemRequest {
+  private static MayamCommand _mayamCommand = new MayamCommand();
```
gausie, line 9: Why the leading _?

gausie (approved):

src/net/sourceforge/kolmafia/request/MayamRequest.java
midgleyc: Python on the brain I guess.

midgleyc:
going to assume you would still approve given I only fixed the comment issue

PR #2329, opened 2024-05-19
Failing tests that show mafia allows the prof to equip items that it shouldn't.

jaadams5:
See https://github.com/kolmafia/kolmafia/pull/2318.

This is an attempt to make a failing test where something does not get equipped by a Mild-mannered professor because the prof lacks stats.

It doesn't fail.

itShouldNotEquipIfNotProfProf (which needs to be renamed) in conjunction with the other test should show that making the path the prof allows the items to be equipped.


jaadams5:
My git-fu is not strong enough to make a branch of #2318 and merge this to see what happens.

There definitely are things the prof cannot equip.  See lined out equip links in the relay browser.  This might be the wrong test but I have not figured out a different path

dsimich:
I put your changes into a local unmergedbranches branch that contains my small wereprof change. Is there anything I should be looking for?

jaadams5:
> I put your changes into a local branch on my machine that contains my small wereprof change. Is there anything I should be looking for?

You could try and run tests but if they pass without your change and fail with it then I am very very confused.

I am hoping to get some of the other devs interested in this - the new path isn't helping - because I think there might be a KoL bug.  

My Prof was at level 13, out of ronin and had a whole bag load of stuff.  Several items had the equip link crossed out in the relay browser.  They were also crossed out when I just logged into KoL.  What does that mean?

There is a KoL bug where stat requirements are enforced on Profs?

There is a KoLmafia bug because mafia's can_equip logic is not aware of constraints on the Prof that KoL is enforcing?

My tests aren't testing what I think they should?

If there is something on KoL's end I suspect @gausie might know whether it is a bug or intended.

midgleyc:
```
assertTrue(EquipmentManager.canEquip(itemAR));
```
This should be false for Treads of Loathing. It's passing as-is because you're testing according to the current behaviour, not the PR behaviour.

Also seal tooth isn't even equipment.

jaadams5:
No more seal tooth.  I was thinking in terms of things that go in slots that have stat requirements and my memory was that it was a weapon and not a combat item.   Duh.  Thank you.

I should consider refactoring itShouldEquipWhatWasRequestedForProf and itShouldNotEquipIfNotProf since they are two tests that should be considered together.  Combined they show that somethings can be equipped by the Prof but not in Standard when the path is the only thing that changed.  Stand by.

jaadams5:
Thank you.  Got some thinking out of a rut.  I added a test that might be clearer.

The wiki says "Although your stats are capped by the Intrinsic, minimum stat requirements are still observed. Although the outfit-equipping commands will not work, you can still equip individual parts if you have the required stat values."

The new test raises some questions.  Is it the Path or the Intrinsic Effect or both that determines the behavior?  Does the implementation deal with beast/not beast or prof/not prof or beast/prof/neither?

How do I access the "hidden" stats and how do I set them in a Test?

The mafia thumb ring requires 40 muscle.  If these are valid tests then the hidden muscle should be 40 if I am reading the wiki right.  Maybe that's how KoL decides to strike out the equip link?  But I don't understand why the test works if the wiki is right.

Pondering.

jaadams5:
Thank you.  I have to deal with Real Life so won't process this for a while.  My initial reaction was that your change was the wrong one but now I wonder if falling (and not exiting if prof) ends up checking the right stats.  But if that is the case then why is my test not failing?  Should I be setting a level for the Prof instead of explicitly setting stats?  I will revisit later.  Thanks again.

jaadams5:
Amazing what a good night's sleep will do.  I've got this backwards.  I thought the problem was that there are some items a Prof should be able to equip but they cannot.  The problem is there are some items that a Prof should not be able to equip and KoLmafia thinks otherwise.  So an inexperienced Prof should not be able to equip the thumb ring.  My test is backwards and the code in the other PR is probably correct :-)  I'll clean up my misunderstandings when I have a better connection.  Duh.

jaadams5:
OBE.  Code copied elsewhere.

PR #2331, opened 2024-05-22
Fix typos in Mayam concoctions

Leo40Git:
Fixes the following messages that appear in the console:
```
Unknown ingredient (magical ice cube) for concoction: sweet potato o' mine
Unknown concoction: thansgiving bomb
```


midgleyc (approved):
Thanks!

I will put together a test for this, I thought we already had one.

PR #2332, opened 2024-05-22
tests: add data file consistency test for concoctions

midgleyc:
Tests as discussed in #2331.


jaadams5:
> I will put together a test for this, I thought we already had one.

Probably not but...

There is a generic datafile structure test that does not include concoctions.txt because it it "too complex".  I have a feeling someone wrote a similar concoctions specific test but I could not find it quickly.  I suspect what I am remembering was never committed.

The gCLI command "checkconcoctions" iterates over concoctions to confirm that Concoction.compareTo() meets its contract for all concoction data.  Sometimes running that turns up inconsistencies in concoctions.txt.  The check was deliberately made a command, and not a test, because it takes time and made running the tests longer than people liked.  I just ran it and there are some things to be fixed but none of them seem to be Mayam.

PR #2335, opened 2024-05-25
Stop enforcing G-Lover restrictions for passives.

heeheehee-kolmafia:
According to the wiki, these work as normal; you just can't actively use any skills unless they contain 'G' in their names.

> However, passive skills appear to function normally.

https://kol.coldfront.net/thekolwiki/index.php/G-Lover

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -864,7 +864,6 @@ public class Modifiers {
                 .filter(SkillDatabase::isPassive)
                 .map(UseSkillRequest::getUnmodifiedInstance)
                 .filter(Objects::nonNull)
-                .filter(UseSkillRequest::isEffective)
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -864,7 +864,6 @@ public void applyPassiveModifiers(final boolean debug) {
                 .filter(SkillDatabase::isPassive)
                 .map(UseSkillRequest::getUnmodifiedInstance)
                 .filter(Objects::nonNull)
```
midgleyc, line 866: can you also get rid of the mapping to `UseSkillRequest`? It looks like the only point of that was `isEffective`, which is only for G-Lover.

heeheehee-kolmafia: Sure, done.

midgleyc (approved):

PR #2336, opened 2024-05-26
perf: short-circuit parameter match if arg count is wrong

oxc:
When trying to match the parmaeters of a function, the _number_ of parameters already is a very good indicator whether the function might match. As a performance improvement, filter out mismatches based on the parameter count before iterating and checking the single parameters.

Initially extracted from #2314

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/Function.java
@@ -75,6 +75,10 @@ public abstract class Function extends Symbol {
   }
 
   public boolean paramsMatch(final Function that, final boolean base) {
+    if (this.variableReferences.size() != that.variableReferences.size()) {
+      return false;
+    }
+
     Iterator<VariableReference> it1 = this.variableReferences.iterator();
     Iterator<VariableReference> it2 = that.variableReferences.iterator();
 
@@ -147,6 +151,10 @@ public abstract class Function extends Symbol {
   }
 
   private boolean paramsMatchNoVararg(final List<? extends TypedNode> params, MatchType match) {
+    if (this.getVariableReferences().size() != params.size()) {
+      return false;
+    }
+
     Iterator<VariableReference> refIterator = this.getVariableReferences().iterator();
     Iterator<? extends TypedNode> valIterator = params.iterator();
 
@@ -194,6 +202,11 @@ public abstract class Function extends Symbol {
   }
 
   private boolean paramsMatchVararg(final List<? extends TypedNode> params, MatchType match) {
+    // on a vararg match, the last parameter can match 0 values, so we need at least X-1 values
+    if (this.getVariableReferences().size() - 1 > params.size()) {
+      return false;
+    }
+
     Iterator<VariableReference> refIterator = this.getVariableReferences().iterator();
     Iterator<? extends TypedNode> valIterator = params.iterator();
     VariableReference vararg = null;


oxc:
Unfortunately, without some kind of benchmark, it's hard to tell whether this is micro optimization. In this case I would expect a significant improvement, but maybe we can come up with some kind of performance benchmark for ASH/JS runtimes?

midgleyc:
I can't comment on unchanged blocks, but I think you can remove both of the

```java
    if (it1.hasNext() || it2.hasNext()) {
      return false;
    }
```

or

```java
    if (refIterator.hasNext() || valIterator.hasNext()) {
      return false;
    }
```

blocks because you've done the check at the start.

Veracity0:
> Unfortunately, without some kind of benchmark, it's hard to tell whether this is micro optimization.

Isn't that true of all optimizations? :)

Looking at the methods you modified, in ASH, they are only called during compilation (with one exception). We look up Function objects at compile time and the compiled parse tree contains references to those objects.

Making compilation faster is nice, but it's only done once per script per execution of KoLmafia; we only compile the script the first time you invoke it, and save the compiled parse tree in an Interpreter object, which we reuse for subsequent invocations of that same script until you exit KoLmafia - even across logins.

The exception is "call TYPE NAME (VALUE...)" which looks up NAME as a function which takes specific parameter types and returns the specified return type.

I have literally never used that myself and I don't remember whose use case convinced me to implement it many years ago, but I expect it's used extremely rarely. Assuming I eventual get back to implementing "functions as first class objects in ASH" - presumably when I get back from my trip to Scotland in August - that will be a vastly cleaner way to accomplish whatever "call" was used for, with the bulk of work done at compile time.

> In this case I would expect a significant improvement, but maybe we can come up with some kind of performance benchmark for ASH/JS runtimes?

Are you saying that the JS runtime repeatedly looks up functions by name and argument types?
By all means - write a benchmark that tests that and get some before and after measurements.

Now, if (you think) my understanding of ASH's runtime regarding function lookup is incomplete, then, by all means - come up with a performance benchmark for ASH that invokes these methods.

oxc:
> Now, if (you think) my understanding of ASH's runtime regarding function lookup is incomplete, then, by all means - come up with a performance benchmark for ASH that invokes these methods.

What you said sounds very reasonable, and I have no reason at all to doubt it. I simply did not consider it. Thanks for the valuable insight.

> Are you saying that the JS runtime repeatedly looks up functions by name and argument types? By all means - write a benchmark that tests that and get some before and after measurements.

This is indeed the case, and also not preventable, due to the dynamically typed nature of ECMAscript. Every function call to a library function or a user defined ASH function has to match the provided parameters with the defined signatures.

Since it is now likely that this only has a reasonable meaning for JavascriptRuntime, it might be worth moving some of those checks into the Javascript runtime part that does this lookup.

Veracity0:
I wonder about combat filters. Those are functions - specified by name - which are executed by Macrofyer.macrofy for each round of a combat. Presumably that needs to search for a function by name and args.

RuntimeLibrary:

```
  public static Value run_combat(ScriptRuntime controller, Value filterFunction) {
     ...
    try {
      String filter = filterFunction.toString();
      Macrofier.setMacroOverride(filter, controller);
      RequestThread.postRequest(FightRequest.INSTANCE);
    } finally {
      Macrofier.resetMacroOverride();
    }
    ...
```

Now I want to investigate setMacroOverride further. :)

I think the optimizations are fine in the methods where you put them

Veracity0:
One more thing.

When we look up a function by name, the list we search is a sorted ArrayList.

A feature of that class I have never seen elsewhere is that you can efficiently get a sublist with exactly all the overloaded functions with that name.

We then do argument type matching on that dramatically pruned list of Functions.

If there are no overloads, you look at one Function.

In the absence of overloaded functions, this is at best a micro optimization.

Not to say you shouldn't do it, since it is cheap, but I want some benchmark metrics.

oxc:
I think this never was important anyway, and with the recent JavaScript changes, it possibly does not help there either anyway. I will re-visit this, if it ever turns out to be a real-world issue.

PR #2337, opened 2024-05-26
Initial 11 Thing I Hate About U support

flaminscotsman:
Work on getting the new [11 Things I Hate About U path](https://kol.coldfront.net/thekolwiki/index.php/11_Things_I_Hate_About_U) into mafia

Covers the primary behaviours
- [x] Remove any foods containing the letter `U`
- [x] Adventures from food is based on number of `I`s in the name
- [x] Adventures from booze are increased by then number of `I`s in the name
- [x] Drunkenness from booze is increased by then number of `U`s in the name
- [x] Bonus enchantments are applied to equipment based on `I`s and `U`s in name
- [x] Potion duration is modified by number of `I`s and `U`s in name

I'm not in love some parts of the code, so happy to change things around 😅 
- The naming of some of the methods is a bit clunky, but I couldn't think of a snappy name for them like `Beeosity` and `Is` is a bit easy to mix with `is`
- The `KolCharacter.applyInebrietyModifiers()` call would more naturally fit in the `Consumable.getRawInebriety()` and `Consumable.getInebriety()` but I didn't think mixing character state with persistence classes was great
- I had several stabs at the potion duration change
  - Updating the `modifiers.txt` to update all the potions to include `[existing+#*path(11 Things I Hate About U)]` - this seemed a maintenance nightmare going forwards
  - Behaving like TRCS where we looped through all potions and applied the effect duration change to each - this lead to parsing a lot of probably unused modifier strings
  - Applying the duration change each place it was used (like `KolCharacter.applyInebrietyModifiers()`) - could lead to future calls not calling the wrapper method
  - (Current) Applying any path changes to modifiers after the modifier string was parsed - feels a little hacky though as all other path based mutations are done in the data file or elsewhere in the code


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -2326,6 +2327,32 @@ public static final boolean hasGs(String name) {
     return name.contains("g") || name.contains("G");
   }
 
+  private static final Pattern I_PATTERN = Pattern.compile("[Ii]");
+
+  public static final int getLetterIs(String name) {
+    return (int) KoLCharacter.I_PATTERN.matcher(name).results().count();
+  }
```
Veracity0, line 2334: I guess. Seems like you could stream over the characters in the string, somehow, and count Is, rather than using Patterns and Matchers.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -2326,6 +2327,32 @@ public static final boolean hasGs(String name) {
...
+
+  public static final boolean hasLetterIs(String name) {
+    if (name == null) {
+      return false;
+    }
+    return name.contains("i") || name.contains("I");
+  }
```
Veracity0, line 2341: This could call your method which counts letter I - iterating once over the characters in the string - rather than 2 calls to contains.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -2326,6 +2327,32 @@ public static final boolean hasGs(String name) {
...
+  public static final int getLetterUs(String name) {
+    return (int) KoLCharacter.U_PATTERN.matcher(name).results().count();
+  }
+
+  public static final boolean hasLetterUs(String name) {
+    if (name == null) {
+      return true;
+    }
+    return name.contains("u") || name.contains("U");
+  }
```
Veracity0, line 2354: More of the same.

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -833,10 +834,25 @@ public static synchronized void availableSkillsChanged() {
...
+          index = i;
+          break;
+        }
+      }
+    }
+
+    if (index < 0) {
+      this.expressions.add(entry);
+    } else {
+      this.expressions.get(index).value.combine(entry.value, '+');
```
Veracity0, line 854: Is this literally the only place you call the new "combine" method on an Expression?
So, your list of operators to "combine" only matters if the operator is '+'?

src/net/sourceforge/kolmafia/KoLCharacter.java
midgleyc: this is how we do it for Bs (`getBeeosity` above that in the file). The `contains` approach has a comment mentioning that it's a perf improvement; not sure how true that is.

I also wrote `LetterBonus.java` for Maximizer.

src/net/sourceforge/kolmafia/Expression.java
```diff
@@ -24,6 +24,12 @@
 public class Expression {
   private static final Pattern NUM_PATTERN = Pattern.compile("([+-]?[\\d.]+)(.*)");
   private static final int STACK_SIZE = 128;
+  private static final char[] binaryOpcodes;
+
+  static {
+    binaryOpcodes = new char[] {'+', '-', '*', '/', '%', '^', '<', '≤', '>', '≥', 'x'};
```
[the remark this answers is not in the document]
flaminscotsman: `combine()` is operating on the bytecodes rather than the expression string. The "input expression" that generates them are `gte(expr1, expr2)` and `lte(expr1, expr2)`

They don't look to be used in any of the data files though!

Interestingly though, when checking, I've also missed `m` (min) as one of the valid binary opcodes

src/net/sourceforge/kolmafia/Modifiers.java
flaminscotsman: Yes, but given the `combine()` method edits the "program code" directly, I didn't want to let people accidentally corrupt the "stack" by using a unary 

I can just swap it to an `add(Expression other)` method - just thought given there's the logic for remapping the literals etc might as well keep it generic!

src/net/sourceforge/kolmafia/Expression.java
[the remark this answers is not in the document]
gausie: I added them within the last year. We're running real short on opcodes!

Veracity0 (commented):
Two comments:

1) getLetterIs and getLetterUs is fine, but Eye think that it'd be catchier if they were "getEyeosity" and "getEweosity", by analogy with "getBeeosity".
Eye don't insist on this and will not withhold approval without it. :)
2) Eye'd like you to confirm that adventure gains for food really does Math.round. It seems likely...

So close. This is nice work, and Eye look forward to doing a run in this path once it is in.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5580,6 +5611,59 @@ public static void addItemAdjustment(
             "shirt power");
         break;
     }
+
+    if (inElevenThingIHateAboutU() && item.priority == AdventureResult.Priority.ITEM) {
+      Lookup source = new Lookup(ModifierType.PATH, Path.ELEVEN_THINGS.getName());
+      int eyes = getLetterIs(item.getName());
+      int ewes = getLetterUs(item.getName());
```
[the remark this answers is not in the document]
flaminscotsman: Done - and much snappier than getLetter_'s

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -384,6 +390,16 @@ else if (consumable.getConsumptionType() == ConsumptionType.DRINK) {
       }
     }
 
+    if (KoLCharacter.inElevenThingIHateAboutU()) {
+      if (consumable.getConsumptionType() == ConsumptionType.EAT) {
+        start =
+            end = (int) Math.round(1.5 * KoLCharacter.getLetterIs(name) * consumable.getFullness());
```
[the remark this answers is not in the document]
flaminscotsman: Both the wiki and the ASS discord seem to agree on `ceil(1.5 * fullness * eyes)`, though given we can only integer fullness and i's round is equivalent

src/net/sourceforge/kolmafia/KoLCharacter.java
flaminscotsman: So I (finally) ran the numbers on doing this

Turns out the `contains` comment about the perf improvement is true - I suspect the JVM is actually offloading the `indexOf` check to a native implementation explaining the two passes being quicker than looping over once in plain Java.

The stream based approach is fairly slow - I suspect that's just due to the overhead of the stream mechanics internally.

The fastest count implementation was a simple for loop testing each character 🤷‍♂️

https://github.com/flaminscotsman/kolmafia-stringutils-performance if you want the full numbers

I can swap the implementation(s) as part of this PR or just change the `eyes` and `ewes` implementations now or change them all in a follow up PR?

Also, not sure if `KolCharacter` is the nicest place for them? Could centralize it all in a `LetterUtils` or something!

Veracity0: This is fine. Perhaps we could move {Beeosity, Eweosity, and Eyeosity} methods to a generalized LetterUtils package, by and by, if we care enough,  but it need not be in this PR.

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -384,6 +390,16 @@ else if (consumable.getConsumptionType() == ConsumptionType.DRINK) {
       }
     }
 
+    if (KoLCharacter.inElevenThingIHateAboutU()) {
+      if (consumable.getConsumptionType() == ConsumptionType.EAT) {
+        start =
+            end = (int) Math.round(1.5 * KoLCharacter.getEyeosity(name) * consumable.getFullness());
```
Veracity0, line 396: You said that Wiki and Discord agree that this is ceil().
Given that, I respectfully request that you use Math.ceil, not Math.round.

Veracity0 (changes requested):

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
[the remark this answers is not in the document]
Veracity0: Equivalent or not, please use ```Math.ceil``` for ```ceil```.

Veracity0 (approved):

PR #2338, opened 2024-05-27
Fix Concoction compareTo contract violation.

jaadams5:
The gCLI command checkconcoction checks the compareTo contract for concoctions.

Without this fix several comparisons fail, for example, "Failed comparing (quasi symmetry) [11107] whet stone and [8262] Mayodiol".  All of the failures involve helpers.

The test fails without the proposed fix and passes with it.  Furthermore, after the fix checkconcoction shows no failures.

This fix will change the behavior of any sorted list of concoctions that contains two or more helpers.

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/Concoction.java
+++ b/src/net/sourceforge/kolmafia/objectpool/Concoction.java
@@ -472,6 +472,9 @@ public class Concoction implements Comparable<Concoction> {
     }
 
     // Sort helpers to the top next.
+    if ((this.isHelper() && (o.isHelper()))) {
+      return nameCheckCompare(o);
+    }
     if (this.isHelper()) {
       return -1;
     } else if (o.isHelper()) {
--- a/src/net/sourceforge/kolmafia/objectpool/Concoction.java
+++ b/src/net/sourceforge/kolmafia/objectpool/Concoction.java
@@ -472,7 +472,7 @@ public class Concoction implements Comparable<Concoction> {
     }
 
     // Sort helpers by name if more than one helper.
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/objectpool/Concoction.java
```diff
@@ -471,6 +471,10 @@ public int compareTo(final Concoction o) {
       return 1;
     }
 
+    // Sort helpers by name if more than one helper.
+    if ((this.isHelper() && (o.isHelper()))) {
```
midgleyc, line 475: you could remove some brackets on this line I think?
```suggestion
    if (this.isHelper() && o.isHelper()) {
```

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2608,8 +2608,9 @@ private static void checkPower(final int itemId, final boolean force) {
     }
   }
 
-  // Helper method to force normalize Concoction comparisons to [-1, 0, 1] before testing
-  private static int sgn(int value) {
+  // Helper method to force normalize Concoction comparisons to [-1, 0, 1] before testing contract.
+  // Visible for unit testing
```
midgleyc, line 2612: You could use `Integer.signum` instead? And also replace this sgn method with it.

src/net/sourceforge/kolmafia/objectpool/Concoction.java
jaadams5: Done

PR #2341, opened 2024-06-01
feat: dump mini kiwi items

midgleyc:
Kiwi shop failed parsing, bit more work needed there.

When the mini kiwi drops eggs, you don't get a "You acquire" message at the moment, but that'll probably get fixed.

### DIFF
--- a/src/data/coinmasters.txt
+++ b/src/data/coinmasters.txt
@@ -1210,3 +1210,18 @@ Elf Guard Toy and Munitions Factory	buy	150	replica Elf Guard medal	ROW1428
 Elf Guard Toy and Munitions Factory	buy	400	Lil' Snowball Factory	ROW1429
 Elf Guard Toy and Munitions Factory	buy	1000	Elf Guard safety bear	ROW1440
 Elf Guard Toy and Munitions Factory	buy	2000	Elf Guard temporary (permanent) tattoo	ROW1430
+
+#Kiwi Kwiki Mart	unknown	mini kiwi bikini	mini kiwi (2)	ROW1477
+#Kiwi Kwiki Mart	unknown	mini kiwitini	mini kiwi	ROW1478
+#Kiwi Kwiki Mart	unknown	mini kiwitini	mini kiwi	ROW1479
+#Kiwi Kwiki Mart	unknown	mini kiwi invisible dirigible	mini kiwi (3)	ROW1480
+#Kiwi Kwiki Mart	unknown	mini kiwi aioli	mini kiwi (5)	ROW1481
+#Kiwi Kwiki Mart	unknown	mini kiwi silicon tiepin	mini kiwi (3)	ROW1482
+#Kiwi Kwiki Mart	unknown	mini kiwi tipi	mini kiwi	ROW1483
+#Kiwi Kwiki Mart	unknown	mini kiwi intoxicating spirits	mini kiwi (3)	ROW1484
+#Kiwi Kwiki Mart	unknown	mini kiwi illicit antibiotic	mini kiwi (3)	ROW1485
+#Kiwi Kwiki Mart	unknown	mini kiwi antimilitaristic hippy petition	mini kiwi (3)	ROW1486
+#Kiwi Kwiki Mart	unknown	mini kiwi whipping stick	mini kiwi (4)	ROW1487
+#Kiwi Kwiki Mart	unknown	mini kiwi icepick	mini kiwi (4)	ROW1488
+#Kiwi Kwiki Mart	unknown	mini kiwi-flavored aspirin-injecting lipstick	mini kiwi (3)	ROW1489
+#Kiwi Kwiki Mart	unknown	mini kiwi digitized cookies	mini kiwi (3)	ROW1490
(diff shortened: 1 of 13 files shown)


src/data/coinmasters.txt
```diff
@@ -1210,3 +1210,18 @@ Elf Guard Toy and Munitions Factory	buy	150	replica Elf Guard medal	ROW1428
 Elf Guard Toy and Munitions Factory	buy	400	Lil' Snowball Factory	ROW1429
 Elf Guard Toy and Munitions Factory	buy	1000	Elf Guard safety bear	ROW1440
 Elf Guard Toy and Munitions Factory	buy	2000	Elf Guard temporary (permanent) tattoo	ROW1430
+
+#Kiwi Kwiki Mart	unknown	mini kiwi bikini	mini kiwi (2)	ROW1477
```
Veracity0, line 1214: So we couldn't tell if it was a "buy" or a "sell", because we don't know that the currency of the Kiwi Kwiki Mart is a "mini kiwi", right? And we can't do that until the Coinmaster is implemented? I did that because the default USED to be "buy", and as a result, the Crimbo Armories - which had both "buy" and "sell" - autogenerated bogus entries for things which were "sell".

Huh.

The Coinmaster should be trivial, and re-visiting the mart after it is implemented should make these all "buy", I assume.

src/data/items.txt
```diff
@@ -11616,6 +11616,20 @@
...
+11594	mini kiwi	940910612	kiwi_minikiwi.gif	food, paste, cook, mix	t,d	5
+11595	mini kiwi bikini	705305695	kiwi_bikini.gif	accessory	t,d	11
+11596	mini kiwitini	263511672	martini.gif	drink	t,d	11
+11597	mini kiwi invisible dirigible	718055433	kiwi_dirigible.gif	familiar	t,d	30
+11598	mini kiwi aioli	863064561	ramekin.gif	usable	t,d	11
+11599	mini kiwi silicon tiepin	701181594	kiwi_tiepin.gif	accessory	t,d	40
+11600	mini kiwi tipi	496300747	kiwi_tipi.gif	usable	t,d	11
+11601	mini kiwi digitized cookies	733008951	10101.gif	food	t,d	11
+11602	mini kiwi intoxicating spirits	774169937	beerbottle.gif	drink	t,d	11
+11603	mini kiwi antimilitaristic hippy petition	779505307	document.gif	potion, usable	t,d	11
```
Veracity0, line 11631: Did KoLmafia's item derivation say this is "potion, usable"?
That now means "this is a potion that you cannot multiuse".
The vast majority of potions are multiusable, so the default should be simply "potion".

I think you should make that the default - and also remove "usable" from the new potions - unless you have already determined that they are not multiusable.

Veracity0 (changes requested):

src/data/items.txt
midgleyc: It was auto-generated, but it doesn't show up on multiuse.php, so that means in this case it indeed isn't multi-usable?

src/data/coinmasters.txt
Veracity0: I have not even hatched my mini kiwi egg yet, so have not even looked at the shop yet.
I am OK deferring coinmaster.txt until it is time to implement the Coinmaster - or mixing method.
I assume that the two ROWs for the mini kiwitini correspond to the bottle of gin and bottle of vodka variants.

Note that this is not a coinmaster with "multiple currencies" - although it presumably is a shop.php where some items take more than one item to make. Like Grandma. We do that via a mixing method.

Perhaps it is time to consider supporting "it's a coinmaster with a few recipes that take additional ingredients".

src/data/items.txt
Veracity0:
```
> test newitem 459688276

Unknown item found: mini kiwi illicit antibiotic (11604, 459688276)
--------------------
11604 mini kiwi illicit antibiotic 459688276 kiwi_pill.gif potion, usable t,d 11
Item mini kiwi illicit antibiotic Effect: "Incredibly Healthy", Effect Duration: 5
--------------------
```

Yes. The "usable" is added by default.
Since "potion" implies "multiple" - UNLESS you manually add "usable" - the default should not include "usable".

Veracity0: You mean that if you have multiple of them in inventory and visit inventory, KoL does not provide "use one" AND "use some"?

Which is more annoying - trying to multi use and having KoL fail it, or using multiple one at a time - and having them all succeed - but taking a lot more requests?

I guess this is good enough for now. :)

Veracity0 (approved):

src/data/coinmasters.txt
Veracity0: Or perhaps we should just shrug and do it as a mixing method - existing technology which should handle this with ease.

My "unified shop.php class" which can be subclassed for CoinMasters, NPCStores, and MixingMethods can remain a pipe dream. :)

midgleyc: To clarify my understanding, the current split is:
* if the shop uses meat, it's npcstores.txt
* if the shop uses exactly one currency per item, it's coinmasters.txt (e.g. Cosmic Ray's Bazaar)
* if the shop uses multiple currency items per item, it's concoctions (e.g. Mad Sliemce)
?

Veracity0: I just saw this. Sorry!
That is exactly correct.

gausie: The unified shop.php class deserves more than to be a pipe dream!

PR #2346, opened 2024-06-05
feat: concoctions for kiwi purchases

midgleyc:


gausie (approved):
Looks fine. Should this not be a coinmaster? I can never remember.

midgleyc:
it's not a coinmaster because the kiwitini takes two currencies. There was discussion on #2341.

PR #2347, opened 2024-06-05
feat: override rhino encoding provider instead of content-types.properties

midgleyc:
Alternative to #784 now that https://github.com/mozilla/rhino/issues/1232 is in.


gausie (approved):
Nice! I had forgotten about this

PR #2356, opened 2024-06-13
Some changes in test with open streams and files.

jaadams5:
These changes were made in response to things discovered while cleaning up the test environment of #1889.  Most of the changes were made as it was discovered a test failed to close or release a file that was deleted as part of cleanup or set up for another test.  Most of these changes are also in #1889 so they will be committed there but isolating them makes sense since they are, IMO, worth keeping while waiting for #1889.


test/net/sourceforge/kolmafia/preferences/AlsoPreferencesTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import net.java.dev.spellcast.utilities.DataUtilities;
+import net.sourceforge.kolmafia.KoLCharacter;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class AlsoPreferencesTest {
```
midgleyc, line 21: why has this been split out?

jaadams5: Because?  :-)  In trying to get test cleanups that actually deleted files I realized that these were the only tests that needed the preference file to persist long enough that the backup (and possibly restore) code kicked in.  What was cleaned up, and when, seemed to differ when I un-nested them.  And, while I don't think it hurt anything the before and after code is different here compared to the rest of the tests.  It also removed a possible source of conflict because it used a different username and explicitly made it lower case compared to the other tests.

I had my reasons but since they were part of an evolutionary process, I will try putting these back in the other file and see what breaks.  I'll mark this as Draft until I have done so or can tell you what is not working.  Thank you.

jaadams5: Merged

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -2,59 +2,75 @@
...
+    // to test global prefs concurrency
+    String EMPTY_USER = "Empty";
     KoLCharacter.reset(EMPTY_USER);
     KoLCharacter.reset(true);
     KoLCharacter.setUserId(0);
     File userFile = new File("settings/" + EMPTY_USER.toLowerCase() + "_prefs.txt");
     File backupUserFile = new File("settings/" + EMPTY_USER.toLowerCase() + "_prefs.bak");
     if (userFile.exists()) {
-      userFile.delete();
+      if (!userFile.delete()) {
```
midgleyc, line 66: why not `verboseDelete` as others in this file?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -2,59 +2,75 @@
...
+        System.out.println("Failed to delete " + userFile);
+      }
     }
     if (backupUserFile.exists()) {
-      backupUserFile.delete();
+      if (!backupUserFile.delete()) {
```
midgleyc, line 71: why not `verboseDelete` as others in this file?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -651,34 +710,46 @@ public void exerciseResetDots() {
...
+        inputStream.close();
         assertThat(contents, not(containsString("\nxyz=abc\n")));
-
-        try (var cleanups2 = withProperty("xyz", "abc")) {
-          contents =
-              new String(
-                  DataUtilities.getInputStream(userFile).readAllBytes(), StandardCharsets.UTF_8);
-          assertThat(contents, containsString("\nxyz=abc\n"));
-        }
+        Preferences.setString("xyz", "abc");
```
midgleyc, line 739: cleanups is preferred

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -651,34 +710,46 @@ public void exerciseResetDots() {
...
       try (cleanups) {
-        File userFile =
-            new File("settings/" + KoLCharacter.getUserName().toLowerCase() + "_prefs.txt");
-        String contents =
-            new String(
-                DataUtilities.getInputStream(userFile).readAllBytes(), StandardCharsets.UTF_8);
+        File userFile = new File("settings/" + USER_NAME + "_prefs.txt");
+        InputStream inputStream = DataUtilities.getInputStream(userFile);
+        String contents = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+        inputStream.close();
```
midgleyc, line 737: try (inputStream) should close, and is preferable to doing it inline.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
jaadams5: verboseDelete was written as a debugging aid but as long as it exists I might as well use it.

jaadams5: Done

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
jaadams5: I might disagree?  The test allows preferences to be written to a file when they are set.  The test reads the file, verifies a preference is not set there, sets a preference, reads the file and verifies that the new preference was written to the file.

Maybe I don't understand the sequencing of cleanups but I didn't use it because I could not guarantee that it would not be in the first read.

jaadams5:
@midgleyc - question about cleanups but everything else appears resolved.  Thank you.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
midgleyc: you could put cleanups on this line, not in the first cleanups.

But this is also fine because this is what you're testing, and "xyz" doesn't need to be reset as cleanups would do.

midgleyc (approved):
Thanks!

PR #2359, opened 2024-06-15
Stateful rejection for aliens in Spaaaaace

gausie:
- **Add moonlight rejection effects to Ronaldus and Grimacia aliens**
- **Round rejection properly**

Based on spading done by a few different people in the community and some abductive reasoning, I think this is how this works. I ran a couple of hundred turns on a multi and it looks right for today and yesterday's light levels at least.


gausie:
Started adding tests and I think this is still slightly wrong.

jaadams5 (approved):

PR #2363, opened 2024-06-23
fix: check for dependencies relative to manifest root

midgleyc:
Fixes an issue with [autoscend](https://github.com/loathers/autoscend).

Currently, we're checking to see if a `dependencies.txt` relative to the root of the repo has changed, instead of relative to the manifest root. It just so happens that everything except autoscend had a build process whereby a `dependencies.txt` in the root was copied to the final folder -- but autoscend only has one in the `RELEASE` folder.

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
@@ -205,7 +205,7 @@ public class GitManager extends ScriptManager {
           }
         }
 
-        if (DEPENDENCIES.equals(diff.getNewPath())) {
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia (approved):

jaadams5:
Does the same issue exist with SVN hosted scripts?  That can be a rhetorical question because even if it is possible to construct a similar case using a SVN accessed repository such a situation is unlikely to occur.  But worth asking. IMO.

midgleyc:
SVN scripts don't allow you to use the `manifest.json` to specify an alternative root, so no.

jaadams5:
Thank you.

PR #2365, opened 2024-06-23
Dispose of chat buffer after test

jaadams5:
The test creates a chat buffer.  The chat buffer has an associated log.  Before this the cleanup code could not delete root/chat because a log file was open.  This disposes of the buffer, thus closing the log file and thus allowing the file and directory to be deleted as part of global cleanup.

(A continuing product with my obsession that everything created in test/root by testing is cleaned up.  root/relay is the last hold out).


midgleyc (approved):
Thanks!

PR #2366, opened 2024-06-23
Allow root/relay to be deleted after testing

jaadams5:
If root/relay is always deleted then only one existing test fails.

This reinitializes root/relay for the test that must have it.

This deletes root/relay at the end of all tests.

The other changes are related to suggestions from static code analysis.

### DIFF
--- a/test/net/sourceforge/kolmafia/request/RelayRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/RelayRequestTest.java
@@ -9,7 +9,6 @@ import java.io.File;
 import net.sourceforge.kolmafia.KoLConstants;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.utilities.FileUtilities;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 public class RelayRequestTest {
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -1,13 +1,32 @@
...
 
 public class RelayRequestTest {
 
+  /**
+   * The global clean up code will delete root/relay since relay is not under Git control. This is
+   * the only test that breaks when that happens. Duplicate the code that loads the relay directory
+   * here so that the directory will be present when the test is run.
+   */
+  @BeforeEach
+  public void initializeRelayFileDirectory() {
```
midgleyc, line 23: this is only need for `findVariousRelayFilesOrNot` so can be moved to inside that method.

jaadams5: Philosophy.  Do I put it in BeforeEach in anticipation of a time when there are other tests that will almost certainly need it or do I inline it and hope that the opportunity to extract and reuse will be noticed?  I'll make the change but BeforeEach was a choice, perhaps not the best one :-)

midgleyc (approved):
Thanks!

PR #2372, opened 2024-07-02
More Roman Candelabra support

gausie:
- **Add Roman Numeral parser**
- **Track roman candelabra casts and recover from consequences.txt**
- **Apply modifiers to Roman Candelabra and track usage in fights**

Is adding the roman numerals replacer there right? Otherwise I could add the spans to the consequence pattern and make a new modifier expression function that does the translatino.


midgleyc:
> Is adding the roman numerals replacer there right? Otherwise I could add the spans to the consequence pattern and make a new modifier expression function that does the translatino.

I think adding the spans would be better for now. Even if there's another item in the future that does the same thing, I think the code is clearer if it's more explicit.

jaadams5 (dismissed):

midgleyc (approved):
Thanks!

PR #2373, opened 2024-07-02
Survival Knife Needs Oasis Unlocked

Alium58:
Fixes 
https://kolmafia.us/threads/first-fight-with-survival-knife-incorrectly-tracks-desertexploration.29821/

Conclusion of spading was first adv in desert (with or without ultrahydrated) doesn't have knife progress. Oasis is always unlocked first adv in desert now, so using existing oasis pref to know if we are on the first adv or not

Question - noticed there seems to be code already which updates `desertExploration` based on a pattern match. If I read it right, that is only for when we visit the desert page right? Confirming there is no parsing of the after combat exploration progress message currently.

Putting as draft as I haven't tested in run yet


Alium58:
Grabbed HTML from first desert adv and made a failing test. Then made correct and test no longer fails, yay!

Marking as ready for review now.

jaadams5 (approved):

PR #2374, opened 2024-07-03
feat: colorize everything looks purple, mark as noremove

midgleyc:


jaadams5 (approved):
Some day we will have an app that can be configured to support different visual abilities  This does not move in that direction but neither does it impede progress so it is Approved.  (I do know with HTML and Java displays such configuration may never be possible.  As someone with friends who are color blind or suffering macular degeneration I do spend time helping them find accommodations in software so I have to comment at least).

Veracity0 (approved):

Veracity0:
Are you sure it's not "colourise"? ;)

PR #2376, opened 2024-07-04
Add conditional skills for candelabra

gausie:
Is this it? Is there anything left outstanding?


jaadams5 (approved):

Malibu-Stacey:
> Is this it? Is there anything left outstanding?

Could do with a property to track if there's a chained copy coming similar to `_relativityMonster`.
`adv1()` handles the chains fine but if you cast the purple ray on wish/locket etc you'd have to handle the extra run_combat manually.
Unless `in_multi_fight()` already handles that I guess. I haven't tested that myself yet.

PR #2378, opened 2024-07-10
Enable building with Java 21.

heeheehee-kolmafia:
Key changes:

- Update gradle to 8.8.
- Update Mockito to 5.12.

For now, this does not force the use of Java 21, but rather enables it for compilation.


heeheehee-kolmafia:
For what it's worth, I had no issues with also changing build.gradle to target Java 21, or subsequently building + running with openjdk22.

```
$ java -version
openjdk version "22" 2024-03-19
OpenJDK Runtime Environment (build 22)
OpenJDK 64-Bit Server VM (build 22, mixed mode, sharing)
```

Veracity0 (approved):

PR #2381, opened 2024-07-22
Throw JS exceptions when passing null / undefined to ASH.

heeheehee-kolmafia:
I threw this commit together a few weeks ago and totally forgot about it.

Still needs tests, but posting as a draft PR in case someone else wants to tinker further.


heeheehee-kolmafia:
Note that I'm using EvaluatorException so it can actually be caught in JS. This is quite possibly an abuse of the API.

heeheehee-kolmafia:
That said, the reason I did it that way was so it would actually catch null / undefined that was nested deeper within some object structure. We'll want to get that working before we merge this PR.

gausie (commented):
Can you add a check to confirm that nulls/undefineds in arrays and objects are fine throughout your code, just not when relayed to the ASH environment. And perhaps update the error message to reflect that?

heeheehee-kolmafia:
> Can you add a check to confirm that nulls/undefineds in arrays and objects are fine throughout your code, just not when relayed to the ASH environment. And perhaps update the error message to reflect that?

Done on both fronts.

Technically the `js` CLI command can represent these types, as the existing tests had demonstrated, but it seems simpler to have uniform enforcement. (We can still print null / undefined at this time. I think adding logic to ban that as well in the interest of uniformity would have so many unintended consequences that it's not worthwhile -- JS functions without explicit return values will return `undefined`, after all)

gausie:
I will try to run some big JS scripts with this branch and then merge if all looks ok.

gausie:
This has worked with all my nonsense for three days, merging.

PR #2387, opened 2024-07-28
Allow manually fixing banishers

gausie:
We manage banishers in memory and persist them to a pref. Because of a listener-related close feedback loop, the moment you edit your banishedMonsters pref to fix a tracking issue, it overwrites it with a representation of the in-memory banisher state. This fixes that, and allows users to manually reload the in-memory state from that pref with `refresh banishes`.


midgleyc (approved):
Nice find!

jaadams5:
Nice.  Are there any other cases where we might look for similar behavior?  I'm wondering if something like this might be involved in the unresolved deadlock and related in #1889.  Thank you.

PR #2389, opened 2024-07-29
Add new monster modifier to recognised set

gausie:
Otherwise users get "Unrecognized monster modifier: mimeo" warnings


midgleyc (approved):
Where is mimeo from?

PR #2392, opened 2024-08-02
Tearaway pants skill tracking

gausie:
- **Allow tracking multiple fightMods**
- **New image for item and skill**
- **Tracking for additional item drop and extra adventures**


midgleyc (approved):
Thanks!

PR #2393, opened 2024-08-12
Mall prices are now longs

gausie:
Feel free to add commits to this - I am doing this by eye first and then will start testing it.


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -192,7 +192,7 @@ public enum Gender {
   private static int currentPP, maximumPP;
 
   // Robots only
-  private static int youRobotEnergy, youRobotScraps;
+  private static long youRobotEnergy, youRobotScraps;
```
libraryaddict, line 195: Bit odd a change

gausie: AdventureResult count fields are used for many things from robot energy to items to meat. Moving it to a long across the board actually feels like a good change.

gausie:
Happy to close for #2394. I am not comfortable with this in one change, though I think it's probably the right move eventually.

PR #2394, opened 2024-08-13
feat: mall max increased

midgleyc:
Alternative to #2393.

Mall max is now 999,999,999,999. Item limits are the same as they were.

I tried to be as minimal as possible. Some changes have occurred due to the blending of items / meat...

There is more work to be done: StoreManageFrame doesn't work perfectly (although does work better). The maximizer max price needs more work to be a long, and various preferences (e.g. autoBuyPriceLimit) should also be longs.

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -474,7 +474,7 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
     return this.getPluralName(this.getPluralCount());
   }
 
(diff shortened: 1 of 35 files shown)


gausie (approved):
Yeah I'm happy for this subset to go in. I think moving from ints to longs across more of the codebase is probably a good move eventually. The integer limits used to be architectural and are now a code decision, I think the item limit will go up eventually too since it's now trivial to do so.

midgleyc:
yes, I wanted a quick fix to the division by zero / mall parsing errors people are getting, and then we can do a proper code base run through in subsequent PRs

PR #2397, opened 2024-08-13
feat: update to new whichitem

midgleyc:


ajcoppa (approved):

midgleyc:
Code not in an ideal state -- e.g. I don't think `TransferItemRequest.ITEMID_PATTERN` should be shared to `MallPurchaseRequest`.

"acquire stick of firewood" doesn't work but that might be a bug somewhere else.

Rinn:
i'll merge this in a moment, mall can't get any more busted so no downside imo

Rinn (approved):

PR #2413, opened 2024-09-04
Update dependencies.

heeheehee-kolmafia:
Most of these are only minor version bumps, but there were a handful of major version changes.

Most notably: the maintainer for shadow has changed, along with the package naming.

Hamcrest 3.0 looks like the main backwards incompatible change is that it's ratcheted up the minimum Java version to require at least Java 8. Since we're building with Java 17, this is a non-issue.

There were a handful of major version changes in dependencies that I'm not sure we're ready for (flatlaf and swingx were the two that I recall), and a couple of dependencies where the latest version is marked as beta or alpha (spotless, slf4j-nop).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,9 @@ plugins {
 	id 'java'
 
 	id 'com.diffplug.spotless' version '6.23.3'
-	id 'com.github.johnrengelman.shadow' version '8.1.1'
-	id 'net.nemerosa.versioning' version '3.0.0'
-	id 'org.ajoberstar.grgit' version '5.2.1'
+	id 'com.gradleup.shadow' version '8.3.0'
+	id 'net.nemerosa.versioning' version '3.1.0'
+	id 'org.ajoberstar.grgit' version '5.2.2'
 	id 'org.panteleyev.jpackageplugin' version '1.6.0'
 	id "com.github.ben-manes.versions" version "0.51.0" // enables ./gradlew dependencyUpdates for outdated
 }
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia:
For context, the non-latest-milestone results from ./gradlew dependencyUpdates:

> The following dependencies have later milestone versions:
>  - com.diffplug.spotless:com.diffplug.spotless.gradle.plugin [6.23.3 -> 7.0.0.BETA2]
>  - com.formdev:flatlaf [1.6.5 -> 3.5.1]
>      https://github.com/JFormDesigner/FlatLaf
>  - com.formdev:flatlaf-intellij-themes [1.6.5 -> 3.5.1]
>      https://github.com/JFormDesigner/FlatLaf
>  - com.formdev:flatlaf-swingx [1.6.5 -> 3.5.1]
>      https://github.com/JFormDesigner/FlatLaf
>  - org.slf4j:slf4j-nop [2.0.16 -> 2.1.0-alpha1]
>      http://www.slf4j.org
>  - org.swinglabs:swingx [1.0 -> 1.6.1]
>      http://www.swinglabs.org/
> 
> Gradle release-candidate updates:
>  - Gradle: [8.8 -> 8.10]

heeheehee-kolmafia:
I'm confused by the test failure. Surely bumping these dependencies shouldn't affect ASH on Windows specifically, right?

(test_finally_return.ash somehow transformed `input` from 200 -> 0. I'm rerunning the failed tasks to see if this was a flake, although even if it is, we should try to understand how reproducible it is and if it's an indicator of a larger bug.)

heeheehee-kolmafia:
Rerunning seems to have addressed the failure.

Deep link, for posterity: https://github.com/kolmafia/kolmafia/actions/runs/10694075396/job/29645403583

Veracity0 (approved):

midgleyc:
That test fails sometimes, I think it's a race.

The last line is 
```
print("I received " + test(200));
```

This leads into a branch that calls `abort`: normally this aborts the ASH file and nothing gets printed, but sometimes it acts as though the function returned 0 (the default value for int) and prints "I received 0".

PR #2369, opened 2024-06-26
Experiment

jaadams5:
This is a Draft.  I am trying to make changes to https://github.com/kolmafia/kolmafia/pull/1889 where I can check them across all operating systems.  If the failing tests go away then there may be some keepers.

PR #2414, opened 2024-09-04
Exported from #2369

jaadams5:
These are test tweaks discovered by #1889 that I decided made some failures less likely to happen.  More specific comments in file annotation comments.


test/net/sourceforge/kolmafia/session/ForbiddenManagerTest.java
```diff
@@ -10,23 +10,17 @@
...
-    KoLCharacter.reset(TESTUSERNAME);
-    MallPurchaseRequest.reset();
-  }
+  private static final String TEST_USERNAME = "ForbiddenManager";
 
   @BeforeEach
   public void beforeEach() {
-    Preferences.reset(TESTUSERNAME);
+    KoLCharacter.reset(TEST_USERNAME);
+    MallPurchaseRequest.reset();
```
jaadams5, line 23: Resetting the character also resets the preferences.  Resetting the MallPurchaseRequest keeps things from leaking since the request is reused.

test/net/sourceforge/kolmafia/swingui/panel/PreferenceWatcherTableTest.java
```diff
@@ -165,9 +165,10 @@ void canRemoveProperty() {
               withProperty("testPrefA", "a"),
               withProperty("testPrefB", "b"),
               withProperty("watchedPreferences", "testPrefA,testPrefB"));
-
+      var table = new PreferenceWatcherTable();
```
jaadams5, line 168: This is believed to fix a concurrent modification exception sometimes thrown in the test.

test/net/sourceforge/kolmafia/webui/CharPaneDecoratorTest.java
```diff
@@ -48,7 +50,9 @@ public void decorateEffects(final String displayMode) {
 
     try (cleanups) {
       String input = html("request/test_charpane_" + displayMode + ".html");
-      CharPaneRequest.processResults(input);
+      assertNotNull(input);
```
jaadams5, line 53: For some reason the file is not processed and so sometimes this throws an NPE.

gausie (approved):
I'm happy to approve these since they seem harmless but I don't quite understand the intention

jaadams5:
Thank you.  Long, long ago #1889 had test failures that made us leery of the code changes.  Looking at the failures (combined with seeing similar failures in other PRs although with significantly less frequency) made me decide most of them were caused by the test infrastructure and not mafia code.  They are in some sense harmless since no mafia code has been changed but it is my belief that they are a small step towards the day when test failure occur for some reason associated with mafia code and not just the timing of test setup and tear down.

PR #2416, opened 2024-09-06
Parse HP / MP / Meat from charsheet.php as longs.

heeheehee-kolmafia:
We already parse these from charpane.php and api.php and store these as longs, but if we parse charsheet.php last, this can still clobber the existing values.

(We probably can / should make the tests cover more than just Meat, but I thought I'd put this PR out there.)


jaadams5 (approved):
Your call about adding tests now, later or not at all.

heeheehee-kolmafia:
> later or not at all.

I'll add them now. My experience is that these two are synonymous, at least until the lack of tests leads to some breakage down the line.

midgleyc (approved):

PR #2420, opened 2024-09-12
fix: use links for sept-ember censer, hulk is once-per-day

midgleyc:


src/net/sourceforge/kolmafia/request/UseItemRequest.java
```diff
@@ -6213,6 +6213,14 @@ else if (responseText.contains("You can't figure out the angle")) {
...
           break;
         }
+      case ItemPool.MINIATURE_EMBERING_HULK:
+
+        // If we are redirected to a fight, the item is
+        // consumed elsewhere. If we got here, it wasn't
+        // actually consumed
+
+        Preferences.setBoolean("_emberingHulkFought", true);
+        return;
```
jaadams5, line 6223: I don't understand the comment or am missing something in context.  Simplistically "if we got here it wasn't actually consumed" suggests to me that if we got here it wasn't consumed so the preference needs to be set to or remain at false.  That is not what the code does.  I won't make this a blocker because I think the problem is my understanding and not the code.

jaadams5 (approved):
A test would be appropriate but I'm not going to require it :-)

src/net/sourceforge/kolmafia/request/UseItemRequest.java
midgleyc: If we use the item when we've already fought a hulk today, the item isn't consumed. If we try to use the item again, it'll fail because we've already fought one (perhaps outside of Mafia).

jaadams5: I think I get it.  Thanks.

midgleyc:
Most of these switches have no tests. Some in `UseItemRequest` have tests, but normally those that have slightly more involved logic than just setting a pref on use (i.e. there is a test for the presence of a string)

There isn't really any logic here, so the tests would be something like "did I write what I wrote", so I didn't write any -- I just checked in the browser manually.

PR #2427, opened 2024-09-14
Adding preference to track slime vials harvested in a day

poshpotato:
Mostly for my own automation use, adds a preference that tracks the ten slime harvests you get in a day.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -2397,7 +2397,7 @@ user	_skateBuff2	false
 user	_skateBuff3	false
 user	_skateBuff4	false
 user	_skateBuff5	false
-user    _slimeVialsHarvested 0
+user	_slimeVialsHarvested	0
 user	_sleazeAirportToday	false
 user	_sloppyDinerBeachBucks	0
 user	_smilesOfMrA	0
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -2397,6 +2397,7 @@ user	_skateBuff2	false
 user	_skateBuff3	false
 user	_skateBuff4	false
 user	_skateBuff5	false
+user    _slimeVialsHarvested 0
 user	_sleazeAirportToday	false
 user	_sloppyDinerBeachBucks	0
 user	_smilesOfMrA	0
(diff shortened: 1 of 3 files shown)


src/data/defaults.txt
```diff
@@ -2397,6 +2397,7 @@ user	_skateBuff2	false
 user	_skateBuff3	false
 user	_skateBuff4	false
 user	_skateBuff5	false
+user    _slimeVialsHarvested 0
```
steveWang, line 2400: These fields should be separated by tabs, not spaces. (in case you're wondering about the test failures)

heeheehee-kolmafia, line 2400: These fields should be separated by tabs, not spaces. (In case you're wondering about the test failure.)

poshpotato: Aye, I've just fixed it and rerun the tests, should pass now. thanks.

src/net/sourceforge/kolmafia/request/VolcanoIslandRequest.java
```diff
@@ -113,8 +114,15 @@ public static void getBreakfast() {
   }
 
   public static void parseResponse(final String urlString, final String responseText) {
-    if (!urlString.startsWith("volcanoisland.php") || !urlString.contains("action=tniat")) {
-      return;
+    if (!(urlString.startsWith("volcanoisland.php")
+        && (urlString.contains("action=tniat") || urlString.contains("action=npc")))) return;
```
midgleyc, line 118:
```suggestion
    if (!urlString.startsWith("volcanoisland.php")
        || !(urlString.contains("action=tniat") || urlString.contains("action=npc"))) return;
```

I think this is what you want -- only parse `volcanoisland.php`, and inside that only parse `tniat` and `npc` actions.

A test would ensure the logic is correct. Something like
```java
var cleanups = new Cleanups(withClass(AscensionClass.SAUCEROR), withProperty("_slimeVialsHarvested", 1));
try (cleanups)
 VolcanoIslandRequest.parseResponse("volcanoisland.php?action=npc", "You ladle some slime out of one of the drums. Fortunately, you had an empty vial on hand for just such an opportunity.");
 assertThat("_slimeVialsHarvested", isSetTo(2));
```
checking the URL is correct and perhaps including the full HTML.

poshpotato: It's worked fine testing in game, but seems to be failing when I use that code for a test. Will have a look at it tomorrow, probably.

midgleyc: I can add a test to your branch -- the first one can be tricky.

midgleyc (approved):
Thanks!

PR #2429, opened 2024-09-16
tests: use hamcrest matcher for maximizer custom assertions

midgleyc:
Main benefit is better messages on failure.


Veracity0 (approved):
What a lot of work. It looks cleaner, as well.

PR #2431, opened 2024-09-17
Refactor and reprioritize ItemFinder.

phulin:
This also prioritizes normal items over their LoL replicas - mafia currently does the opposite, which is very annoying behavior. It's possible there is some other reason this check was originally included, so happy to be corrected.

Edit: see below for new design.


src/net/sourceforge/kolmafia/persistence/ItemFinder.java
```diff
@@ -277,7 +278,8 @@ private static void filterNameList(List<String> nameList, Match filterType) {
           itemId != -1
               && !ItemDatabase.getAttribute(
                   itemId, EnumSet.of(Attribute.TRADEABLE, Attribute.MATCHABLE, Attribute.QUEST))
-              && !NPCStoreDatabase.contains(itemId));
+              && !NPCStoreDatabase.contains(itemId)
+              && !InventoryManager.itemAvailable(itemId));
```
midgleyc, line 282: This doesn't only check for items in your inventory, as the comment says (that would be `hasItem`).

Previous discussion at https://kolmafia.us/threads/change-item-string-fuzzy-matching-to-prefer-owned-items.29044/#post-172700 (I think this is the same thing).

Given

```
    // Never match against (non-quest) untradeable items not available
    // in NPC stores when other items are possible.
```

I am curious about why we are doing this at all. Though given
```
    // If there are multiple matches, such that one is a substring of the
    // others, choose the shorter one, on the grounds that the user would have
    // included part of the unique section of the longer name if that was the
    // item they actually intended.  This makes it easier to refer to
    // non-clockwork in-a-boxes, and DoD potions by flavor.
```
elsewhere in the file, how about adding "matchable" to the non-replica items? I think that would prefer the non-replicas to the replicas.

phulin:
@midgleyc and I were discussing making more extensive changes on Discord, and here's what we came up with for an adjusted design for the item matcher from current behavior. I'm happy to take comments before implementing:
- Find list of possible fuzzy matches based on existing logic (there are some apparent bugs that we should fix also based on my reading of the code)
- If exact match, return
- Eliminate improper matches based on use hint (i.e. equip, food, etc.)
- If the query is the first letter of each word (e.g. JOFPJ for pickle juice, which otherwise also matches jug of porquoise juice), treat as exact match and return *[new behavior]*
- If use hint expects item to be available (i.e. not ANY/CREATE) and only one item remaining that is available based on user settings, return *[new behavior]*
- *[maybe]* If use hint expects item to be available, eliminate any items that can't be acquired from current game state (needs more detail, current rule is untradeable and not in NPC store)
- Eliminate superstring matches, on the theory that if the user wants the superstring they'll include the superstring piece. This will need some algorithmic thought as the current implementation is broken

phulin:
Current design looks like this. PR should now be ready for review.

- Find list of possible fuzzy matches based on existing logic
- Prioritize HP/MP restorative items *[current behavior]*
- If exact match, return
- Eliminate improper matches based on use hint (i.e. equip, food, etc.)
- If only one item left, return
- If the query is the first letter of each word (e.g. JOFPJ for pickle juice, which otherwise also matches jug of porquoise juice), treat as exact match and return *[new behavior]*
- If use hint expects item to be available (i.e. not ANY/CREATE) and only one item remaining that is available based on user settings, return *[new behavior]*
- If use hint expects item to be available, eliminate any items that can't be acquired from current game state (untradeable, non-quest, not in NPC store, not available per user settings)
- Eliminate superstring matches, on the theory that if the user wants the superstring they'll include the superstring piece.
- Return single result, no result, or multiple results

src/net/sourceforge/kolmafia/persistence/ItemFinder.java
```diff
@@ -277,7 +325,9 @@ private static void filterNameList(List<String> nameList, Match filterType) {
           itemId != -1
               && !ItemDatabase.getAttribute(
                   itemId, EnumSet.of(Attribute.TRADEABLE, Attribute.MATCHABLE, Attribute.QUEST))
-              && !NPCStoreDatabase.contains(itemId));
+              && !NPCStoreDatabase.contains(itemId)
```
midgleyc, line 328: `itemAvailable` checks this only if the user has `autoSatisfyWithNPCs`, which I think is what is probably desired here, so this line can be removed

src/net/sourceforge/kolmafia/persistence/ItemFinder.java
```diff
@@ -277,7 +325,9 @@ private static void filterNameList(List<String> nameList, Match filterType) {
...
+              && !InventoryManager.itemAvailable(itemId)
+              && ConcoctionDatabase.getIngredients(itemId).length == 0);
```
midgleyc, line 330: near as I can tell, this checks whether an item can be created, not whether it can be created from items available to the user. This may be what is desired. Am I correct, and was it?

midgleyc:
A general note: "closet put 1 meat" still successfully puts 1 meat into your closet, instead of the item "1 meat". I was worried this wouldn't work.

src/net/sourceforge/kolmafia/persistence/ItemFinder.java
midgleyc: This is a break in behaviour as `autoSatisfyWithNPCs` is false by default, but I think it makes the way `ItemFinder` works less confusing (and also maybe we should adjust the `autoSatisfy` defaults).

src/net/sourceforge/kolmafia/persistence/ItemFinder.java
phulin: Yeah, I want these checks to be conservative about culling the list. If we find there to be a problem or annoyance later we can adjust.

src/net/sourceforge/kolmafia/persistence/ItemFinder.java
phulin: As in my below comment I think we should leave this to be conservative

phulin:
I think this should be good to go now?

midgleyc (approved):
Thanks!

PR #2435, opened 2024-09-19
Remove mimic egg from inventory if you differentiate it successfully

gausie:


midgleyc (approved):

midgleyc:
I've just realised -- I think the behaviour here is wrong, a mimic egg can contain multiple monsters, and you only lose the egg when fighting the last monster.

PR #2438, opened 2024-09-20
Tweak matching rules for Kramco-type items.

phulin:
Kramco behaves a little differently, since its container is not a superstring.


gausie (approved):
Test looks good!

PR #2441, opened 2024-09-21
Track a couple of PirateRealm Sailing Encounters

Malibu-Stacey:
These 2 do advance the sailing turns needed to reach the next island too. Looks like the only one which doesn't is the broken ship encounter.


midgleyc (approved):
Thanks!

gausie:
People insisted to me that this categorically wasn't the case. Huh!

PR #2444, opened 2024-09-27
Update hellion animated monster image

gausie:
Thought this would be fun, feel free to reject if you hold the old one close to your heart!


jaadams5 (approved):
I don't care either way.  If someone likes it, that's fine.

Veracity0:
The original image came from Hellion (#250730). That is his custom avatar image.
I asked his permission to import it into KoLmafia and override the monster's image.
He obliged, and with a fair amount of work, we have the ability to override images.

If you want to update it to be based on KoL's new (as of August 16, 2010!) image, that's fine.

Veracity0 (approved):

PR #2446, opened 2024-09-28
Fix a bunch of warnings

Rinn:
I've suppressed 'incomplete-switch' warnings for any class that had them, there were like a thousand of these and I can't imagine they need to be addressed. The second most class of warnings is 'unused' but I think some of those should get looked at so I haven't touched those yet.

The other warnings I resolved reduces the warning count I'm seeing locally from 1373 -> 261

This will be easier to review if you skip the [first commit](https://github.com/kolmafia/kolmafia/pull/2446/commits/7b3fc71a8a9805d289ef4767c20ca64d532e2df1) and look at the [others](https://github.com/kolmafia/kolmafia/pull/2446/files/7b3fc71a8a9805d289ef4767c20ca64d532e2df1..6bd8d94f15d467e9f063d3b9e2d8692f72a37442) separately.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/TestCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/TestCommand.java
@@ -200,6 +200,7 @@ public class TestCommand extends AbstractCommand {
         RequestLogger.printLine("Inside try: checkpoint known = " + saved.known());
       }
       RequestLogger.printLine("Outside try: checkpoint known = " + saved.known());
+      saved.close();
       return;
     }
 
--- a/src/net/sourceforge/kolmafia/textui/command/TestCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/TestCommand.java
@@ -200,7 +200,6 @@ public class TestCommand extends AbstractCommand {
         RequestLogger.printLine("Inside try: checkpoint known = " + saved.known());
       }
       RequestLogger.printLine("Outside try: checkpoint known = " + saved.known());
-      saved.close();
       return;
     }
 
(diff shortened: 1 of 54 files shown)


src/net/sourceforge/kolmafia/textui/command/TestCommand.java
```diff
@@ -200,6 +200,7 @@ public void run(final String cmd, final String parameters) {
         RequestLogger.printLine("Inside try: checkpoint known = " + saved.known());
       }
       RequestLogger.printLine("Outside try: checkpoint known = " + saved.known());
+      saved.close();
```
midgleyc, line 203: is this one a false positive? if we try-with-resources `checkpoint`, then set `saved` to `checkpoint`, `saved` is always going to be closed after it exits the block?

I'm fine with the change to suppress the warning though... :/

midgleyc (dismissed):
If there's any way to suppress incomplete-switch at project level I think we should, I think that lint is useless and actively bad.

A lot of our switches are "if the enum value is X, Y or Z do something, otherwise do nothing" and that would make us add a default case for "do nothing" when you could simply /not add a case/ (which adds pointless code, and also doesn't save you in the case you forgot to add a case which is exactly what the lint is trying to warn you about).

Rinn:
if there is a way to do it globally i couldn't figure it out, i could have missed something tho

src/net/sourceforge/kolmafia/textui/command/TestCommand.java
Rinn, line 203:
```suggestion
```

Rinn: ah yeah i'll remove it

midgleyc (approved):

PR #2450, opened 2024-09-30
Add tracking for Black & White Apron choice.

phulin:
Tracking preference `bwApronMealsEaten` defaults to -1 for "unknown" and gets set to 0 (known) when you ascend.


midgleyc:
What's the aim in tracking the number of meals eaten per ascension? Does the item change in some way?

jaadams5:
I am confused as to intent and implementation.  If the goal is to count uses then the distinction between known and unknown referred to in comments is somewhat useless.  The usual idiom for a counter that resets at ascension would be to have it default to zero, be incremented upon use and reset to zero at ascension.  Such a counter always runs the risk of being wrong if the user ever does anything in vanilla KOL with mafia.

The purpose is unclear unless there is an in game effect not mentioned on the wiki.  If the goal is spading or some kind of analysis a use count per session can be determined by searching logs for "Took choice 1518/1".  Session log processing will still need to be selective to separate ascensions but that has been solved elsewhere.

This also seems like a good use of a Choice Adventure Script.  For example the script could create a file named with character name and ascension number.  If the file does not exist then the script creates it with a value of zero.  If it exists then it reads the value, increments it, and writes it.

gausie:
Looks good to me. Once we know the relationship between the item and the number of times used, we could remove the -1 default value and recover from missing data on the fly by inferring it backwards.

jaadams5:
"the naive behavior of setting a default of 0 and adding 1 every time something is eaten will look correct but be wrong."

An example or explanation, please.

jaadams5:
Motivation is spading and excavator requires a preference.  Understood.

If a character acquires and uses an apron the count will not be correct until the character ascends, correct?  Or is there some context I am missing?  Thanks.

jaadams5:
OK.  What this does is sets up a counter that will not be correct until the character ascends.

If the counter is one how do you distinguish between a) the character has used it twice but not ascended since switching to this version of mafia and b) the character ascended and has used it once?

jaadams5:
Thank you.  That was a subtlety I missed.  Given that this is not expected to work "immediately" I'm fine with it now that I understand what it is doing.  I do think an opportunity was missed by not spending more time introducing what this was expected to do and why it needed to be done, initially, but ...

jaadams5 (approved):

PR #2456, opened 2024-10-04
feat: various iotm improvements

midgleyc:
* Detect photo booth in clan
* Check Sept Embers on session open, warn if some left when ascending
* Check whether we've received the bat wings Bat Hole adventures
* Add bat skills when the wings are equipped
* Track bat wing free kills


jaadams5 (approved):
Thank you.

PR #2459, opened 2024-10-05
Add tracking for "forcenoncombat" KoL property.

phulin:
Certain zones have apparent periodic noncombats, which are caused by a flag in KoL called "forcenoncombat" that triggers noncombat occurrence after a certain number of turns without one. Examples include the Black Forest, the Skeleton Store, and many others. This PR is a first step towards tracking the last noncombat in each of the forcenoncombat zones. This tracking means that assistant scripts like Guide/TourGuide can predict when your next guaranteed noncombat will occur.

Feedback welcome.


src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -55,6 +56,88 @@
 public class AdventureRequest extends GenericRequest {
   public static final String NOT_IN_A_FIGHT = "Not in a Fight";
 
+  public static final Set<String> ZONELESS_NCS =
```
gausie, line 59: I might want to see these in a datafile. Maybe we could make noncombats.txt to future proof against any other data we want to track about NCs in future. canWalkAway could move there, for example.

gausie: For example, your comment that mafia should have a list of all superlikelies - that maybe lives there also? Though you can have superlikely combats too. Hm

gausie: Really combats.txt needs to turn into encounters.txt, with the attributes after the colon for each entry specifying if it is an NC, and indeed if it is/can be superlikely. Something like that.

phulin: I agree with this 100% - but don't want to expand the scope of this PR too much. It looks like encounters.txt already has some of this. I'll see what I can do to bring those in. I don't know if we want to double down on that design but it does exist.

phulin:
I added an attempt at this based on the existing encounters.txt file. This will work for the current purpose, but I agree that in the future we should combine this into one data source like combats.txt.

I don't know what to about this test failure. I can only reproduce it locally sporadically.

phulin:
This should now be good to go.

src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -619,13 +622,34 @@ public static final String registerEncounter(final GenericRequest request) {
...
+          // Special-case first two NCs in the Upper Chamber, which are superlikely.
+          // In a perfect world, Mafia would have a list of all superlikelies, but this works for
+          // now.
+          && !(AdventureSpentDatabase.getTurns(
+                      AdventureDatabase.getAdventure(AdventurePool.UPPER_CHAMBER))
+                  == 2
+              && encounter.equals("A Wheel -- How Fortunate!"))
+          && !(AdventureSpentDatabase.getTurns(
+                      AdventureDatabase.getAdventure(AdventurePool.UPPER_CHAMBER))
+                  == 5
```
midgleyc, line 641: you can get past the turns spent by e.g. using professor copies to chain fights, right? So just checking the encounter name would be better (I think they are unique).

src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -1529,6 +1533,20 @@ public int get_turns_spent() {
           : 0;
     }
 
+    public int get_last_noncombat_turns_spent() {
+      if (this.content == null) {
+        return -1;
+      }
+
+      var id = ((KoLAdventure) this.content).getAdventureId();
+      // Using getString instead of getInteger for a default of -1.
```
midgleyc, line 1542: I don't understand this comment. Isn't it the case that `Preferences.getString` for a non-existent preference is `""`, and `StringUtilities.parseInt("")` is 0?

src/net/sourceforge/kolmafia/request/AdventureRequest.java
midgleyc: > you optimally would have a tracking property as to whether the first one has occurred

why not do this then? If we've never encountered A Wheel, don't track the NC as it's the superlikely version, otherwise it's a normal NC? This will be wrong for adventures outside of Mafia, but will fix itself when somebody hits the NC for the first time.

midgleyc (approved):

midgleyc:
I've approved and you can decide whether or not to do the tracking property. I think there are arguments for both sides.

PR #2460, opened 2024-10-06
Equip bat wings if you want to cast Rest Upside Down

gausie:
I think I might add a Grants Conditional Skill modifier or something that tidies up both this code and the huge chunk in EquipmentManager


jaadams5 (approved):

Malibu-Stacey:
FYI there is a KoL bug I have reported where multicasting Rest Upside Down eats your charges but only casts it once.

PR #2462, opened 2024-10-08
Attempting a Conditional Skill modifier

gausie:
- **Add modifier for conditional skills on equip to all relevant items/outfits where possible**
- **Implement multiple string modifiers maybe**

So after adding all the conditional skills to modifiers.txt, I realised that having multiple identical string modifiers just means that only the final one parsed is saved. Not wanting to abandon my process, I started moving the string modifiers system over to track a list without much of a plan. This could be useful for other areas of modifiers.txt: many items have two Effect modifiers at the moment anyway and we are just losing that information.

For now, I have just fallen back to returning the first value in the list, which will keep behaviour stable. Though, as I write this I realise true stability would be taking the last one in the list. Anyway.

I'm not really sure whether to carry on with this, or if I should change tack. Suggestions and thoughts appreciated. This is rough and untested for now, so don't worry too much about mistakes.

### DIFF
--- a/src/net/sourceforge/kolmafia/Modifiers.java
+++ b/src/net/sourceforge/kolmafia/Modifiers.java
@@ -119,34 +119,34 @@ public class Modifiers {
     int mys = KoLCharacter.getBaseMysticality();
     int mox = KoLCharacter.getBaseMoxie();
 
-    String equalize = this.getString(StringModifier.EQUALIZE);
-    if (equalize.startsWith("Mus")) {
+    var equalize = this.getString(StringModifier.EQUALIZE);
+    if (equalize.contains("Muscle")) {
       mys = mox = mus;
-    } else if (equalize.startsWith("Mys")) {
+    } else if (equalize.contains("Mysticality")) {
       mus = mox = mys;
-    } else if (equalize.startsWith("Mox")) {
+    } else if (equalize.contains("Moxie")) {
       mus = mys = mox;
-    } else if (equalize.startsWith("High")) {
+    } else if (equalize.contains("High")) {
       int high = Math.max(Math.max(mus, mys), mox);
       mus = mys = mox = high;
     }
 
-    String mus_equalize = this.getString(StringModifier.EQUALIZE_MUSCLE);
-    if (mus_equalize.startsWith("Mys")) {
+    var mus_equalize = this.getString(StringModifier.EQUALIZE_MUSCLE);
+    if (mus_equalize.contains("Mysticality")) {
       mus = mys;
-    } else if (mus_equalize.startsWith("Mox")) {
+    } else if (mus_equalize.contains("Moxie")) {
       mus = mox;
     }
-    String mys_equalize = this.getString(StringModifier.EQUALIZE_MYST);
-    if (mys_equalize.startsWith("Mus")) {
+    var mys_equalize = this.getString(StringModifier.EQUALIZE_MYST);
+    if (mys_equalize.contains("Muscle")) {
       mys = mus;
-    } else if (mys_equalize.startsWith("Mox")) {
+    } else if (mys_equalize.contains("Moxie")) {
       mys = mox;
     }
-    String mox_equalize = this.getString(StringModifier.EQUALIZE_MOXIE);
-    if (mox_equalize.startsWith("Mus")) {
+    var mox_equalize = this.getString(StringModifier.EQUALIZE_MOXIE);
+    if (mox_equalize.contains("Muscle")) {
       mox = mus;
-    } else if (mox_equalize.startsWith("Mys")) {
+    } else if (mox_equalize.contains("Mysticality")) {
       mox = mys;
     }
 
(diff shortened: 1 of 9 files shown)


midgleyc:
Rather than making all String modifiers accept multiple entries, why not make a new modifier type (MultiString?) and corresponding ASH functions that return string[] explicitly? Then shift over the types that can be applied multiple times -- I think for most of the existing string modifiers, only "exactly one" makes sense, and anything else is preferably an error.

I like the idea of making this data instead of lots of code.

gausie:
Perfect. This is the exact right solution. You rock.

gausie:
Ok that is now the solution I've followed. There is an open question around things like `Effect: "x", Effect Duration: 10, Effect: "y", Effect Duration: 15`, but we could address that in a future iteration. Interested in feedback once again.

midgleyc:
First, a point about correctness: for pieces of equipment that offer noncombat skills while equipped, the way we implement that is to always add the skill, then set up some code that equips the item when necessary (e.g. see the code around https://github.com/kolmafia/kolmafia/pull/2465).

You could just remove "rest upside down" from the bat wings (as I did in that PR) but I think it would be nice to be able to remove that code too. You could stick with "Conditional Skill (Equipped)", and then check whether a skill is NC-castable -- if so, it gets the "if we have the item, add it always" treatment, whereas combat skills get the code block you already replaced. 

---

It would be nice to check the provided string is a valid skill -- given that there is `Conditional Skill (Equipped): B". L. A. R. T. Spray (wide)"` I expect that's not tested at the moment.

---

Are there any `Conditional Skill (Inventory)` items?

---

Given
```java
      if (content.getType() == ModifierValueType.STRING) {
        return (MultiStringModifier) content;
      }
```
you might want to give `MultiStringModifier` a different type (e.g. `MULTI_STRING`) given that `StringModifier` also matches, but then the casting won't work.

gausie:
> First, a point about correctness: for pieces of equipment that offer noncombat skills while equipped, the way we implement that is to always add the skill, then set up some code that equips the item when necessary (e.g. see the code around #2465).
> 
> You could just remove "rest upside down" from the bat wings (as I did in that PR) but I think it would be nice to be able to remove that code too. You could stick with "Conditional Skill (Equipped)", and then check whether a skill is NC-castable -- if so, it gets the "if we have the item, add it always" treatment, whereas combat skills get the code block you already replaced.

On this, and your question on `Conditional Skill (Inventory)` later, my plan is to do that once I've got this part working.

> It would be nice to check the provided string is a valid skill -- given that there is `Conditional Skill (Equipped): B". L. A. R. T. Spray (wide)"` I expect that's not tested at the moment.

Yes agreed.

> Given
> 
> ```java
>       if (content.getType() == ModifierValueType.STRING) {
>         return (MultiStringModifier) content;
>       }
> ```
> 
> you might want to give `MultiStringModifier` a different type (e.g. `MULTI_STRING`) given that `StringModifier` also matches, but then the casting won't work.

Yeah I already handled this, either just before or just after you reviewed. Thanks for the review!

gausie:
I need to remember what is left to do here...

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -578,6 +608,14 @@ public void addBitmap(BitmapModifier modifier, int bit) {
     this.bitmaps.add(modifier, bit);
   }
 
+  public boolean addMuiltiString(final MultiStringModifier modifier, String mod) {
```
midgleyc, line 611:
```suggestion
  public boolean addMultiString(final MultiStringModifier modifier, String mod) {
```

test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
```diff
@@ -307,6 +307,181 @@ public void testValidMonstersForCombats() {
     }
   }
 
+  @Nested
+  class ModifiersDotTxt {
+    @Test
+    public void modifiersShouldApplyToValidItems() {
```
midgleyc, line 313: this already exists as a nested class Modifiers way down at the bottom of the file

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -3190,6 +3191,37 @@ private static VariableReference namedParam(String name, Type type) {
             namedParam("modifier", DataTypes.MODIFIER_TYPE));
     functions.add(new LibraryFunction("effect_modifier", DataTypes.EFFECT_TYPE, params));
 
+    params =
```
midgleyc, line 3194: `strings_modifier` that just returns a list of strings (for multistring) should probably also exist

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
```diff
@@ -893,20 +862,18 @@ public static final void optimizeEquipment(final int skillId) {
...
-          SkillPool.SIP_SOME_SWEAT -> equipForSkill(
-          skillId,
-          UseSkillRequest.DESIGNER_SWEATPANTS,
-          UseSkillRequest.REPLICA_DESIGNER_SWEATPANTS);
-      case SkillPool.CINCHO_DISPENSE_SALT_AND_LIME,
-          SkillPool.CINCHO_PARTY_SOUNDTRACK,
-          SkillPool.CINCHO_FIESTA_EXIT -> equipForSkill(
-          skillId, UseSkillRequest.CINCHO_DE_MAYO, UseSkillRequest.REPLICA_CINCHO_DE_MAYO);
-      case SkillPool.REST_UPSIDE_DOWN -> equipForSkill(skillId, UseSkillRequest.BAT_WINGS);
+    if (SkillDatabase.getSkillTags(skillId).contains(SkillDatabase.SkillTag.NONCOMBAT)) {
```
midgleyc, line 865: if you only have the replica powerful glove and try to `cast triple size`, what happens? Does mafia error once and then succeed?

if you only have the normal powerful glove does it succeed, then error?

test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
gausie: Merge thing, thank you

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
gausie: Hm ok I will split this stream and see if we have equipped or can equip anything that matches the filter.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
gausie: Done

PR #2463, opened 2024-10-08
Fixes for Avant Guard council text

Malibu-Stacey:
Tested this with some autoscend runs. All looks good now. Quest tracking properties are appropriately set as expected.


midgleyc (approved):
Thanks!

PR #2466, opened 2024-10-10
Create JSON API for relay scripts.

phulin:
This API allows relay scripts to asynchronously get properties and call functions from the relay browser. I and a few others are working on a library that allows relay scripts to seamlessly call into KoLmafia functions, largely focused on the Guide replacement YORICK. This PR creates the endpoint that that library will use.

Requests to the endpoint take two parameters, `body` and `pwd`. `pwd` needs to match the password hash. `body` is a JSON object like the following:
```json
{
"properties": ["kingLiberated"],
"functions": [{
  "name": "availableAmount",
  "args": [{
    "objectType": "Item",
    "identifierString": "seal-clubbing club"
  }]
}]
}
```

Instead of the existing `org.json` JSON library, this PR adds `com.alibaba.fastjson2`, which is the fastest JAVA JSON library per benchmarks. I plan to file a future PR to convert all other uses of `org.json` over to fastjson2.

This functionality could be accomplished with a relay script (e.g. the existing `yorick.js`), but building it into mafia is much more performant and makes the build/install process much simpler for scripts that would like to use the endpoint.


jaadams5:
Some philosophy comments - they are not intended to be blocking unless some other dev thinks they should be.

This change seems intrusive.  Is that because our JSON support is similarly intrusive or might there be a better way to implement this?  My concept of a plan wonders whether a factory pattern might clean things somewhat.  That might, of course require reimplementing the existing support and that might be to high a price to pay but I can ask the question.

Is there anything besides speed justifying fastjson2 replacing existing code?  Could you share some benchmarks from KoLmafia?

Is there any other use case or benefit from this to a KoLmafia user who is not running scripts from the relay browser?  Perhaps your answer is "No" at the moment but will be "Yes, faster JSON" when everything is replaced by fastjson2?

If I am looking at phase 1 of a 2 phase project would it be better or cleaner to just implement phase 2?

src/net/sourceforge/kolmafia/textui/javascript/ScriptableValueConverter.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @Override
+  protected Scriptable asJavaArray(PluralValue arrayValue) {
+    return cx.newArray(
+        scope, Arrays.stream((Value[]) arrayValue.content).map(this::asJava).toArray());
+  }
+
+  @Override
+  public Value fromJava(Object object, Type typeHint) throws ValueConverterException {
```
midgleyc, line 87: this no longer supports "NativeArray" or "NativeObject" -- what were these and could that cause any problems?

src/net/sourceforge/kolmafia/textui/javascript/ValueConverter.java
```diff
@@ -1,97 +1,38 @@
...
-        ScriptableObject.putProperty(result, key.contentString, asJava(value));
-      } else if (key.getType().equals(DataTypes.INT_TYPE)
-          || DataTypes.enumeratedTypes.contains(key.getType()) && key.contentLong > 0) {
-        ScriptableObject.putProperty(result, (int) key.contentLong, asJava(value));
-      } else {
-        throw new ScriptException(
-            "Maps may only have keys of type string, int or an enumerated type.");
-      }
+public abstract class ValueConverter<ObjectType> {
+  public static class ValueConverterException extends RuntimeException {
```
midgleyc, line 20: given how you've written the code (`throws`, translating this to `EvaluatorException`), why not make this a checked exception?

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -57,4 +72,259 @@ public void exerciseSomeObjectMethodsForCoverage() {
...
+          """);
+        assertThat(rr.statusLine, is("HTTP/1.1 200 OK"));
+        assertThat(rr.responseCode, is(200));
+        assertThat(JSON.parse(rr.responseText), is(expected));
+      }
+    }
+
+    @Test
+    public void returnsFunctions() {
+      var cleanups = withTurnsPlayed(22);
```
midgleyc, line 115: this would correspond to `total_turns_played`, so either don't set this or test that function (preferably the latter).

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -57,4 +72,259 @@ public void exerciseSomeObjectMethodsForCoverage() {
...
+      { "functions": [{ "name": "identity", "args": [{
+        "objectType": "Class",
+        "identifierString": "Seal Clubber"
+      }] }] }
+      """);
+
+        JSONObject expected =
+            JSON.parseObject(
+                """
+      { "functions": [{
```
midgleyc, line 170: is the order guaranteed?

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -57,4 +72,259 @@ public void exerciseSomeObjectMethodsForCoverage() {
...
+      assertThat(rr.responseCode, is(200));
+      assertThat(JSON.parse(rr.responseText), is(expected));
+    }
+
+    @Test
+    public void handlesOverloadedFunction() {
+      var rr =
+          this.makeApiRequest(
+              """
+{ "functions": [{ "name": "elementalResistance", "args": [{
```
midgleyc, line 220: nice to have one test that checks the other overload as well

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -57,4 +72,259 @@ public void exerciseSomeObjectMethodsForCoverage() {
     assertFalse(rr.retryOnTimeout());
     rr.constructURLString("diary.php?textversion=1");
   }
+
+  @Nested
+  class JsonApi {
+    private RelayRequest makeApiRequest(String bodyString) {
```
midgleyc, line 78: some more tests are necessary. Ones I noticed:
* multiple properties requested
* multiple functions called
* objectType is case insensitive, so either test that or make it case sensitive (possibly preferable, considering I expect all users will be using TypeScript anyway)

there are likely more

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+          return dataType.makeValue(identifierNumber, true);
+        }
+      }
+
+      return JSONValueConverter.fromJSON(obj);
+    } else return JSONValueConverter.fromJSON(thing);
+  }
+
+  /**
+   * Handle JSON API request. Format is e.g. { properties: ["kingLiberated"], functions: [{ name:
```
midgleyc, line 3548: This isn't comprehensive enough for me to understand how it's supposed to work.

Please make an actual spec instead of an "e.g." (TypeScript-style is fine, pseudocode is also fine, doesn't need to be overly formal but should be exhaustive).

There are some misspellings ("myTunrcount", "can be request") that I presume will be fixed with a rewrite.

I don't understand what "Enumerated types will be returned in the same format." is supposed to mean. Looking at the tests, it doesn't mean "the same format you passed in", which is what I thought this meant.

You also need to specify how non-enumerated types are returned -- with only the identifierString?

Looking at the tests it seems "objectType" is always capitalized when returned, so these examples are also wrong.

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+      var objectTypeObject = obj.get("objectType");
+      Type dataType;
+      if (objectTypeObject instanceof String objectType
+          && ((dataType = DataTypes.enumeratedTypes.find(objectType.toLowerCase())) != null)) {
+        var identifierStringObject = obj.get("identifierString");
+        var identifierNumberObject = obj.get("identifierNumber");
+        if (identifierStringObject instanceof String identifierString) {
+          return dataType.parseValue(identifierString, true);
+        } else if (identifierNumberObject instanceof Integer identifierNumber) {
+          return dataType.makeValue(identifierNumber, true);
```
midgleyc, line 3539: the way this currently works, a user can pass both identifierString and identifierNumber, and if they conflict it goes for string. I'd like conflicts to go to number (and number to be checked first, for perf). Alternatively, you could error.

Either way, this behaviour should be documented.

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+        }
+      }
+
+      return JSONValueConverter.fromJSON(obj);
```
midgleyc, line 3543: is this ever valid?

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+      var nonString = propertiesArray.stream().filter(name -> !(name instanceof String)).toList();
+      if (!nonString.isEmpty()) {
+        jsonError("Invalid property names " + JSON.toJSONString(nonString));
+        return;
+      }
+      result.put(
+          "properties",
+          ((JSONArray) properties)
+              .stream()
+                  .map(name -> name instanceof String ? Preferences.getString((String) name) : null)
```
midgleyc, line 3580: the previous check confirmed the list contains no non-strings, so this instanceof check is unnecessary

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+    var properties = json.get("properties");
+    if (properties instanceof JSONArray propertiesArray) {
+      var nonString = propertiesArray.stream().filter(name -> !(name instanceof String)).toList();
+      if (!nonString.isEmpty()) {
+        jsonError("Invalid property names " + JSON.toJSONString(nonString));
+        return;
+      }
+      result.put(
+          "properties",
+          ((JSONArray) properties)
```
midgleyc, line 3578:
```suggestion
          propertiesArray
```

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+                  })
+              .toList();
+      if (!nonMatching.isEmpty()) {
+        jsonError("Invalid function calls " + JSON.toJSONString(nonMatching));
+        return;
+      }
+
+      // Everything validated. Transform all function arguments.
+      var functionsResult = new JSONArray(functionsArray.size());
+      for (var func : (JSONArray) functions) {
```
midgleyc, line 3605:
```suggestion
      for (var func : functionsArray) {
```

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3514,156 @@ public boolean pollForCompletion(String type) {
...
+
+        if (name.equals("identity") && args.size() == 1) {
+          functionsResult.add(JSONValueConverter.asJSON(transformedArguments.get(0).asProxy()));
+          continue;
+        }
+
+        Class<?>[] argClasses = new Class[args.size() + 1];
+
+        argClasses[0] = ScriptRuntime.class;
+        Arrays.fill(argClasses, 1, args.size() + 1, Value.class);
```
midgleyc, line 3631: I figure this was copied from `LibraryFunction` -- might be worth extracting into its own little function as the logic needs to be the same here and there

midgleyc (changes requested):
Initial comments.

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: I think so -- if the user does this, it's an error on their part, right? So we should explicitly error.

If you return null both here and on line 3561 that should get caught by your null check on line 3650, which I think should then error as desired.

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3515,191 @@ public boolean pollForCompletion(String type) {
...
+        var identifierNumberObject = obj.get("identifierNumber");
+        if (identifierNumberObject instanceof Integer identifierNumber) {
+          return dataType.makeValue(identifierNumber, true);
+        } else if (identifierStringObject instanceof String identifierString) {
+          return dataType.parseValue(identifierString, true);
+        }
+      }
+
+      return JSONValueConverter.fromJSON(obj);
+    } else return JSONValueConverter.fromJSON(thing);
```
midgleyc, line 3554: want a test for this -- function taking a normal value.

e.g. `numberologyPrize` should work

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3500,6 +3515,191 @@ public boolean pollForCompletion(String type) {
...
+              "Invalid arguments to "
+                  + name
+                  + ": "
+                  + JSON.toJSONString(Arrays.stream(badIndices).mapToObj(args::get)));
+          return;
+        }
+
+        if (name.equals("identity") && args.size() == 1) {
+          functionsResult.add(
+              transformResult(JSONValueConverter.asJSON(transformedArguments.get(0).asProxy())));
```
midgleyc, line 3665: what happens if you call "identity" on something that isn't an enumerated type (e.g. you call it on a string)?

What should happen?

midgleyc (changes requested):
Thanks for the prompt update, mostly done, some more tests and error behaviour to sort.

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: okay, so there is a scenario in which this is not an error: the user passes a non-trivial, non-enumerated type.

An example would be `count({"a": 1})`.

Keep this not erroring, and add that as a test, please?

src/net/sourceforge/kolmafia/request/RelayRequest.java
midgleyc: this was for a function taking a non-object argument, line 3554

midgleyc:
Thanks!

PR #2467, opened 2024-10-12
Fix name mismatches in encounters.txt.

phulin:
#2459 had some mismatches between encounters.txt and adventures.txt, which caused error messages on opening mafia and for some encounters to fail to load.

### DIFF
--- a/src/data/encounters.txt
+++ b/src/data/encounters.txt
@@ -212,7 +212,9 @@ Hero's Field	TURTLE	Blue Monday
 Megalo-City	TURTLE	Blue Monday
 The Fungus Plains	TURTLE	Blue Monday
 Vanya's Castle	TURTLE	Blue Monday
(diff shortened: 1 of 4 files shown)


gausie:
Can you add a regression check for this in DataFileConsistencyTest too?

gausie:
Yeah that works too, though shifts some of the checks onto runtime instead of CI time. I think that's fine.

midgleyc (changes requested):
considering this happened, please add something in DataFileConsistencyTest

midgleyc:
I see that what you've got technically tests it, but if somebody adds a lot of data in the future, they'll have to debug to find which lines fail, whereas with a normal test they could just run the test and know the lines, so I'd prefer that please.

midgleyc (approved):

midgleyc:
Much appreciated!

PR #2468, opened 2024-10-14
Address warnings / errors from `./gradlew build`.

heeheehee-kolmafia:
There were a few more unspecified dependencies in the build graph.

When targeting Java 21, running tests also starts emitting a nag about dynamic agent loading.


midgleyc (approved):
Thanks!

PR #2471, opened 2024-10-14
Maximizer redux

jaadams5:
Attempt to get a test that doesn't equip a weapon.  Based upon https://kolmafia.us/threads/maximizer-didnt-equip-a-weapon.29577/page-4#post-175368

This test passes which is disturbing.

People looking at this for me might look to see whether there is some condition that is being missed.  The scores in the live run are different from the scores in the test.

They might also look to confirm that the test really does what it was expected  to do - both confirm that the Maximizer recommends the cleaver as a weapon and that the maximizer emits the commands to equip the weapon.

It is possible that in a live setting the emitted commands fail in which case this test needs to be modified.

Note also some comments in the code.


test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1451,6 +1457,88 @@ public void itShouldMaximizeAndEquipSelectedWeapon() {
...
   }
 
+  @Test
+  public void itShouldEquipCleaver() {
+    String maxStr =
+        "5item,meat,0.5initiative,0.1da 1000max,dr,0.5all res,1.5mainstat,-fumble,mox,0.4hp,0.2mp 1000max,3mp regen,0.25spell damage,1.75spell damage percent,2familiar weight,5familiar exp,10exp,5Mysticality experience percent,+200bonus spring shoes,+200bonus June cleaver,+200bonus designer sweatpants,2 dump";
+    HttpClientWrapper.setupFakeClient();
+    var cleanups =
+        new Cleanups(
+            withClass(AscensionClass.PASTAMANCER),
```
jaadams5, line 1467: It is possible that some of the withEquipped and withEquippableItem are redundant.  The withEquippableItem  should be deleted when this is recognized.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1451,6 +1457,88 @@ public void itShouldMaximizeAndEquipSelectedWeapon() {
...
+            withEquippableItem("candy cane sword cane"),
+            withEquippableItem("June cleaver"), // 10920
+            withEquippableItem("seal-clubbing club")); // 1
+    String out;
+    String cmd = "maximize(\"" + maxStr + "\", false)";
+    try (cleanups) {
+      out = execute(cmd);
+    }
+    assertFalse(out.isEmpty());
+    System.out.println(out);
```
jaadams5, line 1525: This is done to facilitate debugging and comparison to the live runs with dump 2.  The write to System,out will be removed before this is committed if doing so is worthwhile.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1451,6 +1457,88 @@ public void itShouldMaximizeAndEquipSelectedWeapon() {
...
+    assertTrue(out.contains("Wielding June cleaver"));
+    assertTrue(out.contains("Putting on designer sweatpants..."));
+    assertContinueState();
+    var requests = getRequests();
+    assertFalse(requests.isEmpty());
+    for (var req : requests) {
```
jaadams5, line 1531: An earlier version used filtered streams to get the first instance of a request that equipped the item.  However the current code generates about 50 requests and 45 are get requests.  The stream used getPostRequestBody which fails the test if it is passed a get request.  So this works around that.  There are almost certainly alternative implementations.

jaadams5:
RuntimeLibraryTest > itShouldEquipCleaver() FAILED
    org.opentest4j.AssertionFailedError: Could not find expected equipment request.
    
Seems to be a problem with this commit.  Calling it out so I remember to look at it.

jaadams5:
Previous version of test was flawed.  Success or failure was determined based upon the first POST request encountered.  Code changes elsewhere introduced additional POST requests and the expected one was no longer the first one.

jaadams5:
Undrafting.

jaadams5:
Tried to write a failing test inspired by https://kolmafia.us/threads/maximizer-suggests-breaking-slimesuit-for-a-worse-result.30120/#post-175788

This isn't it but maybe someone else can see what I missed trying to replicate the noted failure.

midgleyc:
By convention we put tests inspired by forum threads in MaximizerRegressionTest instead of MaximizerTest (with a comment linking back to the forum thread).

However, the tests in this PR pass, and so don't show the bug the forum threads report. As such, the tests in this PR are unfocused -- what are they trying to test? What is the value in running them each commit?

If you could produce a failing test for a forum thread, that would be valuable.

jaadams5:
> By convention we put tests inspired by forum threads in MaximizerRegressionTest instead of MaximizerTest (with a comment linking back to the forum thread).
> 
> However, the tests in this PR pass, and so don't show the bug the forum threads report. As such, the tests in this PR are unfocused -- what are they trying to test? What is the value in running them each commit?
> 
> If you could produce a failing test for a forum thread, that would be valuable.

Never heard that convention mentioned before but it can be observed.

These are my attempts to create failing tests based upon forum reports.  My attempts failed, i.e. the tests passed.  I originally posted this to get eyeballs on my nonfailing tests and see if anyone could tell me what was wrong with my tests.  The response was underwhelming but that may have been because the convention on Draft seems to be everyone else ignores the PR until is is no longer draft.

How about I move "your" test into MaximizerRegressionTest, perhaps as another PR, and you look at the other tests in this PR and see if you have any things to check to get them to fail.

Thanks.

jaadams5:
OK.  I see the test is already there.  So feedback on the rest would be nice.  Thanks.

midgleyc:
> you look at the other tests in this PR and see if you have any things to check to get them to fail.

No, I have no idea. I think you have reproduced the original forum posts well, but it does not fail.

jaadams5:
> > you look at the other tests in this PR and see if you have any things to check to get them to fail.
> 
> No, I have no idea. I think you have reproduced the original forum posts well, but it does not fail. I did not respond on this PR before now because I truly have no ideas and thus nothing useful to contribute.

Thank you.

PR #2472, opened 2024-10-14
Finish converting to fastjson2.

phulin:
This PR completes the conversion from org.json:json to fastjson2, which is (per it's name) faster. It also has features to define datatypes with automatic databinding into JSON, which we may want in the future.


src/net/sourceforge/kolmafia/FamiliarData.java
```diff
@@ -575,8 +575,8 @@ private static void parseSoup(final String responseText) {
       if (fam == null) continue;
 
       var data = json.getJSONObject(key);
-      fam.setSoupWeight(data.getInt("times"));
-      var attrs = data.getJSONArray("attr").toList().stream().map(String.class::cast).toList();
+      fam.setSoupWeight(data.getInteger("times"));
```
midgleyc, line 578:
```suggestion
      fam.setSoupWeight(data.getIntValue("times"));
```
avoid boxing / unboxing

src/net/sourceforge/kolmafia/chat/ChatPoller.java
```diff
@@ -350,19 +351,19 @@ public static List<ChatMessage> parseNewChat(
...
+      long mid = msg.getLongValue("mid");
       if (!debug && mid != 0 && mid <= ChatPoller.serverLastSeen) {
         continue;
       }
 
       String type = msg.getString("type");
       boolean pub = type.equals("public");
 
-      String formatString = msg.optString("format");
+      String formatString = msg.getString("format");
```
midgleyc, line 366: `optString` returns empty string if not present. I think this is a problem (/difference) in a lot of places.

midgleyc: in this particular place I guess whoever wrote the code thought it returned null though!

src/net/sourceforge/kolmafia/persistence/DebugDatabase.java
```diff
@@ -2407,12 +2408,12 @@ public static void checkMuseumPlurals(final String parameters) {
...
       var array = getMuseumPluralArray();
-      var length = array.length();
+      var length = array.size();
       for (int i = 0; i < length; i++) {
         var entry = array.getJSONObject(i);
-        var id = entry.getInt("itemid");
+        var id = entry.getIntValue("itemid");
         var name = entry.getString("name");
-        var plural = entry.optString("plural", "");
+        var plural = (String) entry.get("plural");
```
midgleyc, line 2416: this one does need to be empty string if not present

midgleyc (approved):
Thanks!

PR #2473, opened 2024-10-15
Return binary files in relay/ as binary.

phulin:
Mafia currently silently corrupts images in the relay/ directory by trying to read and then write them as UTF-8. This creates a second code path to just read them off disk. This is important for any relay script that wants to distribute its own logo, etc.


src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -733,6 +733,18 @@ private void sendLocalFile(final String filename) {
     } catch (IOException e) {
     }
 
+    // If it's a binary file, send it back without loading it as a string.
+    if (!this.contentType.startsWith("text/") && !this.contentType.startsWith("application/")) {
```
midgleyc, line 737: assuming the latter is to catch `application/json`, might be nicer to just consistently use `text/json`? e.g. `application/octet-stream` should also be sent as bytes

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -733,6 +733,18 @@ private void sendLocalFile(final String filename) {
...
+      this.rawByteBuffer = ByteBufferUtilities.read(override);
+      if (this.rawByteBuffer.length == 0) {
+        this.sendNotFound();
+        return;
+      }
+      this.statusLine = "HTTP/1.1 200 OK";
```
midgleyc, line 743: up a bit in the same file, when sending an image we set Last-Modified so the browser can send 304 if it wants; I think that is a good idea to do.

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -80,6 +85,57 @@ public void exerciseSomeObjectMethodsForCoverage() {
...
+    public void returnsTextFile(String filename) throws IOException {
+      var rr = makeFileRequest(filename);
+      assertThat(rr.statusLine, is("HTTP/1.1 200 OK"));
+      assertThat(rr.responseCode, is(200));
+      assertThat(rr.responseText.trim(), is(html("request/" + filename).trim()));
+    }
+
+    @Test
+    public void returnsPng() throws IOException {
+      var rr = makeFileRequest("test_relay_request_sample.png");
```
midgleyc, line 132: this png is apparently not a valid png as far as GitHub is concerned... could you make it valid?

midgleyc (approved):

PR #2478, opened 2024-10-20
feat: photo booth command

midgleyc:
A casual implementation: if your clan doesn't have the relevant item unlocked, it just silently fails. I figure 99% of users will be going to BAFH which will soon have everything unlocked, so I'm not interested in implementing a more thorough solution.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -2302,6 +2302,8 @@ user	_peteJumpedShark	0
 user	_petePartyThrown	false
 user	_petePeeledOut	0
 user	_peteRiotIncited	false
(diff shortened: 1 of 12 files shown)


jaadams5:
I am disturbed by your decision to justify how you do something because of some assumptions you are willing to make about BAFH but if you had merely said this was phase 1 of two phases and you were not going to implement phase 2 (which adds better conditional and error checking) until there was obvious demand then  I would have thanked you for getting something out there now.  Somedays the reasons for doing something matter even if multiple reasons lead to the same result.

jaadams5 (approved):
Thank you.

PR #2479, opened 2024-10-21
Add items hash methods.

phulin:
This PR adds an ASH method to very quickly hash the contents of various items sources. This is useful for relay scripts that would like to update on changes to the inventory. A script can poll this value using the JSON API and reload whenever the value changes.

Doing this manually via an ASH or JS script is much slower (as is simply serializing the entire inventory), so much so that it takes a nontrivial amount of extra CPU.

Optimally, mafia would have a "push" interface that allowed proactive notification to those scripts, but that doesn't exist right now. This is a second-best solution to that.


midgleyc (changes requested):
Please update `TypescriptDefinition.java` to limit the strings passed to the function.

Would be nice to have a comment in the code as to what this function is supposed to be used for (i.e. script polling as you mentioned in the PR), but non-blocking.

midgleyc (approved):
Much appreciated!

PR #2481, opened 2024-10-22
fix: allow a pre-adventure script to adventure

midgleyc:
A pre-adventure script may want to adventure (e.g. to pick up Shadow Waters before going to a zone requiring high item drop). Under the current implementation, the pre-adventure adventure clears the filter function from the later adventure. Fix this.

### DIFF
--- a/src/net/sourceforge/kolmafia/combat/Macrofier.java
+++ b/src/net/sourceforge/kolmafia/combat/Macrofier.java
@@ -1,5 +1,7 @@
 package net.sourceforge.kolmafia.combat;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.HashSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -30,8 +32,7 @@ import org.mozilla.javascript.Context;
 import org.mozilla.javascript.Scriptable;
 
 public class Macrofier {
-  private static String macroOverride = null;
-  private static ScriptRuntime macroInterpreter = null;
+  private static Deque<MacroOverride> macroOverrides = new ArrayDeque<>();
   private static BaseFunction macroFunction = null;
   private static Scriptable macroScope = null;
   private static Scriptable macroThisArg = null;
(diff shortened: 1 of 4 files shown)


midgleyc:
Okay, I think there is still a bug here if you have a pre-adventure script, written in JavaScript, which sets a filter function, and try to adventure using a different filter function: the pre-adventure script will overwrite. I'll try to think of a fix.

Veracity0 (changes requested):
I don't see a bug report on kolmafia.us discussing the motivation for this.
You know - a place where we-all could discuss the implications and maybe agree on the correct change in existing behavior.
I'd like to see such a bug report from the person who had an issue, rather than fixing something out of the blue that doesn't seem to have any publicly visible support.
Until I see such, in a bug report on kolmafia.us, I'm going to have to block this.

midgleyc:
https://kolmafia.us/threads/mafia-ignores-filterfunc-if-pre-adventure-script-adventures.29985/

src/net/sourceforge/kolmafia/combat/Macrofier.java
```diff
@@ -610,4 +626,25 @@ public static void macroManaRestore(StringBuffer macro) {
...
   }
+
+  private static class MacroOverride {
+    public MacroOverride(String macroOverride, ScriptRuntime macroInterpreter) {
+      this.macroOverride = macroOverride;
+      this.macroInterpreter = macroInterpreter;
+      setJavaScriptVariables(Macrofier.macroFunction, Macrofier.macroScope, Macrofier.macroThisArg);
+    }
+
+    public void setJavaScriptVariables(
```
Veracity0, line 637: Since this is only used in the constructor, I don't see why it couldn't be inline there.
Perhaps preceded with ```// JavaScript variables```
But I'm not going to require it.

Veracity0 (dismissed):

src/net/sourceforge/kolmafia/combat/Macrofier.java
midgleyc: good idea, thanks!

Veracity0 (approved):

PR #2482, opened 2024-10-24
Fix ScriptManager breakage.

heeheehee-kolmafia:
Initially reported by kazgar in
https://kolmafia.us/threads/script-manager-doesnt-launch.29984/

There may still be value in reverting the commit in question if other issues arise.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/ScriptManager.java
+++ b/src/net/sourceforge/kolmafia/persistence/ScriptManager.java
@@ -1,5 +1,6 @@
 package net.sourceforge.kolmafia.persistence;
 
+import com.alibaba.fastjson2.JSON;
 import com.alibaba.fastjson2.JSONArray;
 import com.alibaba.fastjson2.JSONException;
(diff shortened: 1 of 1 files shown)


Veracity0:
I sure wish there had been public discussion of the change which broke this.

Veracity0 (approved):
It would have been nice if the commit that submitted the buggy code had included more tests to verify it wasn't breaking things.
What else has been broken?

heeheehee-kolmafia:
And... one other tiny thing that coverage turned up -- we should set _svnRepoFileFetched so it doesn't actually try to call downloadFile to fetch a new repo file. Not that it'll succeed, since I wouldn't expect the test runner to have network access...

midgleyc:
Looks like anything caught by `public JSONArray(Object... items)` will be broken. The only other call I can see is `DebugDatabase.java:2484`, which means that `checkmuseumplurals` presumably no longer works.

midgleyc (approved):
I never thought of testing `checkmuseumplurals` using the existing mocking, nice one!

PR #2483, opened 2024-10-24
Fix JSON API bug with buffer arguments.

phulin:
Short version: Before this PR, the JSON API fails to take buffer arguments for e.g. bufferToFile because I had never implemented the type hinting system that we use for the JS-to-ASH compatibility layer. This PR makes the JS-to-ASH code generic and adds a _lot_ of tests.

The existing JS-to-ASH code has to resolve several unknown types in order to match a set of JS arguments and a function name to a converted set of ASH arguments and a specific override of an ASH function. First, some JS-to-ASH conversions are undetermined. For example, does JS string go to ASH buffer or string? It needs to be able to convert both based on the receiving function. Complicating things further, some ASH functions - buy, for example - have multiple override signatures for the same argument count. So the JS-to-ASH layer infers both simultaneously and finds the first potentially matching function signature.

I made the mistake of not including that logic in the JSON API, making a number of functions impossible to call. There is also some specialized logic for buffers, since ASH doesn't consider strings and buffers to be coerce-able (but JS has no concept of a native buffer to use). So this PR makes the existing JS-to-ASH argument conversion code a generic `findMatchingFunctionConvertArgs` in `ValueConverter.java` and uses that function for both the JSON API and the internal JS engine. This required some changes to how the exception types work, as well.

I've added a number of test that try to get at all of this behavior.


heeheehee-kolmafia:
I manually triggered a rerun, since some of the tests are flaky (known issue that we need to fix at some point). Can you file a bug on kolmafia.us outlining what the problem is, per @Veracity0's [stated new policy](https://kolmafia.us/threads/r28087-whats-changed-create-json-api-for-relay-scripts-by-phulin-in.29967/#post-175461)?

There are a lot of diffs and it's not immediately obvious to me what the problem is, or how this is fixing it.

jaadams5:
> These tests are passing locally, not sure what's up.

There are known intermittent test failures.  If you see a test failure and cannot relate it to the changes you are working on then re-run the tests.  If you want to be a little more nuanced look at the test failure and code.  If disk I/O is involved - perhaps a file is not found and the failure is due to the inability to read the file - then it is reasonable to try again.  There are also a couple of tests that fail somewhat consistently on Windows that simulate user input.  Dvorak is one and there are a couple others.  We don't think they have the same root cause but don't have any other ideas.

Our clean up is pretty aggressive in deleting disk files created by tests.  That is a lazy response to the fact that not everyone who writes tests bothers to clean up files that are created as side effects.  So there are cases where one test is finishing and deleting (for example) settings and the contents while another test is starting up by writing.  Whether there is an error is a matter of timing.  Telling gradle it can only use one thread pretty much eliminates the problem because the order to test execution is pretty much deterministic and one test does not start until the previous one has cleaned up.  But that makes something that is less than five minutes on my system take 45 minutes.  All that said if you need some assistance in deciding whether a test failure can be ignored tag me and I'll try and help or comment.

jaadams5:
> Thanks. The tests seem to be passing now.

If my memory is correct this is not the first time failing tests have been a concern so I wanted to make sure you know they probably weren't a concern and you could get help if you needed.  It's the least I could do :-)

test/net/sourceforge/kolmafia/textui/javascript/JSONValueConverterTest.java
```diff
@@ -1,82 +1,96 @@
...
-      assertThat(converter.asJava(value), is(true));
+      assertThat(JSONValueConverter.asJSON(value), is(true));
     }
 
     @Test
     public void testBooleanTypeFalse() {
-      Value value = DataTypes.makeBooleanValue(false);
-      assertThat(converter.asJava(value), is(false));
+      Value value = DataTypes.makeBooleanValue(true);
+      assertThat(JSONValueConverter.asJSON(value), is(true));
```
midgleyc, line 63: should be false

test/net/sourceforge/kolmafia/textui/javascript/JSONValueConverterTest.java
```diff
@@ -111,10 +168,298 @@ public void testRecordType() {
...
+      Value result = JSONValueConverter.fromJSON("test buffer", DataTypes.BUFFER_TYPE);
+      assertThat(result.getType(), is(DataTypes.BUFFER_TYPE));
+      assertThat(result.contentLong, is(0L));
+      assertThat(result.content, instanceOf(StringBuffer.class));
+      assertThat(result.content.toString(), is("test buffer"));
+      assertThat(result.contentString, nullValue());
+    }
+
+    @Test
+    public void testStringBufferWithoutBufferHint() {
```
midgleyc, line 233: what's the difference between this and `testStringWithoutBufferHint`?

test/net/sourceforge/kolmafia/textui/javascript/JSONValueConverterTest.java
```diff
@@ -111,10 +168,298 @@ public void testRecordType() {
...
+    public void testDoubleUnsafeIntegerValue() {
+      Value result = JSONValueConverter.fromJSON(22.22d, null);
+      assertThat(result.getType(), is(DataTypes.FLOAT_TYPE));
+      assertThat(result.contentLong, is(Double.doubleToRawLongBits(22.22d)));
+      assertThat(result.content, nullValue());
+      assertThat(result.contentString, nullValue());
+    }
+
+    @Test
+    public void testStringBufferWithBufferHint() {
```
midgleyc, line 223: what's the difference between this and `testStringWithBufferHint`?

test/net/sourceforge/kolmafia/textui/javascript/ScriptableValueConverterTest.java
```diff
@@ -120,4 +136,124 @@ void asJavaFromJavaConvertWithoutDataLoss() {
...
+      ScriptableValueConverter converter = new ScriptableValueConverter(cx, scope);
+      Value value = DataTypes.makeBooleanValue(true);
+      assertThat(converter.asJava(value), is(true));
+    }
+
+    @Test
+    public void testBooleanTypeFalse() {
+      ScriptableValueConverter converter = new ScriptableValueConverter(cx, scope);
+      Value value = DataTypes.makeBooleanValue(true);
+      assertThat(converter.asJava(value), is(true));
```
midgleyc, line 166: should be false again

midgleyc (changes requested):

src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
```diff
@@ -42,107 +35,24 @@ public String getFunctionName() {
...
-
-      if (function == null) {
+    Value ashReturnValue;
+    try {
+      var functionWithArgs =
+          coercer.findMatchingFunctionConvertArgs(getAllFunctions(), ashFunctionName, args);
+      if (functionWithArgs == null || functionWithArgs.function() == null) {
         throw controller.runtimeException(
-            Parser.undefinedFunctionMessage(ashFunctionName, ashArgs));
+            Parser.undefinedFunctionMessage(ashFunctionName, List.of()));
```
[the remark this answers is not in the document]
midgleyc: okay so still a problem here, I can make it crash:

```
> js var a = require("kolmafia"); a.monsterHp(1,{})

Unexpected error, debug log printed.

> js var a = require("kolmafia"); a.monsterHp(1,[])

Unexpected error, debug log printed.

> js var a = require("kolmafia"); a.monsterHp(1,/1/)

Unexpected error, debug log printed.
```

`null` and `undefined` both work which is good!

Test coverage for this would also be appreciated.

test/net/sourceforge/kolmafia/request/RelayRequestTest.java
```diff
@@ -615,115 +625,27 @@ public void identityReturnsSameIdentifier(
...
-    }
-
-    @Test
-    public void testInvalidPropertyNames() {
-      var rr = this.makeApiRequest("""
-      { "properties": [123, true] }
-      """);
+    @ParameterizedTest
+    @CsvSource(
+        value = {
```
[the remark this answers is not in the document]
midgleyc: wasn't quite clear enough here -- I meant 
```java
@CsvSource(
    textBlock =
        """
    { invalid_json };Invalid JSON object in request.
    { "properties": [123, true] };Invalid property names [123,true]
    { "functions": [{ "args": [] }] };Invalid function calls [{\\"args\\":[]}]
    { "functions": [{ "name": "abc" }] };Invalid function calls [{\\"name\\":\\"abc\\"}]
    { "functions": [{ "name": "myTurncount", "args": [null] }] };Invalid function calls [{\\"name\\":\\"myTurncount\\",\\"args\\":[null]}]
    { "functions": [{ "name": "nonExistentFunction", "args": [] }] };Unable to find method nonExistentFunction accepting arguments [].
    { "functions": [{ "name": "itemDropsArray", "args": [{"objectType": "monster", "identifierString": "fluffy bunny"}] }] };Unable to find method itemDropsArray accepting arguments [{\\"objectType\\":\\"monster\\",\\"identifierString\\":\\"fluffy bunny\\"}].
    { "functions": [{ "name": "availableAmount", "args": [{ "objectType": "Item", "identifierString": "nonexistent" }] }] };Failed to convert arguments to ASH: Unidentified object {\\"objectType\\":\\"Item\\",\\"identifierString\\":\\"nonexistent\\"}.
    """,
    delimiter = ';')
```

src/net/sourceforge/kolmafia/textui/javascript/AshStub.java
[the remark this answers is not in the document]
midgleyc:
```
> js var a = require("kolmafia"); a.monsterHp(1,/1/)
```
still prints a debug log.

While you catch a `ValueConverterException` from `coercer.fromJava(arg)`, I don't think that function ever throws the exception -- rather it just returns null. I think converting null to any is the right move here (or that will fix the debug log issue, at least).

midgleyc (approved):
looks good to me now!

PR #2485, opened 2024-10-26
Fix canAdventure for Boss Bat Lair post quest finish

gausie:
- **Cannot adventure in Boss Bat's Lair after completing the quest**
- **Test for Boss Bat Lair access**


midgleyc (approved):
Thanks!

PR #2488, opened 2024-10-27
add attack element none for monsters

ckb11:
This adds 'EA: none' for monsters who have physical attacks as well as elemental attacks


gausie:
Please add tests to this, specifically one for a monster line that has an invalid element listed under `EA`

Also please either give some more information on your PR or link to the kolmafia.us post.

ckb11:
More info here: https://kolmafia.us/threads/allow-monster-ea-none.29374/

Some monsters that are elementally aligned also do physical damage for some attacks.  This update allows for physical as well as elemental attacks.  physical attacks are listed as EA: none

I would like to add tests, but I don't know how.  Is there a guide somewhere?

gausie:
@ckb11 Awesome!

Take a look at https://github.com/kolmafia/kolmafia/blob/main/test/net/sourceforge/kolmafia/MonsterDataTest.java

There's already a test called canHandleElementalAttack. That test probably contains too many cases all at once so don't follow its approach exactly, but it still tells you what to do. Make a new function like that for each case you want to assert the behaviour.

ckb11:
Well, I looked at that java stuff and most of it went over my head (I am a mechanical engineer not a developer so I understand logic and can write some basic scripts but this probably is going to take me more hours to decode than I have).
I guess I don't really understand what tests do, how to make a new test function, or what I should be asserting or how.

I'm not trying to get out of this, but I don't have the basis to do this test thing correctly.

ckb11:
Ah - I see now how a unrecognized element could be problematic.  Looking at MonsterDatabase.java, it looks like there is no 'physical' element, just 'none', which is what I used - and seemingly what Mafia uses now.

Is this a problem (to have mafia default to 'none' for unrecognized attack elements) ?

gausie:
We probably want to have Element.UNKNOWN, or to try to throw an error if we see an unknown EA value. Maybe we also need a DataFileConsistency check. I'll take over this branch maybe quickly

gausie:
I'm happy with this, but someone else should review it now since it contains my changes.

ckb11:
> I'm happy with this, but someone else should review it now since it contains my changes.

Thank you gausie for taking this across the finish line!

jaadams5 (approved):
A somewhat casual review and I definitely would not be surprised if monsters.txt needed correction but everything else looks reasonable.  I might have explored the idea of defining Element.UNKNOWN and using it instead of null but I can't argue that would be better.

PR #2490, opened 2024-11-01
Peace Turkey skeleton

gausie:
- **Peace Turkey skeleton**
- **Add modifiers for Helping Fingers**

Should peace turkey be `drops,other1,hp1,mp1` or should those just be wrapped up in `variable`? Whether it does each of those things is dependent on some state (what step of the apparent cycle it is in). In any case, it will also have `block,delevel,passive` as it has those abilities all the time.

For reference:

```
# none - does nothing
# stat0 - vollyball-like
# stat1 - sombrero-like
# item0 - Item Drop
# item1 - Food Drop
# item2 - Booze Drop
# item3 - Candy Drop
# meat0 - Meat Drop
# combat0 - Physical Attack
# combat1 - Elemental Attack
# drop - special drop
# block - potato-like
# delevel - barrrnacle-like
# hp0 - restore hp during combat
# mp0 - restore mp during combat
# meat1 - drops meat during combat
# stat2 - grants stats during combat
# other0 - does other things during combat
# hp1 - restore hp after combat
# mp1 - restore mp after combat
# stat3 - grants stats after combat
# other1 - does other things after combat
# passive - passive effect
# underwater - breaths underwater
# variable - varies according to equipment or other factors.
```

### DIFF
--- a/src/data/familiars.txt
+++ b/src/data/familiars.txt
@@ -335,5 +335,5 @@
 303	Burly Bodyguard	bodyguard.gif	combat0,block	baby bodyguard	bodyguard badge	0	0	0	0
 304	Doll Moll	molldoll.gif	combat0,meat	Doll Moll violin case	tiny Tina gun	0	0	0	0
 305	Emberiza Aureola	emberbird.gif	combat1	ember egg	tiny ember	0	0	0	0
-306	Peace Turkey	pto_fam.gif	drop,delevel,block,passive	peace turkey outline		0	0	0	0
+306	Peace Turkey	pto_fam.gif	drop,delevel,block,passive,other1	peace turkey outline		0	0	0	0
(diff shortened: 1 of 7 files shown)


src/data/familiars.txt
```diff
@@ -335,5 +335,5 @@
 303	Burly Bodyguard	bodyguard.gif	combat0,block	baby bodyguard	bodyguard badge	0	0	0	0
 304	Doll Moll	molldoll.gif	combat0,meat	Doll Moll violin case	tiny Tina gun	0	0	0	0
 305	Emberiza Aureola	emberbird.gif	combat1	ember egg	tiny ember	0	0	0	0
-306
+306	Peace Turkey	pto_fam.gif	drop,delevel,block,passive,other1	peace turkey outline		0	0	0	0
```
midgleyc, line 338: add tie-dye headband

midgleyc (approved):

PR #2493, opened 2024-11-06
update artifacts api to v4

Rinn:
GitHub is deprecating v3, I think we can just update the version since we're not doing anything weird but I can't really test this without it getting pushed because we only use it for making releases


midgleyc:
By [the migration docs](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md) I think the `Upload binaries` task needs changing because it's currently run on all OSs, and won't modify the existing artifact now that they're immutable -- only the first one will win.

jar is fine because it's only run on ubuntu.

Rinn:
can't look at this anymore someone else will need to, sorry

PR #2498, opened 2024-11-07
Track Cookbookbat Quest Monster

Ignose:
Need HTML from getting a quest for the test. More to come.

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -107,6 +107,7 @@ public class Preferences {
 
   private static final String[] resetOnAscension =
       new String[] {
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -107,6 +107,7 @@ public class Preferences {
 
   private static final String[] resetOnAscension =
       new String[] {
+        "_cookbookbatQuestMonster",
```
gausie, line 110: You don't need to manually reset _prefs on ascension

gausie: Although please add the one for peaceTurkeyIndex that I now realise I forgot

Ignose:
#2499 cares for this and more.

PR #2499, opened 2024-11-08
Cookbookbat Quest + Location tracking v2.0

phulin:
This builds on @Ignose 's PR, and is a separate PR for expediency - would be nice to get this in before the last push of the current challenge path season, which depends heavily on CBB quests.

This adds tracking for the Cookbookbat's ingredient quest monsters and suggested locations.


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -8553,6 +8559,59 @@ private static boolean handleGhostOfCommerce(String text, TagStatus status) {
     return false;
   }
 
+  private static final Pattern[] COOKBOOKBAT_QUEST = {
+    Pattern.compile(
+        "\"As I recall, .*? was common in (?<location>.*?), back in my day\\. +Perhaps if you kill an? (?<monster>.*?), you'll find one\\.\""),
```
midgleyc, line 8564: would be good to run with this a bit (I assume everybody will complain to garbo if it doesn't work), as I think pronouns can be more than just a/an: other places have a/an/the/some/(blank)

I don't know whether specifically CBB ingredient-monsters have such a variety

midgleyc (approved):

src/net/sourceforge/kolmafia/request/FightRequest.java
midgleyc: also would be nice to check that each of the messages matches, as I think we had that problem before with Robortender.

PR #2500, opened 2024-11-08
feat: add remaining video... games? effects

midgleyc:
Also remove hat drop / accessory drop etc. as I don't think these are
in-game.

### DIFF
--- a/src/net/sourceforge/kolmafia/modifiers/DoubleModifier.java
+++ b/src/net/sourceforge/kolmafia/modifiers/DoubleModifier.java
@@ -555,6 +555,8 @@ public enum DoubleModifier implements Modifier {
       "Avoid Attack",
       Pattern.compile("The first attack against you will always miss"),
       Pattern.compile("Avoid Attack: " + EXPR)),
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/modifiers/DoubleModifier.java
```diff
@@ -664,7 +694,7 @@ public static String parseModifier(final String enchantment) {
         String tag = mod.getTag();
 
         if (matcher.groupCount() == 0) {
-          return tag;
+          return tag + ": 1";
```
gausie, line 697: It would be nice to have a unit test for this some time, but going to approve the PR

gausie (approved):

src/net/sourceforge/kolmafia/modifiers/DoubleModifier.java
midgleyc: thanks!

I have more modifier related test changes to do with the tag stuff from #2494, I will add a test for this.

PR #2501, opened 2024-11-08
Complete Cookbookbat tracking.

phulin:
* Rename pref from `_cookbookbatSuggestedLocation` to `_cookbookbatLastLocation`, since the pref still has impact on game state once the CBB quest is over.
* Add tracking for CBB ingredients drops (every 11 combats).
* Add tracking for the "recharging" and "rerolling" behavior of CBB quests - every 5 combats, the quest gets rerolled, or, if you've completed it, you get a new quest.
* Completes the test suite with all quest messages.

PR #2502, opened 2024-11-10
CBB: lastVisitedLocation can be null.

phulin:
Fixes a possible NPE issue with #2501, thanks @libraryaddict.


gausie (approved):
Thank you

PR #2503, opened 2024-11-12
Reset lastNoncombatXXX on ascension.

phulin:
The `lastNoncombatXXX` preferences should reset on ascension, as `turns_spent` also resets to 0.

This PR makes that change, adds an `isResetOnAscension` function, and adds a consistency test to ensure the array is updated in the future.


gausie (approved):
Love it

PR #2507, opened 2024-11-14
feat: upgrade to new upload / download artifacts

midgleyc:
Artifacts are now immutable so different OSs need to download to different places.

Closes #2493.

Tested on https://github.com/midgleyc/kolmafia/actions/runs/11838300139, produced https://github.com/midgleyc/kolmafia/releases/tag/r28125.


gausie (approved):
LGTM

PR #2510, opened 2024-11-14
You can walk away from Adjust Parka and Configure Umbrella

phulin:
Should be self-explanatory. This will prevent Adjust Your Parka and Configure Your Unbreakable Umbrella from showing up in the noncombat_queue.


Veracity0 (approved):
I wonder when KoL changed that?

PR #2511, opened 2024-11-15
Add CBB quest ingredient tracking.

phulin:
Neglected this in the original PRs. Cookbookbat announces which ingredient the quest will result in, which is useful information for the user/scripts to have.

### DIFF
--- a/test/net/sourceforge/kolmafia/request/FightRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/FightRequestTest.java
@@ -222,26 +222,34 @@ public class FightRequestTest {
 
     @ParameterizedTest
     @CsvSource({
-      "test_fight_cookbookbat_quest_new_1.html,false,skullery maid,The Haunted Kitchen",
-      "test_fight_cookbookbat_quest_new_2.html,false,crate,Noob Cave",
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -222,26 +222,34 @@ public void handlesIngredientsCharge(String file, int chargeInitial, int chargeE
...
     public void handlesQuest(
-        String file, boolean isReminder, String monsterName, String locationName) {
+        String file,
+        boolean isReminder,
+        String monsterName,
+        String locationName,
+        String ingredientName) {
+      if (monsterName == null) monsterName = "";
+      if (locationName == null) locationName = "";
+      if (ingredientName == null) ingredientName = "";
```
jaadams5, line 240: These bother me.  The parameters are entirely under your control and if they are null then perhaps you should revisit how they are being set so that you don't have to check and reset.

I'll call this blocking at least until I understand why the fix is not as simple as I think it should be.

jaadams5 (changes requested):

jaadams5 (approved):

PR #2514, opened 2024-11-15
Update several familiar tags

Ignose:
These aren't complete, several familiars have tags we don't know or haven't documented. These are simply the known tags that are as of yet unupdated.


midgleyc:
construct doesn't exist, the tag is object.

Where are you getting these from?

Ignose:
Sure is - I can read I swear.

https://docs.google.com/spreadsheets/d/1gyU5IOjnkQPzw-ntkut8xcwbxVbzHP2ADm4d7ATSncU/edit?gid=1425085206#gid=1425085206

midgleyc:
None of those are confirmed except quantum entangler. You can tell by the tags be bold and the "Confirmed?" column being checked.

Ignose:
Suppose that answers my question; I'll limit this to just the entangler then.

midgleyc (approved):
thanks!

PR #2517, opened 2024-11-17
Deck testing

jaadams5:
This adds a test for DeckOfEveryCardRequest which had not had test before.  The actual tests are somewhat irrelevant because they are somewhat mechanical in nature but at least coverage is increased.

My long term goal is to use the Deck of Every Card in a test in such a way that the test will help debug the threadlock associated with #1889

While writing the tests I added a couple of helper functions to DeckOfEveryCardRequest and cleaned up some lint.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/DeckOfEveryCardRequest.java
+++ b/src/net/sourceforge/kolmafia/request/DeckOfEveryCardRequest.java
@@ -199,16 +199,17 @@ public class DeckOfEveryCardRequest extends GenericRequest {
     return DeckOfEveryCardRequest.buffToCard.get(buff);
   }
 
-  // Only used in testing
+  // Only used in testing.  Note can return null.
   public static EveryCard getCardById(int id) {
-    EveryCard retVal = idToCard.get(id);
-    if (retVal == null) {
-      retVal = new EveryCard();
-    }
-    return retVal;
+    return idToCard.get(id);
   }
 
   private final EveryCard card;
+
+  // Only used in testing.  Note can return null.
+  public EveryCard getRequestCard() {
+    return card;
+  }
 
   public DeckOfEveryCardRequest() {
     super("choice.php");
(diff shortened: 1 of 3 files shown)


test/net/sourceforge/kolmafia/request/DeckOfEveryCardRequestTest.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.List;
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.KoLConstants;
+import net.sourceforge.kolmafia.persistence.MonsterDatabase;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class DeckOfEveryCardRequestTest {
+  private static final String USERNAME = "DeckOfEvertCardRequestTest";
```
midgleyc, line 24:
```suggestion
  private static final String USERNAME = "DeckOfEveryCardRequestTest";
```

src/net/sourceforge/kolmafia/request/DeckOfEveryCardRequest.java
```diff
@@ -199,8 +199,18 @@ public static EveryCard buffToCard(AdventureResult buff) {
     return DeckOfEveryCardRequest.buffToCard.get(buff);
   }
 
+  // Only used in testing.  Note can return null.
+  public static EveryCard getCardById(int id) {
```
midgleyc, line 203: anything used only by tests doesn't have to be public: it can be package-private (i.e. no modifier)

This keeps the public interface for the class minimal.

jaadams5: Done.  I got in the habit of public because once something didn't work until I made it public.  I suspect I needed something from another package in a test but I established the wrong habit :-)

midgleyc (approved):

PR #2518, opened 2024-11-17
Detect changes in fax data

jaadams5:
This computes a hash on data file for a faxbot for the file currently on disk and then for the file after download.  If they are not the same then a global preference is set.  The preference will be set to true if the data for any faxbot has changed.  If there is no file before and after then there are no changes but the condition should be reported elsewhere.

The expected use is to allow a script to see whether the bot files have changed before running a check, for example factoids from Manuel.

To the extent this has been socialized see https://kolmafia.us/threads/some-thoughts-on-faxbots.30002/

Some lint was addressed in FaxBotDatabase.

There are no tests for FaxBotDatabase because it really needs to make a network call and so none were created.  Local testing verified that the preference is set as expected.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
@@ -1,8 +1,13 @@
 package net.sourceforge.kolmafia.persistence;
 
+import static net.sourceforge.kolmafia.preferences.Preferences.setBoolean;
+
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
+import java.math.BigInteger;
+import java.nio.file.Files;
+import java.security.MessageDigest;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -45,12 +50,12 @@ public class FaxBotDatabase {
 
   private FaxBotDatabase() {}
 
-  public static final void reconfigure() {
+  public static void reconfigure() {
     FaxBotDatabase.isInitialized = false;
     FaxBotDatabase.configure();
   }
 
-  public static final void configure() {
+  public static void configure() {
     if (FaxBotDatabase.isInitialized) {
       return;
     }
@@ -101,15 +106,14 @@ public class FaxBotDatabase {
           MafiaState.ABORT,
           "Could not load " + data.name + " configuration from \"" + data.URL + "\"");
       RequestLogger.printLine(FaxBotDatabase.faxBotErrorMessage);
-      return;
     }
   }
 
-  public static final FaxBot getFaxbot(final int i) {
+  public static FaxBot getFaxbot(final int i) {
     return FaxBotDatabase.faxbots.get(Math.max(0, i % faxbots.size()));
   }
 
-  public static final FaxBot getFaxbot(final String botName) {
+  public static FaxBot getFaxbot(final String botName) {
     for (FaxBot bot : faxbots) {
       if (bot == null) {
         continue;
@@ -122,7 +126,7 @@ public class FaxBotDatabase {
     return null;
   }
 
-  public static final List<FaxBot> getSortedFaxbots() {
+  public static List<FaxBot> getSortedFaxbots() {
     // Get preferred faxbot or null
     FaxBot preferred = getFaxbot(Preferences.getString("lastSuccessfulFaxbot"));
     // Use original list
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
```diff
@@ -440,6 +449,20 @@ public void run() {
...
     }
 
+    private static final String NOHASH = "Problem with hash";
+
+    private static String computeHash(File file) {
+      String checksum;
+      try {
+        byte[] data = Files.readAllBytes(file.toPath());
+        byte[] hash = MessageDigest.getInstance("MD5").digest(data);
+        checksum = new BigInteger(1, hash).toString(16);
```
heeheehee-kolmafia, line 459: Rather than parsing this as an integer, why not just output the hex representation, e.g. `HexFormat.of().formatHex(hash)`?

src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
```diff
@@ -440,6 +449,20 @@ public void run() {
       FaxBotDatabase.faxbots.addAll(bots);
     }
 
+    private static final String NOHASH = "Problem with hash";
+
+    private static String computeHash(File file) {
+      String checksum;
+      try {
+        byte[] data = Files.readAllBytes(file.toPath());
+        byte[] hash = MessageDigest.getInstance("MD5").digest(data);
```
heeheehee-kolmafia, line 458: Can we pick a hash algorithm that doesn't have known collision attacks? SHA-256 should be fine, or if you want to be really safe, SHA3-512 should also be broadly available.

heeheehee-kolmafia (changes requested):
> There are no tests for FaxBotDatabase because it really needs to make a network call and so none were created.

Can you mock the network call, similar to DebugDatabaseTest.checkMuseumPlurals?

https://github.com/kolmafia/kolmafia/blob/main/test/net/sourceforge/kolmafia/persistence/DebugDatabaseTest.java#L48

src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
jaadams5: Done

src/net/sourceforge/kolmafia/persistence/FaxBotDatabase.java
jaadams5: Done

jaadams5:
> > There are no tests for FaxBotDatabase because it really needs to make a network call and so none were created.
> 
> Can you mock the network call, similar to DebugDatabaseTest.checkMuseumPlurals?
> 
> https://github.com/kolmafia/kolmafia/blob/main/test/net/sourceforge/kolmafia/persistence/DebugDatabaseTest.java#L48

I'll look.  I didn't find a good example to learn from and the test is complicated because another test (which I wish I had paid more attention to in reviews)  actually uses git managed versions of the faxbot files.  Feel free to approve this and trust me to get to a test "soon"  ;-)

heeheehee-kolmafia:
Admittedly, this approach strikes me as atypical compared to how caching normally works over HTTP -- we're always downloading the file, whether or not there's a new copy, and always computing the hash.

In the typical case, I'd expect us to request the content with an If-Modified-Since header, then just rely on 304 Not Modified as our signal for whether the object was modified.

I don't see that as in-scope for this particular feature, and there's nothing stopping us from replacing the internal implementation details later. I would _prefer_ there to be tests, and I would similarly _prefer_ the various refactoring changes (dropping of "final", replacing `.equals("")` with `.isEmpty()`) to be in a standalone PR, but I don't have any blocking concerns.

jaadams5:
> Admittedly, this approach strikes me as atypical compared to how caching normally works over HTTP -- we're always downloading the file, whether or not there's a new copy, and always computing the hash.
> 
> In the typical case, I'd expect us to request the content with an If-Modified-Since header, then just rely on 304 Not Modified as our signal for whether the object was modified.
> 
> I don't see that as in-scope for this particular feature, and there's nothing stopping us from replacing the internal implementation details later. I would _prefer_ there to be tests, and I would similarly _prefer_ the various refactoring changes (dropping of "final", replacing `.equals("")` with `.isEmpty()`) to be in a standalone PR, but I don't have any blocking concerns.

My use case is an answer to the question - did the local file on disk change as a result of KoLmafia's efforts to download?  I didn't really want to explore caching because timestamps can change even when content doesn't.   My change does not effect whether KolMafia decides to download something or not.  If the second hash is expensive then a subsequent change should make that conditional on whether something was downloaded but that can wait.

I waffle back and forth about when lint needs to be in a separate PR.  Sometimes I make one and sometimes I don't.  No matter what I do there is always someone who suggests they would have preferred the other way.  I tend to make my decision on how many lines lint changes and whether the changes are simple and obvious.  

I am working on a test.  We agree that is a good idea.  I got sidetracked by the fact that some existing tests use files in data that were placed there by git.  I am working on "fixing" that in a separate PR.

Thank you.

PR #2520, opened 2024-11-18
A data file handling change and cleanup

jaadams5:
Prior to this change there were commits that put files in data and tests that depended upon those files being there.  In hindsight that might not have been the best decision if only because it forces all tests to use the same version of the file.  This became an issue in the process of developing an as yet unsubmitted test for FaxBotDatabase.

This change moves the "managed" files out of data and provides a cleanup that will copy files to data.  The tests that were broken by the change were then modified to use the cleanup.


test/net/sourceforge/kolmafia/textui/command/FaxbotCommandTest.java
```diff
@@ -31,10 +33,32 @@ public static void beforeAll() {
     // Set to true so we make http requests
     ChatManager.setChatLiteracy(true);
 
+    // This needs to happen before database configuration.
+    // ignoring cleanup
+
+    var ignoredCleanup =
```
midgleyc, line 39: if you store the cleanup in the class (statically) you can call `close` on it in `afterAll`.

test/internal/helpers/Player.java
```diff
@@ -2694,4 +2698,29 @@ public static Cleanups withoutCoinmasterSellItem(CoinmasterData data, AdventureR
...
+  /**
+   * Copies the source file from root/provided_data to root/data giving it the destination name.
+   * Deletes the copied file as part of cleanup. No error checking, specifically that the source
+   * file is present or that the destination file was written successfully.
+   *
+   * @param sourceName the name of the source file in root/provided_data
+   * @param destinationName the name of the destination file in root/data.
+   * @return deletes the destination file
+   */
+  public static Cleanups withDataFile(String sourceName, String destinationName) {
```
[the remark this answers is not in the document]
jaadams5: Done, or will be.  I was thinking of a test where the source files were something like file1.txt, file2.txt and so on and the destination was always file.txt.  The difference in file contents was the point of the test.  But now we can have it both ways :-)

test/net/sourceforge/kolmafia/textui/command/FaxbotCommandTest.java
jaadams5: Done.  Thanks for the suggestion.

midgleyc (approved):

PR #2522, opened 2024-11-20
Track several charpane elements.

phulin:
Responds to @soolar's request here: https://kolmafia.us/threads/a-request-to-make-some-character-pane-information-acquirable-without-visit_url.29996/

This PR adds a number of elements to be tracked by mafia:
* What specific noncombat forcers are active, via pref `noncombatForcers`.
* Current avatar, which is already tracked, via `get_avatar` function.
* Current custom title, which is not already tracked, via `get_title` function.
* The adventure trail (last 5 visited locations), via `lastAdventureTrail` property. NB maybe this doesn't need to be a preference and should have its own RuntimeLibrary function, would be open to feedback there.
* The container for the last adventure zone (which you get by clicking on the "Last Adventure" link in the native charpane), via `lastAdventureContainer` property - maybe we don't need this, since you can figure this out at runtime based on the zone. Mafia doesn't have existing precise map to the KoL "container" but it could be done in-script or in Mafia.

Open to any and all feedback.


midgleyc:
Can you confirm this also works with the compact character pane? Or minimally doesn't crash?

gausie:
> * What specific noncombat forcers are active, via pref noncombatForcers.

We already have tracking for each of these individually, perhaps it would be better to map these to the existing preferences updating them to a best guess if they don't match.

gausie:
> > > * What specific noncombat forcers are active, via pref noncombatForcers.
> > 
> > 
> > We already have tracking for each of these individually, perhaps it would be better to map these to the existing preferences updating them to a best guess if they don't match.
> 
> We don't have tracking for these, do we? We have e.g. `_claraBellUsed` but that tells you whether it's been used before today, not whether it's currently active.

I know we at least have _stenchJellyUses or whatever, which increments and decrements accordingly.

midgleyc (approved):
I think this is fine as-is, and I don't think you can meaningfully update the other prefs merely from the character pane (e.g. if you use clara's bell before rollover the forcer is active but you can use it again today)

PR #2523, opened 2024-11-21
Test FaxBotDatabase

jaadams5:
In response to a request by @heeheehee-kolmafia at https://github.com/kolmafia/kolmafia/pull/2518

The first part of the test is basically tests that do nothing interesting in terms of testing functionality but according to local checks, increases the coverage.

The second part attempts to test that _faxDataChanged is set.  One test starts out with no files in data, fakes a download and detects that no file to file is a change.  The second starts out with files a download and detects that one file changed.

The first test works when FileUtilities uses a "normal" Http client instead of a resetting one.  The second test has never worked.  

This is being shared as a draft to get help in extending a resetting client so it can be fed a fake response (or making a new cleanup) and get some ideas about why the second test isn't working.


jaadams5:
The test I had envisioned was not possible because there is currently no support in the test framework for inserting data into a response from a ResettingHttpClient.  Given that ResettingHttpClient was introduced to deal with a GOAWAY issue that may be addressed in a future release of Java the work to insert data may very well be discarded.  Experimentation came up with the changes to FileUtilities.  In theory those effect noting unless a programmer opts in to use them so it is hopefully not a future problem.

src/net/sourceforge/kolmafia/utilities/FileUtilities.java
```diff
@@ -209,13 +214,23 @@ private static HttpResponse<InputStream> getResponseFromRequest(
...
     }
 
-    var client = getClient();
     HttpResponse<InputStream> response;
-    try {
-      response = client.send(request, BodyHandlers.ofInputStream());
-    } catch (IOException | InterruptedException e) {
-      StaticEntity.printStackTrace(e);
-      return null;
+    if (testingWithFakeData) {
```
midgleyc, line 218: don't change the source code for the tests -- instead, make a `resetClient`, same as `GenericRequest` has, that you call in your tests when you want future calls to use the fake client instead of a real client, after mocking the fake client

test/net/sourceforge/kolmafia/persistence/FaxBotDatabaseTest.java
```diff
@@ -172,27 +173,31 @@ public void checkPreferenceWhenNoFiles() {
...
+      FileUtilities.setTestingWithFakeData(true);
+      var cleanups = new Cleanups(withNextResponse(200, ""), withProperty(property, false));
       try (cleanups) {
         assertFalse(getBoolean(property));
         FaxBotDatabase.configure();
         // Not sure why not working
-        // assertTrue(getBoolean(property));
+        assertTrue(getBoolean(property));
       }
+      FileUtilities.setTestingWithFakeData(false);
```
midgleyc, line 197: also this set / unset should be in a try / finally as if the test fails the unset won't get called

jaadams5:
Draft while I ponder comments.

src/net/sourceforge/kolmafia/utilities/FileUtilities.java
jaadams5: I'm not understanding the suggestion well enough to implement.  Agree that changing the core code is not the best thing.  What is your proposed resetClient a method of?  Maybe that will get me started.  Thanks.

test/net/sourceforge/kolmafia/persistence/FaxBotDatabaseTest.java
jaadams5: I think you are saying that if I implement your previous suggestion correctly this will not be needed and the test alone would perform the function.

jaadams5:
Undrafting.  The disabled test depends upon injecting test data into a resetting request and I still have not managed to do so.  But at least the rest of the test increases coverage :-)

PR #2528, opened 2024-11-28
Track power pill progress.

phulin:
Tracks Puck/Ms. Puck power pill progress; drop rate is dependent on how many combats you've won with the familiar.


jaadams5 (approved):

midgleyc:
if it resets on ascension, can you add it to the list of relevant prefs?

jaadams5 (approved):

PR #2529, opened 2024-11-29
Fix FoB + Teetotaler typo.

heeheehee-kolmafia:
Teetotalers do benefit from Feast of Boris, but Mafia at some point was attributing this to the "Teetotaller" path.

It may make sense to add checks to prevent similar typos in the future, e.g. if path() or certain other built-in modifier functions fail to resolve.


src/data/modifiers.txt
```diff
@@ -8862,7 +8862,7 @@ Event	Mysticality Day	Experience Percent (Mysticality): [25*interact()]
 Event	Festival of Jarlsberg	Mana Cost: -3
 Event	Lab&oacute;r Day Eve	Adventures: +10
 # Fullness only applies in no path or Teetotaller, and does not apply to AoWoL classes for some reason
```
midgleyc, line 8864:
```suggestion
# Fullness only applies in no path or Teetotaler, and does not apply to AoWoL classes for some reason
```

heeheehee-kolmafia: Done.

midgleyc (approved):

PR #2532, opened 2024-12-01
Reverse Roman Candelabra modifiers.

phulin:
The elemental damage is off, not on, when you have the effects.


jaadams5 (approved):

midgleyc (approved):
force pushed because something went wrong GitHub side

PR #2534, opened 2024-12-04
QT Maximizer and UI Fixes

Alium58:
Fixes https://kolmafia.us/threads/qt-maximizer-issue.30013/
Solution - Ryo's suggested code update to refresh KoL state
Completed 2 autoscend QT runs without issue. Previously was getting multiple failures per day of the run

Fixes https://kolmafia.us/threads/next-fam-in-qt-on-top-of-text.30014/
Solution - change text position


src/net/sourceforge/kolmafia/maximizer/Maximizer.java
```diff
@@ -133,6 +136,11 @@ public static void maximize(
       return;
     }
 
+    if (KoLCharacter.inQuantum()) {
```
midgleyc, line 139:
```suggestion
    // ensure character state is updated
    if (KoLCharacter.inQuantum()) {
```

midgleyc (approved):

PR #2535, opened 2024-12-05
feat: initial dec iotm dump

midgleyc:
Some items remain (e.g. familiar hatchlings) but seems enough for a first skeleton.

### DIFF
--- a/src/data/classskills.txt
+++ b/src/data/classskills.txt
@@ -1092,6 +1092,7 @@
 7531	Summon Cauldron of Bats	batcauldron.gif	combat	1	0
 7532	Assert your Authority	bakebadge.gif	combat	0	0
 7533	Check for photo booth supplies	document.gif	combat	0	0
+7534	Iron Tricorn Headbutt	ts_tricorn.gif	combat	0	0
 
 # Skills granted by books on your Mystical Bookshelf
 # These skills were renumbered.  This section kept for historical purposes only.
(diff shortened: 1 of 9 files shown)


jaadams5 (approved):

midgleyc:
for the test: if we can get the `equalTo` stdout then we can at least find out what exactly is going wrong, which should help. If it's failing consistently in 2471 that looks like a good place to just print the values to the output directly (although I think we should be able to get it out somehow?)

For the file access: I've only seen this when another test fails. we could catch FileNotFoundException and retry

jaadams5:
> for the test: if we can get the `equalTo` stdout then we can at least find out what exactly is going wrong, which should help. If it's failing consistently in 2471 that looks like a good place to just print the values to the output directly (although I think we should be able to get it out somehow?)
> 
> For the file access: I've only seen this when another test fails. we could catch FileNotFoundException and retry

I'll remember this when I dig again.  My hypothesis remains that the test fails because a file is not present.  The absence of the file is not the failure but the lack of content is.  The file is absent because it is in a shared directory and most of the cleanup deletes the directory or entire contents and not just the files that were created by a particular test.  This explains almost all of the test failures and is supported by the absence of failures when Gradle is not allowed to use more than one fork.

There are a couple of tests that simulate keyboard input that sometimes fail but this hypothesis does not seem to apply to them.

I cited #2471 but I have been causally gathering data since #1889 started having failures that initially were believed to not occur elsewhere.

Thanks.

PR #2539, opened 2024-12-13
bbitracking

ckb11:
Some properties to add for the Bugbear Invasion Mothership:

- track # of [handful of juicy garbage] used = juicyGarbageUsed
- track [quantum nanopolymer spider webs] used = nanopolymerSpiderWebsUsed
- track [bugbear autopsy tweezers] used = autopsyTweezersUsed
- track [drone self-destruct chips] used = droneSelfDestructChipsUsed


src/data/defaults.txt
```diff
@@ -3474,4 +3478,4 @@ user	lastNoncombat450	0
 user	lastNoncombat533	0
 user	lastNoncombat539	0
 user	lastNoncombat540	0
-user	lastNoncombat541	0
+user	lastNoncombat541	0
```
jaadams5, line 3481: What changed?  I assume whitespace but since we have tests that should catch some whitespace errors in data files it would be useful to know whether the tests need to be looked at.

jaadams5 (changes requested):
Conditionally approved.  I guess I do want to know about the line in defaults.txt and I would like another viewer.

src/data/defaults.txt
ckb11: I took another look at defaults.txt - it looks like the original has an extra newline at the end and my version does not

ckb11: And fixed

midgleyc (approved):
Thanks!

midgleyc:
One question: what's the point of tracking handful of juicy garbage uses? Wiki suggests the drops are just % chance -- is it for spading if the quest item is forced? Is it?

ckb11:
> One question: what's the point of tracking handful of juicy garbage uses? Wiki suggests the drops are just % chance -- is it for spading if the quest item is forced? Is it?

Partially for spading, partially for tracking how many you have used so far.  In my past 4 run it has been 8 or 9, and you can use resources (vortex, swoop, dupe, etc.) to get more, so it is useful to know about how many more you might need when you are in a fight.

jaadams5 (approved):

PR #2545, opened 2024-12-15
fix: detect choice adventure name

midgleyc:


heeheehee-kolmafia:
There are quite a few failing tests, probably because of testdata files that haven't been updated to the new format.

midgleyc:
Yes. I'm wondering whether it would be easier to support the old version as a fallback.

heeheehee-kolmafia:
I think longer-term we should look at moving away from regex for parsing HTML (in favor of, say, xpath), but yes, I think it'd be simpler to just support both versions for the time being.

midgleyc:
Okay, so now the problem is that NCs and choices put their adventure name in different places, and it was previously using the same logic for both.

Plus whatever's wrong with Dvorak and CMC.

midgleyc:
Okay, the CMC test wasn't returning proper HTML and I guess regex crashes if it's passed a -1 index while string doesn't.

Neither the Dvorak nor Rufus tests fail when I run them alone which means we are in fun territory.

src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -1010,6 +1010,8 @@ private static String parseNoncombatEncounter(final String urlString, final Stri
     return parseEncounter(responseText);
   }
 
+  private static final Pattern BOLD_ENCOUNTER = Pattern.compile("<b(?:| [^>]*)>(.*?)</b>");
```
heeheehee-kolmafia, line 1013: optional: instead of an empty alternative, you could just use `(?: [^>]*)?`.

src/net/sourceforge/kolmafia/request/AdventureRequest.java
```diff
@@ -1029,18 +1031,17 @@ public static String parseEncounter(final String responseText) {
       }
     }
 
-    int boldIndex = responseText.indexOf("<b>", index);
-    if (boldIndex == -1) {
+    if (index == -1) {
```
heeheehee-kolmafia, line 1034: If you wanted, you could probably use `responseText.indexOf("<b", index)` as an approximation of the previous logic.

heeheehee-kolmafia (approved):
A couple of optional comments, but it's fine to merge without either.

src/net/sourceforge/kolmafia/request/AdventureRequest.java
midgleyc: That matches `<body` (although wouldn't with `index + 1`) and `<br` (which I think shouldn't occur in the places we look but can't guarantee)

src/net/sourceforge/kolmafia/request/AdventureRequest.java
midgleyc: At the moment I prefer the empty but we can always change it later

PR #2555, opened 2024-12-17
feat: crimbo zones are 100% combat with NCs every 11 adventures

midgleyc:
Also you can reduce the combat rate to get more NCs like the Spooky
Forest.

Update items from mall, assume rares are 0.1% because it's looking that way.


Veracity0 (approved):
Thanks!

PR #2564, opened 2024-12-21
feat: thanksgiving island

midgleyc:


src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
```diff
@@ -6324,10 +6324,19 @@ public static ChoiceOption findOption(final ChoiceOption[] options, final int de
         1540,
         "Holiday Islands",
         "War is Like Hell: Very Hot",
-        new ChoiceOption("fight ???", 1),
+        new ChoiceOption("fight some underbrush", 1),
```
gausie, line 6327: maybe this should be "and then brando" or whatever to be less confusing. For the next 10 days or whatever

gausie (approved):

src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
midgleyc: I just matched the wiki

PR #2571, opened 2024-12-23
New Crimbo 24 data & ShopRow Coinmaster tweaks

Veracity0:
1) New items from Advent Calendar and Crimbo24 Factory
2) If an existing shop has ShopRows, always print items in it in ShopRow format
3) Otherwise, buyRows print as "buy" and sellRows print as "sell"
4) Put ShopRowPanel in CENTER, not NORTH, which will not restrict it to 6 visible rows


test/net/sourceforge/kolmafia/DataFileMechanicsTest.java
```diff
@@ -30,7 +30,7 @@ public static Stream<Arguments> data() {
         Arguments.of("cafe_booze.txt", 1, 2, 2),
         Arguments.of("cafe_food.txt", 1, 2, 2),
         Arguments.of("classskills.txt", 6, 6, 7),
-        Arguments.of("coinmasters.txt", 2, 4, 5),
+        Arguments.of("coinmasters.txt", 3, 4, 7),
```
midgleyc, line 33: this should probably be removed entirely, as the new format can have an unlimited number of currencies (right?)

Veracity0: In theory, yes, although it looks like shop.php always has 5 costs in it (some of which are empty).
In which case the max should be 8, not 7. :(

midgleyc (approved):

PR #2575, opened 2024-12-23
Finish support for ShopRow Coinmasters

Veracity0:
This should (almost) complete support for ShopRow style Coinmasters

- coinmaster.txt contains new-style shop.php entries with up to 5 currencies per item sold
- (when we visit shop.php, we will detect new items - and currencies - and log such.)
- CoinmastersFrame has a new ShopRowPanel class which supports that kind of Coinmaster.
- As do CoinmasterData and CoinMasterRequest.

I haven't actually tested how/if multi-ingredient CoinmasterPurchaseRequests work.
Those would show up in the Creatable panel of the ItemManager.
I should do that...

Take a look at Crimbo24FactoryRequest.java to see just how easy it is to code this, compared to the previous Coinmasters that have multiple currencies that are NOT specified in coinmasters.txt

I could write tests for various components of this effort (like NPCPurchaseRequest.parseShopInventory).
And, having done that, I should figure out how to parse the Mystic, who has a variant format of shop.php with rows.
And maybe I could make a creation method - like the Kiwi Kwiki Mart - into a CoinMaster.

I'd prefer to do those things in a follow up PR.
I need a break from coding. :)

### DIFF
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -11789,13 +11789,13 @@
 11761	Secrets of the Master Egg Hunters	943549721	book5.gif	usable	t	0	copies of Secrets of the Master Egg Hunters
 11762	Secrets of the Master Egg Hunters (used)	216321799	book5.gif	reusable		0	used copies of Secrets of the Master Egg Hunters
 11763	How to Lose Friends and Attract Snakes	465814056	book3.gif	usable	t	0	copies of How to Lose Friends and Attract Snakes
-11764	How to Lose Friends and Attract Snakes (used)	953745853	book3.gif	reusable		0	used copies of How to Lose Friends and Attract Sna
+11764	How to Lose Friends and Attract Snakes (used)	953745853	book3.gif	reusable		0	used copies of How to Lose Friends and Attract Snakes
 11765	Covert Ops for Kids	484772100	book2.gif	usable	t	0	copies of Covert Ops for Kids
 11766	Covert Ops for Kids (used)	700189532	book2.gif	reusable		0	used copies of Covert Ops for Kids
 11767	Holiday Multitasking	555431420	book3.gif	usable	t	0	copies of Holiday Multitasking
 11768	Holiday Multitasking (used)	639002731	book3.gif	reusable		0	used copies of Holiday Multitasking
 11769	Hoyle's Guide to Reindeer Games	429457903	book4.gif	usable	t	0	copies of Hoyle's Guide to Reindeer Games
-11770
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/request/Crimbo24FactoryRequest.java
```diff
@@ -50,10 +42,11 @@ public static void parseResponse(final String location, final String responseTex
   }
 
   public static String accessible() {
-    return "";
+    return null;
   }
 
   public static final boolean registerRequest(final String urlString) {
+    System.out.println("Crimbo24Factory.registerRequest(" + urlString + ")");
```
midgleyc, line 49: presumably leftover debugging code?

Veracity0: Yeah.

Veracity0:
Upon further thinking, I realize that yes - CoinmasterPurchaseRequests are not complete for this kind of Coinmaster.

ShopRow Coinmasters log requests and parse responses and remove ingredients - whether via visit_url, the Request classes (as used in CoinmastersFrame), or in the browser.

However, even with autoSatisfyUsingCoinmasters, they probably do not work for "acquire", the Create panel in the ItemManager, or in MallSearch results for one reason: all of those use CoinmasterPurchaseRequests, which depend on finding the Coinmaster via an itemId - and shop.php does not require that a particular item be fulfilled by a single row in a single shop.

I'm not sure how to deal with that, yet. I'm going to wait until after Christmas before thinking more about it.

For now, people will have to be satisfied with the CoinmastersFrame.

Veracity0:
What did I do that threw away my entire commit?

midgleyc:
This has already been merged (mysteriously), and now the only thing on this branch is the debug print.

I'm not sure what happened here, but as we don't want the debug print I'm going to close the PR.

midgleyc:
Or you could remove the debug print and I'll merge this. Up to you.

Veracity0:
I screwed up my git.

I had a branch that I was working in.
I wanted to pull in the latest commits before I'd opened a PR.
I ... somehow ... set the upstream branch to allow that.
Later, I committed and pushed and created a PR.

Later still, I wanted to remove the DEBUG print and git would not allow me to push, since "there was code upstream I did not have". I futzed arounds and, somehow, pushed to main.

So, I ended up with a couple commits that were not merged via the PR.
They are reported in ckb's Skill Book PR.

https://github.com/kolmafia/kolmafia/compare/r28188...r28192

So, all my new code - including removing the debug print - is in.

Until I figure out how to set the upstream to the same thing git tells me to do when I try to do my first push, I'll need to push and open a PR and then update through the PR, like I've always done before.

I am sorry.

PR #2580, opened 2024-12-25
feat: warn on ascension with remaining takerspace ingredients

midgleyc:


Veracity0 (commented):
I actually haven't tried this workshed item, yet, but I will when I ascend in January and my model train set is out of Standard.

You list ingredients as (1/0/1/1/0/0), for example.
Are we supposed to know which ingredient is which?
Or does it even matter; the fact that any are non-zero means you should look?

Would it be more informative to have
(1 spice/0 rum/1 anchor/1 mast/0 silk/0 gold) or
(1 spice/1/anchor/1 mast)

I assume that experienced crafters know what each ingredient is for, and it seems like listing the non-0 ingredients might be informative.

midgleyc:
so if any are non-zero you should look, but I do know the order and I'd expect people who use takerspace a lot to know the order.

The primary function of the ingredients was, at a glance
1. how many silks / gold do I have (the last two)
2. how many spices do I have (the first one)

The others aren't so important.

I can add the words to each, but it makes it less at-a-glance-y.

jaadams5:
I was going to take Christmas off, but I too wondered about the message.  It should be useful for people who don't use takespace often, so I would prefer the name of the item added to the list and might consider not reporting zeros, i.e. "(1 spice/1/anchor/1 mast)".  But for most of the items in the ToDo list if the item is present then I do, or do not, do something about it.  So for me "You have unused takerspace ingredients" is sufficient.  

I also wonder about a link.  The message would (again)  be something like "You have unused takerspace ingredients"  and that decorates a link that takes you to someplace where you can see what you have and do something with them.  ChoiceAdv 1537?

midgleyc:
There is already a link taking you to your workspace.

Veracity0 (approved):

PR #2583, opened 2024-12-26
tests: add more tests for functions using xpath

midgleyc:


heeheehee-kolmafia (approved):

midgleyc:
While stepping through the code with the `AdventureRequest` test: the `handleRandomModifiers` function is wrong, but not in a way that affects anything visible for known monsters. It assumes all the modifiers in "var ocrs" are OCRS modifiers, but this isn't necessarily true, so the lines
```java
      remove += (j == count) ? " " : ", ";

      monsterName = StringUtilities.singleStringDelete(monsterName, remove);
```
can result in the last OCRS modifier having ", " appended and so not being matched.

But we get the encounter name from the monster ID, not the monster name, so this winds up not mattering.

PR #2588, opened 2024-12-30
tests: add test for parsing famteam.php

midgleyc:


Veracity0 (approved):
I assume this is because you are thinking of xpathifying the parsing?

midgleyc:
That or Jsouping it. Either way, I think I want a bit more coverage for the pokefam stuff.

midgleyc:
JSoup was also why I was adding tests for the other xpath stuff, but I think that will stay as it is.

PR #2589, opened 2024-12-30
shop.php ROW investigation

Veracity0:
I want to experiment with shop.php using ShopRow Coinmasters.
I decided to validate one of my assumptions - shop rows are unique across shops.
We model such shops as coinmasters, concoctions, and npc stores.

This PR creates a ShopRowDatabase, containing ShopRowData records, stored in shoprows.txt

I generated that file by registering a ShopRowData for all ROW items seen in coinmasters.txt, npcstores.txt, and concoctions.txt.

The "test write-shoprows" commands writes out shoprows.txt containing all the scraped data.

These are the fields:

row - integer, in numeric order in the file
type - conc, coin, or npc
item - item, or CURRENCY (n), if from a Coinmaster "sell" entry, since the data files do not tell us the item
shopName - from data files, which may or may not be exactly what KoL calls it

Counts:

284 concoctions
620 coinmasters
-- 603 buy
-- 17 sell
337 npcstores

minimum row is 3, maximum is 1548
1241 used, so 307 are unknown

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/CoinmastersDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/CoinmastersDatabase.java
@@ -196,22 +196,28 @@ public class CoinmastersDatabase {
           }
         }
 
-        if (row != null) {
-          ShopRowDatabase.registerShopRow(row, "coin", item, master);
-        }
-
         if (type.equals("buy")) {
           List<AdventureResult> list = getOrMakeList(master, buyItems);
           list.add(item.getInstance(purchaseLimit(iitemId)));
 
           Map<Integer, Integer> map = getOrMakeMap(master, buyPrices);
           map.put(iitemId, iprice);
+
+          if (row != null) {
+            ShopRowDatabase.registerShopRow(row, "coin", item, master);
+          }
         } else if (type.equals("sell")) {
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/persistence/CoinmastersDatabase.java
```diff
@@ -199,12 +202,22 @@ public static final Map<Integer, Integer> invert(final Map<Integer, Integer> map
...
+          }
         } else if (type.equals("sell")) {
           List<AdventureResult> list = getOrMakeList(master, sellItems);
           list.add(item);
 
           Map<Integer, Integer> map = getOrMakeMap(master, sellPrices);
           map.put(iitemId, iprice);
+
+          if (row != null) {
+            System.out.println("sell row = " + row + "  for " + master);
```
jaadams5, line 217: Will this print be removed eventually, or wrapped by something?  Not blocking :-)

jaadams5 (dismissed):

src/net/sourceforge/kolmafia/persistence/CoinmastersDatabase.java
Veracity0: Good point. It was a DEBUG print.
I'll remove it now and merger it without further review. :)

PR #2590, opened 2024-12-30
Refactor location-clearing code and add Numberology.

phulin:
This PR refactors common usage of the following code block to clear the stored location to a function in `KoLAdventure`:

```java
        KoLAdventure.lastVisitedLocation = null;
        KoLAdventure.lastLocationName = null;
        KoLAdventure.lastLocationURL = urlString;
        KoLAdventure.setLastAdventure("None");
        KoLAdventure.setNextAdventure("None");
```

Note that `setLastAdventure("None")` immediately sets the first three fields to null, so those sets are immediately overwritten. As a result, the new `clearLocation` function only contains the last two lines of the code block.

The PR also adds Numberology to the postChoice and ChoiceControl handling to make sure 151st infantrymen are not added to combat queues / turns spent.


jaadams5 (approved):
Approved.

PR #2593, opened 2025-01-01
feat: prefs for pirate dinghy, pumpkin spice whorl, infinite jelly

midgleyc:


jaadams5 (dismissed):
Where is _pirateDinghyUsed set?  I didn't see it and asking is easier than downloading and using a search tool.  Non-blocking.

midgleyc:
I had thought that `dailylimits.txt` automatically did that, but it might only do that for combat items. I'll check.

midgleyc:
yeah it doesn't, nice catch.

jaadams5 (approved):

PR #2594, opened 2025-01-01
feat: skeleton standard gear

midgleyc:


Veracity0 (approved):
Sweet. I'll be starting my first to Standard ascensions of the year in a day or two, and will get the pulverization items and the Armory & Leggery shop rows as soon as I have done one Normal and one Hardcore.
I remember in the past having to wait for days before we even knew what the item rewards were.
Are they showing up in Valhalla?

midgleyc:
They are -- they weren't public initially, but were made public some time today.

PR #2597, opened 2025-01-02
feat: devil egg command

midgleyc:


jaadams5 (approved):
Checking for the lack of a deviler seems counter intuitive to me but there is not good reason not to do it that way rather than reversing the check.

PR #2598, opened 2025-01-02
feat: standard effects, more candies

midgleyc:
Candies obtained from the candy egg deviler choice adventure using:
```
js var x = visitUrl("choice.php"); xpath(x, "//option/@value").filter(x
=> x!=0).map(x => Item.get(x)).filter(x => !x.candy)
```


Veracity0 (approved):
Is this parsing a dropdown? Does it show only items you have, or all candies?

gausie:
We should add info handling for a mismatch here like we have for genie
wishes

midgleyc:
It only shows candies in your inventory. Like Sweet Synthesis.

PR #2600, opened 2025-01-02
feat: mchugelarge skills, synergy

midgleyc:
Needs to wait for `noncomforcers` to be populated in `api.php` on the KoL end (and the name here to be corrected if it's not "avalanche"), and for spading to complete on how good Slash is.


Veracity0 (approved):

midgleyc:
eh let's just get it in and fix up the noncomforcers later if something other than "avalanche" is picked

PR #2601, opened 2025-01-03
feat: new ninja items

midgleyc:


jaadams5 (approved):
This is going to break things for any script that uses names and not IDs, correct?  I'll non-blocking approve but it might be appropriate to socialize this. before committing.

Veracity0:
> This is going to break things for any script that uses names and not IDs, correct?
It will break scripts that think they can pull the ninja items. But that's KoL; it's just how things work now.
For that matter, scripts that think the ninja assassin is copyable will also break. Ditto.

midgleyc:
I would expect it to un-break scripts that are tracking obtaining the ninja items from the snowmen directly.

But yeah, they're not pullable any more and attempts to copy the NSA will fail.

PR #2603, opened 2025-01-03
feat: free craft skills, expose ASH functions

midgleyc:
Add the functions 
* free_crafts()
* free_cooks()
* free_mixes()
* free_smiths()
exposing how many of each resource you have.

Support the three new Crimbo skills (one from this year, two from last year).


Veracity0 (approved):
Thanks!

PR #2606, opened 2025-01-04
Add 'to consume' checkbox to UseItemEnqueuePanel

Veracity0:
In the Usables section of the ItemManager, you can queue items to be consumed immediately or to be created.
If the former, we limit the number you can queue to the usable amount, obeying consumption limits.
If the latter, we ignore consumption limits and you can queue as many as you have ingredients for.

A new checkbox - "to consume" - controls this behavior.
A new preference - enqueueForConsumption - (default is true) is linked to the checkbox.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -657,6 +657,7 @@ user	encountersUntilNEPChoice	7
 user	encountersUntilSRChoice	11
 user	enamorangMonster
 user	enamorangMonsterTurn
+user	enqueueForConsumption	true
 user	ensorcelee
 user	ensorceleeLevel	0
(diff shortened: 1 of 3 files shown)


midgleyc (approved):
going to assume you've tested the UI yourself

Veracity0:
I tested with "addCreationQueue" set to false - in which case neither the checkbox nor the "enqueue" button appears.
I tested with that setting set to true, in which case the checkbox and enqueue button appear.
I asked to enqueue items with the checkbox checked - and it limited the number to my consumption limit.
I asked to enqueue items with the checkbox unchecked - and it limited the number to my available ingredients.

PR #2607, opened 2025-01-04
Add Hack Market NPC store

Veracity0:
This includes the fix to make unknown NPC stores log items in npcstores.txt format.


midgleyc (approved):
thanks!

PR #2610, opened 2025-01-04
Parse shops with tokens as currency. Fix Coinmaster bugs.

Veracity0:
1) Hermit was failing because we were not passing in a valid item id. I fixed that.
2) Supposedly Quartersmaster is failing.

I wondered if it is because it has a token, rather than an actual item as currency.
I don't have access to the island at war, at the moment, but the Sept-Ember Censer also uses a token.
I tried that out and it failed because it was passing an item ID instead of a row ID. I fixed that.
There seems to be no issue with tokens...

I think the Quartersmaster was failing for the same reason as the Hermit, but I can't test it.
3) I made ShopRow parsing detect shops that sell items for tokens, rather than items.
I have a test for that, and it can parse the Sept-Ember Censer now.

### DIFF
--- a/src/net/sourceforge/kolmafia/CoinmasterData.java
+++ b/src/net/sourceforge/kolmafia/CoinmasterData.java
@@ -790,6 +790,14 @@ public class CoinmasterData implements Comparable<CoinmasterData> {
     return row;
   }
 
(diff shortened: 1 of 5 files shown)


midgleyc (approved):
thanks!

PR #2611, opened 2025-01-04
missing `break` in numberology case of switch statement

horrible-little-slime:
Looks like this was introduced in #2590 and, scrolling through the PR, no other breaks were missed.


midgleyc (approved):

src/net/sourceforge/kolmafia/session/ChoiceControl.java
```diff
@@ -2882,6 +2882,7 @@ else if (text.contains("takes your nifty new watch")) {
           if (NumberologyManager.numberologyPrize(result).startsWith("fight")) {
             KoLAdventure.clearLocation();
           }
+          break;
```
jaadams5, line 2885: move this out one level please.  The idiom seems to be

`case
{
}
break`
 whereas this looks to be the functionally equivalent 
`case
{
break
}
`

PR #2612, opened 2025-01-04
fix: add breaks to unlikely fall throughs

midgleyc:
Addenda to #2611.


horrible-little-slime:
Is it worth making a workflow / linting rule / third thing analogous to the other two that detects this?

jaadams5 (approved):

midgleyc:
We currently don't have a linter and I don't know how I'd add one for Java.

I looked at migrating the existing switches to the "enhanced" version of switch where you don't have to put the explicit breaks in, but IntelliJ messes up the comments, so I didn't do that.

If it was trivial I would say yes but it's not ;)

jaadams5:
> Is it worth making a workflow / linting rule / third thing analogous to the other two that detects this?

We haven't found a tool that we trust in an automated scenario.  IntelliJ has an Analyze function that points out opportunities to tweak code but not all of them are actually appropriate for our codebase.  The fact that some code was written at least a decade ago is also a factor.

PR #2614, opened 2025-01-05
feat: add checkstyle to prevent uncommented switch fallthroughs

midgleyc:
A follow-up to #2612: ziz helpfully provided a minimal checkstyle config, so I've added a linter to the build.

Deliberate fall-throughs require a preceding comment (e.g. `// Fall through`) to not fail the build.

Checkstyle is configured to only test for exactly this problem we are having (convenient!) and can be run with `gradlew checkstyleMain`, `gradlew checkstyleLib`, and `gradlew checkstyleTest`, and will run as part of `gradlew check`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,12 @@ plugins {
 	id 'org.ajoberstar.grgit' version '5.2.2'
 	id 'org.panteleyev.jpackageplugin' version '1.6.0'
 	id "com.github.ben-manes.versions" version "0.51.0" // enables ./gradlew dependencyUpdates for outdated
+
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
Will this be a failing test for future commits?  (Just asking.  Yes or No is fine).

Has this been applied to the complete codebase or will it discover existing cases when a file is change for some other reason?

midgleyc:
Yes, if a PR introduces an uncommented fallthrough, the build will fail (in exactly the same way the build currently fails for a spotless violation, for example).

This has been applied to the current codebase.

PR #2615, opened 2025-01-05
Refactor: new 'shop' folder for shop.php support

Veracity0:
"shop" folder holds ShopItem.php and ShopRowDatabase.
Thank you IntelliJ.

It may end up with:

ShopRequest - will hold the shop.php related stuff from NPCPurchaseRequest. Superclass for CoinMasterClass?
ShopDatabase - at least, map shopId to shopName. Perhaps sets for npc, coinmaster, concoction categories?
ShopPurchaseRequest - superclass for NPCPurchaseRequst and CoinmasterPurchaseRequest?
ShopCreateItemRequest - subclass for shop.php creation methods?

There are times when I wish Java had multiple inheritance.

I realize I am breaking the "request" and "persistence" folder schema, but this is a big project and I want its core to all be grouped together.


midgleyc (approved):
We already group classes by what they're about, instead of what they do, with the maximizer.

Personally I prefer that style anyway.

PR #2616, opened 2025-01-05
feat: cyberzones in the cyberrealm

midgleyc:


src/data/combats.txt
```diff
@@ -120,6 +120,10 @@ Crimbo's Jelly	100	bottle of cheap whiskey	the abstract concept of poverty	pengu
 Crimbo's Reindeer	100	the Crimborg	the abstract concept of poverty	simple hobo
 Crimbo's Sack	100	endless winter	penguin mafioso	simple hobo
 Crimborg Collective Factory	100	<s>Killer</s> Festive Arc-Welding Elfbot	<s>Killer</s> Festive Decal-Applying Elfbot	<s>Killer</s> Festive Laser-Calibrating Elfbot	<s>Killer</s> Festive Weapons-Assembly Elfbot	Rudolphus of Crimborg
+# all the cyberzones have the common viruses, then one of the rare viruses and one hacker
```
Veracity0, line 123: There are 3 rare viruses and 3 hackers, and their distribution among the three zones is randomized? Or what?

Veracity0 (approved):

src/data/combats.txt
midgleyc: There are 5 hackers and 5 rare viruses, and each zone has one hacker and one rare virus.

Which one is explained by the bottom left drawer in the server room, so we'll want to parse that and write it to prefs as support improves.

PR #2619, opened 2025-01-05
fix: pirate dinghy unlocks island on use, not on obtain

midgleyc:
It's easy to find bugs when you actually ascend ;)


Veracity0 (approved):
I'm about to ascend. All done with Crimbo on all characters, and I've done enough coding for today, modulo new bugs detected.

PR #2621, opened 2025-01-05
feat: make $item[0] match the item called '0'

midgleyc:
This will also change the behaviour of `$item[1]` to be `1` instead of `seal-clubbing club`, which I figure is fine... but I'll think more overnight.


Veracity0 (approved):
I think this is fine, too.

PR #2625, opened 2025-01-06
fix: don't allow bat wings restore to be cast with +hp back item

midgleyc:
Need to test this actually works properly, but parking here.


midgleyc:
This works but means if you want to restore MP using the bat wings it fails, which isn't great.

gausie (approved):
I still think it's better than the alternative for now

PR #2628, opened 2025-01-06
Fix 'Use another A-Boo clue' decoration

Veracity0:
Yet another KoL change that broke a Relay Browser decoration.


midgleyc (approved):
having ids on elements? This is positive!

PR #2630, opened 2025-01-06
Coinmasters can live in zones, if zone is Removed coinmaster is automatically inaccessible

gausie:
As discussed in a previous PR

- **Allow coinmasters to live in zones, and use zone removal for accessibility unless overwritten**
- **Use coinmaster zone to eliminate crimbo24 coinmasters**
- **Do not rely on blank string meaning null:**


jaadams5 (approved):

Veracity0 (approved):
I think the convention would be withZone, but inZone is fine.

gausie:
I had some vague notion that "with" was a descriptor of the workings of the coinmaster, and verbs like "in" could act as descriptors of the coinmaster in the wider context of the game. Happy to change it in future.

PR #2645, opened 2025-01-09
feat: upgrade to Rhino 1.8

midgleyc:
Rhino 1.8 was released a few days ago: it has `String.matchAll` and a whole bunch of other modern JavaScript stuff: default parameters, null coalescing, findLast, most of proxy and reflect and promises, generators, symbol keys and many other things most scripters won't use.


gausie (approved):

midgleyc:
> some of the weird things people have to do in their browser console (like bulk using items that aren't multi-use) finally doable in the CLI

if you `use 20 ICEbox` in the CLI, it uses them one at a time, so what is this referring to?

PR #2647, opened 2025-01-09
Dedigitizer coinmaster

Veracity0:
I got the recipes from the Wiki and the ROWs from "/buy xxx".
The Wiki does not have the recipes for ```3d printed server room key``` or ```STATS+++ rom chip```, but I got the ROWs.

We'll need to add those two items in a followup PR.

I put the Dedigitizer in the "Aftercore" section of the CoinmastersFrame.
Perhaps it should go in the IOTM section.

Also, we'll need to fix "accessible" once we figure out how to handle the 3d printed server room key.
It's a day pass, so it will require the sort of "access granted today" code that other day passes have.

### DIFF
--- a/src/data/coinmasters.txt
+++ b/src/data/coinmasters.txt
@@ -1300,5 +1300,24 @@ The Dedigitizer	ROW1558	visual packet sniffer	1 (16)
 The Dedigitizer	ROW1577	synthetic coffee	1 (8)	0 (8)
 The Dedigitizer	ROW1578	iDrops	1 (16)	0 (16)
 The Dedigitizer	ROW1567	cybeer	1 (32)	dedigitizer schematic: cybeer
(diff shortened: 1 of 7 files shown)


midgleyc (approved):
It is present in "coolitems" as "cyberrealm", so it's exactly like FantasyRealm / PirateRealm.

You also don't need the server key in inventory -- I keep mine in my display case.

Veracity0:
> You also don't need the server key in inventory -- I keep mine in my display case.

Interesting. I bought it before I ascended, and KoL automatically put it into inventory for me.
So, I assumed it needed to be there.

Well, ick. Where else can it be? Closet? Storage? (not since KoL auto pulls it).

OK. I'll do a followup PR to add "crAlways" and "_crToday" preferences and will re-code accessible() to use those settings.

PR #2650, opened 2025-01-10
feat: cyber monsters are nocopy and resist non-cyber damage

midgleyc:


Veracity0 (approved):
You also cannot flyer them. I tried it during the island war, and, although I couldn't actually read the green text, it was something something can't flyer something something.

Or was it simply because I could not use items against them?

PR #2651, opened 2025-01-10
Add CyberRealm prefs for free fight tracking

Ignose:


Ignose:
This is missing a test; the page for the test does not currently contain FREEFREEFREE, so I'll leave this as a draft until it does and add the test then.

src/data/defaults.txt
```diff
@@ -1952,6 +1952,7 @@ user	_crimboTree	false
 user	_currentDartboard
 user	_cursedKegUsed	false
 user	_cursedMicrowaveUsed	false
+user	_cyberRealmFreeTurnsUsed    false
```
horrible-little-slime, line 1955:
```suggestion
user	_cyberRealmFreeTurnsUsed	false
```
this needs to be a tab and not spaces

Veracity0:
My code from AdventurePool.java:
```
  public static final int CYBER_ZONE_1 = 585;
  public static final int CYBER_ZONE_2 = 586;
  public static final int CYBER_ZONE_3 = 587;
```
From defaults.txt
```
user	_cyberFreeFights	0
user	_cyberZone1Turns	0
user	_cyberZone2Turns	0
user	_cyberZone3Turns	0
```
You should use these properties, rather than introducing your own.

PR #2652, opened 2025-01-10
add tearaway pants count to daily deeds

Malibu-Stacey:


midgleyc (changes requested):
also add the pref to the constructor with a listener

midgleyc (approved):
Thanks!

PR #2654, opened 2025-01-10
crAlways and _crToday properties

Veracity0:
1) crAlways property tracks  if you have permanent access to CyberRealm
2) _crToday property tracks if you have one-day access to CyberRealm
3) Add "Back to the Network Map" link if you've hacked the system in that zone today.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/coinmaster/shop/DedigitizerRequest.java
+++ b/src/net/sourceforge/kolmafia/request/coinmaster/shop/DedigitizerRequest.java
@@ -51,7 +51,7 @@ public class DedigitizerRequest extends CoinMasterRequest {
     if (Preferences.getBoolean("crAlways") || Preferences.getBoolean("_crToday")) {
       return null;
     }
-    return "You don't have a server room key in inventory.";
+    return "You can't access the server room.";
   }
 
   public static final boolean registerRequest(final String urlString) {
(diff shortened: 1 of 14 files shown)


src/net/sourceforge/kolmafia/request/coinmaster/shop/DedigitizerRequest.java
```diff
@@ -48,11 +48,10 @@ public static void parseResponse(final String location, final String responseTex
...
 
   public static String accessible() {
-    int serverRoomKey = SERVER_ROOM_KEY.getCount(KoLConstants.inventory);
-    if (serverRoomKey == 0) {
-      return "You don't have a server room key in inventory.";
+    if (Preferences.getBoolean("crAlways") || Preferences.getBoolean("_crToday")) {
+      return null;
     }
-    return null;
+    return "You don't have a server room key in inventory.";
```
midgleyc, line 54: perhaps more like "You can't access the server room."?

midgleyc (dismissed):
Thanks!

src/net/sourceforge/kolmafia/request/coinmaster/shop/DedigitizerRequest.java
Veracity0: Sure.

Veracity0:
I also updated QuestManagerTest and KoLAdventureValidationTest to test crAlways and _crToday, a la the PirateRealm and FantasyReal properties are handled, with regards to Zone access.

We'll have to add code to handle individual locations being done for the day - which will require their own tests, obviously.

midgleyc:
Preferences for turns spent (like Gingerbread city) would be helpful, so you can see how far away you are from the choice adventures.

midgleyc (approved):

Veracity0:
I agree. Tracking turns for the three zones will come in another PR.

PR #2659, opened 2025-01-12
Add tracking for mini kiwi aioli

Malibu-Stacey:


midgleyc (changes requested):
Please add a test for this.

You can test session log output using `RequestLoggerOutput.startStream(); [...] var text = RequestLoggerOutput.stopStream();` as seen in `FightRequestTest.java`.

HTML is preferred.

Look at `EatItemRequestTest.java`'s milk of magnesium tests for examples.

midgleyc (approved):
Thanks!

I didn't realise how weird the magnesium test was, so I've changed it to something more like what I was hoping for.

PR #2662, opened 2025-01-12
fix: fix kgb parsing

midgleyc:
Broken since #94.

Presumably an editor set up to trim trailing whitespace broke the line.


Veracity0 (approved):
This is funny.

One could argue that having a consequence without a regex and/or actions - i.e. only 2 fields - should be OK (as in this case) and change the parsing of the file to allow it.

But simply keeping the requirement that there be at least 4 fields and adding back the two tabs, as you did here, is fine.

midgleyc:
Yeah, that would also be fine. There's a test now so it can be caught, anyway.

PR #2664, opened 2025-01-12
Add MutexI for watches.

heeheehee-kolmafia:
This fixes the maximizer wanting to equip a second watch in ACCESSORY1 even when an existing watch is equipped in a later accessory slot.


heeheehee-kolmafia:
Note that with this change, we could probably kill off the special handling for NONSTACKABLE_WATCH in the maximizer. That said, I don't know if any scripts rely on the existence of the modifier, so we might not be able to remove the modifier entirely.

midgleyc (approved):
Thanks!

While autoscend doesn't currently use Nonstackable Watches, I was planning to add it: at the end of the day it wants to pull items that gives +adventures, and it needs to be aware of which ones are watches. If the list of Mutexes was exposed somehow (and presumably they had names so you could figure out which ones were watches without parsing all of them) that would also do it.

PR #2665, opened 2025-01-13
Add some files to DataFileMechanicsTest

jaadams5:
The typo in kgb parsing fixed in #2662 raised a question - why didn't a data file test find it?  The answer is that there is no data file test for consequences.txt but investigating that identified several data files that were candidates for the mechanics test.  This PR adds tests for them.  

A debug print identified several files that had a version number hard-coded in a test and the current version is different.  Those were fixed.  

That same debug showed that journeyman.txt, TCRS.astral_consumables.txt and TCRS.astral_pets.txt have no version in the file yet they are being read by a versioned reader in tests.  Adding a version number to the files was not investigated nor was the use of versioned readers in the application code itself.


midgleyc (approved):
Thanks!

Is there a point to the versioned readers, if we just ignore the versions? Are they used for anything?

PR #2667, opened 2025-01-13
feat: add Jsoup, translate FightRequest to Jsoup

midgleyc:
Slightly less weird cleaned HTML, easier debugging, similar speed. I hope to convert some of the regex matching to HTML parsing. We need to keep HtmlCleaner for the exposed `xpath` function; it also seems faster for the other xpath cases.

Good thing I wrote the pokefam tests, those functions have changed quite a lot.

Converted robotically with no attempt at using extra features:
```
new HtmlCleaner().clean -> Jsoup.parse
TagNode -> Element
TagNode.getName() -> Element.tagName()
TagNode.getAllChildren() -> Element.childNodes() // children() is just Elements
TagNode.getAttributeByName(X) -> Element.attr(X) // returns blank string instead of null on missing
TagNode.getAttributes() -> Element.attributes() // not a map, but can be iterated over
TagNode.getElementListByName("X", true) -> Element.children().select("X") // need children() to avoid matching current element
TagNode.getElementListByName("X", false) -> Element.select(">X")
TagNode.getElementsByName("X", true) -> Element.children().select("X") // need children() to avoid matching current element // need .size() instead of .length
TagNode.getElementsByName("X", false) -> Element.select(">X") // need .size() instead of .length
TagNode.getChildTags() -> Element.children()
TagNode.getText().toString() -> Element.wholeText() // text() collapses spaces
TagNode.getParent() -> Element.parent()
TagNode.getParent().removeChild(X) -> X.remove()
TagNode.findElementByName("X", true) -> Element.children().select("X").first()
TagNode.findElementByName("X", false) -> Element.selectFirst(">X")
TagNode.findElementByAttValue("X", "Y", true, false) -> Element.children().select("[X=Y]").first()
TagNode.findElementByAttValue("X", "Y", false, false) -> Element.selectFirst(">[X=Y]")
TagNode.evaluateXPath(X) -> Element.selectXpath(X, Node.class) // jsoup cannot select attributes (e.g. @href), html cleaner will accept invalid xpath (e.g. input[@type='checkbox']@checked)
BaseToken -> Node
BaseToken.getName() -> Node.nodeName() // in case you don't know whether you've got a comment or an actual node
ContentNode -> TextNode
ContentNode.getContent() -> TextNode.getWholeText() // text() collapses spaces
CommentNode -> Comment
CommentNode.getContent() -> Comment.getData()
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -92,6 +92,7 @@ dependencies {
 	}
 
 	implementation 'net.sourceforge.htmlcleaner:htmlcleaner:2.29'
+	implementation 'org.jsoup:jsoup:1.18.3'
 	implementation('org.eclipse.lsp4j:org.eclipse.lsp4j:0.23.1') {
 		exclude group: 'org.eclipse.xtend', module: 'org.eclipse.xtend.lib'
 	}
--- a/src/net/sourceforge/kolmafia/request/FamTeamRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FamTeamRequest.java
@@ -17,9 +17,8 @@ import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.session.ResultProcessor;
 import net.sourceforge.kolmafia.utilities.HTMLParserUtils;
 import net.sourceforge.kolmafia.utilities.StringUtilities;
-import org.htmlcleaner.CleanerProperties;
-import org.htmlcleaner.HtmlCleaner;
-import org.htmlcleaner.TagNode;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Element;
 
 public class FamTeamRequest extends GenericRequest {
   private static final Pattern ACTIVE_PATTERN =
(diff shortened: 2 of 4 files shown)


midgleyc:
Pokefam would also need more adjustment to avoid `unspecified macro action??` with no longer removing the nodes (near the `done = false` bit).

Veracity0:
I went looking for this and discovered you'd closed it.

What are your current thoughts on this project? How can it proceed?

midgleyc:
I need to fix the failing tests, and (more importantly IMO) I need to create more tests targeting the areas that fail, to ensure everything works the same way.

So far I've noticed:
* HTML entities like `&quot;` are converted to `"`
* the cleaned HTML is different... whether it is more correct or less correct I do not know. The KoL HTML is exclusively quirks mode, which is uncommon.
* the round logging is different, I think due to the HTML being different, for example
  * the trainset test fails because the image isn't where we expect
  * pokefam logs additional "unspecified macro action"
  * when fighting a hermetic seal with a patriotic eagle it logs `Round 3: Honor Wing, Release the Patriotic Screech! (0 Muscularity Points)` which I presume is from the combat skills list. Weird!

midgleyc:
Trainset failed because now <p> tags can contain <table> tags (which is correct in quirks mode), so the bit of the function that early returns stops parsing.

PR #2668, opened 2025-01-13
Find defense element from encounter if filedrawer not parsed

Veracity0:
The Half-Way choices have multiple encounters for each possible element.
If we've parsed the file drawer (part of breakfast), we can deduce the element from the _cyberZoneXDefense.
If we've not yet parsed it, we can attempt to deduce it from lastEncounter.

The list of such encounters is incomplete; it only has ons that I have personally seen.
We can fill it out in future PRs.


src/net/sourceforge/kolmafia/session/ChoiceAdventures.java
```diff
@@ -9079,16 +9079,39 @@ else if (KoLCharacter.isEd()) {
...
+          Map.entry("firewall", "hot"),
+          Map.entry("ICE barrier", "cold"),
+          Map.entry("corruption quarantine", "stench"),
+          Map.entry("parental controls", "sleaze"),
+          Map.entry("null container", "spooky"));
+
+  // This is incomplete
+  private static Map<String, String> encounterToElement =
+      Map.ofEntries(
+          Map.entry("Ice Cream Antisocial", "cold"),
```
midgleyc, line 9093: There are more on the wiki at https://kol.coldfront.net/thekolwiki/index.php/Cyber_-_Halfway

Veracity0: Sweet. I updated the source and tests to have everything they listed.

That list is still almost certainly incomplete: there are 2 hot, 4 cold, 2 stench, 2 sleaze, and 4 spooky.
But, we can always add more later.

midgleyc (approved):

PR #2669, opened 2025-01-13
feat: spacegate drop rates

midgleyc:
Cannon devleaked the spacegate drop rates: 

https://discord.com/channels/466605739838930955/1272891891674779749/1328447052719456277

The primitive alien items are also conditional (i.e. each planet's aliens can only drop one of the five items, at 1/1000 rate) but given we can only specify one drop type, "fixed" is more important.


gausie (approved):

src/data/monsters.txt
```diff
@@ -2051,17 +2051,17 @@ LOV Engineer	2010	lovengineer.gif	FREE NOCOPY Scale: 5 Cap: ? Floor: ? Init: -10
...
-exotic hostile animal	2018	sganimalc1.gif,sganimalc2.gif,sganimalc3.gif	NOCOPY Scale: [10+25*pref(_spacegatePlanetIndex)] Cap: 10000 Floor: ? Init: [60+10*pref(_spacegatePlanetIndex)] P: beast EA: stench EA: none Article: an	2-3 alien meat (m100)	2-3 alien toenails (m100)	alien animal goo (0)	alien animal milk (c0)
+large hostile animal	2017	sganimalb1.gif,sganimalb2.gif,sganimalb3.gif,sganimalb4.gif,sganimalb5.gif,sganimalb6.gif,sganimalb7.gif,sganimalb8.gif,sganimalb9.gif,sganimalb10.gif	NOCOPY Scale: [10+25*pref(_spacegatePlanetIndex)] Cap: 10000 Floor: ? Init: [60+10*pref(_spacegatePlanetIndex)] P: beast EA: stench EA: none Article: a	alien meat (100)	alien meat (20)	alien toenails (100)	alien toenails (20)	alien animal goo (20)
+exotic hostile animal	2018	sganimalc1.gif,sganimalc2.gif,sganimalc3.gif	NOCOPY Scale: [10+25*pref(_spacegatePlanetIndex)] Cap: 10000 Floor: ? Init: [60+10*pref(_spacegatePlanetIndex)] P: beast EA: stench EA: none Article: an	2-3 alien meat (m100)	2-3 alien toenails (m100)	alien animal goo (100)	alien animal milk (f1)
```
Veracity0, line 2059: This is not new in your changes, but what  does (m100) mean?
The "m" prefix is not documented at the top of the file.
You don't have to add a comment, but it'd be nice.

Veracity0 (dismissed):

src/data/monsters.txt
midgleyc: multi-drop: it's when multiple of the same item drop in the same line. Like flaming monstera's leaf drops.

These are distinct from dropping the same item multiple times in weird ways (e.g. if you pickpocket the item, you get exactly 1, and it doesn't drop after combat).

Veracity0 (approved):

PR #2670, opened 2025-01-13
Add CyberRealm support to canAdventure()

Veracity0:


midgleyc (approved):

gausie (approved):
I think you can technically do this in one test with a single @CartesianTest here but this is probably better!

PR #2674, opened 2025-01-14
Support for the hashing vise

Veracity0:
Thanks, Erosion, for the hashing vise.

I have a request class, which will be used for a command.
I have smashed an item into bits and have HTML for visiting the vise and for smashing a schematic.
I verified - and have a test - that we remove the schematic from inventory after it is smashed.

1) I have a HashingViseRequest class which looks complete, but I'd like some tests.
It accepts an AdventureResult of a schematic, complete with count, and will iterate calling the vise to smash one at a time.
At the moment, the actual call to the choice.php is commented out, since I wanted to debug the interaction between the command and the request class without smashing all my Stuff,

2) Speaking of the 'vise' command:
```
> vise

[count] <item> [, <another>]... - use your hashing vise to smash schematics into bits.

> vise fish

'fish' matches no schematics

> vise digit

'digit' matches 22 schematics

> vise rom

'rom' matches 3 schematics

> vise 2 beer, 2 burger

(hashable quantity of dedigitizer schematic: cybeer is limited to 1 by availability in inventory)
Hashing dedigitizer schematic: cybeer
Hashing complete.
Hashing dedigitizer schematic: cyburger
Hashing dedigitizer schematic: cyburger
Hashing complete.
```

- It does substring matching on the set of 22 schematic names
- You can specify a count
- You specify multiple schematics - with counts - separated by commas

I have tests! I think this is good to go.


Veracity0:
Isn't this special.
```
This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`.
Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

I "learned more" that this automatic failure will happen starting January 30, 2025.
Which is to say, this is a github bug.

Regardless, we should presumably upgrade before that date,,,

Veracity0:
Workaround provided by Chris, referenced in this bug report.
https://github.com/awalsh128/cache-apt-pkgs-action/issues/145

midgleyc (approved):

PR #2679, opened 2025-01-17
feat: add Jsoup, translate FightRequest to Jsoup

midgleyc:
Take 2, after #2667.

Slightly less weird cleaned HTML, easier debugging, similar speed. I hope to convert some of the regex matching to HTML parsing. We need to keep HtmlCleaner for the exposed `xpath` function; it also seems faster for the other xpath cases.

Thanks to the extensive tests for finding various issues: the HTML is cleaned slightly differently (but I think more correctly, for quirks mode).

Converted robotically with no attempt at using extra features:
```
new HtmlCleaner().clean -> Jsoup.parse
TagNode -> Element
TagNode.getName() -> Element.tagName()
TagNode.getAllChildren() -> Element.childNodes() // children() is just Elements
TagNode.getAttributeByName(X) -> Element.attr(X) // returns blank string instead of null on missing
TagNode.getAttributes() -> Element.attributes() // not a map, but can be iterated over
TagNode.getElementListByName("X", true) -> Element.children().select("X") // need children() to avoid matching current element
TagNode.getElementListByName("X", false) -> Element.select(">X")
TagNode.getElementsByName("X", true) -> Element.children().select("X") // need children() to avoid matching current element // need .size() instead of .length
TagNode.getElementsByName("X", false) -> Element.select(">X") // need .size() instead of .length
TagNode.getChildTags() -> Element.children()
TagNode.getText().toString() -> Element.wholeText() // text() collapses spaces
TagNode.getParent() -> Element.parent()
TagNode.getParent().removeChild(X) -> X.remove()
TagNode.findElementByName("X", true) -> Element.children().select("X").first()
TagNode.findElementByName("X", false) -> Element.selectFirst(">X")
TagNode.findElementByAttValue("X", "Y", true, false) -> Element.children().select("[X=Y]").first()
TagNode.findElementByAttValue("X", "Y", false, false) -> Element.selectFirst(">[X=Y]")
TagNode.evaluateXPath(X) -> Element.selectXpath(X, Node.class) // jsoup cannot select attributes (e.g. @href), html cleaner will accept invalid xpath (e.g. input[@type='checkbox']@checked)
BaseToken -> Node
BaseToken.getName() -> Node.nodeName() // in case you don't know whether you've got a comment or an actual node
ContentNode -> TextNode
ContentNode.getContent() -> TextNode.getWholeText() // text() collapses spaces
CommentNode -> Comment
CommentNode.getContent() -> Comment.getData()
```


jaadams5 (approved):

midgleyc:
Played a few days, looks okay, going for the merge, will await the bug reports from things I missed ;)

PR #2680, opened 2025-01-18
digitizer schematic: familiar-in-the-middle

Veracity0:


midgleyc (approved):
could have a test for coinmasters that confirms all the provided items are valid

Veracity0:
This is a strange one. Since the dedigitizer only shows you rows for things you schematics for, I didn't get the rows from visiting the shop. I saw the order - and the price - from the wiki, and used /buy to get the ROWs and hand-generated the row data. So this was my error.

I discovered it today when I got the schematic at the end of zone 1 and the CoinmastersFrame had it greyed out when I went to buy it.

CoinmastersDatabase uses ShopRow.fromData(data) to parse the entries.
That uses AdventureResult.parseItem(data[index], true) for each ingredient.
Why is that not reporting an error? The "true" is "pseudoAllowed".
I think it is getting an AdventureResult with an itemId of -1.

I think that we should log errors at coinmasters.txt parse time, so this should probably be detected and logged in ShopRow.parseData. That is an easy enough place to write a test for - and we'll see immediately if there are other issues in coinmasters.txt.

Veracity0:
Last comment. :)

With that logging, when I log in (which loads coinmasters.txt) I see this in the gCLI:
```
The Dedigitizer: unknown cost dedigitizer schematic: familiar-in-the-middle wrapper
```
So that is the only data error.

PR #2682, opened 2025-01-18
CyberRealm combat skills

Veracity0:
Support for CyberRealm combat skills in FightRequest: RAM cost, object usage, etc.

CyberRealm has eight new combat skills that are useable only in the three adventuring zones - and only against processes; not against hackers. They use "RAM"  and are available - or not - based on equipment you are wearing or items in inventory.

Sounds like a mess to calculate which are available, but FightRequest parses KoL's "use skill" dropdown. You can see this in the stationary buttons: when fighting a CyberRealm process, all of your normal combat skills are not available, and only the specific Cyber skills you can use will be present.

Therefore, I THINK no special effort will be needed to track whether a skill is available or not.

1) Verify this with combat_skill_available($skill) - YES
2) Does that depend on current RAM? - YES

Speaking of RAM, you start with 3 and can have up to 10. As you use skills in combat, it is consumed and can drop as low as 0. It will be refreshed at the start of the next combat in a Cyber zone.

1) A script may want to know the currently available RAM. Add my_ram() to get the currently available RAM
2) Does a script care what the value is outside of combat? Perhaps. A script should know, since It's derived from equipment and effects, but it's meaningless until you enter combat with a process. my_ram() outside of combat could return the currently configured amount.

Here are the skills.

```
7542	Throw Cyber Rock
RAM Cost: 0

7543	Brute Force Hammer
RAM Cost: 3
brute force hammer equipped

7544	Inject Malware
RAM Cost: 1
malware injector equipped

7545	Encrypted Shuriken
RAM Cost: 2
encrypted shuriken equipped

7546	Refresh HP
RAM Cost: 1
wired underwear equipped

7547	Launch Logic Grenade
RAM Cost: 0
logic grenade in inventory

7548	Deploy Glitched Malware
RAM Cost: 0
glitched malware in inventory

7554	Thrust your geofencing rapier
RAM Cost: 7
geofencing rapier equipped
```
I spent about a day's worth of VMF profit, and I now have all of the following:

cybervisor (+3 RAM)
datastick (+1 RAM)
logic grenade
glitched malware
wired underwear -> Refresh HP
encrypted shuriken -> Encrypted Shuriken
brute force hammer -> Brute Force Hammer
malware injector -> Inject Malware
geofencing rapier -> Thrust your geofencing rapier

1) I now have tests with actual HTML for all of the skills, except for Throw Cyber Rock.
That one is always available - no required equipment or items or RAM.

2) I also found some more  elemental encounters for Half-Way NCs.

3) I added my_ram() to ASH.

4) And lastly, CDM now puts in FREEFREEFREE for OVERCLOCK(10) free fights. He did NOT include text indicating that it was that specific source, so if there is something else - like a free runaway -which did it, we can't tell. I also don't know if he fixed the "The (free) Half-Way NCs consume a n OVERCLOCK(10) charge".

But, I adjusted the test in FightRequest and uploaded new HTML with the FREEFREEFREE.

This should be good to go.


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -851,6 +854,19 @@ public static final boolean canRockFlyer() {
...
   }
 
+  public static int getCurrentRAM() {
+    // RAM starts out at a particular value and decrements through the
+    // fight as cyber skills are used.
+    return (currentRound == 0) ? calculateInitialRAM() : currentRAM;
+  }
+
+  public static int calculateInitialRAM() {
+    return 3
```
midgleyc, line 864: `3 + KoLCharacter.currentNumericModifier(DoubleModifier.RAM)` should do it I think

Veracity0: It needs a cast, but otherwise, right. Done.

PR #2683, opened 2025-01-18
fix: haiku dungeon encounter

midgleyc:


jaadams5 (approved):

gausie:
What was this for out of interest?

midgleyc:
If you adventure in the Haiku Dungeon, you get in the logs:
```
[475] The Haiku Dungeon
Encounter: <font size=+2>77</font>
```
because we're not picking the right monster name because haiku dungeon HTML is not the same as normal.

PR #2685, opened 2025-01-19
fix: decorate nuns again after dec 2024 changes

midgleyc:


jaadams5 (approved):

jaadams5:
Philosophical question - now that KoL HTML has changed, how many tests might we have that would now fail with up to date HTML instead of the "archived" HTML.  Does it matter?  Is there an easy way to check?

midgleyc:
There is no way to check, so ???.

We have the URLs for some, but they also depend on player state. So there is no way to get the pages as they would be if the HTML changes were in at the time.

PR #2686, opened 2025-01-21
Prepare codebase for warning about future Java minimum supported version increments

libraryaddict:
Edit: https://kolmafia.us/threads/add-a-pending-change-in-required-java-version-nag-message.30245/

This PR is marked as a draft as it could probably stand to be reshuffled around, and the java version is hardcoded as 24.
So, perhaps merge it in when 24 is out in a few months?

image
image

And when run in headless, the warning appears after login stuff apparently.
image


gausie:
The chunk of code that checks version and produces a list of strings should be a function shared across these two uses.

jaadams5:
Ignoring my basic concerns about the feature, I think the nag message should be short and sweet.  It should tell what version is being used and what version KoLmafia will be expecting after a specific date.  "will soon" really imparts no useful information.  Similarly this is a nag and not a "how to" so the URL for one of the places to get a Java update is just clutter.  I'm fine with all the verbosity in the gCLI but not on the opening screen or on a command window that might not have a very large buffer to hold scrollable output.

If the required future version is embedded in the code then there are people who will never know about it (or changes to it) until they get the latest version of mafia.  That might be too late.  As an alternative I might consider placing the future required version and the date it will become the required version in a file.  The file could be downloaded (like the faxbot file or mall prices file or a couple of others) and the nag would just display what was in the file.

gausie:
> If the required future version is embedded in the code then there are people who will never know about it (or changes to it) until they get the latest version of mafia. That might be too late. As an alternative I might consider placing the future required version and the date it will become the required version in a file. The file could be downloaded (like the faxbot file or mall prices file or a couple of others) and the nag would just display what was in the file.

I disagree completely. If they don't update mafia, they don't need to worry about version requirement changes.

I think the contents of the message are mostly fine. If we had an internal date to make the change it would make sense to communicate it. But if we don't, that's fine.

jaadams5:
> I disagree completely. If they don't update mafia, they don't need to worry about version requirement changes.

This is why I regret that this change was not discussed before a PR was submitted or regret that I was not aware of/part of such discussion.  But I absolutely agree with you.  This has no relevance to someone who never upgrades mafia.

If this is deemed A Good Thing and implemented then I still believe a better implementation than the proposed one would be to acquire the data in the message from someplace other than code that is committed.  For example, if we decide we are going to go to Java 23 on the way to 24 then maintaining that information elsewhere would allow a user (who gets nag messages because of their version) to get the change without having to update mafia.  Perhaps Github is the source augmented by a creative use of the tags/badges/shields displayed?

libraryaddict:
> > I disagree completely. If they don't update mafia, they don't need to worry about version requirement changes.
> 
> This is why I regret that this change was not discussed before a PR was submitted or regret that I was not aware of/part of such discussion. But I absolutely agree with you. This has no relevance to someone who never upgrades mafia.

Sorry too much going on atm to look at other things. But for this

https://kolmafia.us/threads/proposal-require-java-21.30044/#post-175968

Probably should have made it's own post yeah

jaadams5:
Yeah, that was buried.  But I will expose it in its own thread so that I know that I am at least a vocal minority in thinking it is not helpful/needed.  Maybe I will learn something :-)

libraryaddict:
FYI, Further discussion can be found here

https://kolmafia.us/threads/add-a-pending-change-in-required-java-version-nag-message.30245/

gausie:
In an effort to merge or close this PR, I have implemented the proposed changes, as well as sorted out the styling to work with FlatLaf. I have used one of the two wordings provided, but I propose we merge this and can adjust the wording in future, since no one will see this code for a while.

Screenshot 2025-03-04 at 13 17 17

Separately, I would like to avoid upgrading our Java version if possible because https://cheerpj.com/ say they will support Java 21 in a year's time. Being able to run a full instance of mafia including GUI in your web browser would be very interesting.

gausie:
We can also close this. I don't hate the code now though as written. If you think that multitool will replace this in the next 6 months, feel free to close.

jaadams5:
OK.  I will start something at KoLmafia.us and close or agree to commit this depending upon reaction.  Thanks.

src/net/sourceforge/kolmafia/swingui/LoginFrame.java
```diff
@@ -1,14 +1,10 @@
...
-import javax.swing.JPasswordField;
-import javax.swing.SwingConstants;
+import java.awt.*;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import javax.swing.*;
 import net.java.dev.spellcast.utilities.JComponentUtilities;
 import net.java.dev.spellcast.utilities.SortedListModel;
 import net.sourceforge.kolmafia.KoLConstants;
```
jaadams5, line 10: I blame the IDE but our convention is to not allow * includes.

jaadams5 (changes requested):

jaadams5 (approved):

PR #2687, opened 2025-01-21
Set gradle's java toolchain, fixes some workspaces being unable to compile

libraryaddict:
Simple enough change, without this gradle will fail to compile the project for me.

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileLibJava'.
> error: invalid source release: 21
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -281,6 +281,10 @@ task gitUpdate() {
 java {
 	sourceCompatibility = JavaVersion.toVersion(javaSourceCompatibility)
 	targetCompatibility = JavaVersion.toVersion(javaTargetCompatibility)
+
+	toolchain {
+		languageVersion = JavaLanguageVersion.of(javaTargetCompatibility)
+	}
 }
 
 tasks.withType(JavaCompile) {


midgleyc (approved):
okay, sure. Sorry you're having this problem.

libraryaddict:
> okay, sure. Sorry you're having this problem.

Oh don't feel bad, I had this in my setup for ages but while I was making the other pr occurred to me that I should probably do this incase someone else gets the same problem but doesn't know how to fix.

libraryaddict:
> This means I can no longer build mafia in Java 23. Is there a way to indicate that the toolchain version must be "at least" 21, instead of exactly 21?

Wait I don't get it. Are you saying that you're using jdk 23 and now you can't compile it. Or that its compiling to java 21 when you want it to compile to 23?

gausie:
I have JDK 23 installed and when I run `./gradlew runShadow` I get 

```
FAILURE: Build failed with an exception.

* What went wrong:
BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 67
> Unsupported class file major version 67
```

gausie:
Adding

```groovy
plugins {
  id("org.gradle.toolchains.foojay-resolver") version "0.9.0"
}

toolchainManagement {
  jvm {
    javaRepositories {
      repository("foojay") {
        resolverClass = org.gradle.toolchains.foojay.FoojayToolchainResolver
      }
    }
  }
}
```

to the beginning of `settings.gradle` has sorted it. Should we merge that?

PR #2696, opened 2025-01-26
When fight the darkness, decorate with actual monster name

Veracity0:
KoL includes a MONSTERID comment for the real monster when you fight "the darkness".
We know this, because we decorate it with the element, drops, HP, etc in the Relay Browser.

This PR also decorates the monster name:

You are fighting the darkness

becomes

You are fighting The Bush hiding in the darkness


midgleyc (approved):
Nice idea! I think at some point the comment didn't appear when blind...

Veracity0:
Thanks! I noticed it when I was using the Amulet of Perpetual Darkness to boost my ML and noticed that KoLmafia decorated it the the monster drops

PR #2702, opened 2025-01-31
If multiple (available) NPC stores sell an item, show all in Mall Search results

Veracity0:
For example, from ```test row-duplicate-items```:
```
Degrassi Knoll Bakery and Hardware Store	ROW600	wad of dough	50 Meat
Bugbear Bakery	ROW678	wad of dough	50 Meat
Madeline's Baking Supply	ROW737	wad of dough	40 Meat
```
If you are in a Knoll sign, have the Bugbear Costume, and have met Madeline, all three shops are available.
Any of those conditions NOT true, the corresponding shop is not available.

This PR shows all available NPC purchase requests in the Mall Search frame when you search for "wad of dough".

Hmm. The previous behavior showed the last unavailable purchase request, if none were available.
Is that right? If you don't have Degrassi Knoll available, nothing you can do will open it. It seems like luck - or perhaps the order of shops in npcstores.txt - that determines what you will get.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
@@ -110,22 +110,22 @@ public class NPCStoreDatabase {
   }
 
   public static final List<NPCPurchaseRequest> getAvailablePurchaseRequests(final int itemId) {
+    List<NPCPurchaseRequest> result = new ArrayList<>();
+
     List<NPCPurchaseRequest> items = NPCStoreDatabase.NPC_ITEMS.get(itemId);
     if (items == null || items.size() == 0) {
-      return null;
+      return result;
     }
-
-    List<NPCPurchaseRequest> result = new ArrayList<>();
 
     for (var item : items) {
       boolean canPurchase = canPurchase(item.getStoreId(), item.getShopName(), itemId);
-      item.setCanPurchase(canPurchase);
       if (canPurchase) {
+        item.setCanPurchase(true);
         result.add(item);
       }
     }
 
-    return (result.size() > 0) ? result : null;
+    return result;
   }
 
   public static final PurchaseRequest getPurchaseRequest(final int itemId) {
--- a/src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
@@ -2,6 +2,7 @@ package net.sourceforge.kolmafia.persistence;
 
 import java.io.BufferedReader;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
```diff
@@ -108,6 +109,25 @@ public static final String getStoreName(final String storeId) {
...
+    List<NPCPurchaseRequest> items = NPCStoreDatabase.NPC_ITEMS.get(itemId);
+    if (items == null || items.size() == 0) {
+      return null;
+    }
+
+    List<NPCPurchaseRequest> result = new ArrayList<>();
+
+    for (var item : items) {
+      boolean canPurchase = canPurchase(item.getStoreId(), item.getShopName(), itemId);
+      item.setCanPurchase(canPurchase);
```
midgleyc, line 122: if we can't purchase, this modifies the original item. did you want to do that, or only if it gets added to the list?

Veracity0: The routine which used to be called - getPurchaseRequest(itemId) - would return the first available purchase request and, if none were available, would return the last found unavailable one - and set it unavailable.

As this is coded, it probably should modify the unavailable ones, although it shouldn't hurt.
But if I wanted (somewhat) similar behavior, should it return a list of the last-found unavailable one?

src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
```diff
@@ -108,6 +109,25 @@ public static final String getStoreName(final String storeId) {
...
+      if (canPurchase) {
+        result.add(item);
+      }
+    }
+
+    return (result.size() > 0) ? result : null;
```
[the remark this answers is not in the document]
Veracity0: That is what MallSearchRequest already does. OK, sure.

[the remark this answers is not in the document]
Veracity0: (Looking at MallSearchRequest, it does something very similar for CoinMasterPurchaseRequests. Perhaps CoinmastersDatabase.getAllPurchaseRequests should similarly return an empty list.)

src/net/sourceforge/kolmafia/persistence/NPCStoreDatabase.java
Veracity0: I adjusted it to only set canPurchase for items it will return.
I did not do anything about returning ... something ... if none of the requests are available.
Considering the method is called getAvailablePurchaseRequests, it is behaving as advertised. :)

midgleyc (approved):
Yes, I think this is better.

PR #2704, opened 2025-02-01
centralize request logging for shop.php

Veracity0:
Major refactoring that I have been building up to all month.

ShopRequest.registerRequest can handle any shop.php request.

A request with no "action=buyitem" is a simple visit.
We will log those, for shopIds that are type COIN (since that is how every single CoinMasterRequest class did its registerRequest), otherwise, no.

If there is "action=buyitem", this is a transaction.

For CONC shops:
```
Use A X, B Y, C Z to make N ITEMS
```

For all other shops - including type NONE (i.e. we just learned it by visiting a new shop):

If there is a single cost, and it is Meat:
```
buy N ITEMS for Y each from SHOPNAME
```
(The same way NPC purchases log right now.)
(For consistency, ```Buy N ITEMS for Y Meat each from SHOPNAME``` might be better.)

Otherwise, log as a coinmaster transaction:
```
Trade A X, B Y, C Z for N ITEMS
```

I removed the registerRequest methods from all shop.php coinmasters and concoctions.
I removed all the special casing in RequestLogger.
I removed all the special casing in NPCPurchaseRequest.

------

Next refactoring will be looking at parsing inventories when you visit a shop and logging new rows.
We already do that, but I want to make sure that what it logs is exactly what we want.
Also that it registers shops and shoprows correctly such that, having visited the shop, you can actually use it, and my ShopRequest stuff will Do The Right Thing when you buy things.
I'm sure I can simplify stuff with new Shop and ShopRow database stuff.

------

Next BIG refactoring will attempt to do something similar for ResponseTextParser which calls NPCPurchaseRequest.parseShopResponse which has a lengthy method calling different classes based on shopid

I think ShopRequest will be able to handle all the "remove used costs" and such, but there's other stuff in most classes` parseResponse methods.

We'll see what I can come up with.


jaadams5 (approved):
I looked at everything but focused on what was added.  No comments :-)

PR #2707, opened 2025-02-03
Add information on Just the Facts results to Location Details/Combat

HippoKingKoL:
If you have the skill Just the Facts (from the Book of Facts IOTM) you get an effect after most combats, dependent on monster/class/path.

Mafia knows this information, but it's not trivial for a player to access outside a scripting environment.

This PR adds that information to the fight.php page and the location details pane, if you have the skill and it is unrestricted. See screenshots attached.

location_details
fight


jaadams5:
In my casual glance I notice two methods called appendFact that appear to be pretty close to identical.  Perhaps a refactor so that there was one method (perhaps a static in FactsDatebase?) and it was referenced in two places.  You could also tell me that they are not similar and I should have my eyesight checked,

As always tests would be appropriate.

gausie:
Yeah I would move this into one function in MonsterData, probably, that returns `null` if nothing should be displayed

HippoKingKoL:
The reason I ended up with duplication is that one had to operate on a StringBuffer and the other on a StringBuilder, which seem very similar but not identical (thread safety) and not convertible.

So I need one function that does both. Would this return a regular string?

eg in `MonsterData`:
`public String getFact(...)
{
 StringBuffer buffer;
 buffer.append("mystuff");
 return buffer.toString
}`

and then have each of the current append functions just append the string output of this to their respective buffer type.

HippoKingKoL:
I've removed the duplicate code, and added 3 tests I think work. Please check they're sane - thanks.

src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -1858,6 +1860,31 @@ private String makeItemCount(MonsterDrop drop) {
     return drop.itemCount().isEmpty() ? "" : drop.itemCount() + " ";
   }
 
+  public String getFact() {
+    if (!(KoLCharacter.hasSkill(SkillPool.JUST_THE_FACTS)
+        && StandardRequest.isAllowed(RestrictedItemType.SKILLS, "Just the Facts"))) {
+      return null;
```
midgleyc, line 1866: returning "" will save a null check and simplify the method down to `buffer.append(monster.getFact())`

src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -1858,6 +1860,31 @@ private String makeItemCount(MonsterDrop drop) {
...
+    }
+    StringBuffer buffer = new StringBuffer();
```
midgleyc, line 1868: there are such a low number of concatentations in this method that not using a buffer at all is likely better. Just return `"<br />Just the Facts:" + f`.

src/net/sourceforge/kolmafia/MonsterData.java
```diff
@@ -1858,6 +1860,31 @@ private String makeItemCount(MonsterDrop drop) {
...
+        FactDatabase.getFact(KoLCharacter.getAscensionClass(), KoLCharacter.getPath(), this, false);
+    buffer.append("Just the Facts: " + f.toString());
+    return buffer.toString();
+  }
+
+  void appendFact(StringBuilder buffer) {
+    this.appendFact(buffer, false);
+  }
+
+  public void appendFact(StringBuilder buffer, boolean stateful) {
```
midgleyc, line 1880: stateful isn't used, so you can eliminate the parameter.

src/net/sourceforge/kolmafia/MonsterData.java
HippoKingKoL: Done

src/net/sourceforge/kolmafia/MonsterData.java
HippoKingKoL: I added the null because Gausie specifically mentioned it [here](https://github.com/kolmafia/kolmafia/pull/2707#issuecomment-2631488596)

I'll leave it for now until hopefully you two agree?

test/net/sourceforge/kolmafia/MonsterDataTest.java
```diff
@@ -425,6 +429,56 @@ void itemDropsAreRenderedProperly(final String monsterName, final String itemDro
...
+    }
+
+    @Test
+    void factsAreRenderedIfUnrestricted() {
+      var monster = MonsterDatabase.findMonster("fluffy bunny");
+
+      var cleanups =
+          new Cleanups(
+              withSkill(SkillPool.JUST_THE_FACTS),
+              withNotAllowedInStandard(RestrictedItemType.SKILLS, "just the facts"),
```
jaadams5, line 454: Since the canonical name of the skill is "Just the Facts" perhaps the same capitalization should be used in the calls to withNotAllowedInStandard?  Not a blocker.

midgleyc (approved):

src/net/sourceforge/kolmafia/MonsterData.java
midgleyc: I don't agree but don't care enough to block on it.

test/net/sourceforge/kolmafia/MonsterDataTest.java
midgleyc: the key is lowercased when passed to the standard restricting function so I think this is fine

jaadams5: It's more of a cognition thing. If the canonical name is Just the Facts then why not use it rather than assume the reader knows it will be lowercased before use?  And since I am lazy I will copy the string, thinking it is canonical and use it in a context where it must be canonical and so fuzzy matching or a case sensitive compare will mean my code is wrong.  But not a blocker so...

PR #2708, opened 2025-02-04
Remove unnecessary constructors for CoinMasterRequest classes

Veracity0:
1) Gaining (or using) a worthless item did not fire the (coinmaster) signal.
That's because the HermitRequest did not initialize its buyItems until The Hermit was visited for the first time.
2) AltarOfBonesClass had a bug: it provided a parseResponse method - which is called by ResponseTextParser for an "external" request - but it did not have a processResults which would call it for an "internal" request.
3) After I freed the king in my first 30 turns today, I spend the turns to train the last two familiars.
```
Results for Burly Bodyguard after 12 trials using 144 turns:

Contest         XP[1]   XP[2]   XP[3]  Original Rank Derived Rank
Cage Match      24      39      55           0            
3Scavenger Hunt 25      40      56           0            3
Obstacle Course 25      35      54           0            3
Hide and Seek   25      42      55           0            3

Results for Emberiza Aureola after 12 trials using 109 turns:

Contest         XP[1]   XP[2]   XP[3]  Original Rank Derived Rank
Cage Match       0       0       0           0            0
Scavenger Hunt  61      36       5           0            1
Obstacle Course 59      62      20           0            2
Hide and Seek   26      46      59           0            3
```
4) CoinMasterRequest constuctor simplification:

Every Coinmaster must provide two constructors:
```
  public CoinMasterRequest(final CoinmasterData data)
```
 for a simple visit to the "buy" URL.

And either this:
```
  public CoinMasterRequest(final CoinmasterData data, final boolean buying, final AdventureResult[] attachments)
```
for coinmasters with "buy" and "sell" items, or:

```
  public CoinMasterRequest(final CoinmasterData data, final ShopRow row, final int quantity)
```
for a new-style "ROW" coinmaster.

Those three constructors can be created via reflection for use by user requests, via the "coinmaster" command or the CoinmastersFrame.

For the handful of coinmasters with extra actions (Mr. Store, Bounty Hunter Hunter):
```
  public CoinMasterRequest(final CoinmasterData data, final String action)
```

If your CoinMasterRequest class wants an additional constructor, the class can implement it in terms of one or the other transaction constructors.

I got rid of the never-used constructors in CoinMasterRequest and all its subclasses.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/coinmaster/AltarOfBonesRequest.java
+++ b/src/net/sourceforge/kolmafia/request/coinmaster/AltarOfBonesRequest.java
@@ -34,12 +34,9 @@ public class AltarOfBonesRequest extends CoinMasterRequest {
     super(ALTAR_OF_BONES, buying, attachments);
   }
 
-  public AltarOfBonesRequest(final boolean buying, final AdventureResult attachment) {
-    super(ALTAR_OF_BONES, buying, attachment);
-  }
-
-  public AltarOfBonesRequest(final boolean buying, final int itemId, final int quantity) {
-    super(ALTAR_OF_BONES, buying, itemId, quantity);
+  @Override
+  public void processResults() {
+    parseResponse(this.getURLString(), this.responseText);
   }
 
   public static void parseResponse(final String urlString, final String responseText) {
--- a/src/data/familiars.txt
+++ b/src/data/familiars.txt
@@ -332,9 +332,9 @@
 300	Mini Kiwi	kiwi_kiwi.gif	stat0,meat0,drop	mini kiwi egg	aviator goggles	3	3	3	1
 301	Proto-Protozoa	protoproto.gif	combat1	proto-proto-protozoa	extra ooze	0	0	0	3
 302	Evolving Organism	bacteria.gif	combat1,hp1,mp1,block,variable	unevolved organism	extra DNA	3	1	2	0
-303	Burly Bodyguard	bodyguard.gif	combat0,block	baby bodyguard	bodyguard badge	0	0	0	0
+303	Burly Bodyguard	bodyguard.gif	combat0,block	baby bodyguard	bodyguard badge	3	3	3	3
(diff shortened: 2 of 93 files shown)


src/net/sourceforge/kolmafia/request/coinmaster/AltarOfBonesRequest.java
```diff
@@ -34,12 +34,9 @@ public AltarOfBonesRequest(final boolean buying, final AdventureResult[] attachm
...
   }
 
-  public AltarOfBonesRequest(final boolean buying, final AdventureResult attachment) {
-    super(ALTAR_OF_BONES, buying, attachment);
-  }
-
-  public AltarOfBonesRequest(final boolean buying, final int itemId, final int quantity) {
-    super(ALTAR_OF_BONES, buying, itemId, quantity);
+  @Override
+  public void processResults() {
```
midgleyc, line 38: was this a deliberate change as part of this PR? It wasn't called out.

Veracity0: It was deliberate: it was a bug in the obsolete class.

ResponseTextParser has
```
      case "bone_altar.php" -> {
        AltarOfBonesRequest.parseResponse(location, responseText);
      }
```
but that is invoked only if this was an "external" request.
If you instantiate and run the class, without a processResults, it would never parse the response, since that is not an external request.

Veracity0:
Give me a sec... I am going to restore the constructors to CoinMasterRequest - since they are useful for a few specific subclasses - like TrapperRequest and HermitRequest.

The large bulk of subclasses do not need them and will not override them.

src/net/sourceforge/kolmafia/request/coinmaster/AltarOfBonesRequest.java
Veracity0: I added it to the PR description. Sorry!

jaadams5 (approved):

PR #2709, opened 2025-02-04
Track Cupid Bow actions and activities

Ignose:


jaadams5:
Some discussion at https://kolmafia.us/threads/toy-cupid-bow-preferences.30274/#post-176168

Progress will be halted until conflicts are resolved.  If you need help or want someone else to do that, please ask.

It looks like the code is not spotless so spotless should be run after resolving conflicts.

I think there was some more HTML posted at KoLmafia.us  It would be nice to use it in a test also.

I will circle back and check the logic since there might be a way to do it with two preferences.

Thanks.

Ignose:
Pulling this one and opening a new one, I had to re-clone the project to get recent updates for some reason.

PR #2710, opened 2025-02-05
Toy Cupid Bow - Clean

Ignose:


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7091,6 +7095,32 @@ private static void processP(Element node, TagStatus status) {
       FightRequest.logText(str, status);
     }
 
+    if (KoLCharacter.hasEquipped(ItemPool.TOY_CUPID_BOW)) {
```
horrible-little-slime, line 7098: This is identical to checking `status.cupidbow`, which is what you check down below. Why are these different?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7091,6 +7095,32 @@ private static void processP(Element node, TagStatus status) {
...
+      if (Preferences.getInteger("cupidBowLastFamiliar")
+          == KoLCharacter.getEffectiveFamiliar().getId()) {
+        Preferences.increment("cupidBowFights");
```
horrible-little-slime, line 7101: is it worth capping this at 5?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7091,6 +7095,32 @@ private static void processP(Element node, TagStatus status) {
       FightRequest.logText(str, status);
     }
 
+    if (KoLCharacter.hasEquipped(ItemPool.TOY_CUPID_BOW)) {
+      if (Preferences.getInteger("cupidBowLastFamiliar")
+          == KoLCharacter.getEffectiveFamiliar().getId()) {
```
horrible-little-slime, line 7100: Technically I don't think we want effective familiar here--that accounts for stuff like doppelshifter and chameleon. I guess chameleon can never equip a TCB, but doppelshifter can.

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -3705,6 +3705,23 @@ public void canDetectTimePrankMessage() {
...
+    void canDetectCupidBow() {
+      var cleanups =
+          new Cleanups(
+              withFamiliar(FamiliarPool.MINI_KIWI),
+              withEquipped(ItemPool.TOY_CUPID_BOW),
+              withProperty("_cupidBowFamiliars", ""),
+              withFight());
+      try (cleanups) {
+        parseCombatData("request/test_cupid_bow.html");
+        assertThat("_cupidBowFamiliars", !isSetTo(""));
```
horrible-little-slime, line 3720: I don't think this is a valid assertion, but I don't know a lot about assertions. Either way it feels more prudent to check that it _is_ set to `"300"` (the mini kiwi's famid) rather than making sure that it isn't an empty string.

Ignose: It isn't, I updated to 380. Tests are hard.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -7091,6 +7095,31 @@ private static void processP(Element node, TagStatus status) {
       FightRequest.logText(str, status);
     }
 
+    // looks askance at the toy bow you've provided and shoots the plunger-arrow over the horizon,
+    // then proceeds to drag something more appropriate back.
+    if (status.cupidbow) {
+      if (Preferences.getInteger("cupidBowLastFamiliar")
+              == KoLCharacter.getEffectiveFamiliar().getId()
+          && Preferences.getInteger("cupidBowFights") < 5) {
```
horrible-little-slime, line 7103: don't cap it like this, use the 3 arg version of `Preferences.increment`

jaadams5:
Test failure is yours and not one of the intermittent failures that have nothing to do with new code.

At kolmafia.us you said "Preference tracking for how many combat turns in a row".  I don't think this code does that.  If my familiars for each combat are A, A, B, A and I do not equip the bow on B but do on all turns with A I think the preference will be 3 but it does not mean three turns in a row.  Since I have not looked at the mechanic I don't know whether you misspoke, I misunderstood or the code is not correct.

Ignose:
> At kolmafia.us you said "Preference tracking for how many combat turns in a row". I don't think this code does that. If my familiars for each combat are A, A, B, A and I do not equip the bow on B but do on all turns with A I think the preference will be 3 but it does not mean three turns in a row. Since I have not looked at the mechanic I don't know whether you misspoke, I misunderstood or the code is not correct.

The update of the pref is predicated on the boy being equipped. If it's not equipped, the counter doesn't change. This is an explanatory failure on my part.

You can run 4 turns with the bow equipped, 10 turns without, and 1 more turn with and still get your drop, as long as the [Bow, Familiar] combination doesn't change. If that makes sense.

> Test failure is yours and not one of the intermittent failures that have nothing to do with new code.

Understood but I have no idea what I'm doing wrong. :)

test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -3705,6 +3705,26 @@ public void canDetectTimePrankMessage() {
...
+          new Cleanups(
+              withFamiliar(FamiliarPool.MINI_KIWI),
+              withEquipped(Slot.FAMILIAR, ItemPool.TOY_CUPID_BOW),
+              withProperty("_cupidBowFamiliars", "111"),
+              withFight());
+      try (cleanups) {
+        parseCombatData("request/test_cupid_bow.html");
+        var text = RequestLoggerOutput.stopStream();
+        assertThat(text, containsString("looks askance at the toy bow"));
+        assertThat("_cupidBowFamiliars", isSetTo("111; 380"));
```
jaadams5, line 3723: I would consider using Preferences.getString and print what it returns for _cupidBowFamiliars.  That would tell you whether the problem is the code or the test.  Maybe temporarily
`assertEquals(Preferences.getString("_cupidBowFamiliars), "111; 380");` or whatever is syntactically correct?

I am suspicious about the test because I don't have how isSetTo works is memorized so

Ignose: I tried that; specifically:

`        System.out.println(Preferences.getString("_cupidBowFamiliars"));
        System.out.println(Preferences.getString("cupidBowLastFamiliar"));
        System.out.println(Preferences.getString("cupidBowFights"));`

And received
`111

0`

So while it could be the test, I think I'm missing something in the code.

Ignose: `System.out.println(KoLCharacter.hasEquipped(ItemPool.TOY_CUPID_BOW));`

Is returning false. I am unsure how to fix this.

jaadams5: Agreed.  It doesn't look as if the code is setting any of the preferences.  I'll let you try and resolve because I can't really walk through it with the debugger right now.

src/net/sourceforge/kolmafia/objectpool/ItemPool.java
```diff
@@ -3892,6 +3892,7 @@ public class ItemPool {
   public static final int STATS_ROM_CHIP = 11823;
   public static final int HASHING_VISE = 11826;
   public static final int GEOFENCING_RAPIER = 11827;
+  public static final int TOY_CUPID_BOW = 11836;
```
horrible-little-slime, line 3895: > js Item.get("11836")

Returned: new-in-box toy Cupid bow
id => 11836

horrible-little-slime: Once this is fixed it's worth adding additional tests:
- one for successfully incrementing the pref when you don't get a drop
- one for resetting the pref when you swap familiars with the TCB equipped
- one for _not_ resetting the pref when you swap familiars and _don't_ have the TCB equipped
- one for setting the `_cupidBowFamiliars` pref when it's empty, rather than when it currently has a thing in it
(these should all be pretty straightforward; I'd recycle html for an existing test for ones where you don't want to get the drop. That being said I'm not a mafia dev so maybe this isn't necessary. but those are the cases I'd test, as just some guy)

jaadams5 (dismissed):
Opportunities for more testing but I will approve now and push later if no one beats me to it.

Ignose:
> Opportunities for more testing but I will approve now and push later if no one beats me to it.

I want to write a few more tests, if you're okay with that.

Ignose:
Additional tests are failing. Looks like prefs are being double incremented.

Ignose:
One remaining test is failing; counter doesn't increment on free run.

horrible-little-slime:
I'm out of ideas for where the TCB processing code can go to trigger on all combat resolutions, including wins, losses, and runs.

jaadams5 (approved):

jaadams5:
If your familiars are A,A,B,A each with the bow, I think the code would end with a count of 1 for A,  But if you are trying to track when equipping on A will no longer be useful on A.  Is this wrong?  Again, I have not studied the mechanic so I may be missing something.

Ignose:
> If your familiars are A,A,B,A each with the bow, I think the code would end with a count of 1 for A, But if you are trying to track when equipping on A will no longer be useful on A. Is this wrong? Again, I have not studied the mechanic so I may be missing something.

We wanna track both when a familiars daily drop is used up, and how many turns we've spent "charging" a familiar. The "charge" resets on even a single combat with another familiar, so A,A,B,A puts you at A,1

Veracity0:
I sort of wish you'd tried my "familiar fetches a heart necklace" html, but since you are not actually checking for WHICH item was retrieved, eh. It is the same (start of the) message you are looking for.

Ignose:
> I sort of wish you'd tried my "familiar fetches a heart necklace" html, but since you are not actually checking for WHICH item was retrieved, eh. It is the same (start of the) message you are looking for.

I did end up using your HTML instead of mine, if it's any consolation.

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -5650,6 +5670,7 @@ public static class TagStatus {
     public boolean serendipity;
     public boolean mildManneredProfessor;
     public boolean batwings;
+    public boolean cupidbow;
```
midgleyc, line 5673: is this used any more?

src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4336,6 +4336,26 @@ else if (responseText.contains("a horrible grinning clown head emerges")) {
...
+      } else {
+        Preferences.setInteger("cupidBowLastFamiliar", currentFamiliarId);
+        Preferences.setInteger("cupidBowFights", 1);
+      }
+      if (responseText.contains("looks askance at the toy bow")) {
+        String currentFams = Preferences.getString("_cupidBowFamiliars");
+        Preferences.setString(
+            "_cupidBowFamiliars",
+            (currentFams.isEmpty() ? "" : (currentFams + ";")) + currentFamiliarId);
+        Preferences.setInteger("cupidBowFights", 0);
```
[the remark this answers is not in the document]
Ignose: Yep! Ange1ade tested it. You can pre-charge tomorrow's equipment.

src/net/sourceforge/kolmafia/request/FightRequest.java
Ignose: I suppose not; I'll remove it.

src/net/sourceforge/kolmafia/request/FightRequest.java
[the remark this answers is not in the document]
Ignose: Wait are you asking about the reset to 0? My understanding is, you have to do another 5 fights to get the next equip, regardless of rollover.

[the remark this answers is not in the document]
midgleyc: right, but in this case you shouldn't be setting `cupidBowFights` to 0 after you get a drop, because if you wait for tomorrow without moving the bow, you can get a drop again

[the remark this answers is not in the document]
Ignose: https://discord.com/channels/466605739838930955/474736566246572042/1335850741348438096

Correcting the logic.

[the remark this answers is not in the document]
jaadams5: > https://discord.com/channels/466605739838930955/474736566246572042/1335850741348438096
> 
> Correcting the logic.

I can't see the link because Discord wants me to log in, something I can't do :-)  Any chance you could copy/past the relevant bits for those who are Discord impaired?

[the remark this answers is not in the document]
Ignose: ange1ade — 2/3/2025 12:53 AM
this is minor, but I'm in day 3 of a HC run, first turn of the day and it drops a normal fam equip for my chest mimic
Image
so maybe it drops after 5 turns on the fam, but you can get a new fam equip for each fam per day? (and the counter accumulates?)

plungerfire40  (#1955580) — 2/3/2025 1:06 AM
yeah. It drops if you haven't gotten it today and turns-with-current-fam are >=5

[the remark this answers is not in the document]
jaadams5: Thank you.

PR #2711, opened 2025-02-05
fix: fix end of fight tracking with both old and new combat bars enabled

midgleyc:
Ref: https://kolmafia.us/threads/enabling-show-old-form-with-combat-action-bar-breaks-end-of-fight-tracking.29484/

Don't incorrectly end the fight on round 0.

Do correctly end the fight when running away.

For existing users, you have to view the account -> combat page once to set the preferences correctly.


Veracity0 (approved):
Thank you!

PR #2717, opened 2025-02-06
Apriling band quad tom used without worm-riding hooks summons a giant sand worm

Veracity0:
1) I refactored a couple of switch statements to the modern form.
2) Apriling band quad toms with worm riding hooks in inventory advance desert exploration.
3) Apriling band quad tom with no worm riding hooks fights a (zoneless) sandworm as a free fight.


src/net/sourceforge/kolmafia/request/GenericRequest.java
```diff
@@ -2484,17 +2484,22 @@ private static void checkItemRedirection(final String location) {
...
-        break;
-
-      case ItemPool.DOLPHIN_WHISTLE:
+      }
+      case ItemPool.APRIL_BAND_TOM -> {
+        if (InventoryManager.hasItem(ItemPool.WORM_RIDING_HOOKS)) {
+          return;
+        }
+        itemName = "Apriling Band Quad Tom";
+        consumed = true;
```
midgleyc, line 2500: does this still get hit given the url is `inventory.php?iid=11567&action=aprilplay` instead of straight use?

also I don't think the item itself is ever consumed

Veracity0: Obviously I need to try it out myself and write a test.

I had no idea this was a thing until it was mentioned on the CyberRealm thread.

Veracity0: I tested both with and without worm riding hooks in inventory and wrote tests for both.

jaadams5 (approved):

PR #2718, opened 2025-02-07
mall search results with CRLF in shop names no longer break parsing

Veracity0:
Malibu Stacey noticed that if you do a mall search for "sprocket", it takes many seconds and eventually comes back with no mall stores at all. Only NPC shops.

I got the HTML for this search - which I include in the PR - and discovered that the result contained the following:
```
<b>
The Darksteel Forge
</b>
```
Yes. A store name with a literal CRLF before and after the name.

Our Search pattern trying to extract the HTML ```<table class="ittemtable"...STUFF...</table>``` did not have a Pattern.DOTALL. And apparently the Matcher took a really really long time to discover it had no matches.

I stuck in a DOTALL and now the mall search of a sprocket parses as fast as ever.

Astonishing that we went so many years before being presented with a store name like this.


Veracity0:
I'm going to add some tests...

jaadams5 (approved):
I wonder if there is any other player created data that might have a similar problem?

Malibu-Stacey:
My entire contribution was just debugging why someone else was failing to create a clockwork counterclockwise dome when doing `retrieve_item($item[clockwork maid])` because the extent of their debugging was to paste their log & do literally nothing else other than muse about switching to using buy instead of retrive_price.

It's amazing how far the tiniest bit of reading the error message and trying things yourself will take you when debugging.

PR #2723, opened 2025-02-09
Need to add more fields to CoinMasterShopRequest that uses ShopRows

Veracity0:
Which would be just the Kiwi Kwiki Mart, among coinmasters that are currently accessible.


jaadams5 (dismissed):
>who needs a code review for a fix this simple? :)

:-) indeed.

jaadams5:
Of course I should have waited to see the tests pass...

Veracity0:
> Of course I should have waited to see the tests pass...
Ha ha.

I added ajax=1 - which works fine and makes the responseText much smaller - and the test was assuming no ajax in the urlString.

I fixed the test.
I should change the response text it parses...

jaadams5 (approved):
Second time's the charm.

PR #2725, opened 2025-02-11
Add test fixtures for the Floral Mercantile Exchange and Gene Fiddling shops

Veracity0:
The Central Loathing Floral Mercantile Exchange is a shop.php - shopid=flowerexchange - that we do not support.
Presumably because KoL changes the price for tulips every thirty minutes.

We don't parse the rows successfully - just as we don't parse the Crackpot Mystic's shop successfully.

We have a test fixture for the latter - test_shop_mystic.html

This PR submits a test fixture for the Floral Mercantile Exchange - test_shop_flowertradein.html

Ooooh! gausie provided test fixtures for shopId = mutate, which trades rads for skills!

I added a vestigial coinmaster for that shop: GeneFiddlingRequest.
I used the "visit" html to generate the shop row data.
I suppose I could use the "bought skill" html to write a test.

Not yet. There is more CoinmasterData and ASH RuntimeLibrary stuff needed to support skills.
For now this has all the ShopRow and CoinmastersDatabase support.


Veracity0:
I am hoping somebody will review this before gausie goes in to modify the same files.
It'd be nice if he started with ShopRequest.parseInventory understanding skills as possible AdventureResults.

Veracity0:
FWIW, my policy: If it's been 24 hours without a review, it goes in without a review.

gausie:
It could always be me who reviews it ;)

gausie (approved):
Oh wow you actually added the Skill priority to AdventureResult - nice!

Mental note that we have to be "careful" of name clashes between items and skills in the shop row database, since skills will take priority. This is probably fine, but knowing KoL it could one day not be. A problem for then!

PR #2726, opened 2025-02-11
Remove accidentally included test code from boxing daycare

spaghetti-squash:
Sorry for leaving this in, I added it when I was writing tests and it seems that I and all reviewers missed it.


spaghetti-squash:
Not 100% sure why/how this wasn't covered by my tests, but also it is basically always immediately eaten by being properly updated

gausie (approved):

PR #2728, opened 2025-02-12
properly parse trainset when only one lap remains

spaghetti-squash:


jaadams5 (dismissed):

src/net/sourceforge/kolmafia/session/TrainsetManager.java
```diff
@@ -85,7 +85,8 @@ public static TrainsetPiece getByName(String name) {
   private static final Pattern CURRENT_STATION =
       Pattern.compile("<br>Your train is about to pass station (\\d)\\.<");
   private static final Pattern LAPS_BEFORE_RECONFIGURE =
-      Pattern.compile("Let the train finish (\\d) more laps before rearranging it.</p>");
+      Pattern.compile(
```
gausie, line 88: I would have considered doing

```regex
Let the train finish (\d|this)(?: more)? laps? before rearranging it.</p>
```

To make the group matching a little easier to process mentally

gausie: Or perhaps better
```regex
Let the train finish (\d|this) (?:more laps|lap) before rearranging it.</p>
```

PR #2729, opened 2025-02-13
JSON API should return error on bad continuation state.

phulin:
When a user somehow puts mafia into an error state, the JSON API function execution stops working due to the fact that `LibraryFunction.execute` pays attention to the continuation state.

I'm not sure if this is the right way to address the problem - open to feedback.


gausie:
I think this is the right way to achieve this. I worry about letting the relay browser continue to execute arbitrary scripting functions while mafia is in an error state, even though I appreciate in your specific case it is likely to be limited to "readonly" functions

midgleyc:
I think this is also okayish leaning bad. What was the use-case that triggered this?

From what I understand, the user can put Mafia into "abort" state by, say, hitting escape. This is a trigger that whatever script is running should stop; you can use it to stop relay scripts. With the relay API we introduced a new type of relay script that does its functions as and when, and this PR is to also refuse to stop when requested.

I also don't think the script should be able to send an API request to reset the error state. Technically it can also reset the error state with the ability to run arbitrary functions -- e.g. I think `cliExecute("relay")` should do it.

jaadams5:
I am one of those people who makes a script related mistake - such as starting the wrong one - and panics, mashing escape until everything stops and I clean up the mess I made.

Anything that ignores my button mashing is A BAD IDEA.

I drove cars with "idiot lights".  If the light went on the operator was supposed to pull over, check the owner's manual for possible causes and take appropriate action.  Some times the action was "have the car checked by a mechanic before driving it again" and some times it was just "change your oil soon".  So I am conditioned to believe that the designer of a system triggered some kind of "warning" so that I could take action, or not.  Some of those warnings definitely saved me time and money.

I run many scripts that I did not write and some that I do not understand.  My confidence in the safety of those scripts is undermined if I know sometimes they will ignore things KoLmafia thinks are an error and worthy of bringing to my attention.  I note that some of those scripts trap errors and take action to correct them and that is OK.  It is the error that is just ignored that causes my concerns.

My sense is folks are somewhat lukewarm about this.  I'd like to see a little more detail making the case that this is a Good Idea.  I'd be interested in knowing how this would handle panicked escape mashing and whether it makes scripts I didn't write more trustworthy.

midgleyc (approved):
I think this is reasonable.

jaadams5:
I approve of the changes.  But I don't understand the tests.  If there is an error state then why should there be any JSON?  One test expects JSON to be null but the other expects it to contain the response code.

jaadams5 (approved):
Approved but I don't quite understand the tests.

jaadams5:
> > Approved but I don't quite understand the tests.
> 
> It returns an error object saying the reason the service is unavailable. Is that an issue?

No.  One test checks for a response that is null and AFAIK the other checks for a response that is not null and contains a message about the error state.  Since handleJSONApi returns void there may just be a lot going on with shared values that I am missing.  But this is OK whether I understand or not.

PR #2730, opened 2025-02-13
Make place names clickable in relay browser after a /go command"

gausie:
- **Decorate /go output in chat pane (so that I can click to go again)**
- **Add test**

Allows the placename to be clicked here to repeat the action of navigation. This is really useful for me to be able to repeat an action over and over, maybe others will find it useful.

Screenshot 2025-02-13 at 15 51 30

EDIT: As the tests here touch aspects of chat for the first time it inadvertently brings our coverage over 40%!


Veracity0 (approved):
Maybe I will use a /go command someday. :)

PR #2731, opened 2025-02-13
Some familiars were appropriately capitalised yesterday

gausie:


Veracity0 (approved):
When I did the familiar training for the quantum entangler, I noticed and wondered about this.

PR #2733, opened 2025-02-13
More support for coinmasters that sell skills

Veracity0:
Using the Genetic Fiddling coinmaster (shopid=mutate), pretty much finished supprt for coinmasters that sell skills.

1) added boolean availableSkill(int skillId) to CoinmasterData. 
Assumption is only new ShopRow coinmasters will support it.
There must be a ShopRow with the result item being a skill with that skillId.
And if you already have the skill, it is no longer "available".
2) If a coinmaster has addition requirements, it can use "withAvailableSkill" to override it.
GeneticFiddlingRequst doesn't need to to that. I assume.
3) added AdventureResult skillBuyPrice(int skillId) to CoinmasterData
Assumption is that only new ShopRow coinmasters will support it.
Assumption is there is only one "cost".
This returns that cost.
4) CoinMasterRequest only calls purchasedItem for a purchase.
It only calls RequestProcessor.learnSkill in processResults; an external request (like a GenericRequest) learns it elsewhere.
5) No need for ResponseTextParser to have special code to remove rads when you learn a Nuclear Autumn skill.
6) I added a Genetic Fiddling panel to the CoinmastersFrame. It LOOKS good, but is otherwise untested.
7) Added ASH functions:
```
boolean sells_skill($coinmaster[Genetic Fiddling], $skill[Boiling Tear Ducts]) returns true
int sell_price($coinmaster[Genetic Fiddling], $skill[Boiling Tear Ducts]) returns 30.
```
8) I have tests for visiting and buying from "whichshop=mutate" using either a CoinMasterRequest or a GenericRequest.
9) I have tests for the new ASH functions.
10) I have tests for CoinmasterData.availableSkill

I suppose that ASH would like a
```
boolean buy(coinmaster, skill)
```

Later. Maybe.

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -511,6 +511,13 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
     return -1;
   }
 
+  public int getSkillId() {
+    if (this.priority == Priority.SKILL) {
+      return this.id;
+    }
+    return -1;
+  }
+
   /**
    * Accessor method to retrieve the total value associated with the result. In the event of substat
    * points, this returns the total subpoints within the <code>AdventureResult</code>; in the event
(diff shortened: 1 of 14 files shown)


gausie (approved):
This is awesome. I hope TPTB remember skill shops exist and this is used again some time.

PR #2736, opened 2025-02-14
bjorn enchantments for new familiars

spaghetti-squash:
I just bug reported (to kol) a bunch of unbjornable familiars; will add empty lines for them in about a week if they haven't been added

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -4856,6 +4856,7 @@ Throne	Bulky Buddy Box	Hot Resistance: +2, Cold Resistance: +2, Stench Resistanc
 Throne	Casagnova Gnome	Meat Drop: +20, HP Regen Min: 16, HP Regen Max: 20, MP Regen Min: 9, MP Regen Max: 11
 Throne	Cat Burglar	Item Drop: +10
 Throne	Chauvinist Pig	Experience (Muscle): +2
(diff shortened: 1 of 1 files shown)


src/data/modifiers.txt
```diff
@@ -4856,6 +4856,7 @@ Throne	Bulky Buddy Box	Hot Resistance: +2, Cold Resistance: +2, Stench Resistanc
 Throne	Casagnova Gnome	Meat Drop: +20, HP Regen Min: 16, HP Regen Max: 20, MP Regen Min: 9, MP Regen Max: 11
 Throne	Cat Burglar	Item Drop: +10
 Throne	Chauvinist Pig	Experience (Muscle): +2
+Throne	Chest Mimic	Experience Maximum HP: +20
```
midgleyc, line 4859:
```suggestion
Throne	Chest Mimic	Maximum HP: +20
```
also wiki says this is +5

also also we should test the modifiers are actually valid but that doesn't need to be this PR

spaghetti-squash: Both you and wiki are correct, bad copy-paste error on my part

spaghetti-squash:
```suggestion
Throne	Chest Mimic	Maximum HP: +5
```

midgleyc (approved):

PR #2738, opened 2025-02-15
Require a Throne line in modifiers.txt for every familiar (and do some things to support that)

gausie:
Add data file consistency test to ensure all familiars have a Throne description.
Derive FamiliarData.canCarry() from data.
Differentiate between a blank modifier (unspaded) and a modifier of "none" (cannot be carried)).

### DIFF
--- a/test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
+++ b/test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
@@ -22,10 +22,12 @@ import java.io.IOException;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.function.Function;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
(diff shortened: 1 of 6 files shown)


test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
```diff
@@ -621,4 +623,40 @@ public void monsterElementsAreValid() {
...
+        FamiliarDatabase.entrySet().stream()
+            .map(Map.Entry::getKey)
+            // Ignore Pokefam-exclusive familiars
+            .filter(id -> !FamiliarDatabase.hasAttribute(id, "pokefam"))
+            // Ignore familiars with no hatchling (currently April Fools familiars, but may also
+            // catch future weirdos
+            .filter(id -> FamiliarDatabase.getFamiliarLarva(id) > 0)
+            .collect(Collectors.toSet());
+    String file = "modifiers.txt";
+    int version = 3;
```
jaadams5, line 639: We should probably revive the versioned file discussion or at least consider reporting version mismatches in a more visible way..  There might be a case for bumping the version of modifiers.txt.  Neither point is blocking, however.

jaadams5 (approved):

PR #2741, opened 2025-02-15
fix: flip .equals as this.name may be null

midgleyc:
As reported in https://kolmafia.us/threads/checking-sell_price-of-cyburger-results-in-a-npe.30320/#post-176341

### DIFF
--- a/src/net/sourceforge/kolmafia/AdventureResult.java
+++ b/src/net/sourceforge/kolmafia/AdventureResult.java
@@ -369,19 +369,19 @@ public class AdventureResult implements Comparable<AdventureResult>, Cloneable {
   }
 
   public boolean isHP() {
-    return this.name.equals(AdventureResult.HP);
+    return AdventureResult.HP.equals(this.name);
   }
 
   public boolean isMP() {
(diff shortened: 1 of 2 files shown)


heeheehee-kolmafia (approved):
Thanks! Still not sure if this.name being null is a separate bug or not...

midgleyc:
As it wasn't happening before I think it is, but I also think this is a good idea regardless.

Veracity0:
It is a separate bug.
The issue is that old-style coinmasters have a "token" which is the currency.
This will be in data.item or data.tokenItem

What about coinmasters that have multiple currencies?
And what about coinmasters with new ShopRow format, which don't have an "item", even if they only have a single "currency".

data.getBuyPrice assumes that data.buyPrices is not null and returns an integer which is the number of "item" or "token" needed to buy the thing - or 0 if it is not sold.

data.itemBuyPrice returns an AdventureResult of that many item or token.

I think it should return null if the item is not sold and all caller of that method whould understand that.

The issue here is that since there was no "item", it assumed there was a token and was getting an instance of the token - which was null.

I think that's the root casue, and I will make a PR for it.

PR #2746, opened 2025-02-17
Command and JSON APIs should return 401 if the password hash is incorrect.

phulin:
These APIs currently just reset the connection, which is difficult to handle from the script side. Instead, they should respond RESTfully, which means sending back a 401 Unauthorized status code.


gausie:
Minor but shouldn’t this be 401? Since the solution is to reauthenticate

gausie (approved):

PR #2747, opened 2025-02-17
RelayAgent should respond to malformed requests.

phulin:
This continues my quest to have Mafia respond reasonably with RESTful API responses. RelayAgent also currently has a nasty habit of just terminating the connection if a malformed request comes in. This PR makes it return 400/401 status codes depending on the issue, and adds some basic testing to RelayAgent.


gausie (approved):
Love this

PR #2750, opened 2025-02-18
Attempt to handle mall purchase parsing failure

Veracity0:
This bug report:

https://kolmafia.us/threads/mafia-seems-to-spend-meat-but-not-buy-an-item.30300/

show examples of KoLmafia making a mall purchase but not recognizing that it succeeded, and therefore attempting to purchase from the next cheapest store, ad infinitum, until the user runs out of Meat.

Obviously, that almost never happens: when we detect that there is an item with a "rel" string, it was a successful purchase.

If we then fail to parse it, log the text to the DEBUG log and put KoLmafia into an ERROR state, so makePurchases doesn't try to move to the next mall store.

I wonder what happens if I manually attempt to buy something I cannot afford?
I'll check that and mark this ready to review if things work as expected.

Yeah. So, check for "You can't afford that item" before proceeding to parse the result of the purchase.


gausie (approved):
This is the perfect way forward, as ever

PR #2751, opened 2025-02-18
Stop character sheet breaking Zootomist stats

HippoKingKoL:
At the moment, every time the char page request is hit for Zootomist, it gets the wrong stats and sets your character back to level 1. This uses the code from Grey You where a similar problem existed to fix the issue.


midgleyc (approved):
thanks!

PR #2753, opened 2025-02-19
Add sell_cost function to get costs for an item as a map of item -> quantity

gausie:
As discussed in the shop.php thread on the forums. Veracity might have an idea of shop types that still need to be tested here.


Veracity0 (approved):
This is lovely.

Something we absolutely don't need to consider now, because we don't support it, is that shop.php can support Meat costs, as well - but our paradigm is to put those items in npcstores and non-Meat costs into coinmasters - and a shop can be both.

If KoL ever releases a shop.php with an item that costs both Meat and non-Meat, we'll have to think about this here - and elsewhere.

PR #2758, opened 2025-02-20
feat: always read level from API, don't calculate it

midgleyc:
Zootomist uses grafts for levels, so easier to just always use the API.

This introduces a race on startup in Gelatinous Noob, which isn't great, but can be fixed
by refreshing the charpane as we'll definitely have the level by then.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -185,8 +185,6 @@ public abstract class KoLCharacter {
   public static int AWOLtattoo = 0;
 
   private static int currentLevel = 1;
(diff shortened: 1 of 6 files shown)


jaadams5 (approved):
I accept the sense of urgency in order to support the new path.  I would like to know if there are plans to address the race condition and whether disabling the test rather than adding  withLevel is a better way to remind someone to do so?

midgleyc:
I don't plan on addressing the race condition: it occurs only on the first time you view the charpane in a given session, only if Mafia manages to parse the charpane faster than it parses api.php, and can be fixed by viewing the charpane again.

It could be fixed by parsing level from the charpane, but I'm not sure that would work with custom titles.

PR #2765, opened 2025-02-21
Use @RetryingTest for our common failure cases

gausie:
This allows us to benefit from the tests without having to immediately work out how to fix them.


jaadams5 (approved):

jaadams5:
I note that we have a number of intermittent failures that to the best of my knowledge are caused by running tests in parallel.  Test one aggressively cleans up the contents of a shared directory when test two has just created files there, that it needs.  The test failure is due to the absence of an expected file or the absence of content in a file.  Just mentioning because my opinion is that a RetryingTest would be the wrong solution in those cases.

PR #2769, opened 2025-02-22
Grafted familiars can't be used in Zootomist.

phulin:
This is causing various script failures. This PR designates them as not "usable", so `have_familiar` will return false in scripts. This works like QT or G-Lover.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -13,6 +13,8 @@ import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import net.java.dev.spellcast.utilities.LockableListModel;
 import net.java.dev.spellcast.utilities.SortedListModel;
 import net.sourceforge.kolmafia.AscensionPath.Path;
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/KoLCharacterTest.java
```diff
@@ -297,6 +297,35 @@ public void familiarsWithoutGsAreStillOwnedInGLover() {
...
+            withProperty("zootGraftedButtCheekLeftFamiliar", FamiliarPool.MOSQUITO));
+
+    try (cleanups) {
+      var fam = KoLCharacter.usableFamiliar("mosquito");
+      assertThat(fam, nullValue());
+    }
+  }
+
+  @Test
+  public void graftedFamiliarsAreOwnedInZootomist() {
```
midgleyc, line 322: I would rather these two tests be combined to make it obvious they have the same setup; I think this makes clearer the intent that grafted familiars are owned but unusable.

src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -4436,13 +4438,32 @@ private static FamiliarData usableFamiliar(final Predicate<FamiliarData> familia
...
+            "zootGraftedHandLeftFamiliar",
+            "zootGraftedHandRightFamiliar",
+            "zootGraftedHeadFamiliar",
+            "zootGraftedNippleLeftFamiliar",
+            "zootGraftedNippleRightFamiliar",
+            "zootGraftedShoulderLeftFamiliar",
+            "zootGraftedShoulderRightFamiliar")
+        .map(pref -> Preferences.getInteger(pref))
+        .filter(id -> id > 0)
+        .collect(Collectors.toUnmodifiableSet());
```
midgleyc, line 4456: optional comment: returning the stream directly and doing `anyMatch`  feels better to me in some sense (probably perf or simplicity)

midgleyc (approved):
two comments, optional.

midgleyc (approved):
Thanks!

PR #2770, opened 2025-02-22
Disable unavailable resonances in the maximizer.

phulin:
This disables not-currently-available resonances in the maximizer (by graying them out).

image


Veracity0 (approved):
Perfect! I was just presented with "reminiscences of a cheesier age" when I had already used all 3 mayam selections today, and was thinking "that is unavailable. why are you suggesting it?"

Veracity0 (changes requested):
Although ... I will point out I never used that responance, but it was not available to me for other reasons.
- Having used all the resonances
- Having used other resonances that made that one impossible.

The second of those subsumes the first.
Can you accommodate that? Disallow a resonance if it is not allowed because the four specific symbols are not all available?

Veracity0 (approved):
You are correct - the code handles it.
I don't think you need another test.

PR #2774, opened 2025-02-25
Last Available modifier for items

gausie:
EDIT: This now includes all items, and is somewhat rigorously tested.

I think the only thing left to go is skills.

While I was in here, the modifier tools kept annoying me about the avatar IotMs that are only free pullable / only usable in hardcore in their special paths. I decided to make boolean modifiers support evaluating conditions, which was surprisingly easy. I retroactively applied that to the Time-Twitching Toolbelt, with which we had struggled before.


gausie:
There are 2216 more modifiers to add, but they are all for items that currently do not make an (uncommented) appearance in modifiers.txt. That makes them harder to add because I need to make sure they go to the right section in the right order.

PR #2775, opened 2025-02-26
Avant Guard points and Burly Bodyguard ancestral weight

gausie:
- **Avant Guard has points (points^2 experience for your bodyguard familiar after its first combat)**
- **Track burly bodyguard prefs and exp and add tests**


midgleyc (approved):
Thanks!

PR #2777, opened 2025-02-27
feat: fam tags from zoot

midgleyc:
The order can be determined from the order the nipple skill effects appear in. I assume the sheet had these right.

The order for "animal" and "technological" is wrong, presumably because "Regenerates X" always appears at the bottom of the effect.


Veracity0 (approved):
Does zoot allow you to graft familiars which are out of Standard?
If so, I should be able to check various OLD fams, when I try this path this weekend.
How do I do that?

Veracity0 (approved):

midgleyc:
If you break Ronin, you can graft familiars which are out of standard.

PR #2783, opened 2025-02-28
feat: add function for checking item data from museum, and update item data

midgleyc:


Veracity0 (approved):
I had never heard of loathers/museum.
How was the data collected for it?
Anyway, cool. Thanks for doing this!

midgleyc:
Like ePeterso2's Jicken Wings before it, it receives a data dump from the KoL devs of item and collection data. gausie exposed plurals a while back, and recently agreed to expose item data too (for Mafia).

PR #2786, opened 2025-03-01
feat: update use / combat from museum

midgleyc:
Some items newly marked as "usable" require further user interaction to actually do anything. This was already the case for e.g. ancient cure-all, so I think this is fine.

Some items marked as "food helper" were actually just used automatically from being in inventory.

Some items that send you to a shop are "usable", and some are not. Fun!

### DIFF
--- a/src/data/items.txt
+++ b/src/data/items.txt
@@ -33,7 +33,7 @@
 5	pasta spoon	249844320	pastaspoon.gif	weapon, smith	t,d	1
 6	ravioli hat	882973843	ravioli.gif	hat	t,d	1
 7	saucepan	323344201	saucepan.gif	weapon, smith	t,d	1
-8	spices	266103456	spice.gif	none, combat reusable, smith, cook, mix	t,d	1	spices
+8	spices	266103456	spice.gif	message, combat reusable, smith, cook, mix	t,d	1	spices
 9	disco mask	449052630	discomask.gif	hat	t,d	1
 10	disco ball	371973562	discoball.gif	weapon, smith	t,d	1
 11	stolen accordion	199623005	accordion.gif	weapon, smith	t,d	1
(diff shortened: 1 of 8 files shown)


Veracity0:
So, things that send you to a shop used to be marked "none", but many of them are now marked "reusable".
Does this mean that they now show up in the "General" section of the Item Manager Frame and you can click the "use item" button on them? If so, what does it do?

src/data/items.txt
```diff
@@ -7993,10 +7993,10 @@
 7965	Holy MacGuffin	534962786	macguffin.gif	none	q	0
 7966	Ka coin	826932303	kacoin.gif	none	q	0
 7967	World's Best Adventurer sash	584125368	bestsash.gif	accessory	q	0	World's Best Adventurer sashes
-7968	topiary nugglet	491943392	nugglet.gif	none	t	0
+7968	topiary nugglet	491943392	nugglet.gif	reusable	t	0
 7969	beehive	905003610	beehive.gif	none, combat reusable	q	0
 7970	electric boning knife	580217520	elecbone.gif	none, combat reusable	q	0	electric boning knives
-7971	Aggressive Carrot	839021006	carrot.gif	none, curse	t	0
+7971	Aggressive Carrot	839021006	carrot.gif	usable	t	0
```
Veracity0, line 7999: Is it not still a "curse" item? I.e., you need to give it a user name to afflict.

midgleyc: trying to use it in `curse.php` doesn't work. It sends you to a choice adventure.

midgleyc:
Using the item sends you to the shop. The 2002 Mr. Store Catalog worked that way already. The odd silver coin was also marked as usable, but actually wasn't usable.

For the item manager frame: I don't know, I would guess it sends the player to the shop in the background but does nothing visible?

We overrode most of the use links manually in the `UseLinkDecorator`. For items like FDKOL commendation that we didn't override, now uncloseting it adds a "use" link that sends you to the shop, whereas before it had nothing.

Veracity0 (approved):

PR #2788, opened 2025-03-01
Add getaway campsite command.

phulin:
This way we can display cloud-talk in the maximizer, which I always forget.


midgleyc (approved):
Thanks!

PR #2792, opened 2025-03-02
Leprecondo Support

gausie:
This is all done. A `leprecondo` command for info and configuration would be great and will come in a future PR. I could work on it later in the week unless someone else is interested. The furniture -> location data is all there ready, and I added it to AreaCombatData's toString() command since that was easy enough.

- [x] Track discoveries
- [x] Update discoveries when visiting the leprecondo
- [x] Track installed furniture
- [x] Track needs
- [x] Learn need order
- [x] Track turn on which last need changed
- [x] Track daily reconfigurations
- [x] Make sure it all applies on action as well as visit


Veracity0:
This is good, so far.

I didn't open my Leprecondo until I started this run, so, as of right now, having spent one turn on day 3, I have 3 of the 6 basic furnishings.

I specifically went to a location with something better on turn one today - and I got a basic furnishing I didn't have..
This is going to require careful planning to fill up.

I think the person who said "oh, I got all six basics on my first day" was lying. :)

I suggest that a command like "accordions" would be useful to see all the furnishings, where you find them, whether or not you have them, which needs they fulfill, rewards, etc.

Once I get the 6 basics, I expect I'd like to spend my early turns every day in a location to get something I don't have, and a reference like that in KoLmafia would obviate going to look at a spreadsheet.

gausie:
> I suggest that a command like "accordions" would be useful to see all the furnishings, where you find them, whether or not you have them, which needs they fulfill, rewards, etc.

Yes I agree. I think I may make a leprecondo manager and move some of this code around.

I also heard that some grammar was fixed in the discovery messages at some point in the last 24 hours, so I'll need to make sure the regex holds up (it won't).

src/net/sourceforge/kolmafia/session/LeprecondoManager.java
```diff
@@ -0,0 +1,242 @@
...
+          Map.entry("plops down and reads some old comics", "dumb entertainment"),
+          Map.entry("fires up his second-hand hot plate", "food"),
+          Map.entry("heads to his filthy mattress for a nap", "sleep"),
+          Map.entry("grabs a beer from the beer cooler", "booze"),
+          Map.entry(
+              "wanders over to the karaoke machines and sings a few popular songs",
+              "dumb entertainment"),
+          Map.entry(
+              "sets his karaoke machine to dance mode and careens around the room", "exercise"),
+          Map.entry("spends a while running on his cupcake treadmill", "exericse"),
```
midgleyc, line 90: "exericse"

Might be worth making an enum to avoid typoes.

src/net/sourceforge/kolmafia/session/LeprecondoManager.java
```diff
@@ -0,0 +1,242 @@
...
+          Map.entry("fine upholstered dining table set", "The Hidden Apartment Building"),
+          Map.entry("whiskeybed", "The Castle in the Clouds in the Sky (Ground Floor)"),
+          Map.entry("high-end home workout system", "The Degrassi Knoll Gym"),
+          Map.entry("complete classics library", "The Haunted Library"),
+          Map.entry("ultimate retro game console", "Megalo-City"),
+          Map.entry("Omnipot", "Cobb's Knob Laboratory"),
+          Map.entry("fully-stocked wet bar", "The Purple Light District"),
+          Map.entry("four-poster bed", "Dreadsylvanian Castle"));
+
+  private static final Map<String, String> FULFILMENT_TO_NEED =
```
midgleyc, line 76: needs to be filled out with the other messages

midgleyc (approved):

PR #2796, opened 2025-03-04
Pretty up logging of combat skills with embedded vars

Veracity0:
I noticed that StationaryButtons prettify dart skills - "darts: throw at butt" - but not Zootomist skills "left %N kick".
That also reminded me that logging combat skill use prettifies neither.
```
[258] The Hidden Park
Encounter: pygmy janitor
Round 1: Veracity casts LEFT %N KICK!
You acquire an effect: Everything Looks Yellow (30)
Round 2: Veracity wins the fight!

[258] The Hidden Park
Encounter: pygmy blowgunner
Round 1: Veracity casts DARTS: THROW AT %PART4!
Round 2: pygmy blowgunner takes 451 damage.
Round 2: Veracity wins the fight!
```

1) Create ``` String SkillDatabase.getPrettySkillName(int skillId)```.
2) Use this for fight logging:
```
[415] The Smut Orc Logging Camp
Encounter: smut orc screwer
Round 1: Veracity casts DARTS: THROW AT BUTT!
Round 2: smut orc screwer takes 394 damage.
Round 2: Veracity wins the fight!

[419] Twin Peak
Encounter: spider (duck?) topiary animal
Round 1: Veracity casts LEFT QUANTUM ENTANGLER KICK!
You acquire an effect: Everything Looks Yellow (30)
Round 2: Veracity wins the fight!
```
3) StationaryButtons already did that for darts. Now, it calls the new method and uses it for darts and Zootomist combat skills.
4) While I was at it, I removed all the magic numbers from the StationaryButtons switch statement that chose appropriate button text for skills by number, and made it a modern switch statement. I think we'd not invented SkillPool when that was first implemented...
5) I removed the RelayWarning for equipping spring shoes when looking at giant coffee grounds if you are a Zootomist, since the reason you'd want to do that is to gain lots of substats, which doesn't happen in this path.

I added some tests.
And I opened an "issue" at ScotchLog, mentioning this and asking if it will be a problem.

https://github.com/docrostov/ScotchLog/issues/9


gausie:
This is good for reading logs, but not great for parsing logs. Do you think we want to make sure that people involved in tools like https://kolmafioso.app like (or at least know about) this change?

Veracity0:
> This is good for reading logs, but not great for parsing logs.

Well, perhaps I lack some imaginations, but considering that dart parts change every single round (and while you are in battle, you can use _currentDartBoard to see the mapping), but the log doesn't show that. For me, at least.
Hmm. I think some scripts use verbose "log every preference change" and I suppose they could use that and deduce which skill was used. But it seems that changing the name logging like this PR does could simplify things by obviating that mapping.

For the zootomist skills, again, if the log ... somehow ... had the grafting properties in it, you could deduce what to expect.

> Do you think we want to make sure that people involved in tools like https://kolmafioso.app like (or at least know about) this change?

I think it is reasonable to tell them about it.

gausie:
I asked @docrostov and he says

> i just looked at the tests i think it will be fine i just will need to change bullseye tracking

So I was probably just being overly cautious

PR #2797, opened 2025-03-04
feat: more leprecondo items / effects

midgleyc:


gausie:
Hmmm do we not pick up Last Available automatically. I thought I made it so we did.

midgleyc:
Certainly doesn't look like it!

Veracity0:
So what about making that work and re-committing the items, etc., with that data?

gausie:
It's probably because the tag is in a comment at the end of the doc and not in the main enchantment block, and I've missed the place that generates the gcli report. We can merge this and I'll just have to add manually.

gausie:
Here is the issue: https://github.com/kolmafia/kolmafia/pull/2799

gausie (approved):

PR #2801, opened 2025-03-05
Allow 'leprecondo furniture' as a goal (condition)

Veracity0:
I currently have 16 pieces of Leprecondo furniture.

If I set "1 leprecondo furniture" as a goal before adventuring, it will adventure in the selected zone until your leprechaun drags a piece of furniture back to its condo.

I originally did zone checking and would count the goal as done if there was no more furniture available in that zone. However, since there are 6 furniture pieces that can be found in essentially any zone, that can't work.

I am in-run and out of turns.
After rollover, I will bomb the frat house back to the stone age and confirm that this (still) works as expected and will mark this ready for review.

Edit: works as expected.


gausie (approved):
Great!

PR #2806, opened 2025-03-07
Try and fix "Max" command in maximizer

Ignose:


gausie:
Please add a description of what you're trying to do as well as some tests for your change.

PR #2808, opened 2025-03-08
synchronize loadUserPreferences with saveToFile

Rinn:
This is a speculative fix, but what I think may have been happening for the preference reset is during the process for a time-in login:
1. The user is timed out from being idle, laptop goes to sleep, etc.
2. An action occurs where the user is timed back in
3. Preferences.reset(null) is called, the current user's preferences are flushed to disk and the pref collections are cleared
4. Preferences.reset(user) is called, which calls loadUserPreferences()
5. Preferences.userPropertiesFile is set to the user's preference file, before it is loaded from disk, this happens while the pref collections are still empty and the preferences are set later in this function
6. Preference collections are reloaded from the user file contents

I speculate that between step 5 and 6:
1. Another thread receives a property change notification (maybe _concoctionDatabaseRefreshes? that preference is updated all the time)
2. Due to the saveSettingsOnSet global property being true this calls writeToDisk() and cleared or partial settings are saved to disk
3. loadUserPreferences() continues and reloads settings from the bad file just written

This uses the same lock object around the entire body of loadUserPreferences() that is synchronized in writeToDisk().

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -226,95 +226,100 @@ public class Preferences {
         new File(KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.txt");
     File backupFile =
         new File(KoLConstants.SETTINGS_LOCATION, Preferences.baseUserName(username) + "_prefs.bak");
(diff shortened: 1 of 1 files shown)


gausie (approved):
I like this and am happy to merge it. Even if it doesn't fix the problem the logs will be helpful.

PR #2812, opened 2025-03-09
feat: generate spoons in breakfast

midgleyc:


src/net/sourceforge/kolmafia/objectpool/SkillPool.java
```diff
@@ -132,6 +132,8 @@ public class SkillPool {
   public static final int OLD_SCHOOL_COCKTAILCRAFTING = 230;
   public static final int HOLIDAY_MULTITASKING = 240;
   public static final int OVERCLOCK10 = 243;
+  public static final int GENERATE_IRONY = 245;
+
```
gausie, line 136: I assume this is one of those newlines IntelliJ makes you think you should add because of weird spacing. Or something totally different

gausie:
I haven't followed - do we know that this would generate more spoons vs be cast more times in a future season? Shall I ask CDM

src/net/sourceforge/kolmafia/objectpool/SkillPool.java
midgleyc: no, it's because I find myself going to the end of 200 and 7000 a lot, so I want to add a newline so I can get there easier (i.e. in vim it's pressing `}` a few times)

midgleyc:
we don't know, I just assumed it would work this way.

src/net/sourceforge/kolmafia/objectpool/SkillPool.java
gausie: 👍

PR #2813, opened 2025-03-09
Added test from #1889

jaadams5:
This test was originally written for https://github.com/kolmafia/kolmafia/pull/1889

It is believed that https://github.com/kolmafia/kolmafia/pull/2808 is an alternative solution to the same problem that so far has not had the intermittent race condition when using Deck of Every Cards which is the primary reason 1889 has not be committed.

Capturing the test is a step towards closing 1889 without being committed.


jaadams5:
Making it draft until understand why test fails.

jaadams5:
#1889 and #2808 don't fix the same problem.  This test failing without #1889 gives hope to the possibility that this is a reasonable test for #1889.  So this can be closed without committing.  The lint changes here were pushed to #1889.

jaadams5:
Closing.  See above.

PR #2832, opened 2025-03-14
Sweat out Some Booze needs to have the designer sweatpants equipped"

gausie:
We have previously not needed to equip the pants for this. Apparently this is no longer the case...


gausie:
I think this is just not true. Weird. Closing for now

PR #2833, opened 2025-03-15
Passing test that Draws from Deck of Every Card 

jaadams5:
The goal is to write a test that draws a card from the Deck of Every Card.  A passing test could presumably be merged into #1889.  If it fails there, as expected, then we have one more tool to use to fix #1889 and allow it to be merged.  A secondary goal was to use the run method of DeckOfEveryCardRequest since there is currently very little test coverage of that method.

The HTML comes from a debug log that was started when the Draw daily deed was primed to pick the Mickey Mantle card and stopped when the card was drawn.

The pwd hash in the log was replaced (because I could not recall whether publishing it was a risk or not).  

The log showed 8 requests and six non empty responses.
Requesting: https://www.kingdomofloathing.com/inv_use.php?whichitem=8382&cheat=1
Requesting: https://www.kingdomofloathing.com/choice.php?forceoption=0
Requesting: https://www.kingdomofloathing.com/api.php?what=status&for=KoLmafia
Requesting: https://www.kingdomofloathing.com/choice.php?whichchoice=1086&option=1&which=58
Requesting: https://www.kingdomofloathing.com/choice.php?forceoption=0
Requesting: https://www.kingdomofloathing.com/api.php?what=status&for=KoLmafia
Requesting: https://www.kingdomofloathing.com/choice.php?whichchoice=1085&option=1
Requesting: https://www.kingdomofloathing.com/api.php?what=status&for=KoLmafia

I extracted the corresponding responses and wrote the test to submit them.  I used an empty file as the response when there was no response data.

I have tried several variations including not providing a response with an empty file.  At various times I have used the debugger to look at builder.client.getRequests() and found requests to item_desc and guild.  Those are not in the log file so there may be some character data that is needed that is not yet being set up by the test.

I am stuck.  My hunch is that I have a fundamental misunderstanding about HTML requests and responses, GET and POST and how to use builder.client.addResponse.

I did look at the test in #2831 which seems to be similar but I have not gleaned my mistakes.

The log is attached.

[DECK_DEBUG_20250311.txt](https://github.com/user-attachments/files/19263436/DECK_DEBUG_20250311.txt)

Thank you.


Veracity0:
"At various times I have used the debugger to look at builder.client.getRequests() and found requests to item_desc and guild. Those are not in the log file so there may be some character data that is needed that is not yet being set up by the test."

Bingo.

- If you have a password hash, KoLmafia looks at your vinyl boots to determine your gender. Use ```withGender(Gender.FEMALE)```
- I ran into ...something... that called KoLCharacter.getGuildStoreOpen() when I wrote a test for using an item summoning a giant sandworm. I simple did ```KoLCharacter.setGuildStoreOpen(false);```. Perhaps I should have created ```withGuildStoreOpen(boolean)``` :)

Veracity0:
Here are some hints on how to develop/debug this kind of test.
```
import static internal.helpers.Networking.printRequests;
```

Then, declare the builder - and the client, so you don't have to constantly do builder.client:

```
   var client = builder.client;
```

Then, after you've run the requests and want to validate which requests were commmitted

```
        printRequests(requests);
```

comment out ALL the assertions.
I.e., your test only runs the requests and print them.
If it is not doing exactly what your DEBUG log shows, perhaps you are missing a call to api.php, or something.

Other observations: This:
```
  builder.client.addResponse(302, html("request/empty.html"));
```
isn't correct.

First, ```html("request/empty.html")``` can simply be ```""```; you don't need to load an empty file to get an empty string. :)

That's what I use when KoLmafia is requesting api.php and I don't want to include an actual responseText for that - which causes an immense pile of worms, as it will set all sorts of character state.

Second, that is a redirect. If you really want to reproduce KoLmafia's processing to match what you see in your DEBUG log, you should force the redirect. Your DEBUG log will tell you exactly what the redirect should be. In this case, it is surely to choice.php
```
        client.addResponse(302, Map.of("location", List.of("choice.php?forceoption=0")), "");
```

Veracity0:
OK, I think I remember the issue with the guild store.
If you load a real api response, it will do things including setting your class.
It will also do a ConcoctionDatabase refresh.
Depending on your class, mixing methods depend on having access to the Guild - chefstaff, still, malus, ...
KoLCharacter.guildStoreOpen() will make a request to see if it is open,, if it doesn't know.

It is really in your best interest to NOT use an actual api.php response.

Here's a test from KoLAdventureValidationTest.java which has what I consider to be best practices that I mentioned above:

```
    @Test
    public void canAdventureIfHalfAstralWithTripUnSelected() {
      var client = builder.client;
      var cleanups =
          new Cleanups(
              withHttpClientBuilder(builder),
              withEffect("Half-Astral", 5),
              withProperty("currentAstralTrip", ""),
              withLimitMode(LimitMode.ASTRAL),
              withPasswordHash("astral"),
              // If you have a password hash, KoL looks at your vinyl boots
              withGender(Gender.FEMALE));
      try (cleanups) {
        client.addResponse(302, Map.of("location", List.of("choice.php?forceoption=0")), "");
        client.addResponse(200, html("request/test_visit_astral_travel_agent.html"));
        client.addResponse(200, ""); // api.php
        client.addResponse(200, html("request/test_choose_great_trip.html"));
        client.addResponse(200, ""); // api.php

        assertTrue(BAD_TRIP.canAdventure());
        assertTrue(MEDIOCRE_TRIP.canAdventure());
        assertTrue(GREAT_TRIP.canAdventure());
        assertTrue(GREAT_TRIP.prepareForAdventure());
        assertThat(Preferences.getString("currentAstralTrip"), is("Great Trip"));

        assertThat(requests, hasSize(5));

        assertPostRequest(requests.get(0), "/adventure.php", "snarfblat=97&pwd=astral");
        assertGetRequest(requests.get(1), "/choice.php", "forceoption=0");
        assertPostRequest(requests.get(2), "/api.php", "what=status&for=KoLmafia");
        assertPostRequest(requests.get(3), "/choice.php", "whichchoice=71&option=3&pwd=astral");
        assertPostRequest(requests.get(4), "/api.php", "what=status&for=KoLmafia");
      }
    }
```

jaadams5:
Thank you.  Trying some things now.  Closing store and setting gender reduced the number of responses and I am in the process of cleaning up my supplied responses.

jaadams5:
Thank you.  It works now.  I did create the Cleanup withGuildStoreOpen and used it in the only other place that would use it.

Veracity0:
I'm going to look at this tomorrow, after I free the king and have access again to my Deck.
But, I find it hard to believe that the requests you log are precisely the ones that KoL processes.
I see this:
```
      assertThat(requests, hasSize(6));
      assertPostRequest(requests.get(0), "/inv_use.php", "whichitem=8382&cheat=1");
      assertPostRequest(requests.get(1), "/api.php", "what=status&for=KoLmafia");
      assertPostRequest(
          requests.get(2), "/choice.php", "whichchoice=1086&option=1&which=58&pwd=cafebabe");
      assertPostRequest(requests.get(3), "/api.php", "what=status&for=KoLmafia");
      assertPostRequest(requests.get(4), "/choice.php", "whichchoice=1085&option=1&pwd=cafebabe");
      assertPostRequest(requests.get(5), "/api.php", "what=status&for=KoLmafia");
```

Looking at DeckOfEveryCardRequest.run, I see this:
```
  @Override
  protected boolean shouldFollowRedirect() {
    return true;
  }
```
Cool. That means when this request - a choice.php subclass - is executed, if it redirects, it wants KoL to follow the redirect. GenericRequest does that too.

So, when the run() method does this:
```
    GenericRequest useRequest = new GenericRequest("inv_use.php");
... add form fields
    useRequest.run();
```
I expect to see a POST request to inv_use, which will redirect to choice.php.
I expect to see a GET request to choice.php
And, presumably a call to api.php.
Then:
```
    super.run();
```
which will run the DeckofEveryCard request - which is choice.php.
I expect to see a call to choice.php.
And, presumably, a call to api.php.

And then, since we are drawing a specific card:
```
      // Otherwise, need to confirm the draw
      this.constructURLString("choice.php?whichchoice=1085&option=1", true);
      super.run();
```
Another call to choice.php
And, presumably, a call to api.php.

For some reason, your test is not following the redirect - and allowing KoL to parse the initial response and do its normal choice processing - leaving ChoiceManager.handlingChoice at true, for example.
I told you how to do that.
I expect you to do that.

Your test may succeed, but I don't think it is actually doing exactly the same processing KoLmafia does.

I'll try this out tomorrow.

FWIW, you also ignored every one of my "best practice" style comments:
- You don't HAVE to test with a password hash.
- You don't HAVE to use an actual api.php response test - unless you have actually set up your test environment to agree with whatever state you really had when you got the HTML. Otherwise, all sorts of state change will happen in KoLCharacter - above and beyond the specific expected Preference changes expected from DecKOfEveryCardRequest.

I have to wonder: is there a Daily Deed for DeckOfEveryCard Preferences? The issue you are trying to figure out is a Preferences clash/hang/something, if I recall.

I'll get my own DEBUG log tomorrow and confirm that there is actually a redirect that you are skipping the processing for.

jaadams5:
Thank you.  I did not and do no understand how to handle a redirect.  Since there is an educational component to this I will review the provided answer more closely.  This is not the test I want if it doesn't follow the same code path.  Thank you for your patience.

Veracity0:
```
client.addResponse(302, Map.of("location", List.of("choice.php?forceoption=0")), "");
```

jaadams5:
Test works and AFAIK matches the DEBUG log.  

I can detect redirection in a DEBUG log by finding response code 302 and no associated response body.

Each request in the log should have a corresponding response in the test.

If the logged response has status 302 and no corresponding response body then the mimicked response in the test should also have a status of 302.  The request contents should be a name, value pair where the name is "location" and the value is the value of the location field in the header of the logged response.

If the logged response has a status of 200 then the response body form the log should be used as the mimicked response.

When writing the test some request results will be Gets and others Posts so the proper assertion needs to be used.

There are cases where character status will effect the number and sequence of requests.  Specifically, api calls may generate additional requests.. The character's Gender and whether the Guild store are two known examples.

The presence or absence of a password hash may also change the sequence or number of requests.

Above are some of the things I (just) learned.  

Thank you.

jaadams5:
> Looks good!

Thank you.  I appreciate your help and patience.

PR #2835, opened 2025-03-15
removed coinmasters can be accessible

Veracity0:
In particular, the Cheer-o-Vend 3000, even though Crimbo17 is Removed.

The "Removed" check was intended to allow coinmasters in a zone to have accessible methods and not have to edit the method, if the zone itself is Removed, but, we can't quite do that. We CAN delete the custom method, usually.

```
> inv crystalline cheer

> ash is_accessible($coinmaster[Cheer-o-Vend 3000])

Returned: false

> ash inaccessible_reason($coinmaster[Cheer-o-Vend 3000])

Returned: You need some crystalline cheer.

> pull 1 crystalline cheer

Pulling items from storage...
Requests complete.

> ash is_accessible($coinmaster[Cheer-o-Vend 3000])

Returned: true
```

### DIFF
--- a/src/net/sourceforge/kolmafia/CoinmasterData.java
+++ b/src/net/sourceforge/kolmafia/CoinmasterData.java
@@ -1469,15 +1469,14 @@ public class CoinmasterData implements Comparable<CoinmasterData> {
   }
 
   public String accessible() {
-    // Returns an error reason or null
-    if ("Removed".equals(this.getRootZone())) {
-      return "Zone is no longer accessible";
-    }
-
     return this.accessible.get();
   }
 
   public String accessibleInternal() {
+    // Returns an error reason or null
+    if ("Removed".equals(this.getRootZone())) {
+      return "Zone is no longer accessible";
+    }
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


gausie:
To me the way around this was by overriding accessible on the coinmasters that are accessible despite their location being removed. But happy to do this if you think it's better

Veracity0:
Isn't that what this is doing?

Cheer-o-Vend 3000 is accessible, if you have the item that redirects to the shop, even though its location was removed.
It has an "accessible" method which checks for the item - not the fact that Crimbo17 is closed. I did not add that; it was existing code.

However, as coded in CoinmasterData, the "location is removed" check happened before the Cheer-o-Vend "accessible" method was called and in fact, early-returned without calling the overridden method.

This PR moves the "location is removed" check into the default "accessible" method - the one which is called if the coinmaster doesn't override "accessible".

Isn't this what you are suggesting?

gausie (approved):
Oh. Maybe? I haven't looked at it in context. What you've said sounds right though

PR #2836, opened 2025-03-16
fix: warbear foil hat uses a hardcoded list of robots

midgleyc:
For example, Dataspider and Pocket Professor count for this but aren't robots in Zoot.


gausie:
Sounds good. Maybe also post a bug report and see if we can get this fixed the same way Zombie Slayer was fixed.

gausie (approved):

PR #2839, opened 2025-03-17
Fix issue with Mayam calendar showing uncreatable items as free to retrieve

gausie:
- **Create NPC and mall pricing cleanups**
- **Move existing mall and npc store tests to new cleanups**
- **Fix behaviour of retrieve_price when requesting an item that has no ingredients and is not creatable**
- **Clearer comments**

This does seem right logically, but I don't know this area of the codebase particularly well so would be keen on some review.


jaadams5 (dismissed):

Veracity0:
(This is why I haven't reviewed, yet: I USED to be familiar with this part of the codebase, but, it's been years since I worked on it. I'll re-familiarize myself with it, by and by. Sorry!)

Veracity0:
I was wondering what kind of mixing methods have no ingredients. Turns out there are several:

MAYAM
PHOTO_BOOTH
TAKERSPACE (which does have ingredients, but they are all virtual)
GNOME_PART

Every single one of those is a "permitNoCost" mixing method.

Which is to say, there is no inherent cost to using the method - as opposed to COMBINE, say, which costs "10" - for meat paste.

Veracity0:
I see what you are doing, now.

Notice that the final return value is ```price * quantity / (yield * madeQuantity)```.
```price``` is sum of inherent cost to make + costs of all the ingredients.

For all current no-ingredient mixing methods, the inherent cost is 0. Can't assume that will always be the case, but it might be clearer to refactor like this:
```
    CraftingType method = ConcoctionDatabase.getMixingMethod(item);
    long price = ConcoctionDatabase.getCreationCost(method);
    AdventureResult[] ingredients = ConcoctionDatabase.getIngredients(itemId);

    if (ingredients.length == 0) {
      // This is a concoction with no ingredients, so if creatable == 0,
      // we can be sure that we cannot concoct it right now.
      var conc = ConcoctionPool.get(itemId);
      if (conc == null || conc.creatable == 0) {
        return Long.MAX_VALUE;
      }
      if (price == 0) {
        return 0;
      }
    }

    long yield = ConcoctionDatabase.getYield(itemId);
    long madeQuantity = (quantity + yield - 1) / yield;
    price *= madeQuantity;

    for (AdventureResult ingredient : ingredients) {
      long needed = ingredient.getCount() * madeQuantity;

      long ingredientPrice =
          ingredient.isMeat()
              ? needed
              : InventoryManager.priceToAcquire(
                  ingredient.getInstance(needed), exact, mallPriceOnly, level + 1);

      if (ingredientPrice == Long.MAX_VALUE) {
        return ingredientPrice;
      }

      price += ingredientPrice;
    }

    return price * quantity / (yield * madeQuantity);
```

Note also that for MAYAM, at least, conc.creatable will be either 0 or 1, since you cannot re-use symbols.
I don't know about PHOTO_BOOTH or GNOME_PART, but TAKERSPACE could allow more than 1.

What happens if somebody wants to acquire 2 of them? Where is that filtered out?
I  assume farther up in InventoryManager will try to make some and then buy the rest - for tradeable items.

spaghetti-squash:
GNOME_PART should definitely be capped at 1; PHOTO_BOOTH should be capped only by `3 - Properties.getInteger("_photoBoothEquipment")` (that is, you can get the same item multiple times per day)

gausie:
GNOME_PART, at least, is limited by GnomePartRequest's `canMake`

Veracity0:
You are waiting for an actual review.
I looked closely and made suggestions, but I neither approved it nor said "make these changes".
I was waiting for you to respond to my suvgestions, but no.
Do you want me to respond with either "approved" or "needs changes"?
I could do either, but I am actually waiting for your response my comments/review/suggestions.
Thanks!

gausie:
No, have just been busy and not had the opportunity to absorb and act on the review. Thank you for it

gausie:
I have applied the feedback from Veracity and am at least comfortable that this fixes the specific issue tested without breaking anything else along the way, so it's probably worth merging. If we find that some of the other methods have the issue, we can reassess where changes need to happen to support that.

midgleyc (approved):

PR #2840, opened 2025-03-17
no changing familiar nag in Zootomist

Veracity0:
The first time in a run you change from a familiar to a new one, KoLmafia gives you a nag, in case you had been attempting a 100% familiar run.

That's not a thing in Zootomist. Skip the nag.


midgleyc (approved):

Tokoeka:
Fully recognise this has already been done and whatnot (and that I'm not a mafia dev, so have very little (if any) say) but I disagree that 100% familiar runs are not a thing in Zoot, having already completed one prior to this change, and the familiar nag actually helped me do so.

gausie:
I think that's a reasonable enough argument to revert this.

PR #2841, opened 2025-03-18
Revert "no changing familiar nag in Zootomist"

gausie:
Reverts kolmafia/kolmafia#2840

As @Tokoeka [mentioned](https://github.com/kolmafia/kolmafia/pull/2840#issuecomment-2732037297), completing a 100% familiar run is possible in this path and the nag was useful in doing so.

As someone who doesn't do 100% familiar runs, the feature is not useful to me at all, so proposing a revert as a neutral observer :)

### DIFF
--- a/src/net/sourceforge/kolmafia/request/RelayRequest.java
+++ b/src/net/sourceforge/kolmafia/request/RelayRequest.java
@@ -2510,8 +2510,7 @@ public class RelayRequest extends PasswordHashRequest {
 
     if (!KoLCharacter.getPath().canUseFamiliars()
         || KoLCharacter.inPokefam()
-        || KoLCharacter.inQuantum()
-        || KoLCharacter.inZootomist()) {
+        || KoLCharacter.inQuantum()) {
       return false;
     }
 


midgleyc:
Okay so the way you would do this is:
* use a familiar that drops +familiar exp items
* use the out-of-combat +fam exp items over multiple days to get enough XP
* after this falls out of standard, use the shortest-order cook (perhaps with the blue plate) plus the out-of-combat stuff to get everything to the right level in one day

midgleyc (approved):
I think this will be more common once it leaves standard.

But if folk are planning on doing it now, fair enough.

Tokoeka:
> Okay so the way you would do this is:
> 
>     * use a familiar that drops +familiar exp items
> 
>     * use the out-of-combat +fam exp items over multiple days to get enough XP
> 
>     * after this falls out of standard, use the shortest-order cook (perhaps with the blue plate) plus the out-of-combat stuff to get everything to the right level in one day

Yeah, was able to get all familiars levelled and attached in a single day in Normal, reckon Hardcore would take 2-3 whilst in Standard due to inability to pull turngen items

PR #2842, opened 2025-03-18
feat: add types for zoot fams, split delevel before and during combat

midgleyc:
Heat Wave increases init, but apparently only in combat (doesn't show up on talking spade). I have bug reported and hopefully CDM will fix it to be like the Woim, which is more useful (e.g. for modern zmobies).

I have split "delevels before combat" from "delevels during combat". There should not be any user-visible changes from this yet.


gausie (approved):
Great change!

PR #2844, opened 2025-03-19
feat: add modifier for 'lantern'

midgleyc:
Lanterns add extra spell damage equal to the largest hit of a spell, in a particular element (or elements).

Add two modifiers: "Lantern" giving the number of sources, where 1 is +100% damage; and "Lantern Element", giving the element(s) of the spell (omitted if nonconstant).

Porcelain porkpie is absent due to only working on PM spells. Novelty Belt Buckle is absent as the damage comes as an extra hit, and doesn't work for the wall of bones (and matters less for other monsters as it's only 10-20%).


jaadams5 (approved):

gausie:
We really ought to have `element_modifier` (and now `elements_modifier`)

midgleyc:
I think this is the first time it would appear.

Why do we have all these functions anyway? Why can't you use `string_modifier(x).to_element()`?

PR #2860, opened 2025-03-28
update after casting skill incorrectly unselects skill

Veracity0:
Previous PR correctly deselected skill which was disabled, but also deselected skill which was not disabled after a cast.
Turns out that is a side effect of updating skill list after casting - which was new in that PR - and is unnecessary.
Remove that and only skill which is out of casts gets deselected.


midgleyc (approved):

Veracity0:
Swing is hard.

PR #2864, opened 2025-03-30
feat: track getting blood bag from cloake feasting

midgleyc:
I wanted to log the message too but it's really hard to get it in the right place and not log something twice.

We have
```html
<table>
 <tbody>
  <tr>
   <td>
    <p>You feast on your foe's nourishing, life-giving blood, stealing 100 hit points.</p>
    <center></center>
    <p>Some of the blood soaked into your cloake -- you wring it out into a plastic bag you found in the trash and create a makeshift blood bag.</p>
    <center></center></td>
  </tr>
 </tbody>
</table>
```

which is handled in `processTable`, which detects the table has a damage message, logs it, and then finishes. But in addition to the damage message there's the blood bag acquisition message, in another `<p>` tag. And in addition to both of those, there's the HP gain message and the blood bag acquisition (which are deleted by a previous function which deletes tables inside tables, and tracks them separately).

So the full HTML is
```html
<table>
 <tbody>
  <tr>
   <td>
    <p>You feast on your foe's nourishing, life-giving blood, stealing 100 hit points.</p>
    <center>
     <table>
      <tbody>
       <tr>
        <td><img src="https://d2uyhvukfffg5a.cloudfront.net/itemimages/hp.gif" height="30" width="30"></td>
        <td valign="center" class="effect">You gain 125 hit points.</td>
       </tr>
      </tbody>
     </table>
    </center>
    <p>Some of the blood soaked into your cloake -- you wring it out into a plastic bag you found in the trash and create a makeshift blood bag.</p>
    <center>
     <table class="item" style="float: none" rel="id=10168&amp;s=0&amp;q=1&amp;d=0&amp;g=0&amp;t=0&amp;n=1&amp;m=0&amp;p=0&amp;u=u">
      <tbody>
       <tr>
        <td><img src="https://d2uyhvukfffg5a.cloudfront.net/itemimages/bloodbag.gif" alt="blood bag" title="blood bag" class="hand" onclick="descitem(205374773)"></td>
        <td valign="center" class="effect">You acquire an item: <b>blood bag</b></td>
       </tr>
      </tbody>
     </table>
    </center></td>
  </tr>
 </tbody>
</table>
```
which gives us the order we want: 
```
Round 1: Ryo_Sangnoir casts DARK FEAST!
Round 2: Steve Belmont takes 100 damage.
Round 2: You gain 125 hit points
Some of the blood soaked into your cloake -- you wring it out into a plastic bag you found in the trash and create a makeshift blood bag.
You acquire an item: blood bag
```
but we would need to change how we handle tables to get that, at least.

I decided to just add the property in the end.


Veracity0:
I agree with "I'd like to log the message too". The "it's really hard" comment makes me think that perhaps the code should be refactored - again.

What if we didn't extract tables and process them at the end?
If we processed things strictly in the order KoL presents them, we'd see everything in that order.

An issue might be that there are places where we decide "this message is all there is to process" - and return early - and other places where we don't do an early return.

So, it likely not be a trivial refactoring, but is probably worthy of consideration in the future.
I'd expect to see a lot of tests, verifying that everything we expect to log is still logged - possibly in a better position.

Sounds like a project.

I suppose I should look at your code now. :)

Veracity0 (approved):
See other comment. This is fine, for setting property.

midgleyc:
> What if we didn't extract tables and process them at the end?

Yes, I think this is the move. We would need to never early-return from processing tables, and instead process children as normal. This would also catch the `<p>` tags after the damage.

PR #2867, opened 2025-04-01
feat: dump april shower items

midgleyc:


test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
```diff
@@ -185,7 +185,8 @@ public void testFamiliarHatchlings() throws IOException {
       if (ItemDatabase.getConsumptionType(id) == ConsumptionType.FAMILIAR_HATCHLING) {
         var name = ItemDatabase.getItemDataName(id);
         // replica familiars from Legacy of Loathing are special
-        if (!name.startsWith("replica ")) {
+        // wet paper tiger familiar is not public at the moment
```
jaadams5, line 188: When is that expected to change?  Perhaps delay this until it is public?

jaadams5 (approved):
Comment is not blocking.  Having this available soon may justify having to undo a portion later.

test/net/sourceforge/kolmafia/DataFileConsistencyTest.java
midgleyc: I get the impression it's going to be a while.

PR #2872, opened 2025-04-02
Suppressed unchecked cast warning and cleaned up some lint

jaadams5:
Everything changed was flagged by IntelliJ's analysis but not everything flagged was corrected.  Not that it matters unless an error is actually logged but assertEquals wants the first parameter to be what was expected and the second to be what the test generated.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
@@ -368,7 +368,7 @@ public class MallPriceManagerTest {
 
       // This item is a NPC item available for the 5th cheapest price
       long mallPrice = MallPriceManager.getMallPrice(item);
-      assertEquals(mallPrice, 400);
+      assertEquals(400, mallPrice);
     }
   }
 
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
```diff
@@ -737,7 +738,8 @@ public void canSkipForbiddenStoresWithActualRequest() {
         // MallSearchRequest will give MallPriceManager a full set of results.
         // MallPriceManager will filter them.
         List<PurchaseRequest> results = MallPriceManager.getSavedSearch(item.getItemId(), 5);
-        assertEquals(results.size(), 30);
+        assert (results != null);
```
midgleyc, line 741:
```suggestion
        assertNotNull(results);
```

midgleyc: I think IntelliJ should not complain if there's a test here, but not certain

midgleyc (approved):

PR #2874, opened 2025-04-03
Parameterized test

jaadams5:
This test was intended as a canary for preference file corruption.  This parameterizes it so that all the combinations of two related preferences are checked.  This may be a step towards tests with Swing unavailable in #1889


gausie (approved):
Unusual to see a parameterised test where the assertions are not influenced by the parameters, but code looks fine

Purely as a comment, no need to change this: this could have been a [@CartesianTest](https://junit-pioneer.org/docs/cartesian-product/), which are always fun

PR #2881, opened 2025-04-07
feat: add modifiers for replacing / additional buff effects

midgleyc:
This is step 1 for a project to support items which change buff skills by adding / replacing the effect they give.

The desired use-case is to make Lubricating Sauce / Tubes of Universal Meat / Thoughtful Empathy show up in maximizer. My current plan is to make a new command `ecast <num> <skill> @ <effect>` that I can add as a souce for these new skills / the effects that get replaced.


gausie (commented):
I know we have it elsewhere but I wonder if there's a way we can do this in a way that makes the modifier order not matter.

Could, for example, we do a single modifier that is comma separated and isntead make effect_modifier and skill_modifier (etc) know how to deal with them appropriately?

midgleyc:
Presumably we could do that, but I like the order mattering solution. Can you see a problem with it?

gausie:
I just don't think it's obvious at all that order matters for modifiers, and it's begging for problems in the future.

midgleyc:
Well, order does matter, so we can make that a requirement and note it at the top of the file?

I can't see what problems could occur, unless you tried to sort the modifiers alphabetically or something? But we also did that for monster item drops, and item drop order turned out to matter. So the fix is to not do that.

midgleyc:
This doesn't need to be exposed as a modifier at all.

PR #2885, opened 2025-04-10
fix: combat rate is capped at 35%

midgleyc:


spaghetti-squash:
Will there still be a way to access your uncapped combat rate for Community Service tests?

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -322,10 +322,16 @@ private double derivePrismaticDamage() {
   private double cappedCombatRate() {
     // Combat Rate has diminishing returns beyond + or - 25%
     double rate = this.doubles.get(DoubleModifier.COMBAT_RATE);
+    if (rate > 75.0) {
+      return 35.0;
```
jaadams5, line 326: If there is a cap at 35 then why not make this test and line 332 below a test against 35 and not 75?

jaadams5 (approved):

jaadams5 (approved):

PR #2891, opened 2025-04-17
Update zootomistPoints max to 11

Rinn:
@gausie said 11 is probably correct


midgleyc:
When you go to inject CCs at the Specimen Preparation Bench, on your first Zoot ascension you can do it once (unless this got fixed) and the cap is 11.

So I capped zootomistPoints at 10, as then the number of times you can inject is 1 + zootomistPoints.

Rinn:
Ah okay, sounds good then nm.

PR #2901, opened 2025-04-30
Add some Items to Item Pool (discovered while Using the test Framework to explore some Maximizations)

jaadams5:
The Item Pool changes may be replicated and committed elsewhere but I am not expecting to commit the new file unless someone thinks it is useful.

I tried to set up some simple maximizations to see whether there was an item combination where no weapon was the choice.  I then had way too much fun with Cartesian tests. 

I was looking for a case where no item was equipped but an offhand was selected.  I did not find one.  

In all cases the weapon and offhand were the only things that contributed to the score.  Weapons were chosen even when the contribution was zero.

The intended us is to make a local branch and change the code to generate test cases of interest.  If the cases are of more general interest than copy the test with changes, add it as a new test and then push for collaboration.


jaadams5:
> Why do we want to add unused item pool integers?

Convenience for me to support my habits and preferences?  If you want we can discuss my bad habits and workflows that date from when RCS was the version control system of choice and network access to the remote repository was not always available :-) This commit lets me continue to play with the test code locally while only manipulating one file.

By my quick count there are about 400 unused items in ItemPool so if that is a concern perhaps this should be replaced by a PR that deletes the unused ones?

jaadams5:
> If it's really helpful to you, happy to merge. But my genuine answer to your question is that I would delete the 400 integers

At best this is a personal convenience and similar to implementing ash functions because the submitter did not want to write the function in ash, although they could, or changing behavior for users on the dev server.  

This raises some questions and I think I will start discussion elsewhere before I close or commit.  Thanks.

jaadams5:
Closing.  It can be trivially reopened but it looks like cleaning unused ItemPool definitions is likely to be prefered over adding new and unused ones.

PR #2922, opened 2025-05-16
feat: tickle-me emilio tags

midgleyc:


src/data/items.txt
```diff
@@ -11942,3 +11942,4 @@
 11914	imposing pilgrim's hat	589422935	pilghat.gif	hat	t,d	23
 11915	crown of thorns	150089161	hat_thorns.gif	hat	t,d	23	crowns of thorns
 11916	stylish bycocket	574057027	hat_bycocket.gif	hat	t,d	23
+11917	Thwaitgold mat hatterpillar statuette	270247826	thwaithatter.gif	none		0
```
jaadams5, line 11945: I see that "Thwaitgold mat hatterpillar" agrees with the wiki but I have to wonder if it is a typo and "mad" or "hat" is intended.

jaadams5 (approved):
Comment is not blocking although I might suggest checking the wiki or some other source before committing just in case there has been an update.

midgleyc:
I suspect it is a typo, but that's how it is for the moment.

Also I added it to the wiki, so it would be consistent with this.

gausie:
Someone brought it up in /dev and I think Jick is changing it as we speak

PR #2924, opened 2025-05-17
feat: support hats in Hat Trick

midgleyc:
Add a new HATS slot to allow for Hat Trick, which requires multiple hats, to work.

We are fortunate in that the path forbids removing any hats you equip, which means e.g. that modifier code does not have to be extensively rewritten, and can instead just read the user's current hats.

It is likely there are bugs in this implementation, which will be hashed out as folk run the path.

Maximizer has hats explicitly disabled in Hat Trick until I figure out what to do here. Currently it has issues because the hats are 'available' (because they're equipped on a slot), but it's failing on simRetrieve so I need to do some updates there. In time.


gausie (approved):
Fantastic work. Let's get it to the people

midgleyc:
For maximizer: I think, as we can't unequip the hats, not equipping any hats is the right move here (e.g. if you auto-equip crown of thorns for wall of skin, you're messed up for wall of bones).

I should implement something like gelatinous noob absorbs where it runs through the hats and proposes you equip them all, but doesn't do any of that automatically.

PR #2926, opened 2025-05-18
Handle articles other than a or an in Cookbookbat fight text

Malibu-Stacey:
I noticed some CBB quests weren't being parsed e.g. Iiti Kitty has no article in KoL & thus wasn't matching the regex. Pygmy orderlies article is `some`. Cabinet of Dr. Limpieza & Bubblemint Twins article is `the`
This appears to fix it.
I don't know if this is good regex for this or not because I can't stand regex.


gausie:
This doesn't look like it would work for Iiti Kiti. Maybe you could you provide some HTML for a test to confirm.

midgleyc:
CrystalBallManager.java uses `Pattern.compile("<li> +(?:an?|the|some)? ?(.*? in .*?)</li>");`.

Malibu-Stacey:
> This doesn't look like it would work for Iiti Kiti. Maybe you could you provide some HTML for a test to confirm.

According to my logs, it worked perfectly fine.

Malibu-Stacey:
> CrystalBallManager.java uses `Pattern.compile("<li> +(?:an?|the|some)? ?(.*? in .*?)</li>");`.

Thanks, I'll update this to be the same. The only difference between my change and that is using the wildcard for a & an together.

jaadams5 (approved):

gausie:
> > This doesn't look like it would work for Iiti Kiti. Maybe you could you provide some HTML for a test to confirm.
> 
> According to my logs, it worked perfectly fine.

Ah I didn't realise the double space was in the source too. Makes sense.

PR #2931, opened 2025-05-21
feat: update nohookah / notcrs effects, php random methods

midgleyc:
Extract status effect updates and extract PHP random methods from #1767.

These are useful to have in main.


gausie (approved):
LGTM. I will rebase that branch from this. We'll crack it eventually.

PR #2935, opened 2025-05-30
correct names for cola battlefield zones

ianknowles:
correct names for cola battlefield zones


ianknowles:
Checked via last adventure link. Will recheck with autumnaton tomorrow

ianknowles:
Location Cola Wars Battlefield (Cloaca Uniform) with id 86 is known to Mafia as Battlefield (Cloaca Uniform)
Location Cola Wars Battlefield (Dyspepsi Uniform) with id 87 is known to Mafia as Battlefield (Dyspepsi Uniform)
Location The Cola Wars Battlefield with id 85 is known to Mafia as Battlefield (No Uniform)

jaadams5:
Could we possible let @Veracity0 comment on this?

I seem to recall trying to do something similar and being told the mismatch was both deliberate and necessary.  I may be remembering something that didn't happen or something that does not apply to these locations but I would prefer to know @Veracity0 's opinion before I approve or commit.

jaadams5 (approved):

PR #2941, opened 2025-06-05
Functions for predicting Beret Busking effects

gausie:
- **Track limits for beret skills**
- **Track total power**
- **First bash at beret_busking_effects command**

`int [effect] beret_busking_effects()` reports what you would get if you cast it right now
`int [effect] beret_busking_effects(int power, int castsUsed)` reports what you might get with the data provided. Note that the second param is 0-indexed.

The return value if `effect => turns of effect`, with an extra `$effect[none] => meat gain`


src/net/sourceforge/kolmafia/KoLCharacter.java
```diff
@@ -5731,6 +5731,25 @@ public static void addItemAdjustment(
...
   }
 
+  private static final AdventureResult HAMMERTIME = EffectPool.get(EffectPool.HAMMERTIME);
+
+  public static int getTotalPower() {
+    // These are the *extra* power supplied, so we can do the maths easier
+    var tao = KoLCharacter.hasSkill(SkillPool.TAO_OF_THE_TERRAPIN) ? 1 : 0;
+    var hammertime = KoLConstants.activeEffects.contains(HAMMERTIME) ? 3 : 0;
+
+    var hat =
```
midgleyc: By Semenar:

```
However, in Hat Trick, if you have Tao of the Terrapin, the strongest hat counts triple instead of double.
```
so I think the hat trick power wants adjusting (unless this has changed since)

gausie: Oh! How weird

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -11087,6 +11100,55 @@ public static Value voting_booth_initiatives(
...
+  public static Value beret_busking_effects(
+      ScriptRuntime controller, final Value power, final Value cast) {
+    MapValue value = new MapValue(DataTypes.EFFECT_TO_INT_TYPE);
+
+    // Calculate the capped power
+    var cappedPower =
+        Math.min(power.contentLong, 1100)
+            + (long) Math.floor(Math.pow(Math.max(0, power.contentLong - 1100), 0.8));
+
+    // $effect[none] will indicate the meat gained
```
midgleyc, line 11118: I think this is an inherently confusing way for this to work. Meat gain should probably be another function, or this should return an object, or something other than "the return value is effects mapped to duration, except that if the effect is none it's meat gained instead".

gausie: It's a pretty normal pattern in mafia to have meat occupy the [none] spot. This pretty much directly reflects AdventureResults where the count would be quantity of meat or turns of an effect. I would strongly advocate for keeping it

src/net/sourceforge/kolmafia/KoLCharacter.java
gausie: CDM is fixing this

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
gausie: I have aligned the implementation of this with other places this happens, so it now uses AdventureResult to collect the expected results

PR #2948, opened 2025-06-10
fix: unequip a 2-handed weapon when force-equipping an offhand

midgleyc:


gausie (dismissed):
This is a nice place to have a test, but it's optional.

gausie (approved):
⛹

PR #2951, opened 2025-06-11
fix: make maximizing stinky cheese actually work

midgleyc:
Follow-up to #2950; my test passed for unrelated reasons (maximizer will always try to equip something unless you specify `-tie`).


gausie (approved):
Ahha! Nice

PR #1724, opened 2023-05-24
Sot's Parcel Location

jaadams5:
A sense of deja vu is expected.

The code that extracts the location string is believed to be correct.  The string is a KoL name for a unique location.  In some cases the mapping from the KoL string to a unique KoLmafia location depends upon conditions.

The preference will now be a KoLmafia location string and not what was extracted.  If KoL string has an exact match to a KoLmafia location then that will be used on the assumption that KoL already knows it is a place that can be adventured in.  Otherwise this PR will attempt to map the KoL string to a unique adventerable location and return that if successful.

Right now there is only data for mapping "The Hippy Camp" and "The Frat House".  Furthermore that is embedded in the code and as such feels inelegant.

Tests are only for The Hippy Camp.  If this PR looks like it is going to survive then tests for The Frat House could be added.

Opening as Draft for feedback.

PR #2954, opened 2025-06-12
fix: names for base hippy camp, frat house, video game dungeons

midgleyc:
Knowing the base game names for locations is useful for the Cookbookbat quest and the Sot's quest.

The Wartime versions (disguised and undisguised) have the same name as the undisguised standard versions, which is unhelpful, so these are left as they are.


midgleyc:
Okay, ref #1724, 
* The Hippy Camp can mean the base one or either wartime
* The Orcish Frat House (/not/ The Frat House, which doesn't exist) can mean the base one or either wartime

I will fix up the implementation.

midgleyc:
Ref #2695.

PR #2955, opened 2025-06-13
Fix Sleazy Back Alley combat rate

Semenar:
Wiki says 80, LAR data confirms it is definitely above 50, Phill did not hit any NCs when running capped +combat (suggesting it is at least 65). Changing to 80 to coincide with the Wiki but could spade it if needed.


gausie (approved):
I am inclined to agree

PR #2956, opened 2025-06-14
Perform fuzzy matching for hermit CLI command.

heeheehee-kolmafia:
I had tried using ItemFinder.getFirstMatchingItem(), but that apparently tries to find matching items, independent of the values in the provided source list.

Note that I removed 11-leaf clover from being added to hermitItems with a count of -1 -- nothing seems to rely on this existing. However, the presence of this entry was causing cloverCount() to return -1, which was less than the 0 clovers needed.


heeheehee-kolmafia:
Context: https://kolmafia.us/threads/hermit-command-cannot-grab-jabanero-peppers.30623/

heeheehee-kolmafia:
And, for completeness, since I hadn't checked this explicitly:

```
> hermit? jabanero

Visiting Hermit...
jabañero pepper
Requests complete.

> hermit jabanero

Visiting the Hermit...
You acquire an item: jabañero pepper
Hermit successfully looted!
```

jaadams5 (approved):

PR #2957, opened 2025-06-17
advance quests by defeating bosses, not by parsing dropped items

Veracity0:
If dropped items are quest items, that works, but the Crown of the Goblin King is not, for example.
Nor is goblin water, but we had a kludge to check if it was from the Aquagoblin.

Checking for end of island war also checked only for The Man (and one path variant) and The Big Wisniewski (and one path variant).

MonsterData MonsterDatabase.getQuestBoss(MonsterData) - given a path variant boss - will return the "base" boss and end-of-fight processing will do quest management using that, rather than hardcoded lists of path variants.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/ResultProcessorTest.java
+++ b/test/net/sourceforge/kolmafia/session/ResultProcessorTest.java
@@ -147,7 +147,7 @@ public class ResultProcessorTest {
         // That does  advance the quest
         assertTrue(
             QuestDatabase.isQuestFinished(QuestDatabase.Quest.GOBLIN),
-            "Defeating Acquagoblin finishes the L05 Quest");
+            "Defeating Aquagoblin finishes the L05 Quest");
       }
     }
 


test/net/sourceforge/kolmafia/session/ResultProcessorTest.java
```diff
@@ -121,27 +123,58 @@ public void gettingCosmicBowlingBallInCombatResetsReturnCombats() {
...
+        assertFalse(
+            QuestDatabase.isQuestFinished(QuestDatabase.Quest.GOBLIN),
+            "Getting Goblin Water does not finish L05 Quest");
+
+        // Handle fight completion
+        QuestManager.updateQuestData(acquisition, testMonster);
+        // That does  advance the quest
+        assertTrue(
+            QuestDatabase.isQuestFinished(QuestDatabase.Quest.GOBLIN),
+            "Defeating Acquagoblin finishes the L05 Quest");
```
heeheehee-kolmafia: typo? (no "c")

heeheehee-kolmafia (approved):
Looks reasonable, just one small spelling nitpick.

Feel free to merge after adding that change.

test/net/sourceforge/kolmafia/session/ResultProcessorTest.java
Veracity0: Done.

PR #2961, opened 2025-06-30
no McHuge mods for Noobs

ckb11:
https://kolmafia.us/threads/gelatinous-noob-modifiers.30637/

Mafia adds McHugeLarge bonuses as 'Outfit' bonuses, but they are not actually outfits in KoL because those bonuses do not apply from that equipment in GN.  This removes McHugeLarge bonuses in GN.
Also, tweak description of the CLI 'absorb' command

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLCharacter.java
+++ b/src/net/sourceforge/kolmafia/KoLCharacter.java
@@ -6007,6 +6007,9 @@ public abstract class KoLCharacter {
   }
 
   private static int getMcHugeLargeLevel(Modifiers mods) {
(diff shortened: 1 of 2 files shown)


ckb11:
I do not understand enough of Mafia to fully know how the McHugeLarge pseudo-outfit bonus works, but this fix seems to do what I want it to in Noobcore.

midgleyc (approved):
Looks good to me!

It's an outfit bonus because that way Maximizer works without having to add additional code to handle McHugeLarge specifically. It can't be a Synergy because of the slightly weird way the level works.

PR #2962, opened 2025-07-01
feat: skeleton cooler yeti

midgleyc:
Item drop has no familiar message so not sure if that works


Ignose:
Cold res is fixed now per https://discord.com/channels/466605739838930955/474736566246572042/1389653901606785024

PR #2969, opened 2025-07-18
Allowing for arbitrarily renaming "/cli" chat command

yojimboslaw:
Jimmyking asked earlier today if there was a way to shorten that command to something like "/l", but it seemed like there were only three hardcoded commands to use ("/do", "/run", and "/cli"), so I tried to make that possible for people to do without having to compile their own build.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1434,6 +1434,7 @@ user	recentLocations	5
 user	recoveryScript
 user	redSnapperPhylum
 user	redSnapperProgress	0	roa
+user	relayChatCLITrigger	cli
 user	relayCounters
 user	relayPort	0
 user	relayShowSpoilers	false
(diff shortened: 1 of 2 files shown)


yojimboslaw:
So I've run that command and it seems to have generated a bunch of ancillary files and didn't actually change anything in the files that I've formatted incorrectly. I assume those new files just get applied in the compiling process or something along those lines? I was sort of hoping there would be a procedural way to amend the formatting in the source code without expanding this minor project beyond the two files I've edited, since this is my first time using any of these tools.

Accommodating the default behavior would be really easy to do, since it's just a matter of adding in a fourth boolean in each of the two lines I've edited, so sure I can make that change. I was just assuming anyone who would actually change their `/cli` command wouldn't want to also use the `/cli` command.

And yeah, you're right that "/l" isn't the best idea, but it was the example I was given for what they wanted to do.

PR #2970, opened 2025-07-18
Introduce pref deprecation mechanism, make new Kokomo resort summon counter pref, deprecate old kokomo pref

gausie:
- **Add the concept of deprecation for preferences**
- **Deprecate old kokomo resort pass pref and introduce new one that counts up to the skill level**

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -1435,6 +1435,7 @@ user	recentLocations	5
 user	recoveryScript
 user	redSnapperPhylum
 user	redSnapperProgress	0	roa
+user	relayChatCLITrigger	cli
 user	relayCounters
 user	relayPort	0
 user	relayShowSpoilers	false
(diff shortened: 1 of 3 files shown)


Veracity0:
I've thought about this is the past, both in the context of KoLmafia itself and in scripts with custom preferences.
There is an additional stage in this, which is not addressed in this PR - and probably need not be - which, perhaps, you could comment on.

At what point can we remove the deprecated preference? In other words, declare it completely obsolete?

My scripts have used an "upgrade properties" function which will look for deprecated properties, generate equivalent replacements - and remove the deprecated one. That last step is key, if you care about not having dead properties in your pref files.

KoLmafia could have something like that; when we remove a property from defaults.txt, also run code to remove that specific property from the prefs file. It has the wrinkle that scripts might go for years without ever updating a deprecated property. For that matter, when KoLmafia removes a property, a script could re-use the name.

So this may not even be possible for us.

I'm going to approve this, because it looks good for what it does, but, I'd be interesting in your thoughts on a full active -> deprecated -> obsolete (and removed) life cycle for properties.

Veracity0 (approved):

gausie:
This also might end up being too chatty. We can just fix forward if so.

Veracity0:
That's fine. This is good for what it does, so, let's get it in and decide how we want to proceed - eventually - to actually removing obsolete properties.

PR #2994, opened 2025-07-30
feat: make radios in breakfast, if preference is set

midgleyc:
You might not want to do this because you can't get grey text from radios.


jaadams5 (approved):

jaadams5:
Out of scope and I see you are just using the existing idiom but appending breakfastType to make a preference string would be more robust if breakfastType were an enum or something more obviously intended to be a constant.

Does this bother me enough to fix it myself?  Probably not but I did want to understand what breakfastType could be before I understood the new code.

PR #2997, opened 2025-07-31
chore: prefer newer choice switches in postChoice1 and postChoice2

midgleyc:
Ref #2643.

Noticed that for the last few additions I was adding a block, so we may as well always have a block. Fall-through was only used in two places.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ChoiceControl.java
+++ b/src/net/sourceforge/kolmafia/session/ChoiceControl.java
@@ -1929,20 +1929,18 @@ public abstract class ChoiceControl {
 
       case 852:
         // Shen Copperhead, Jerk
-        // Deliberate fallthrough
       case 853:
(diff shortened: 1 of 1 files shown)


Veracity0 (changes requested):
This was a lot of work.
Rather than looking through the diff, I examined the file and looked at postChoice1 and postChoice2.

I notice that in postChoice1, most choices used to have a break, and you added a block and removed the break.
There were quite a few choices that had a return, and you added a block and kept the return.

Considering that there is nothing in postChoice1 following the big switch statement, there was no functional difference between ending a case with a return vs. a break.

Can I suggest that you get rid of all the end-of-case return statements in the postChoice1 switch statement?

midgleyc:
Right, in the middle I wasn't certain there was nothing after the switch. Removed the early returns.

PR #3016, opened 2025-08-08
Updates to Fishy and beret busking pool

gausie:


midgleyc (approved):
Context: https://discord.com/channels/466605739838930955/581155825385472001/1403143988562755605

gausie:
> * It was noticed that busks were shuffled
> * The effect removed was found to be fishy (fishy is no longer wishable)
> * Fishy was added back into Busks (but only gives 1 turn)
> * Fishy was probably removed because of the upcoming path, which is clearly lake related
- @Shiverwarp

PR #3017, opened 2025-08-09
feat: log pokefam fights better

midgleyc:
MONSTERID is not present, log the enemy team, log moves again.

Logging the enemy team gives more evidence towards determining tags for the April Fools familiars. For example, it looks like hippies have pokefams that are "vegetable" or "food" (or possibly "edible" because the crossover is huge) or their element, almost always "stench".


Veracity0 (approved):
I really enjoyed Pokefam, although I haven't done it in years. Perhaps I should give it another shot after you've finished souping up support. 😀

PR #3018, opened 2025-08-10
Add [turn] numbers to Trick-or-Treating

ckb11:
This adds a [turn] message to Trick-or-Treating, which is helpful for log parsing and parsing of fight Encounters that happen while trick-or-treating.


ckb11:
Does something more need to be done to get this finalized?
(I don't understand the 'spotless' comment)

midgleyc:
The checks are failing due to spotless (required formatter). You need to run
```
./gradlew :spotlessApply
```
from the root of the repository.

ckb11:
Thank you - I don't know what that did, but checks passed!  Yea for obscure coding things!

PR #3021, opened 2025-08-11
Update to gradle 9

gausie:


midgleyc (changes requested):
I'm getting SHA conflicts on the wrapper jar, need to figure out what's happening here

so for context on this, when I run `./gradlew wrapper --gradle-version=9.0.0 && ./gradlew wrapper` I get a different jar file downloaded

the page https://gradle.org/release-checksums/ says the sha256 checksum should be `76805e32c009c0cf0dd5d206bddc9fb22ea42e84db904b764f3047de095493f3`. I don't get that for either this jar, or the jar that gets downloaded for me.

midgleyc (approved):
okay, must have been a mistake on my end

PR #3031, opened 2025-08-13
chore: fix clean tasks after migration

midgleyc:
Just rewrite them in Kotlin.

`outputs.upToDateWhen { true }` prevented them from running, but even after that was fixed they still didn't work. In lieu of finding out why I just changed them until they did.


gausie (approved):
Very nice

midgleyc:
Yes, whichever is picked will obsolete the other.

PR #3034, opened 2025-08-14
Revert "chore: fix clean tasks after migration"

heeheehee-kolmafia:
Reverts kolmafia/kolmafia#3031

This seems to have broken the build under macos:
```
> Could not create task ':pruneDist'.
   > listFiles(...) must not be null
```


midgleyc:
I can fix this instead: a null check followed by an empty list should do it.

PR #3038, opened 2025-08-14
feat: add mobius ring mods

midgleyc:


gausie (approved):
Embrace addition of new modifier expression variables!

PR #3043, opened 2025-08-15
fix: remove zone sea penalty, just have locations

midgleyc:
Also fix the Mer-kin zone penalties, they are -150% not -200% by the wiki.

Don't have a default penalty, this makes it harder to compare the penalties with the wiki.


gausie:
This is failing because of a dumb test

gausie (approved):

PR #3048, opened 2025-08-16
Sea maps are now permanent

gausie:


midgleyc (dismissed):
This means the preferences aren't quite accurate any more, because we set them to true if we see the zone, but the zone unlock could have come from Little Brother

But there's no way to know whether it did or not.

gausie:
We could even mark them all when seen except your class-aligned one

midgleyc:
Yes, that sounds like a good idea.

jaadams5 (approved):

PR #3049, opened 2025-08-16
Strip duplicates from _perilLocations on write

gausie:
Cool that doing .sorted() and .distinct() is optimised by Java (https://stackoverflow.com/a/43807329)


midgleyc (approved):
cool

PR #3061, opened 2025-08-19
feat: make pruneDist clear appropriately for real this time

midgleyc:
Ref #3031.

So what I think was happening was that `build/revision.txt` was read at configuration time, then the `delete` registered the files with the old revision, then the new revision was created, then the files that existed at the time of the old registration were deleted.

Extract getting the revision to something that runs at configuration time (I sure hope this is how it works) to ensure it's created at the right time.

I sure am confused as to how stuff like `tasks.jar` manages to pick up the right revision though.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -156,12 +156,15 @@ tasks.register<Delete>("pruneDist") {
   onlyIf {
     dist.exists()
   }
-  delete(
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia (approved):

jaadams5:
Worked as expected for me

PR #3077, opened 2025-08-26
feat: more effects that depend on duration

midgleyc:


gausie:
What was the missing "a" breaking?

gausie (approved):

PR #3078, opened 2025-08-26
Update statuseffects.txt with yeshookah

DamianDominoDavis:


gausie:
I think merging this will throw off any functions like the beret one that rely on reading the effect data in order.

We should work out why that function is working if these are indeed hookahable.

midgleyc:
Anything that was `good nohookah` needs to become `good notcrs`.

I think the intrinsics are probably either blacklisted or coded such that their `code` contains the word `duration` and therefore match like the effects that vary based on turns left.

DamianDominoDavis:
> I think the intrinsics are probably either blacklisted or coded such that their `code` contains the word `duration` and therefore match like the effects that vary based on turns left.

I had someone who didn't know Flavour wish for Spirit of Garlic. You can get (20) of it.

midgleyc (approved):
Thanks!

PR #3079, opened 2025-08-28
Fix Zootomist Level 1 council text handling

RESPRiT:
## Bug location

https://github.com/kolmafia/kolmafia/blob/16eb54d2761ea8ec415448b9572075a184b2bf89/src/net/sourceforge/kolmafia/persistence/QuestDatabase.java#L1049

The logic of `handleCouncilText()` is:

- Parse the council message HTML, break it up by paragraph (`<p>`), and then, for each paragraph:
  - Go through "questscouncil.txt" and, for each quest message:
    - Check if any paragraph of the council message HTML _contains_ any paragraph of a questcouncil.txt quest message
    - If so, update the quest state by setting respective preferences.

## The bug

https://github.com/kolmafia/kolmafia/blob/16eb54d2761ea8ec415448b9572075a184b2bf89/src/data/questscouncil.txt#L498

The Level 2 quest message of Z is for Zootomist contains two (identical) very sneaky, very short paragraphs - `...`:

```
Ah, you're back. And you've brought one of your familiars with you! How cute! ...Wait, is it... have you...
<p>
...   <- here, and...
<p>
...   <- here!
<p>
...Okay, well. Well, now that we've established what your deal is -- and it's quite a deal, yes, that's really a whole thing, isn't it -- I can definitely say that this first task I've got for you is right up your alley. I need you to go to the Spooky Forest, and bring us back a mosquito larva. Please don't come any closer, just hold up your map and I'll point to where it is. 
```

At Level 1, [two council messages](kol.coldfront.net/thekolwiki/index.php/The_Council_of_Loathing/Z_is_for_Zootomist) have paragraphs containing `...`:

- Initial text:
> Look, why don't you just head over to the Toot Oriole's place, see if he's left a letter for you. While you're gone I'll see what sort of quests we've got for an adventurer of your**...** whatever it is.

- And, after reading King Ralph's Letter:
> Hi, okay, well, I'm still looking for suitable quests for an adventurer of your**...** let's say, stature. We really don't get a lot of mad scientists or extremely unlicensed medical practitioners coming through here. If you could, you know, bulk up a little? That would help a lot.

This causes these two messages to set the property `questL02Larva` to `started` upon visiting the council URL.

## Main consequence

https://github.com/kolmafia/kolmafia/blob/16eb54d2761ea8ec415448b9572075a184b2bf89/src/net/sourceforge/kolmafia/KoLAdventure.java#L551

The `questL02Larva` property is used to check if the woods are open in `KoLAdventure.java`, which causes `can_adventure()` to erroneously return `true` at Level 1 for Distant Woods zones (which, in turn, breaks some of autoscend's logic).

## The fix

A few solutions come to mind:

1. The simplest fix with the fewest side effects would be to just remove the offending paragraphs from `questscouncil.txt`.
2. A more robust fix could be to change the `.contains()` call to `.equals()` - but, `questscouncil.txt` does not contain exact text for all quest messages (e.g. Exploathing `canParseInitialCouncilPage` test fails with this change), and would need to be overhauled/checked for each path; also, `handleCouncilText()` would need to account for more than just paragraphs since elements such as `<center>` can be present in council text.
3. Some midway fix that doesn't do an exact match, but uses a slightly stricter heuristic.

I chose the first option for this PR, because it's the least amount of effort.

(A comment could also be added to the top of `questscouncil.txt` that warns about removing similar types of paragraphs from council text for future additions.)

## A test

For posterity, I also added a test case against the HTML of the post-King Ralph Letter Zootomist Level 1council text.


midgleyc (approved):
Thanks!

PR #3080, opened 2025-08-28
fix: hippy camp is not accessible at verge of war

midgleyc:
This would be a behavioural change.

Currently mafia treats `can_adventure($location[hippy camp])` as "if I try to adventure at the hippy camp, will it work", and it will work, because adventuring at the hippy camp redirects you to the verge of war version. But scripts want to treat it as "if I try to adventure at the hippy camp, will I see the monsters from `appearance_rates`, like the dirty hippy", and they won't, because it's a different location.


jaadams5 (approved):

gausie:
Agree that it is a behaviour change, but also agree that it is a necessary one.

PR #3083, opened 2025-08-29
clear item monster when retrieve adventure.php

Veracity0:
Before this, using a dolphin whistle, fighting the item monster, and then adventuring and getting a non-combat didn't log correctly.

Using an item to get a fight or choice logs the location and sets itemMonster.
Adventuring again and redirecting to a fight or choice clears itemMonster in checkItemRedirection or checkChoiceRedirection.
Getting an NC does not redirect and we didn't log the location, since itemMonster non-null indicated it was already logged.
Getting a response from adventure.php is an NC. Clear itemMonster in that case.


midgleyc (approved):
nice, this is also the mimic egg -> NC log problem

PR #3089, opened 2025-08-31
change good+notcrs effects to neutral+none if you can't get them from the hookah

nasurte:
also remove notcrs from effects marked neutral, also mark some of the new effects good


nasurte:
Yeah, this was mostly the shield (~28k shield effect procs vs ~500 hookah hits), though both of them seem to have the same criteria, as you can get Blue Swayed from both and most +ML/+/-combat buffs from neither.

PR #3096, opened 2025-09-01
feat: monodent skeleton

midgleyc:


midgleyc:
"skeleton", for me, means that I dump the items / monsters / effects / whatever into the data files, and don't implement any of the logic that would be associated with that.

e.g. we need some monster replacement logic, conditional skills, counting construct kills, modifier differences etc etc.

PR #3097, opened 2025-09-01
feat: sea path pull blacklist

midgleyc:


Veracity0:
Holy shit. This is why I made a point of finishing my 11th Normal run in The Sea on Saturday. :)

PR #3100, opened 2025-09-01
Defeating Seaceress increments seaPoints, not freeing king

Veracity0:


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -4348,6 +4349,18 @@ else if (responseText.contains("a horrible grinning clown head emerges")) {
...
+        case "Nautical Seaceress" -> {
+          // You feel your dolphin whistling endurance improve.
+          Path path = Path.UNDER_THE_SEA;
+          if (responseText.contains("dolphin whistling endurance")) {
+            int points = KoLCharacter.isHardcore() ? 2 : 1;
+            path.incrementPoints(points);
+          } else {
+            // I don't know if the message is omitted if you are at max points,
+            // but it seems logical
+            path.setPoints(11);
```
midgleyc, line 4365: if you don't have a dolphin whistle at all, you receive one and don't receive the message

Veracity0: Thanks! It would be nice to get responseTexts for getting the durable dolphin whistle the first time and for having one that is maxed out, but I will accommodate this.
I'm done with this path, otherwise I could test the second, at least.

Veracity0: OK. I fabricated two additional responseTexts:

- one with no message about "improving your dolphin experience" but in which you receive a dolphin whistle
- one with neither a message about "improving your dolphin experience" or receiving a dolphin whistle

It would be nice to get actual responseTexts for those situations, but if KoL actually behaves as expected, we will behave sensibly.

midgleyc (approved):

Issue #3092, opened 2025-08-31
Flagellate flagon pref


bosh-code:
A new preference should be added to track if the user has used their [Flagellate flagon](https://kol.coldfront.net/thekolwiki/index.php/Flagellate_flagon) today. It's not destroyed on use and the stat gain can be decent. I'll probably look to add this if no one is opposed.

The reason I want a preference is to remind myself in TourGuide or Yorick. Also maybe to use in garbo.

PR #3102, opened 2025-09-02
feat: add flagellate flagon pref

bosh-code:
Adds the pref as detailed here: https://github.com/kolmafia/kolmafia/issues/3092

Also added it a Daily Deed

### DIFF
--- a/src/data/monsters.txt
+++ b/src/data/monsters.txt
@@ -1682,7 +1682,7 @@ spider-goblin conspirator	1285	gourd_spidergob.gif	NOCOPY Atk: 32 Def: 26 HP: 35
 tin spider conspirator	1284	gourd_canspider.gif	NOCOPY Atk: 28 Def: 28 HP: 35 Init: 50 Meat: 20 P: construct Article: a	razor-sharp can lid (0)	spider web (0)	sharp tin strip (0)	wad of spider silk (0)
 Fnord the Unspeakable	1286	gourd_fnord.gif	NOCOPY Atk: 38 Def: 38 HP: 50 Init: 75 P: horror	razor-sharp can lid (0)	razor-sharp can lid (0)	Knob Goblin firecracker (0)	Knob Goblin firecracker (0)	spider web (0)	spider web (0)	sharp tin strip (0)	wad of spider silk (0)	goblin collarbone (0)
 fearsome giant squid	1320	bath_octopus.gif	NOCOPY Atk: 125 Def: 125 HP: 140 Init: 500 P: fish Article: a	old chum (10)
-ferocious roc	1322	bath_pelican.gif	NOCOPY Atk: 120 Def: 130 HP: 140 Init: 500 P: beast Article: a	old chum (0)
+ferocious roc	1322	bath_pelican.gif	NOCOPY Atk: 120 Def: 130 HP: 140 Init: 500 P: beast Article: a
 giant man-eating shark	1321	bath_whale.gif	NOCOPY Atk: 130 Def: 120 HP: 140 Init: 500 P: fish Article: a	old chum (10)
 Bristled Man-O-War	1325	bath_bubble.gif	NOCOPY Atk: 135 Def: 135 HP: 160 Init: 500 P: fish Article: a	old chum (15)
 Deadly Hydra	1326	bath_showerhead.gif	NOCOPY Atk: 140 Def: 140 HP: 180 Init: 750 P: beast Article: the	Hundred Headed IPA (25)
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/request/UseItemRequest.java
```diff
@@ -6292,6 +6292,11 @@ else if (responseText.contains("You can't figure out the angle")) {
       case ItemPool.YAM_BATTERY:
         Preferences.setBoolean("_yamBatteryUsed", true);
         break;
+
+      case ItemPool.FLAGELLATE_FLAGON:
+        Preferences.setBoolean("_flagellateFlagonUsed", true);
+        break;
```
gausie: You need to `return` here if the item is not consumed

gausie:
We should also have a `flagellateFlagonActive` pref that tracks that we have used the flagon but not yet benefitted from it. We'll want to spade whether the charge lasts over rollover. And if it does, whether they stack.

jaadams5:
After the coffee kicked in I realized we need both "has the effect been granted?" and "has the granted effect been used/consumed?"  If we can figure it out it would be nice to set those correctly if use and/or consumption occurred outside of KoLmafia.  

We can could could check for a failure message after use to set one of them.  We could learn something if we drink and don't get the "You pour" message.

gausie:
> After the coffee kicked in I realized we need both "has the effect been granted?" and "has the granted effect been used/consumed?"  If we can figure it out it would be nice to set those correctly if use and/or consumption occurred outside of KoLmafia.  
> 
> 
> 
> We can could could check for a failure message after use to set one of them.  We could learn something if we drink and don't get the "You pour" message.

I already covered this 😉

bosh-code:
Sorry in advance for all the quotes, trying to keep these points organized so I can come back to it.

> I'm fine with the preference. I don't think it should be a Daily Deed. It would make more sense on the consumption panel...

Sounds good to me. I'll remove the deed and add it to the queue panel.

> We should also have a `flagellateFlagonActive` pref that tracks that we have used the flagon but not yet benefited from it.

OK, I'll add that too.

> We'll want to spade whether the charge lasts over rollover. And if it does, whether they stack. And if they stack, whether it's multiple times on a single drink or on successive drinks (I'm sure it'll be the latter)

I'll buy a second one and do the spading tomorrow and update here and the wiki if it needs it.

How are the values for preferences like this usually reconciled if the item was used outside of mafia? Does mafia just assume it hasn't been used and when it sees the failure message, it updates the value? Or is there a check at login for preferences like this?

gausie:
> > We'll want to spade whether the charge lasts over rollover. And if it does, whether they stack. And if they stack, whether it's multiple times on a single drink or on successive drinks (I'm sure it'll be the latter)
> 
> I'll buy a second one and do the spading tomorrow and update here and the wiki if it needs it.

You shouldn't need to buy a second one. Just use the flagon before rollover, and use it after rollover. See if you get the charge twice on one drink, once on two drinks, or just once on one drink.

> How are the values for preferences like this usually reconciled if the item was used outside of mafia? Does mafia just assume it hasn't been used and when it sees the failure message, it updates the value? Or is there a check at login for preferences like this?

There's no check at login for things like this because the data is not exposed without using the item. What you've done works - we just mark it as used regardless of the output. This works because either it successfully worked or it fails from already having been used. Both should result in the pref being set to true.

bosh-code:
> You shouldn't need to buy a second one. Just use the flagon before rollover, and use it after rollover. See if you get the charge twice on one drink, once on two drinks, or just once on one drink.

Using the flagon before rollover, then again after rollover, I got one use text on my first two drinks, "You pour your drink into your flagellate flagon. Pouring it while the handle wiggles is an experience in and of itself." However, my first drink gave me 0 stats, while my second gave me +25x stats.

```
Verifying ingredients for yam martini (2)...
Creating yam martini (2)...
You acquire yam martini (2)
Successfully created yam martini (2)
You gain 0 Strongness
You gain 0 Magicalness
You gain 0 Roguishness
You gain 4 Adventures
You gain 15 Sarcasm
Preference _mimeArmyShotglassUsed changed from false to true
You gain 25 Strongness
You gain 27 Enchantedness
You gain 37 Sarcasm
You gain 5 Adventures
You gain 13 Cheek
You gain 1 Drunkenness
You lose some of an effect: Ode to Booze (-1)
```

So while it lasted over rollover, it didn't actually do anything. It only really works once per day.

jaadams5:
Fixed the tests.  There was a space instead of a tab between the name and value in defaults and it took me two commits to realize and fix that.

jaadams5 (dismissed):
I'll approve but I would hope someone else will look at this and decide it is OK to be pushed.

It is possible the "rollover stacking" will work differently from what is coded and there is always room for tests ;-)

gausie (changes requested):
We need to have:

`_flagellateFlagonUsed`: boolean, default to false, present in daily limits for the flagon itself
`flagellateFlagonsActive` or something: integer, defaulting to zero, reset on ascension, , incremented each time the a flagon is used, decremented each time the flagon charge procs

You seem to have added `_flagonsUsed` as well as the boolean which seems totally useless to me since it is reset daily.

src/net/sourceforge/kolmafia/request/UseItemRequest.java
```diff
@@ -6292,6 +6292,17 @@ else if (responseText.contains("You can't figure out the angle")) {
       case ItemPool.YAM_BATTERY:
         Preferences.setBoolean("_yamBatteryUsed", true);
         break;
+
+      case ItemPool.FLAGELLATE_FLAGON:
+        {
+          // You get your flagellate flagon ready for your next drink.
+          if (responseText.contains("You get your flagellate flagon ready for your next drink.")) {
+            Preferences.setBoolean("_flagellateFlagonUsed", true);
```
gausie: We probably want to set this boolean regardless, since if it fails it means we've also used the flagon

PR #3109, opened 2025-09-04
track Hatch opening

ckb11:
track Hatch opening at The Wreck of the Edgar Fitzsimmons
This adds a property noting the turncount of the last time the hatch was opened


jaadams5 (dismissed):

gausie:
Do we do this turncount pattern elsewhere? Why not just have a boolean that resets on ascension?

gausie:
> You need the turncount pattern because the hatch stays open (monsters in the zone are replaced) for 20 turns since you last opened it.
> 
> Are you sure this should be a daily property?
> 
> Also AreaCombatData might want updating.

Ahh I didn't understand the mechanic. Ok cool. Yeah we want this to reset on ascension but not daily. I'll maybe add some changes to this branch if you don't mind, ckb

PR #3114, opened 2025-09-12
fix: ink bladder does not give effect

midgleyc:
Ref #3112.


jaadams5:
Confusion between the two ink bladders?

PR #3116, opened 2025-09-14
Track Nautical Seaceress defeat in Under the Sea path

ckb11:
Track Nautical Seaceress defeat in Under the Sea path
I don't know if this is the best way to do this, but it seems reasonable from how other FINAL quest tracking is done in other paths.


Veracity0 (changes requested):
This is OK. My response to your forum post indicates some other refactoring I want to do, but doing the monster map and setting Quest.FINAL to FINISHED this way is fine.

How about a test for that? "test_fight_nautical_seaceress_on.html" exists and could be used.

In fact, QuestManagerTest.UnderTheSea nested class uses that html file and the "defeatingNauticalSeaceressIncrementsSeaPoints" test verifies that seaPoints are incremented correctly. That test could also use "withQuestProgress" and verify it progresses.

ckb11:
And from my last run with this change:
[159] Mer-kin Temple (Center Door)
Encounter: The Nautical Seaceress
...
Preference questL13Final changed from unstarted to finished
Preference seaPoints changed from 8 to 9

Veracity0 (approved):
Looks good to me!
Thanks.

Veracity0 (changes requested):
Wait: 

You want:

PR #3117, opened 2025-09-16
Liberating the King is not proof of Desert access

ckb11:
In the Sea Path, you don't get Desert access even after Freeing the King, so remove this from the desertBeachAccessible() check.

Is there another situation where we DO need to set "lastDesertUnlock" when the King is free?  I can't think of one.


Veracity0:
Removing a previously passing test which fails after a code change is almost never the correct solution. :)
It used to check that desertBeachAccessible is true in Kingdom of Exploathing aftercore - but now it fails?

You did not have to do the Meatcar quest in Under the Sea (although you can).
Are there other paths like that? Perhaps Community Service - which never tried, since it looks boring to me.

Perhaps an additional solution would be to:
- identify all paths where the Desert Beach is available without a Meatcar equivalent during the path and carries over after the path. (Perhaps only Kingdom of Exploathing and Ed)
- For those in which you free the King, in liberateKing, call KoLCharacter.setDesertBeachAvailable()
- For paths in which you free the king without being required to open the desert beach, your current patch will suffice.

I dunno. if a test fails, fix the thing that makes it fail, rather than removing the test.

Veracity0:
Here's what I figured out when reading about paths in the Wiki:

1) Paths where you don't have to "open" areas to do standard council quests:
Under the Sea
(I ran all my runs as an AT - free guild access - and Gnoll sign - Meatcar components - so I, sometimes, for no particular reason spent Meat but no turns to build a Meatcar before I freed the king, and for  me, the desert beach remained available.)
Grey Goo
Community Service

2) Paths where you automatically get access to certain areas - including the desert
Kingdom of Exploathing
Ed the Undying.

A better way to do this:

1) You modified KoLCharacter.desertBeachAccessible to remove kingLiberated().
Also remove isEd() and isKingdomOfExplothing(); it will only check for possession of a Meatcar equivalent.
2)  KoLCharacter.setPath sets your ascension path.
In that method, if (path == Path.ACTUALLY_ED_THE_UNDYING || path == Path.KINGDOM_OF_EXPLOATHING), call setDesertBeachAvailable(), since the desert beach is available and remains available even after finish the path, for those two paths.

That should automatically fix the test, since withPath()  calls KoLCharacter.setPath()

ckb11:
I think I did all the things, but the test still seems to fail

ckb11:
Closer maybe... but:
KoLAdventureValidationTest > Exploathing > thatExploathingHasBeachAccess() FAILED

Veracity0:
OK. I checked out the PR and experimented and did this:

1) Moved BeachAccess tests into their own Nested class and made them test specifically desertBeachAccessible(), rather than canAdventure - since access to specific Beach areas depends on more rather than the beach simply being available.
2) withPath(EXPLOATHING) followed by withPath(NONE) seems to have side effects that affect other tests?
I have no idea why. I changed it into setPath(EXPLOATHING) followed by setPath(NONE) and everything works.

I think this is good to go.

Veracity0 (approved):

Veracity0:
> > 2. withPath(EXPLOATHING) followed by withPath(NONE) seems to have side effects that affect other tests?
> >    I have no idea why. I changed it into setPath(EXPLOATHING) followed by setPath(NONE) and everything works.
> 
> Is this worth exploring? I am amused by tracking down and fixing test side effects. Is the side effect consistent or is it possibly an intermittent test failure that is most likely caused by tests running on multiple threads and the cleanup of shared spaces by one test destroys the initialization for another?

The tests that failed were in a different Nested test class, so probably executed in parallel.
Considering that we have:
```
  @BeforeEach
  public void beforeEach() {
    KoLCharacter.reset("KoLAdventure");
    Preferences.reset("KoLAdventure");
    KoLConstants.inventory.clear();
    EquipmentManager.resetEquipment();
  }
```
I don't really understand why there can be test leakage - unless "parallel" test executions that modify the same static globals are sharing data in memory?

Which would make all of our test structure which uses Nested test classes pretty brittle.

withPath(Path.KINGDOM_OF_EXPLOATHING) will save the current path and set it to KINGDOM_OF_EXPLOATHING.
when the "try" exits and the Cleanup closes, it restores the old path.
So: NONE -> KINGDOM_OF_EXPLOATHING -> NONE

withPath(Path.NONE) behaves similarly.
So: KINGDOM_OF_EXPLOATHING -> NONE -> KINGDOM_OF_EXPLOATHING

If you chain those:
(initial)NONE
(1st cleanup)->KINGDOM_OF_EXPLOATHING
(2nd cleanup)->NONE
within the body of the try.

When the try exits, if the cleanups close in the same order:
(initial)NONE
(first cleanup closes)->NONE
(2nd cleanup closes)->KINGDOM_OF_EXPLOATHING

Which is not correct.

Simply doing 2 setPaths in the try body:
(initial)NONE
(1st setPath)->KINGDOM_OF_EXPLOATHING
(2nd setPath)->NONE

which is as desired, with no Cleanups needed.

The side effect is that the lastDesertUnlock property will be changed - but we have a withProperty() Cleanup for that.

PR #3121, opened 2025-09-20
Pick some line

jaadams5:
Implemented suggestions from IntelliJ's analyze command.

Swapped parameter order on some test assertions (so that error message will be correct).

Adjusted some static and final modifiers.

Replaced get(0) with getFirst() in some iterations.

Did NOT remove any code flagged as unused


Veracity0:
> Replaced get(0) with getFirst() in some iterations.

Here is our convention:
```
          assertThat(requests, hasSize(1));
          assertPostRequest(requests.get(0), "/place.php", "whichplace=mountains");
```
and
```
        assertThat(requests, hasSize(2));
        assertPostRequest(requests.get(0), "/place.php", "whichplace=manor2&action=manor2_ladys");
        assertPostRequest(requests.get(1), "/api.php", "what=status&for=KoLmafia");
```
It appears that IntelliJ wants to change get(0) in the first case to getFirst().
I don't think it wants to do that in the second case, thank goodness; getFirst() followed by get(1) would be jarring.

But, I really don't see the point of having a different convention for lists with a single element vs. lists with more than one element.

Veracity0 (approved):

PR #3122, opened 2025-09-22
fix leave choice number for Interstellar Trade

ckb11:


ckb11:
choiceAdventure1243
Interstellar Trade
1: Buy the thing
2: Don't buy the thing

PR #3130, opened 2025-10-04
Match exact item names in zap groups

gausie:


jaadams5 (approved):

jaadams5:
What was observed to be broken and made you decide this had to be corrected?

gausie:
Good question! I have a project called Data of Loathing that exposes a GraphQL API of mafia data (but relationalised) and this issue becomes clear when you try to match up names to ids

You can play with it here https://data.loathers.net/graphiql

Issue #3093, opened 2025-08-31
Hypercubed Pref


bosh-code:
A new preference should be added to track if the user has gotten their [Hypercubed](https://kol.coldfront.net/thekolwiki/index.php/Hypercubed) today. It's only granted on the first rest in a [Residence-Cube](https://kol.coldfront.net/thekolwiki/index.php/Residence-Cube) each day.

TourGuide (and maybe Yorick, not sure) has a line about it in the free rests, but it would be nice to hide it once the player has gotten Hypercubed that day:

<img width="473" height="159" alt="Image" src="https://github.com/user-attachments/assets/67a557e8-9c0e-4449-b68c-0909eeb4827e" />

I'm happy to try and add it if no one is opposed.

gausie:
Do you get it once per day or once per hypercube per day? i.e. if you change dwelling from and then back to the hypercube can you get it a second time? I assume not.

bosh-code:
> Do you get it once per day or once per hypercube per day? i.e. if you change dwelling from and then back to the hypercube can you get it a second time? I assume not.

I just tried it and no it didn't give me Hypercubed a second time

PR #3135, opened 2025-10-17
add and track rest effect prefs

bosh-code:
For issue #3093

Added prefs to track when an effect is gained by resting.

Skipped the pilgrim hat as its random.
Wasn't sure what to do with the mushroom house as there is no special text when the get the effect vs not, so I just assumed the first rest will give the effect.


jaadams5:
> all `preferences` should be `Preferences`
> 
 Just because I read that and did not immediately realize the significance, that change needs to be made in the code and is the reason it failed to compile and test.

gausie:
@bosh-code this has gone almost two months with no update. We are still interested in the contribution, but I'll close the PR for now to reduce noise. Please feel free to reopen once you have it working!

PR #3138, opened 2025-10-19
Clear error state if falling back to replica deck.

heeheehee-kolmafia:
retrieveItem sets Mafia's state to ERROR when it fails. However, in this case, we expect the failure and wish to fall back to an alternative.

Context: https://kolmafia.us/threads/31406/


jaadams5:
> Closes #3136.

Any reason why this is a better solution?  Since I have not pushed my test yet it is arguably a better PR but since I was implementing your suggestion...

PR #3139, opened 2025-10-20
Upgrade dependencies, gradlew.

heeheehee-kolmafia:
Per dependencyUpdates, we have the following diffs remaining:

```
The following dependencies have later milestone versions:
 - com.diffplug.spotless:com.diffplug.spotless.gradle.plugin [6.23.3 -> 8.0.0]
 - com.formdev:flatlaf [1.6.5 -> 3.6.2]
     https://github.com/JFormDesigner/FlatLaf
 - com.formdev:flatlaf-intellij-themes [1.6.5 -> 3.6.2]
     https://github.com/JFormDesigner/FlatLaf
 - com.formdev:flatlaf-swingx [1.6.5 -> 3.6.2]
     https://github.com/JFormDesigner/FlatLaf
 - org.junit.jupiter:junit-jupiter [5.11.0 -> 6.0.0]
     https://junit.org/
 - org.junit.jupiter:junit-jupiter-api [5.11.0 -> 6.0.0]
     https://junit.org/
 - org.junit.jupiter:junit-jupiter-params [5.11.0 -> 6.0.0]
     https://junit.org/
 - org.junit.platform:junit-platform-launcher [1.11.0 -> 6.0.0]
     https://junit.org/
 - org.slf4j:slf4j-nop [2.0.17 -> 2.1.0-alpha1]
     http://www.slf4j.org
 - org.swinglabs:swingx [1.0 -> 1.6.1]
     http://www.swinglabs.org/

Gradle release-candidate updates:
 - Gradle: [9.1.0 -> 9.2.0-rc-2]
```

spotless 7.0+ introduces quite a few formatting diffs. If we choose to accept these, that should be done in a standalone PR. It's not obvious that (for instance) the treatment of switch expressions is desirable with respect to continuations, but in general newer spotless versions will have better support for newer Java features.

flatlaf, swingx continue to require legacy versions, as was the case in https://github.com/kolmafia/kolmafia/pull/2413.

JUnit 6.0 seems to introduce a number of test breakages, possibly due to changes in how the tests are run. This warrants further investigation.

Gradle, slf4j-nop have been upgraded to the latest stable (non-alpha, non-release-candidate) version.

Nominal motivation is https://kolmafia.us/threads/31425.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -72,14 +72,10 @@ dependencies {
 
   testImplementation("org.hamcrest:hamcrest:3.0")
   testImplementation("com.spotify:hamcrest-optional:1.3.2")
-  testImplementation("org.junit.jupiter:junit-jupiter-api:5.11.0")
-  testImplementation("org.junit.jupiter:junit-jupiter-params:5.11.0")
-  testImplementation("org.mockito:mockito-core:5.20.0")
-  testImplementation("org.eclipse.xtext:org.eclipse.xtext.xbase.lib:2.41.0.M1") {
   testImplementation("org.junit.jupiter:junit-jupiter-api:5.14.0")
   testImplementation("org.junit.jupiter:junit-jupiter-params:5.14.0")
-  testImplementation("org.mockito:mockito-core:5.13.0")
-  testImplementation("org.eclipse.xtext:org.eclipse.xtext.xbase.lib:2.36.0") {
+  testImplementation("org.mockito:mockito-core:5.20.0")
+  testImplementation("org.eclipse.xtext:org.eclipse.xtext.xbase.lib:2.41.0.M1") {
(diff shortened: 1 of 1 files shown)


heeheehee-kolmafia:
I'm probably thinking of this particular report from our first attempts at using Gradle: https://kolmafia.us/threads/migrating-away-from-ant-gradle-or-other.26249/page-2#post-163694

That said, this is a big enough jump (two major versions over the past 4 years) that I think we'd want to push it into its own PR as well. Based on a quick skim of https://github.com/JFormDesigner/FlatLaf/releases, I don't see any relevant incompatibilities, but I'd still prefer to isolate that change in case it does break some UI bits.

midgleyc (approved):
as last time, get it in and we can revert if something breaks (although from looking at the code I think it'll be safe).

PR #3142, opened 2025-10-21
A new test for coverage and some lint

jaadams5:
A test for the play command was started as part of a PR that was closed before completion and comital.  Since there was previously no PlayCommandTest the code was extracted and added here.  Tests were written for cases that increased coverage.


jaadams5:
> I think that for the tests which are supposed to not error in the command, it would be better to have the deck in inventory, and confirm the request is sent.

Wow.  That was fast.  I will look at that.  Thanks

midgleyc:
If you look at e.g. AlliedRadioCommandTest, the last test in that file is an example.

midgleyc (approved):

PR #3144, opened 2025-10-22
tests: add test for closet parsing

midgleyc:


test/internal/helpers/Player.java
```diff
@@ -2115,9 +2127,32 @@ public static Cleanups withResponseMap(final Map<String, FakeHttpResponse<String
...
+  }
+
+  /**
+   * Sets next response to a GenericRequest Note that this uses its own FakeHttpClientBuilder so
+   * getRequests() will not work on one set separately
+   *
+   * @param responseFunc Function returning responses, keyed by request.
+   * @return Cleans up so this response is not given afterwards.
+   */
+  public static Cleanups withResponseFunc(
```
gausie: Nitpick is that I would probably call this `withResponse` and allow overloading to handle the fact that we are supplying a function. Could even turn `withResponseMap` into an overload of `withResponse`

midgleyc: Updated both to `withResponses`.

gausie (approved):

PR #3146, opened 2025-10-23
fix: update beret combat skill preferences

CheeseyPickle:
See https://kolmafia.us/threads/beret-combat-skill-preferences-not-updating.31423/

I think this should fix it? First time contributing, so pls take a look.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/FightRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FightRequest.java
@@ -10872,12 +10872,12 @@ public class FightRequest extends GenericRequest {
       }
       case SkillPool.BERET_BLAST -> {
         if (responseText.contains("You focus your decades") || skillSuccess) {
-          Preferences.increment("_beretBlastUses");
+          skillSuccess = true;
         }
       }
       case SkillPool.BERET_BOAST -> {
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -10870,6 +10870,16 @@ private static void payActionCost(final String responseText) {
...
+      case SkillPool.BERET_BLAST -> {
+        if (responseText.contains("You focus your decades") || skillSuccess) {
+          Preferences.increment("_beretBlastUses");
+        }
+      }
+      case SkillPool.BERET_BOAST -> {
+        if (responseText.contains("You brag about your beret until") || skillSuccess) {
+          Preferences.increment("_beretBoastUses");
+        }
+      }
```
gausie, line 10882: I think you want to just set `skillSuccess = true` here so that the dailylimits check can handle it. This implementation will double count casts when you are in anapests/haiku/DMT

CheeseyPickle: Thanks for the lookover! Fixed.

jaadams5 (approved):

gausie:
Let's merge. Thank you for the contribution!

gausie (approved):

PR #3147, opened 2025-10-26
feat: support skills that grant multiple effects

VerdantPurple:
Implement a general `$skill.to_effects()` method, returning an aggregate of all relevant effects a given skill can/does grant.

Also default to the more relevant effect for April Shower-related skills, e.g. Sauce Contemplation currently returns only Lubricating Sauce, even if one doesn't own the April Shower Thoughts Shield at all, never mind having it equipped.

```
> ash $skill[Sauce Contemplation].to_effect()

Returned: Lubricating Sauce
...
```

```
> ash $skill[Sauce Contemplation].to_effect()

Returned: Saucemastery
...

> ash $skill[Sauce Contemplation].to_effects()

Returned: aggregate effect [2]
0 => Saucemastery
1 => Lubricating Sauce
```

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/EffectPool.java
@@ -12,6 +12,12 @@ public class EffectPool {
   public static final int BLOODY_HAND = 15;
   public static final int LEASH_OF_LINGUINI = 16;
   public static final int GHOSTLY_SHELL = 18;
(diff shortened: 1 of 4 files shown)


midgleyc (changes requested):
something in RuntimeLibraryTest confirming that `to_effect` does what you want with the relevant item equipped and `to_effects` does what you want with a relevant skill and `$skill[none]` would be nice

but neat idea in general!

PR #3149, opened 2025-10-28
Change config to deal with imports

jaadams5:
As per comment in https://github.com/kolmafia/kolmafia/pull/3148


src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -1804,7 +1810,7 @@ public static final void unlockGoatlet() {
    * After we win a fight, some quests may need to be updated. Centralize handling for it here.
    *
    * @param responseText The text from (at least) the winning round of the fight
-   * @param monsterName The monster which <s>died</s>got beaten up.
+   * @param monster The monster which <s>died</s>got beaten up.
    */
```
jaadams5, line 1814: This is a JavaDoc error which was easier to fix than to convince my IDE to ignore it.

jaadams5:
See https://github.com/kolmafia/kolmafia/pull/3150

The expectation is that this PR and that one will not both be committed.

config/checkstyle/checkstyle.xml
```diff
@@ -31,6 +31,9 @@
     <module name="FallThrough">
       <property name="reliefPattern" value="(?i)falls?[ -]?thr(u|ough)"/>
     </module>
+<module name="AvoidStarImport">
```
heeheehee-kolmafia: The indentation on this line looks off. I'm a little surprised spotless isn't complaining, but maybe it's not checking these XML files.

jaadams5: Better?

heeheehee-kolmafia: Better (in that the start and end have the same indentation now), but it's nested under <module name="TreeWalker"> but at the same indentation. It should be at the same indentation as all of the other elements at the same depth.

https://github.com/kolmafia/kolmafia/blob/format/config/checkstyle/checkstyle.xml#L33-L34

Namely, as written, lines 34-36 should be indented two more spaces.

jaadams5: There is some irony here.  Maybe someday we will have a style checker for XML files :-)

I really have no clue what I was doing beyond the comment in #3148 and wasn't sure whether I was creating a sibling or a child.

midgleyc: You could add a spotless block to format xml files, but that's outside the scope of this PR.

Something like
```
format("xml") {
  target("**/*.xml")

  trimTrailingWhitespace()
  indentWithSpaces(2)
  endWithNewline()
}
```

jaadams5: I'm going to resolve because this is out of scope and in 5 minutes I could not figure out where to add a spotless block so this is not going to happen quickly if I seize the opportunity to learn.

midgleyc (approved):

PR #3151, opened 2025-11-01
feat: shrunken head skeleton

midgleyc:


gausie (approved):
Impressive - to make shrunken head skeleton you need to have a little skull

PR #3152, opened 2025-11-01
feat: persist maximizer filter selections

VerdantPurple:
When the maximizer is closed and reopened, all filter options get reselected, which can be annoying. Wish in particular provides many boosts that tend to swamp the others, and I find myself deselecting it often.

Add an option to remember the filters that were selected last time. Filters for full organs will still be temporarily disabled regardless of persistence.


midgleyc (changes requested):
both `maximizerPersistFilters` and `maximizerLastFilters` should be added to `src/data/defaults.txt`: I would default the former to `true` and the latter to blank.

Also, you could eliminate `maximizerPersistFilters` entirely and just always have it true: I can't really think of a benefit in turning it off.

midgleyc (approved):
Thanks!

I think it's fine to change default behaviour. It can always be changed back if there's an outcry. We do releases extremely often, and upgrading and downgrading are both trivial.

PR #3154, opened 2025-11-03
feat: modify pvp command to allow attacking random tougher opponents

CheeseyPickle:
One of the options for pvp in-game is attacking a random tougher opponent for more fame. Right now, there's no kolmafia command that covers this, so add an optional argument to the "pvp" command to allow for it?


CheeseyPickle:
Testing time!

New descriptions show up
image

Trying to fight tougher opponents
image
Confirmed tougher based on fame gains
image

Fighting randos
image
image
triggerofsol confirmed less fame than me. (50 vs. 358)
Lucelle Ball confirmed the opposite.

No new argument
image
Randomness is confirmed

Errors are as expected as well
image

I might rewrite the second error actually, since that doesn't really make sense.

Just to be sure, I ran 8 fights with the tougher argument at the end of the day and really only fought tougher opponents, so we're good!

Anyway, that's it for the testing. Tl;dr, everything works as expected, might rewrite an error message. What do y'all think?

src/net/sourceforge/kolmafia/textui/command/PvpStealCommand.java
```diff
@@ -67,6 +81,17 @@ public void run(final String cmd, String parameters) {
       }
       mission = "lootwhatever";
     } else {
+      spaceIndex = parameters.indexOf(" ");
```
midgleyc: this isn't right, e.g. the above checks for starting with loot and the lower checks for equal

I would rather just keep the message "What do you want to steal?" to avoid the complexity of this extra check

CheeseyPickle: Fair, removed the new error message.

I couldn't think of a better way to implement that one unfortunately. :(

CheeseyPickle:
> Rather than the manual testing you've done above (which confirm that it works now), we'd much rather have unit testing (to confirm that it works for good). Is that something you'd be able to do?

I can certainly try. I've used JUnit in class, so now seems like a good time to apply that practically. 😅 

Should I stick the tests in this PR? Since it seems like the pvp/steal command doesn't have a test in general so I'd have to add that.

gausie:
In this PR ideally, yes

test/net/sourceforge/kolmafia/textui/command/PvpStealCommandTest.java
```diff
@@ -0,0 +1,243 @@
...
+    var setup = setupClient();
+    var client = setup.client;
+    var cleanups = setup.cleanups;
+
+    try (cleanups) {
+      String output = execute("tougher fame 1");
+
+      assertThat(
+          output,
+          containsString("Use all remaining PVP attacks to steal fame via Quality Assurance"));
```
CheeseyPickle, line 145: For some reason, whenever I assertRequest here, it fails due to PvP Manager spamming requests to place=fight (i.e. the initial fight menu).

When I delete this test, the successful fuzzy match test fails for the same reason. When I delete both, the Cartesian attacksSuccessfully _doesn't_ fail. I've even tried editing the test html to change the number of attacks it says I have left. 

I just don't know why PvpManager is making all the extra requests, just because the attacks parameter is being passed as 0 instead of 1.
https://github.com/kolmafia/kolmafia/blob/0077378ca8580eeb1dcb72b8e676380c2f2df6eb/src/net/sourceforge/kolmafia/session/PvpManager.java#L114-L156

Regardless, wherever it fails, it fails after PvpStealCommand hands everything off to PvpManager, so I say it should be dealt with on that end. Mostly because I'm tired of dealing with this :(

jaadams5: I will look at this more closely when I get time.  But a failing test is either a test that is not correct or a sign of a KoLmafia bug.  The former is much more likely :-)  If the test is correct for the request that generated the HTML then the test should have the same number of GETS and POSTS and in the right order.  If that is absolutely correct and the test generates something different then a next step would be to run the test in the debugger and see if there is something in the character state that is being checked  by KoLmafia that is different between the test and the conditions that created the debug log.  These can be obscure but I have seen cases where the test had to include a seemingly unrelated item in inventory, where the character's gender needed to be set and where the guild store had to be open.  You might try writing a test for the PVP command without your changes and get it to work against unmodified code.  Using a working test as the basis for the test for your changes might make things easier.

CheeseyPickle: Alright, I think I may have found the error. PvPManager isn't getting reset between tests, so it only checks available stances once on the first test. The specific variable seems to be PvPManager.stancesKnown, which is not being reset to false (probably because it's a static variable :/).

Looking at other tests, it looks like it's ok if I just set the static variable to false directly? At the very least, the tests now pass that I've done so :D

midgleyc: yeah, setting it in `beforeEach` is fine -- for ones with heavier reset logic we tend to add a `Cleanups` to `Player` which would reset it, but this is also fine.

midgleyc (approved):

PR #3158, opened 2025-11-06
feat: don't add items stolen by ewes

midgleyc:
Improve support for 11TIHAU.

As reported in https://kolmafia.us/threads/mafia-interprets-items-stolen-by-ewes-as-being-acquired-11tihau.30709/

### DIFF
--- a/test/internal/extensions/TimeTests.java
+++ b/test/internal/extensions/TimeTests.java
@@ -17,7 +17,7 @@ public class TimeTests
   private static final String TEST_METHOD_START = "TEST_METHOD_START";
 
   @Override
-  public void beforeAll(ExtensionContext context) throws Exception {
+  public void beforeAll(ExtensionContext context) {
     getClassStore(context).put(TEST_START, System.currentTimeMillis());
   }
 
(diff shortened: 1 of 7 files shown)


test/net/sourceforge/kolmafia/request/FightRequestTest.java
```diff
@@ -4317,4 +4317,19 @@ public void canDetectShrunkenHeadZombieCollapse() {
       }
     }
   }
+
+  @Test
+  public void shouldNotAcquireItemsStolenByEwes() {
+    RequestLoggerOutput.startStream();
+    var cleanups = withFight();
+
```
jaadams5, line 4325: Sometimes when dealing with acquisition or loss confirming a count before and after the action makes the test clearer.

So you could add assertThat(InventoryManager.getCount(ItemPool.LION_OIL), equalTo(0)); here but this is just an observation.

midgleyc: I feel like that statement would be more testing the setup code (i.e. that we start without any items) than the test logic.

To be thorough, I suppose all these tests that gain items should have a `withNoItems()` to remove them.

PR #3161, opened 2025-11-07
Add a eweItem pref, copying dolphinItem

Ignose:


midgleyc (changes requested):
Ewes can steal multiple items at once, so this isn't quite right.

Also update the test.

midgleyc:
They can steal one per combat, but you can have multiple combats before fighting a ewe.

midgleyc (dismissed):

Ignose:
> They can steal one per combat, but you can have multiple combats before fighting a ewe.

Let me ask, because the code doesn't make sense to me... how does Mafia reset the DolphinItem pref? I don't have anything resetting the eweItem pref.

midgleyc:
Oh right, good spot: after you use a dolphin whistle / durable dolphin whistle, we reset `dolphinItem` to blank.

In the switch around line 4300 in FightRequest, add a case for ewe and clear the eweItem pref there.

midgleyc (approved):

PR #3163, opened 2025-11-08
Cleaned up some questionable test code

jaadams5:
In an attempt to generate 100% coverage I wrote some tests that did things like compare equal against itself and apply equals to mismatched types.  These were flagged by the analyzer.  Removing them did not change the coverage for FaxBotDatabase so the code might as well be cleaned up.


test/net/sourceforge/kolmafia/persistence/FaxBotDatabaseTest.java
```diff
@@ -91,21 +91,16 @@ public void coverageForInvalidMonster() {
 
     @Test
     public void exerciseSomeFaxbotOverridesForCoverage() {
-      FaxBotDatabase.FaxBot nullBot = null;
       FaxBotDatabase.FaxBot botA = new FaxBotDatabase.FaxBot("A Bot", 1999);
       FaxBotDatabase.FaxBot botAlsoA = new FaxBotDatabase.FaxBot("A Bot", 1999);
       FaxBotDatabase.FaxBot botB = new FaxBotDatabase.FaxBot("B Bot", 1999);
       FaxBotDatabase.FaxBot botInvalid = new FaxBotDatabase.FaxBot(null, 1999);
-      assertTrue(botA.equals(botA));
```
midgleyc, line 99: this is valid contract testing -- `equals` should be reflexive.

You could implement `equals` such that this wasn't true, e.g. by implementing it as `return false`.

midgleyc: same for the other deleted lines

PR #3164, opened 2025-11-08
Add XML to spotless and then run

jaadams5:


build.gradle.kts
```diff
@@ -127,7 +127,13 @@ spotless {
     indentWithSpaces(2)
     endWithNewline()
   }
+  format("xml") {
+    target("**/*.xml")
```
jaadams5, line 133: @midgleyc - help, please.  

I'm having two issues.

A case could be made that xml in test/root/provided_data should not be changes since those are files added to the repo to be tested against.  The case could be expanded to mean nothing in test/root should be changed and, as it turns out there isn't xml in test/   There are also no xml files in src/net/sourceforge/kolmafia.  So one option would be to exclude both src/ and test/.  Alternatively just exclude test/root.  Which do you recommend and how do I change the target to implement that recommendation?

Second, I have experimented with various targets and even with I edit checkstyle.xml so that it should not pass running gradlew spotless passes and gradlew spotlessapply doesn't change anything.

Thanks.

midgleyc: For the first, yes, documentation is bad.

https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#quickstart mentions a `targetExclude` parameter, and you can add `targetExclude("test/**/*.xml")` to omit the test files.

For the second, I can mostly reproduce: it doesn't do the indenting that we want it to do. However, if you add many lines to the bottom of the file, it will remove them, so it is both detecting the file and doing something.

What precisely is going on here I don't understand.

midgleyc: Okay, the default formatter doesn't like the XML file because of the doctype and comments. The eclipseWTP formatter can handle it, but deletes the doctype and comments. Not great.

jaadams5: Thank you.  Stand by.

jaadams5: Added exclusions.  

checkstyle.xml  remains unresolved.

jaadams5: .idea is local, not under git control and used by my IDE.  We may need a similar exclusion for users of other IDEs.

jaadams5: What's the next step?  Do we close because it is minor?  Do we impose an additional restirction on "our" XML files so this can be done as is?  Do we inject a different formatter?  Do we submit a bug report and pause?

jaadams5: @midgleyc ?

util/rebel.xml
gausie, line 1: Do we actually use this file?

build.gradle.kts
midgleyc: I guess we close because there's no easy way to do what we want.

jaadams5: Closed.  Thanks.

util/rebel.xml
jaadams5: It is associated with JRebel which allows code changes to be reloaded more quickly.   Hola added it in 2012 with the comment " If a developer wants to use JRebel when working on KoLmafia, make it easier for them to do so."

I'm thinking we don't use/need it although a poll of the devs would be needed to confirm.

Moot point since we can't do what we want and so will be closing but not committing this.

PR #3180, opened 2025-12-02
Knucklebone tracking

gausie:
- **Track knucklebone drops**
- **Update spotless and apply new changes (sorry for including in this changeset)**
- **Track knucklebones dropped from resting**

Sorry again for the spotless update being included here; browse by the commits either side to read around

### DIFF
--- a/src/net/sourceforge/kolmafia/FamiliarData.java
+++ b/src/net/sourceforge/kolmafia/FamiliarData.java
@@ -1252,7 +1252,7 @@ public class FamiliarData implements Comparable<FamiliarData> {
         new DropInfo(
             FamiliarPool.SKELETON_OF_CRIMBO_PAST,
             ItemPool.KNUCKLEBONE,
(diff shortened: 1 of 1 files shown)


src/net/sourceforge/kolmafia/session/ResultProcessor.java
```diff
@@ -3328,6 +3328,10 @@ private static void gainItem(boolean adventureResults, AdventureResult result) {
           Preferences.increment("_miniKiwiDrops", 1);
         }
         break;
+      case ItemPool.KNUCKLEBONE:
+        if (adventureResults && KoLCharacter.currentFamiliar.getId() == FamiliarPool.SKELETON_OF_CRIMBO_PAST) {
+          Preferences.increment("_knuckleboneDrops", 1, 100);
+        }
```
midgleyc, line 3335: break

PR #3181, opened 2025-12-02
chore: update spotless to 8.1.0

midgleyc:
Extract from #3180.


jaadams5 (approved):
Did not read everything but trusting that it breaks nothing :-)  Spotless is much safer in that regard then code analysis "lint".

PR #3189, opened 2025-12-03
Add api support to detect shrunken head ownership and add hasShrunkenHead property

Rinn:
This is not ready to merge, I'm assuming the coolitems api string but it has not been added. @gausie is following up


gausie (approved):
You were correct in your guess

PR #3191, opened 2025-12-04
Preserving some changes from #1889

jaadams5:
Capturing some minor changes from #1889 in anticipation of its closure,


jaadams5:
> I agree. I think it's time to piss or get off the pot with that old PR (not saying that it's not useful).
> 
> Not a lovely idiom though. In Yiddish we have "shpil oder forshpil" which just means "play or lose".

I started (again) a thread in the Mod Lounge but I think it is time to let it go.

My grasp of Yiddish is limited to a few spoken words and there are definitely times when Yiddish is more expressive than English.

PR #3193, opened 2025-12-05
feat: track zombies

midgleyc:
You can tell them by the wriggling worm drops, which are spadable.

See also https://github.com/loathers/excavator/pull/123.


gausie:
Do you want to add the parsing for the tag? I don't think it necessarily does anything for now but for completeness sake

midgleyc:
sure, good idea

PR #3194, opened 2025-12-05
Skeleton of Crimbo Past coinmaster

gausie:
Adds coinmaster for the SoCP including daily special. I added a check on session refresh before seeing the [`visit($coinmaster)`](https://wiki.kolmafia.us/index.php?title=Visit) function which suggests that perhaps was previously considered an exercise left to the user? I'd rather keep it though and potentially change other rotating shops to follow this pattern. Thoughts welcome.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
+++ b/test/net/sourceforge/kolmafia/session/ChoiceControlTest.java
@@ -1762,6 +1762,7 @@ class ChoiceControlTest {
       assertThat("_crimboPastSmokingPope", isSetTo(true));
       assertThat("_crimboPastPrizeTurkey", isSetTo(true));
       assertThat("_crimboPastMedicalGruel", isSetTo(true));
+      assertThat("_crimboPastDailySpecial", isSetTo(false));
       assertThat("_crimboPastDailySpecialItem", isSetTo(ItemPool.ELVEN_SOCKS));
       assertThat("_crimboPastDailySpecialPrice", isSetTo(442));
 


src/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequest.java
```diff
@@ -0,0 +1,239 @@
...
+
+    if (itemId != -1) {
+      CoinMasterRequest.buyStuff(data, itemId, 1, false);
+    }
+
+    return true;
+  }
+
+  public static String accessible() {
+    var ownsFamiliar = KoLCharacter.ownedFamiliar(FamiliarPool.SKELETON_OF_CRIMBO_PAST).isPresent();
```
midgleyc, line 195: general question: what happens if you try to access the shop with a comma pretending to be a crimbo past spirit, but you don't have a crimbo past spirit?

test/internal/helpers/Player.java
```diff
@@ -877,6 +877,18 @@ public static Cleanups withFamiliarInTerrarium(final int familiarId, final int e
     return new Cleanups(() -> KoLCharacter.removeFamiliar(familiar));
   }
 
+  /**
+   * Adds familiar to player's terrarium, but does not take it out
```
midgleyc: Removes

test/internal/helpers/Player.java
```diff
@@ -877,6 +877,18 @@ public static Cleanups withFamiliarInTerrarium(final int familiarId, final int e
...
+   *
+   * @param familiarId Familiar to add
```
midgleyc: to remove

src/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequest.java
gausie: I guess it probably works. We should test and I can either update here or in a later PR

PR #3207, opened 2025-12-13
refactor: convert Jarlsberg's Cosmic Kitchen to coinmaster 

midgleyc:


gausie:
It seems alright to me if it works

PR #3259, opened 2025-12-28
chore: update rhino library version in classpath from 1.7.15 to 1.8.1

xn:
I was poking around looking for what version of rhino y'all were using and I saw the classpath wasn't updated in [3199](https://github.com/kolmafia/kolmafia/pull/3199)

I know it's teeny and inconsequential, but it's a start.

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -22,7 +22,7 @@
 	<classpathentry kind="lib" path="src/jar/svnkit-cli-1.10.2.jar"/>
 	<classpathentry kind="lib" path="src/jar/svnkit-javahl16-1.10.2.jar"/>
 	<classpathentry kind="lib" path="src/jar/trilead-ssh2-1.0.0-build222.jar"/>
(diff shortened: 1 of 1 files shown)


gausie (approved):
Thank you!

gausie:
> Looks like `.classpath` is a file used only by Eclipse. Question for the dev team: is anyone still using Eclipse?
> 
> 
> 
> Given we didn't notice the file hadn't been updated, perhaps we should consider deleting it entirely.

Didn't see this. Happy to just keep it for now in any case, but we can also remove

midgleyc:
Sure, happy to merge (as a strict improvement) and we can consider deleting at a later date.

jaadams5:
> Sure, happy to merge (as a strict improvement) and we can consider deleting at a later date.

This would not be the first time an Eclipse user committed an Eclipse specific configuration file so when we have time and interest we should investigate it.  That said I do think one of the core devs does use Eclipse at least occasionally so we should check.

PR #3263, opened 2025-12-31
Skeleton modifies concoction database when it visits

### LINKED ISSUE #3243: Items "available" from multiple coinmasters may use an inaccessible one through "create"
Reported on or after r28808, https://discord.com/channels/466605739838930955/1225972355730247800/1451687677358248049

Image

bphillab:
I think this resolves #3243 
image

### DIFF
--- a/test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
@@ -86,14 +86,13 @@ class SkeletonOfCrimboPastRequestTest {
 
   @Test
   public void dailySpecialPopulatesInConcoctionPool() {
-    // Item 10883 is "Crimbo snack mix" (descid 957545937)
     String responseText =
-        "value=\"Daily Special: candy egg deviler\"></td><td valign=center><img style=\"margin-right: 1em;  cursor:pointer;\" onclick=\"descitem(477053908)\" alt=\"candy egg deviler\" title=\"candy egg deviler\" src=\"/images/itemimages/deviler.gif\" height=\"30\" width=\"30\" />(1315 knucklebones)";
+        "Daily Special: candy egg deviler> onclick=\"descitem(477053908)\"  />(1315 knucklebones)";
 
     SkeletonOfCrimboPastRequest.visit(responseText);
 
     // Verify it is in the ConcoctionPool
-    var concoction = ConcoctionPool.get(ItemPool.RETHINKING_CANDY_BOOK);
+    var concoction = ConcoctionPool.get(ItemPool.CANDY_EGG_DEVILER);
     assertThat(concoction, notNullValue());
     assertThat(concoction.getPurchaseRequest(), notNullValue());
     assertThat(


test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
```diff
@@ -54,4 +65,39 @@ void exitsTheShopChoice() {
       assertPostRequest(requests.getFirst(), "/choice.php", "whichchoice=1567&option=5");
     }
   }
+
+  @Test
+  public void canRegisterDailySpecial() {
+    // Daily Special: <item description link> (10 knucklebones)
+    String responseText =
+        "Daily Special: <a class=\"nounder\" href=\"descitem(352309634)\"><b>rethinking cnady</b></a> (10 knucklebones)";
```
gausie, line 73: cnady?

test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
```diff
@@ -54,4 +65,39 @@ void exitsTheShopChoice() {
...
+    assertThat(Preferences.getInteger("_crimboPastDailySpecialPrice"), is(10));
+    assertThat(
+        Preferences.getBoolean("_crimboPastDailySpecial"),
+        is(false)); // false means "not yet bought"
+  }
+
+  @Test
+  public void dailySpecialPopulatesInConcoctionPool() {
+    String responseText =
+        "Daily Special: candy egg deviler> onclick=\"descitem(477053908)\"  />(1315 knucklebones)";
```
gausie, line 90: Please use real tect from the game, otherwise how useful is the test?

test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
```diff
@@ -54,4 +65,39 @@ void exitsTheShopChoice() {
...
+    SkeletonOfCrimboPastRequest.visit(responseText);
+
+    // Verify it is in the ConcoctionPool
+    var concoction = ConcoctionPool.get(ItemPool.CANDY_EGG_DEVILER);
+    assertThat(concoction, notNullValue());
+    assertThat(concoction.getPurchaseRequest(), notNullValue());
+    assertThat(
+        concoction.getPurchaseRequest().getShopName(),
+        is(SkeletonOfCrimboPastRequest.SKELETON_OF_CRIMBO_PAST.toString()));
+    assertThat("Has more than 1 ingredient", concoction.getIngredients().length, is(1));
```
gausie, line 101: Is this more than 1 or 1?

test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
```diff
@@ -54,4 +65,39 @@ void exitsTheShopChoice() {
...
+    assertThat(
+        Preferences.getInteger("_crimboPastDailySpecialItem"), is(ItemPool.RETHINKING_CANDY_BOOK));
+    assertThat(Preferences.getInteger("_crimboPastDailySpecialPrice"), is(10));
+    assertThat(
+        Preferences.getBoolean("_crimboPastDailySpecial"),
+        is(false)); // false means "not yet bought"
+  }
+
+  @Test
+  public void dailySpecialPopulatesInConcoctionPool() {
```
gausie, line 88: Perhaps it might be better to test this more end-to-end. It simply being in the concoction pool doesn't verify that "create" works properly

test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
midgleyc: That message is what gets printed if the assertion fails, so it's supposed to be 1.

src/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequest.java
```diff
@@ -128,7 +130,9 @@ private static boolean applySpecial() {
     prices.entrySet().removeIf(e -> e.getKey() < ItemPool.SMOKING_POPE);
     prices.put(item.getItemId(), price);
     SKELETON_OF_CRIMBO_PAST.registerPurchaseRequests();
-
+    Concoction conc = ConcoctionPool.get(item);
+    conc.resetIngredients();
```
midgleyc, line 134: this is perhaps a bit hacky

the research to realise this is the problem is good, but also the existing concoction is using an inaccessible shop (right?), so "create" shouldn't consider it available. So there's still something wrong in the code.

bphillab: Yeah, and this was the piece I was concerned about. Maybe I should have just started more of a discussion.

The solution then is probably completely different then. (ie targeting invalid concoctions at the start)

test/net/sourceforge/kolmafia/request/coinmaster/SkeletonOfCrimboPastRequestTest.java
gausie: Ah nice ok

PR #3269, opened 2026-01-01
feat: club 'em skills limits

midgleyc:


midgleyc:
You'd have to tag it with the skill name and the limit and the type, which is practically all that goes in dailylimits.txt anyway. So I don't think it's worth combining the files.

PR #3273, opened 2026-01-02
feat: angelbone, devilbone, wire-crossin' elf

midgleyc:
wire-crossin' elf has been made available by gausie and should be in eggnet shortly.


jaadams5:
What is eggnet?  I can get some background from https://wiki.kingdomofloathing.com/Tools but does it do anything for a player who doesn't have a chest mimic?

midgleyc:
No, but it means that it'll be an easy source for many people in the future.

It's also available through OnlyFax.

PR #3275, opened 2026-01-03
feat: convert unarmed to an expression

midgleyc:
Likely the cause of issues around passive skills not being recognised, see e.g. https://kolmafia.us/threads/mafia-sometimes-loses-track-of-liver-capacity-modifiers-when-changing-equipment.32301/


midgleyc:
Not particularly a fan of all the statics dotted around on Modifiers but changing that would be a more extensive refactor.

Nice that maximizer has enough tests to feel confident changing Unarmed in the first place.

PR #3287, opened 2026-01-06
Expose more advanced maximizer function to ASH, and current_score function

soolar:
I believe this PR is ready for proper review now! It introduces a new form of the `maximize` ASH function that allows the user to specify `EquipScope` and maximizer filters and get the text that follows the `display` string, as well as a `current_maximizer_score` function that allows arbitrary `Evaluator` usage from ASH.

Also fixes a bug where the wish filter for the maximizer didn't actually do anything, and wishes were instead categorized under the "other" filter.

Here is the forum thread where I requested this feature change, by the way: https://kolmafia.us/threads/minor-change-to-ash-maximize-detailed-function-output.32353/

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
@@ -4008,6 +4008,7 @@ public class ItemPool {
   public static final int PRIZE_TURKEY = 12053;
   public static final int MEDICAL_GRUEL = 12054;
   public static final int THE_ETERNITY_CODPIECE = 12067;
+  public static final int ANGELBONE_TOTEM = 12069;
   public static final int LEGENDARY_SEAL_CLUBBING_CLUB = 12134;
 
   private ItemPool() {}
--- a/src/net/sourceforge/kolmafia/request/UseSkillRequest.java
+++ b/src/net/sourceforge/kolmafia/request/UseSkillRequest.java
@@ -156,6 +156,7 @@ public class UseSkillRequest extends GenericRequest implements Comparable<UseSki
         new BuffTool(ItemPool.PRIMITIVE_ALIEN_TOTEM, 25, false, null),
         new BuffTool(ItemPool.FLAIL_OF_THE_SEVEN_ASPECTS, 15, false, null),
         new BuffTool(ItemPool.CHELONIAN_MORNINGSTAR, 10, false, null),
+        new BuffTool(ItemPool.ANGELBONE_TOTEM, 6, false, null),
         new BuffTool(ItemPool.MACE_OF_THE_TORTOISE, 5, false, null),
         new BuffTool(ItemPool.OUIJA_BOARD, 2, true, null),
         new BuffTool(ItemPool.TURTLE_TOTEM, 0, false, null),


soolar:
> The duration is only displayed if the preference "verboseMaximizer" is true, so you probably want to expose the duration directly.

Hmm, tough call, because you don't get any of this info without verboseMaximizer actually, I changed that setting so long ago I forgot it existed. I might add a check that removes the afterdisplay string if verboseMaximizer isn't true... Or I could make this function always act as if verboseMaximizer is on (setting it back to false if it was before returning ofc).

Also, I added some more code (that also still needs tests for the time being while I also test it out a bit in action locally).

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -9111,11 +9119,24 @@ public static Value maximize(
...
     }
 
     return value;
   }
 
+  public static Value last_maximizer_score(ScriptRuntime controller) {
+    if (Maximizer.eval == null) {
+      return DataTypes.makeFloatValue(0);
+    }
+    double current = Maximizer.eval.getScore(KoLCharacter.getCurrentModifiers());
```
midgleyc: this will return the last evaluation for your current modifiers (i.e. not the maximizer modifiers), which is not really the same thing as the last score.

test/root/scripts/maximizer_result.ash
```diff
@@ -28,3 +50,12 @@ print("score = " + result.score);
 print("effect = " + result.effect);
 print("item = " + result.item);
 print("skill = " + result.skill);
+
+MaximizerResultFull resultFull = do_maximize_full("fishing skill", 0, 0, true, "");
```
midgleyc: can you make a nontrivial test, e.g. something that shows `wish` working the way you expect it to?

you'll probably have to move the code out into `RuntimeLibraryTest.java` so you can add items to the inventory with `Cleanups`.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
soolar: Mmmm, true, it would update as you change your modifiers. That's still good enough for my personal use case, so I could either rename the method (something like `last_maximizer_string_current_score`) or look in to getting the behavior I actually named it for.

Really though the ideal function would be something like `float maximizer_score(string max)` where you can specify the maximizer string instead of needing to rerun the maximizer, I just... didn't see a super straightforward way to achieve that. I'll dig a little more and see if I can achieve that.

soolar: Oh wait, I do see a straightforward way to achieve that now that I'm digging more.

midgleyc: does `Maximizer.best.getScore()` not work?

soolar: To make it take an arbitrary string I made Evaluator's constructor that takes a string public and used that to avoid the need to run a full maximization pass for this function.

test/root/scripts/maximizer_result.ash
soolar: I'm in the process of doing so (having difficulty figuring out how to get wishes to show up in the maximizer in a test though) but I'm going to keep the trivial case test here in addition because the original form of this test case is how I quickly realized the problem with changing the return type of the existing method.

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -2567,6 +2585,18 @@ private static VariableReference namedParam(String name, Type type) {
...
+        List.of(
+            namedParam("maximizerStringValue", DataTypes.STRING_TYPE),
+            namedParam("maxPriceValue", DataTypes.INT_TYPE),
+            namedParam("priceLevelValue", DataTypes.INT_TYPE),
+            namedParam("equipScope", DataTypes.INT_TYPE),
+            namedParam("filters", DataTypes.STRING_TYPE));
+    functions.add(new LibraryFunction("maximize", maximizerResultFullArray, params));
+
+    params = List.of(namedParam("evaluationString", DataTypes.STRING_TYPE));
+    functions.add(new LibraryFunction("current_score", DataTypes.FLOAT_TYPE, params));
```
midgleyc: `current_maximizer_score` instead of `current_score`, please

`current_score` is too vague

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
```diff
@@ -98,29 +98,43 @@ public class Maximizer {
...
     KoLmafiaCLI.isExecutingCheckOnlyCommand = false;
 
-    Maximizer.maximize(equipScope, maxPrice, priceLevel, false, EnumSet.allOf(filterType.class));
+    Maximizer.maximize(equipScope, maxPrice, priceLevel, false, filter);
 
     if (!KoLmafia.permitsContinue()) {
       return false;
     }
 
+    if (Maximizer.best == null) {
```
midgleyc: I don't think this is right. It was never null before your change, so how could it be null afterwards?

soolar: I don't fully understand either tbh. But when I try to run the new form of the maximize ash function prior to any other maximizer runs in the session without this change, it's always null. I agree this is probably not the right way to fix that though... Just not 100% sure what is.

test/net/sourceforge/kolmafia/textui/RuntimeLibraryTest.java
```diff
@@ -1790,6 +1790,88 @@ public void itShouldUnequipThenEquip(String whichCase) {
...
+        out2 = execute(cmd2);
+        out3 = execute(cmd3);
+      }
+
+      assertFalse(out1.isEmpty());
+      assertTrue(out1.contains("equip back makeshift cape"));
+
+      assertFalse(out2.isEmpty());
+      assertTrue(out2.contains("equip back makeshift cape"));
+      assertTrue(out2.contains("make &amp; equip hat Mark IV Steam-Hat"));
```
soolar, line 1863: Huh, for some reason in this test it wants to pull the ice pick when running as part of all tests, but when just running maximizer tests it works properly. Presumably a problem with my cleanups? I've tried specifying `withRonin(true)` as an addition though and that didn't help...

jaadams5: It could be that some test that used an ice pick did not clean up properly.  I'll try and take a look.

jaadams5: Never mind.  I looked at the test and superficially it looks like the problem is not the test setup.

Still on my list, though.

jaadams5: I need to take a break.  But I added code that called internal.helpers.Maximizer.getBoosts(); before each execution.  I printed the size of the boost array.  When I run the test alone the array is empty before the first call to the maximizer.  But when I run all the tests it already has one entry on the first call.  Since the boosts are a static member something is clearly leaking.  Not sure whether that was always the case but...

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
soolar: I THINK this is the last real problem to solve for this PR, but I'm not sure what to do for it...

soolar: Huh. I got rid of these lines and am no longer seeing the issue. I guess some other change in main fixed the underlying issue, or something...?

PR #3292, opened 2026-01-08
Make Coinmasters list searchable

gausie:
- **Add option for searchable selector panels, and enable for the coinmaster panel**
- **Make clicking things open the right Coinmaster**


midgleyc:
`StringUtilities.getMatchingNames`, which we use elsewhere, uses fuzzy matching but also a bunch of other stuff

Fuzzy matching does e.g. let you search "crb" for Cosmic Ray's Bazaar.

PR #3295, opened 2026-01-09
feat: decorate valhalla with codpiece gems

midgleyc:


src/net/sourceforge/kolmafia/webui/ValhallaDecorator.java
```diff
@@ -571,6 +574,35 @@ private static void switchFolderHolder(StringBuffer buffer) {
...
+    StringBuilder builder = new StringBuilder();
+
+    if (!InventoryManager.equippedOrInInventory(ItemPool.THE_ETERNITY_CODPIECE)) {
+      return;
+    }
+
+    builder.append("Eternity Codpiece: ");
+    for (var slot : SlotSet.CODPIECE_SLOTS) {
+      AdventureResult gem = EquipmentManager.getEquipment(slot);
+      if (gem != null) {
```
gausie, line 587: Rather than print nothing here we could prompt to fill a slot, might help someone out who accidentally only adds 4 gems. And will make the ui less weird if someone has an empty codpiece.

midgleyc: it prints `(none)`, the `name` of `UNEQUIP`.

gausie: Slay

PR #3298, opened 2026-01-10
fix: You can buy/access stores without a buy action

bphillab:
Trying to fix this issue:
https://kolmafia.us/threads/coinmaster-cli-command-does-not-work-with-socp-shop.32035/

It appeared to also impact other choice.php based coinmasters.

It looks like we rely on getBuyAction to gate whether we can buy from a store. This fix removes that requirement and adjusts some others to look at more than just getBuyAction.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/TaleOfDreadCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/TaleOfDreadCommand.java
@@ -80,7 +80,7 @@ public class TaleOfDreadCommand extends AbstractCommand {
             "inv_use.php?pwd&which=3&whichitem=" + ItemPool.TALES_OF_DREAD + "&ajax=1");
     RequestThread.postRequest(request);
 
-    if (!request.responseText.contains("<b>Tales of Dread</b>")) {
+    if (!request.responseText.contains(">Tales of Dread</b>")) {
       KoLmafia.updateDisplay(MafiaState.ERROR, "You don't own the Tales of Dread");
       return;
     }


src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -1972,7 +1972,7 @@ public boolean get_buys() {
     }
 
     public boolean get_sells() {
-      return this.content != null && ((CoinmasterData) this.content).getBuyAction() != null;
+      return this.content != null;
```
midgleyc: for these two, instead of checking `getBuyAction` and `getSellAction`, check `getBuyItems` and `getSellItems` not being null or empty

src/net/sourceforge/kolmafia/textui/command/CoinmasterCommand.java
```diff
@@ -60,11 +60,11 @@ public void run(final String cmd, String parameters) {
 
     if (isBuy) {
       action = data.getBuyAction();
-      if (action == null) {
+      List<AdventureResult> buylist = data.getBuyItems();
+      if (action == null && buylist == null) {
```
midgleyc: check buylist null or empty, don't check action at all

src/net/sourceforge/kolmafia/textui/command/CoinmasterCommand.java
```diff
@@ -81,7 +81,8 @@ public void run(final String cmd, String parameters) {
       }
     } else {
       action = data.getSellAction();
-      if (action == null) {
+      List<AdventureResult> selllist = data.getSellItems();
+      if (action == null && selllist == null) {
```
midgleyc: check selllist null or empty, don't check action at all

midgleyc (changes requested):
see comments, some tests would also be nice

testing the proxy record is easiest in RuntimeLibraryTests, but you could also do it in CustomScriptTests. You want to confirm the following, e.g.:
* dimesmaster: buy and sell
* big brother: sell true, buy false

no shops only sell -- technically the Central Loathing Floral Mercantile Exchange only sells, but it's implemented in such a way that Mafia doesn't know that (using shop rows, which don't distinguish between buy and sell)

for the command test, look at AutumnatonCommandTest or SaberCommandTest. You don't need to test the emitted HTML, I think, so SaberCommandTest is a good pick. If you think you do, look at AutumnatonCommandTest to see one way of setting up the fake HTML client.

src/net/sourceforge/kolmafia/textui/command/CoinmasterCommand.java
bphillab: I think I'm going to remove this check based on what's happened so far with the store that only sells.

bphillab:
I hope you don't mind, but decided to use skeleton of crimbo past for the proxy test rather than big brother since that was the thing that alerted us to this stuff.

I merged in a set of tests for coinmaster command that are likewise skeleton of crimbo past centric for a similar reason.

I wanted to thank you for your patience and help on this.

midgleyc (approved):
Thanks!

PR #3303, opened 2026-01-11
fix: Updating standard rewards for 2026

bphillab:
Fixing bug:
https://kolmafia.us/threads/2026-standard-gear-pulverizes-into-twinkly-stuff-according-to-mafia.32376/


bphillab:
A knock against this pull is that I didn't update the shoprows for the 2025 standard rewards since I don't have that at the moment.

midgleyc (approved):
whoops, I thought just changing standard-pulverization.txt would be enough. Thanks!

PR #3307, opened 2026-01-13
Update perishable Crimbo consumables

oshi97:
I'm not sure if this is the source of truth for these things but I'm assuming it is
Updated using wiki values


oshi97:
nvm this is a dupe

PR #3317, opened 2026-01-20
fix: Gem stones not being decremented when equipped in codpiece

bphillab:
Addressing: https://kolmafia.us/threads/gems-equipped-to-the-eternity-codpiece-are-not-deducted-from-inventory.32396/
Looks like we have something that catches the "You acquire ..." but misses the "You lose a ..."

I think the card sleeve approach works so copied that over.


midgleyc:
I think this is right, in keeping with the way the folder / boot works.

Notably if you're doing this you shouldn't need or want to run the API check after parsing the response.

ResultProcessor some time along the way lost the ability to check for item loss (e.g. it fails to detect knucklebones). It would be nice to fix that, but I think this and that are separate. In particular, we're not losing the items we equip.

PR #3320, opened 2026-01-20
fix: gem stone equip/unequip issue

bphillab:
Addressing: https://kolmafia.us/threads/gems-equipped-to-the-eternity-codpiece-are-not-deducted-from-inventory.32396/

### DIFF
--- a/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
+++ b/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
@@ -944,6 +944,7 @@ public class EquipmentRequest extends PasswordHashRequest {
     if (urlString.startsWith("choice.php") && urlString.contains("whichchoice=1588")) {
       // instead of parsing the page, get our updated gems from api.php
       parseCodpiece(responseText);
+      ApiRequest.updateStatus();
       return;
     }
 
--- a/test/net/sourceforge/kolmafia/request/EquipmentRequestTest.java
+++ b/test/net/sourceforge/kolmafia/request/EquipmentRequestTest.java
@@ -6,7 +6,10 @@ import static internal.helpers.Networking.html;
 import static internal.helpers.Player.withEquipped;
 import static internal.helpers.Player.withFamiliar;
 import static internal.helpers.Player.withHatTrickHat;
+import static internal.helpers.Player.withItem;
 import static internal.helpers.Player.withPath;
+import static internal.helpers.Player.withUnequipped;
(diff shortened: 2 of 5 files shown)


midgleyc:
(ref #3317)

src/net/sourceforge/kolmafia/request/EquipmentRequest.java
```diff
@@ -943,7 +943,7 @@ public void processResults() {
 
     if (urlString.startsWith("choice.php") && urlString.contains("whichchoice=1588")) {
       // instead of parsing the page, get our updated gems from api.php
-      ApiRequest.updateStatus();
+      parseCodpiece(responseText);
```
midgleyc, line 946: I was wrong here, your `parseCodpiece` doesn't check slots, it just updates inventory counts, so we do still need to get the gems from api.php

src/net/sourceforge/kolmafia/request/EquipmentRequest.java
```diff
@@ -1194,6 +1194,29 @@ public static void parseFolders(String responseText) {
...
+    AdventureResult newItem = lost != null ? ItemPool.get(lostId) : EquipmentRequest.UNEQUIP;
+    AdventureResult oldItem =
+        acquired != null ? ItemPool.get(acquiredId) : EquipmentRequest.UNEQUIP;
+
+    if (newItem != EquipmentRequest.UNEQUIP) {
+      AdventureResult remove = oldItem.getInstance(-1);
+      AdventureResult.addResultToList(KoLConstants.tally, remove);
+      AdventureResult.addResultToList(KoLConstants.inventory, remove);
+    }
+    EquipmentRequest.switchItem(oldItem, newItem);
```
midgleyc, line 1218: I think this isn't right -- switchItem also updates inventory counts.

Would be best to have some tests: equipping something over nothing, nothing over something, something over something, and erroring (e.g. trying to equip a non-gem or gem you don't have)

bphillab: The way I understand it is that we catch the "you acquire..." and miss the "you lose an item..." my calculation here is to remove the acquired item then switch which adds 1 to acquired item and subtracts 1 from removed item. 

I can write this as just subtracting the removed item above.

bphillab: Hm... Equipping a non-gem/non-owned gem just pops out what was in that slot.

PR #3321, opened 2026-01-20
Fix importing hyphenated modules in JavaScript

gausie:
- **Maybe fix rhino making invalid method names**
- **Add test for hyphenated module**

Since merging rhino upgrade, people noticed that JS modules with hyphens were causing Rhino to produce invalid method names

### DIFF
--- a/test/root/expected/require_hyphenated_module.js.out
+++ b/test/root/expected/require_hyphenated_module.js.out
@@ -1,0 +1,1 @@
+hyphenated module loaded successfully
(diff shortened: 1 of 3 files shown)


gausie:
This apparently doesn't quite do it. Will investigate today.

gausie:
Yes ok. This does technically fix *a* problem, but doesn't fix the output of some built scripts who have methods with hyphens in them (because the method is a value in an object and their key has a hyphen or other illegal character)

libraryaddict:
One solution is to use the snapshot builds via file reference, jitpack or github packages.

PR #3323, opened 2026-01-24
maximizer 1H weapon test and rescore patch

BadHorseMonkey:
I set gpt-5.2 codex after the “maximizer did not equip a weapon” problem. The bot came up with a theory, made a failing test, and then made a fix.  

I strongly encourage a thorough review.  I will upload the transcript of the bot convo in comments.

### DIFF
--- a/src/data/monsterparts.txt
+++ b/src/data/monsterparts.txt
@@ -2333,12 +2333,12 @@
 2510	The Unquiet Bones of Claus	giant bone	giant claw	giant rib	giant skull
 2511	beef bodyguard bat	steak	vein	wing
 2512	Basted Boss Bat	head	torso	wing
-2513	Gabogooblin King
-2514	The Maety Bonerdagon	bone	flesh chunk
-2515	Feeble Dr. Awkward, El Beef
-2516	Lard Spookyraven
-2517	Protector Speck-ter
-2518	Groarst
-2519	The Big Mac Wisniewski
-2520	The Manwich
+2513	Gabogooblin King	arm	head	leg	torso
+2514	The Maety Bonerdagon	bone	flesh chunk	skull
+2515	Feeble Dr. Awkward, El Beef	arm	head	leg	torso
+2516	Lard Spookyraven	lard
+2517	Protector Speck-ter	head	tendril
+2518	Groarst	roast	wing
+2519	The Big Mac Wisniewski	arm	cheeseburger	leg	torso
+2520	The Manwich	arm	leg	sandwich	torso
 2521	Naughty Sorceress, all sausage	head	sausage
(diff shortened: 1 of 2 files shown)


test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -2350,6 +2624,29 @@ public void doesNotRecommendAprilBandWithoutItem() {
     }
   }
 
+  @Test
```
midgleyc, line 2627: already a test for this in `ModifiersTest.java`

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -813,6 +819,274 @@ public void swordModifierFavorsSword() {
     }
   }
 
+  @Nested
+  class OffhandSynergy {
```
midgleyc, line 823: none of these tests fail even without the evaluator code change

also they don't really meet the way we normally test: we try to use real items, not override modifiers, and there's no need to test items exist

and also the multiple tests testing effective / not effective doesn't matter because the character has moxie highest and is considering the yam cannon, which is a ranged weapon favouring moxie, so it'll be picked no matter what anyway

based on what you say, you want a 1-handed weapon with a negative score, an offhand with a large positive score, and a 2-handed with a mediocre score. but you can equip an offhand with no mainhand, and I think if the only mainhand you have available reduces your score, that's what you want to do?

BadHorseMonkey: I’ll push back on matching the style or write the test myself. I was definitely getting a fail with the new tests, but it may’ve cleaned it up poorly after fixing the code. 

I was hoping more to capture the cause than to get a clean patch. 

I both like and dislike the “however we got to none+offhand, re-evaluate all 1H options”. It doesn’t find ‘why’, but the symptom gets cleared. 

I really appreciate you and Frono looking at this.

midgleyc: Sometimes none + offhand is right. If you're maximizing "-ml", and the only weapon you have available is +ML, it's correct to not equip it.

Similarly, if you're explicitly requesting "effective", you have higher muscle than moxie, and you have only ranged weapons available, it's correct to not equip a weapon.

The "maximizer did not equip a weapon when it was supposed to" problem was probably caused by the Unarmed modifier leading maximizer to ignore all enchants when the bug hit, which was fixed by moving the modifier to an expression.

BadHorseMonkey: The example in the tests was none + august scepter scored lower than yam cannon + august scepter, but it selected none + august scepter.

The tests output the scores it expected test for this, and the speculation was that something caused yam cannon to have already been rejected.

Base score: 603.3000000000001
 Offhand only score: 1853.3 
 1H + offhand score: 1875.8 
 2H  score: 1605.55
 
 Where  Offhand was the august scepter and 1H was the yam cannon, while 2H was also in there so that yam cannon wasn't the selection of the first pass.
 
 I'll try to get the test to conform.  It failed every time It ran it before the fix, but it might've optimized the failure away.

BadHorseMonkey: > Sometimes none + offhand is right. If you're maximizing "-ml", and the only weapon you have available is +ML, it's correct to not equip it.

I'm aware that sometimes none + offhand is correct.  I think that's not the case in the data I have from my own example.

In the example [I posted to the thread](https://kolmafia.us/threads/maximizer-didnt-equip-a-weapon.29577/post-176370), I have the session log, debug log, full inventory, and skills list of my HCAT who ended up with none+august scepter instead of literally any ranged 1H weapon + august scepter.

But when we make a test case from that, it passed.  Which is super frustrating.  

It was an Autoscend maximizer string (also prominently in that post) so it's reasonably complex.  

It's a lot of data, but I hoped it would be enough to show an issue.  It's concerning that it can't.

And of course, we've changed the maximizer in the past year, so it might have been cleared up anyway.

So, the AI effort was to get something to try to test.  I'm slightly annoyed at myself for letting it build a synthetic failing test that didn't work with the real data.

PR #3326, opened 2026-01-29
Testing suite for maximizer classes

gausie:


gausie:
Oops meant to open this in my repo to check it out

PR #3331, opened 2026-02-03
feat: more conditional skills

midgleyc:


src/net/sourceforge/kolmafia/session/EquipmentManager.java
```diff
@@ -525,6 +513,20 @@ private static void manageConditionalSkills(
     }
   }
 
+  private static boolean shouldApplySkill(Integer id) {
```
gausie, line 516: Smart

midgleyc: there may be a link to the new heartstone skills...

but this needs slightly more work, because this function only runs when an item is equipped / unequipped, and you can gain the skill while adventuring. like monkey paw, which I also omitted.

midgleyc: and also the colosseum weapon skills which we have no support for at all because you just use spells to kill the monsters ;)

PR #3334, opened 2026-02-05
Revert "Upgrade to rhino v1.9.0"

Tokoeka:
Reverts kolmafia/kolmafia#3318

Causes breaking issue with some scripts.


midgleyc (approved):
Thanks!

Note: causes problems with e.g. https://github.com/Kasekopf/loopstar due to https://github.com/mozilla/rhino/pull/2241

PR #3337, opened 2026-02-08
heartstone skills are not single cast

C2Talon:
I'm not sure why these were added as single cast skills, but I've had zero troubles using theses skills with the `quantity` parameter on `runskillz.php` having a value more than `1`. And mafia seems to recognize the result fine:

```
> ashq visit_url(`runskillz.php?pwd={my_hash()}&action=Skillz&whichskill={$skill[heartstone: %pals].id}&targetplayer={my_id()}&quantity=2`,false,true)

You acquire an effect: Best Pals (50)
Preference _heartstonePalsUsed changed from 1 to 3
```


gausie:
They were not properly multicast on release

PR #3354, opened 2026-02-17
feat: add function for computing next available mobius strip noncombat

midgleyc:
Add preferences for whether and when the mobius ring was primed.

`turns_until_mobius_noncombat_available` tells you how many turns you have to spend until there's a chance of getting the "Time is a Möbius Strip" encounter.

`_mobiusRingPrimed` tells you whether you have "primed" your Möbius ring today, which kicks off the first countdown for getting the first NC. You have to prime the ring once per day, which you do by completing a combat with it equipped. 

`_mobiusRingPrimedTurn` tells you which turn you primed your ring at, so we can do the initial 4-turn countdown.

### DIFF
--- a/src/data/monsters.txt
+++ b/src/data/monsters.txt
@@ -1458,7 +1458,7 @@ Nautical Seaceress	2498	seaceress.gif	BOSS NOCOPY Atk: 2000 Def: 2500 HP: 4000 I
 
 # Adventurer Meat World
 
-beef bodyguard bat	2511	beefbat.gif	Atk: 25 Def: 22 HP: 50 Init: 50 Meat: 200 P: construct Article: a	beefy pill (0)
+beef bodyguard bat	2511	beefbat.gif	Atk: 25 Def: 22 HP: 50 Init: 50 Meat: 200 P: construct Article: a	beefy pill (10)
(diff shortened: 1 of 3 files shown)


src/net/sourceforge/kolmafia/request/FightRequest.java
```diff
@@ -3100,7 +3100,7 @@ public static void updateFinalRoundData(
     if (responseText.contains("The mulled wine you drank") && KoLCharacter.getCurrentHP() > 0) {
       Preferences.decrement("getsYouDrunkTurnsLeft", 1, 1);
     } else if (garbledCombat) {
-      Preferences.decrement("ghostPepperTurnsLeft", 1, 0);
+      Preferences.decrement("getsYouDrunkTurnsLeft", 1, 0);
```
jaadams5, line 3103: Presumably something you noticed was wrong and fixed?

midgleyc: correct, the two surrounding ones were both `getsYouDrunkTurnsLeft`.

Before, it would have double-decremented `ghostPepperTurnsLeft` in the case of garbled combat.

PR #3361, opened 2026-02-21
Add 'get_no_pulls' to retrieve 'No pull' items from storage

libraryaddict:
Adds a way to get `No pull` items from storage, identical logic as `get_free_pulls`.

Also fixes a bug where `No pull` items was not cleared when storage is emptied.


midgleyc (approved):
sure, but what's your use-case here?

libraryaddict:
> sure, but what's your use-case here?

Step by step tracking of items & meat gained / lost during a scripted run during different stages of the day.

With the recent crates of prize winning cheese/booze from crimbo which are no pull, I'm getting a dumb "lost items on ascension, gained items on prism break", so a jump in value gained/lost.

Ideally, it could be something like `get_storage(boolean show_everything)` which is still a possibility I suppose.

PR #3365, opened 2026-02-24
Allow escaping commas on "call <script> (<parameters>)"

libraryaddict:
Current "call" command behavior is that arguments can be surrounded by () which will be split by comma and fed as multiple arguments.

This PR allows the user to escape the commas and to escape the escapes. Includes tests.

Current behavior: `call script.ash (arg\\, arg\, arg)`
```
arg\\
arg\
arg
```

New: `call script.ash (arg\\, arg\, arg)`
```
arg\
arg, arg
```

This also includes a change to how scripts tests can be invoked, via a new file that ends with the extension ".params" in the scripts folder. Maybe belongs in a new folder for consistency.


midgleyc (approved):
Sure.

I think having them in the same folder is fine. I kind of want to make it one folder per script, which would let you e.g. provide a json with extra options like the parameters instead of having to make a file for them, but this is fine.

PR #3371, opened 2026-03-02
Add Effect / Effect Duration to more items.

midgleyc:
The flipside of #3265.

For data, I used the command
```
js Effect.all().map(y => [y, y.all.filter(x => x.startsWith("use 1")).map(x => x.substring(6)).map(x => toItem(x)).map(x => [x, stringsModifier(x, "Effect")]).filter(x => !x[1].includes(`${y}`))]).filter(y => y[1].length > 0).map(x => `${x[0]}|${x[1]}`)
```

and then went through manually to add effects / durations where necessary.

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -412,7 +412,7 @@ Item	orange peel hat	Spooky Resistance: +3, Stench Resistance: +3, Hot Resistanc
 Item	orange traffic cone	Moxie: +3, Spooky Resistance: +2, Last Available: "2005-03", Familiar Effect: "cap 15"
 Item	Orcish baseball cap	Muscle: +3, Mysticality: -3, Familiar Effect: "1xVolley, 1xBarrr, cap 15"
 Item	oriole-feather headdress	Maximum HP: +1, Maximum MP: +1, Familiar Effect: "atk, 3xVolley, cap 3"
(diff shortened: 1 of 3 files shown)


src/data/modifiers.txt
```diff
@@ -11452,7 +11454,7 @@ Item	lotion of stench	Effect: "Stinky Hands", Effect Duration: [R+5]
 Item	LOV Elixir #3	Effect: "The Power of LOV", Effect Duration: 40, Last Available: "2017-02"
 Item	LOV Elixir #6	Effect: "The Magic of LOV", Effect Duration: 40, Last Available: "2017-02"
 Item	LOV Elixir #9	Effect: "The Moxie of LOV", Effect Duration: 40, Last Available: "2017-02"
-Item	Love Potion #0	Effect: "Tainted Love Potion", Effect Duration: 20
+Item	Love Potion #XYZ	Effect: "Tainted Love Potion", Effect Duration: 20
```
gausie, line 11457: I recall this flipflopping before - what's going on here?

midgleyc: I changed it manually to match the name in `items.txt`, which is the one we got from Museum.

jaadams5 (approved):

PR #3372, opened 2026-03-02
feat: limit workflow permissions

midgleyc:
Build needs `contents: write` to create releases.

Check needs `checks: write` to publish the test reports.


midgleyc:
Despite the `checks: write` it seems to not actually set `checks: write`:
```
GITHUB_TOKEN Permissions
  Checks: read
  Contents: read
  Metadata: read
```
Weird.

PR #3407, opened 2026-03-23
Runtime library documentation in ashref/jsref

gausie:
## Summary
- Add descriptions to runtime library functions and parameters via `.withDescription()` on `LibraryFunction` and `VariableReference`
- Display docs as JSDoc-style `/** */` blocks above the function signature in `ashref` and `jsref` output, with theme-aware grey colouring
- Use TypeScript-style signatures in `jsref` (`function name(param: type): returnType;`)
- Fix wiki link colour for dark mode themes
- Generate `@param` tags in TypeScript definition output
- Extract shared doc formatting into `RefCommand` utility

At some point in the future I will make a gradle task that outputs HTML/JSON so that we can host live documentation for the runtime library too

## Test plan
- [x] `AshRefCommandTest` — doc block above signature, param descriptions, undocumented functions unchanged
- [x] `JsRefCommandTest` — same as above plus TS-style signature format
- [x] `TypescriptDefinitionTest` — JSDoc output, param tags, all overloads have descriptions
- [x] `LibraryFunctionTest` / `VariableReferenceTest` — `.withDescription()` chaining


src/net/sourceforge/kolmafia/KoLmafiaASH.java
```diff
@@ -276,37 +278,49 @@ private static void showFunctions(
...
+      if (func instanceof LibraryFunction lf) {
+        var docBlock = RefCommand.formatDocBlock(lf);
+        if (docBlock != null) {
+          RequestLogger.printHtml(docBlock);
+        }
+      }
 
-      description.append(func.getType());
-      description.append(" ");
+      StringBuilder signature = new StringBuilder();
```
midgleyc: does this (+showExistingFunctions) perhaps want to move into `AshRefCommand`?

The formatting at least has an obvious parallel with JsRefCommand. Maybe that wants to be its own class. Not sure.

test/net/sourceforge/kolmafia/textui/parsetree/LibraryFunctionTest.java
```diff
@@ -29,6 +29,53 @@ void beforeEach() {
...
+    assertThat(fn, is(org.hamcrest.Matchers.instanceOf(LibraryFunction.class)));
+    var lf = (LibraryFunction) fn;
+    assertThat(lf.getDescription(), nullValue());
+
+    var result = lf.withDescription("test description");
+    assertThat(result, is(lf)); // returns this for chaining
+    assertThat(lf.getDescription(), is("test description"));
+
+    // clean up
+    lf.withDescription(null);
```
midgleyc: minimally do this as try-finally, but preferably like VariableReferenceTest create a new function instead of changing an existing one

test/net/sourceforge/kolmafia/textui/parsetree/LibraryFunctionTest.java
```diff
@@ -29,6 +29,53 @@ void beforeEach() {
...
+  }
+
+  @Test
+  void seededDescriptionsExist() {
+    var printFn = allFunctions.findFunctions("print")[0];
+    assertThat(
+        ((LibraryFunction) printFn).getDescription(),
+        is("Prints a message to the CLI and session log."));
+
+    var adventureFn = allFunctions.findFunctions("adventure")[0];
```
midgleyc: only need to do one of these.

some would say that you should check non-empty and presence etc instead of [0] but tbh I think just crashing is fine there.

libraryaddict:
No idea what the status of this is, but to me it feels like this should be supporting `@Annotations` instead of hardcoding names in. Like its doubling down on code debt.

Then at compile or runtime, reflection to get the arguments and their names and annotations.

```
@Function(description = "Get session logs")
public Value get_session_logs(ScriptRuntime runtime, @Argument(description = "How many days to return") Value days)
```

midgleyc:
Why do you think annotations is better, or that this way is debt? What is this way making it harder to do in the future?

Runtime reflection has both a perf and complexity penalty, so I'm biased against it.

libraryaddict:
I'm a bit interested in the performance hit, I'll check that tomorrow. Easy to simulate.

And correct me if I'm wrong, annotations is designed for metadata. 

https://www.oracle.com/technical-resources/articles/hunter-meta.html

https://www.geeksforgeeks.org/java/annotations-in-java/

The only issue becomes when it's bulky. In which case should probably be using identifiers and writing the meta elsewhere I suppose. Kind of like translations or resource strings. Could probably do that to begin with.

If performance hit is a serious concern, it can be generated at compile time and bundled in.

midgleyc:
So the downside is that currently, when adding a function the user has to add it to the master functions list and /also/ add a separate appropriately named function for the implementation, whereas with the annotation style only the function (plus annotations) would be required, so it's only one place compared to two?

libraryaddict:
Quick test indicates the annotation would double the speed of loading, though that's about 4.5ms and 9ms. I don't think there's any real cost afterwards.

The downside of the current implementation is the twice declared yes. The upside is that it's twice as fast.

Regardless, this wasn't the right place for me to bring it up sorry.

midgleyc:
> that's about 4.5ms and 9ms.

Actually pretty good! From my experience with Spring / Hibernate the reflective indirection was adding several minutes to initial startup. Might depend on hardware too.

Still, I don't think adding the metadata in two places is a large burden on us, and I'd rather not make things worse for users (hundreds) in favour of a slightly easier time for developers (~ten). Also the complexity penalty of having to maintain code that deals with an annotation system (less familiarity on my part) compared to a class system (very familiar).

> Regardless, this wasn't the right place for me to bring it up sorry.

It's cool, I learned something. I never considered adding the metadata in two places to be an issue at all, so I was confused at first.

gausie:
> I'd rather not make things worse for users (hundreds) in favour of a slightly easier time for developers (~ten).

I strongly disagree with this in this context, for what it's worth.

gausie:
I'm going to close this probably as I don't think I'm going anywhere with it, but anyone is welcome to come back to it in future. I am somewhat tempted to go down the reflection route.

PR #3417, opened 2026-04-12
feat: stop bulk purchasing sgeeas and tiny houses

midgleyc:
For tiny houses, I don't see any reason to favour them over over healing items.

For sgeeas, the current mall situation is that they cost mall-min, then 3.5k for 10, then 4.5k, and I don't think it's justifiable to buy 30 every time you want to use 1.

For antidotes they're NPC-buyable, so that's fine.


jaadams5 (approved):
I'm guessing this is a residual from a time when reducing server hits on KOL was a success metric for KoLmafia.

PR #3419, opened 2026-04-15
Update the number of recent options shown in the maximizer when maximizerMRUSize is updated

VeeArrKoL:
Addresses issue described at https://kolmafia.us/threads/maximizermlist-does-not-correctly-handle-changes-to-maximizermrusize.32511/

Cursory inspection indicates that the change fixes the issue for the most part. There is one remaining issue: if the maximizer frame is open when the preference is changed, the number of MRU entries populated doesn't update immediately but the location of the separator does. This visual blemish only remains until the next time the frame is opened or until the "update" button actually causes an update of the Swing elements. 

There is also a separate issue where the MRU list seems to get padded with duplicates if its size is increased, but that's a matter of the actual preference getting duplicates put into it, not a display issue.


midgleyc (approved):
Thanks!

PR #3422, opened 2026-04-16
Fix maximizer recommending campaway cloud if not owned or while under Standard restrictions.

VeeArrKoL:
Fixes issue described at: https://kolmafia.us/threads/maximizer-suggests-using-campaway-cloud-when-under-standard-restrictions.32503/


midgleyc (approved):
Thanks!

PR #3424, opened 2026-04-19
add [turn] to skeleton spade fights

ckb11:
This adds [turns] to sessions similar to other encounters and fights, useful for log parsing


midgleyc (approved):
Thanks!

PR #3432, opened 2026-05-01
Implement replaceable effect mutexes

VeeArrKoL:
I took a stab at implementing something for the request in: https://kolmafia.us/threads/request-support-for-mutually-exclusive-buffs-in-maximizer.32504/

When speculating about casting a skill, consider not only the added effect from the skill but also the loss of any existing replaceable mutually-exclusive effects.

Looking for some feedback on the following:

- Is the implementation within `ModifierDatabase` okay? I opted to directly cache the set of effects that are removed when casting a skill for quick lookup, but am open to feedback on whether there's any better mechanism for doing this directly in mafia already.
- The names `MutexER` and "replaceable effect mutex" are both sort of terrible.
- This change is technically not completely correct: there are a few replaceable effect mutex groups whose effects are *wishable* and wishing for one of the effect *doesn't* remove the other effects in its mutex group. (For example, you can get both Coffeesphere and Oilsphere by having one and wishing for the other.) 
  - This... probably won't come up much--neither the Dread songs nor the Crimbo shanties are wishable, and it seems a bit unlikely anyone would be interested in wishing in order to get multiple effects from the other groups. And it seems even less likely that both effects would actually be relevant to whatever maximizer string is being queried.

### DIFF
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
@@ -20,6 +20,7 @@ import static internal.helpers.Player.withItemInStorage;
 import static internal.helpers.Player.withLocation;
 import static internal.helpers.Player.withMCD;
 import static internal.helpers.Player.withMeat;
+import static internal.helpers.Player.withMoxie;
 import static internal.helpers.Player.withMuscle;
 import static internal.helpers.Player.withNotAllowedInStandard;
 import static internal.helpers.Player.withOverrideModifiers;
@@ -30,6 +31,7 @@ import static internal.helpers.Player.withSign;
 import static internal.helpers.Player.withSkill;
 import static internal.helpers.Player.withStats;
 import static internal.matchers.Maximizer.recommends;
+import static internal.matchers.Maximizer.recommendsEffect;
 import static internal.matchers.Maximizer.recommendsSlot;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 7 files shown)


test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -1165,6 +1167,74 @@ public void equipAtMostOneHalo() {
...
+              withEffect("Quiet Desperation"),
+              withSkill("Disco Smirk"),
+              withSkill("Quiet Desperation"));
+
+      try (cleanups) {
+        assertTrue(maximize("moxie"));
+
+        assertThat(getBoosts(), not(hasItem(recommendsEffect("Disco Smirk"))));
+      }
+    }
```
midgleyc, line 1202: for these two tests, put the two assertions in the same test.

same for the lower two tests.

you want to test "given a replacable mutex, we recommend the better effect only" and combining the assertions shows that intent more clearly

midgleyc (changes requested):
one comment, looks good otherwise.

For the wish / cast, you would have to send the effects all the way through, and then split on whether the command starts with "cast" or not. It is viable but probably not worth it.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
VeeArrKoL: Unless I'm misunderstanding something about the testing framework, combining these tests in this way won't work. In the first test, we HAVE Disco Smirk (+10 Moxie) and want to ensure we DO recommend Quiet Desperation (+25% Moxie, a net gain of 15 Moxie); whereas in the second test, we HAVE Quiet Desperation and want to ensure we DO NOT recommend Disco Smirk (because while in a vacuum it would be +10 Moxie, it's a net loss because it would also result in losing 25 Moxie from Quiet Desperation being removed).

midgleyc: oh yeah, you're right, I missed that the setups were different. my bad. I thought they were both "given we have both skills, suggest only the better effect"

midgleyc (approved):

PR #3433, opened 2026-05-01
feat: cache prices from mallprices.txt if from same day

midgleyc:
This works, but does not fix the caching problem.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/MallPriceDatabase.java
@@ -91,13 +91,12 @@ public class MallPriceDatabase {
         Price p = MallPriceDatabase.prices.get(id);
         if (p == null) {
           MallPriceDatabase.prices.put(id, new Price(id, price, timestamp));
-          ++count;
-          ++MallPriceDatabase.modCount;
         } else if (timestamp > p.timestamp) {
           p.update(price, timestamp);
-          ++count;
-          ++MallPriceDatabase.modCount;
         }
+        MallPriceManager.cachePriceIfFromCurrentRolloverDay(id, price, timestamp);
+        ++count;
+        ++MallPriceDatabase.modCount;
       }
     } catch (IOException e) {
       StaticEntity.printStackTrace(e);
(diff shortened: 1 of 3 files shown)


jaadams5:
What is the caching problem?

jaadams5 (approved):

midgleyc:
Many scripts, such as [garbage-collector](https://github.com/loathers/garbage-collector), will open the day by gathering prices for all items, using the "allitems" category search.

However, later `mall_price` calls (with no age specified) will search the mall again, instead of using the cached data.

And we can tell that immediately after searching "allitems", the data are cached into MallPriceManager.mallPrices. But somehow, later on that cache must be cleared or something, because the `mall_price` calls search the mall directly.

When the session is cleared for other reasons (e.g. log out and close Mafia), the category search will be run again -- that's expected behaviour due to how [mall-check](https://github.com/loathers/mall-check) works. That's sort of an issue, in that there's no need for it to run again, but practically not an issue.

This was surfaced by an implicit restriction of 1 mall search per minute, targeted at bots, hitting some players. That's a limit of 1440 mall checks per day per character, and the allitems calls hit 281 of those at the moment. When checking the session logs, I found that despite the allitems searches being made, we still search the mall manually afterwards sometimes, for reasons I can't fathom.

jaadams5:
Thank you.

I remember things that didn't happen and I am not looking at code to confirm or deny any evidence of my memory.  But I recall a time when one of the variations of getting a mall price was deliberately coded to avoid using cached results.  There was a belief, unsupported by data, that mall prices changed rapidly enough that the server hit should always be made.  There was the hope that players who could live with historical prices used those instead.  There was also a hope, existing as a scripting idiom, that buy(someItem, 1, mall_price(someitem)) would "always" succeed.

https://wiki.kolmafia.us/index.php?title=Mall_price says

> Returns the current mall price of the given item, ignoring the first five items listed to compensate for stores with limits and min-priced sales. To prevent abuse, this will only perform an actual Mall search once per item per session; subsequent calls for the same item will return a cached value. The cache will be updated after attempts to purchase the item using the CLI or ASH buy() command (even unsuccessful ones), but not in any other conditions, including purchasing with the purchase tab.

so that may help find why caching is not working as expected or may expand the definition of "works".

PR #3434, opened 2026-05-01
Reset olfactedMonster on ascension

jaadams5:
https://kolmafia.us/threads/settings-not-being-updated-at-the-right-time.32539/#post-179693


midgleyc (approved):
Thanks!

PR #3442, opened 2026-05-09
Limiting Wet Crap For Sale to characters in 11,037 Leagues Under the Sea

VeeArrKoL:
Fixes an issue where the maximizer would recommend buying scrolls from the Wet Crap For Sale shop, even if the user is not in the appropriate path.


midgleyc (approved):
Thanks, good spot!

PR #3443, opened 2026-05-09
Correctly consider equipping items with unarmed bonuses

VeeArrKoL:
Currently, the maximizer will not correctly consider items with bonuses for unarmed characters under typical circumstances. Because the list of candidate items is created by only replacing the new item and seeing if that results in an improvement, items with unarmed bonuses will usually not be considered (as they will not provide their benefits unless other things happen as well). This case is similar to the cases of outfits and other synergies which the maximizer handles. This change adds another case to that special handling to address items with unarmed bonuses.

The added test describes the kind of situation where the current implementation fails:
- The player is currently armed and has item(s) in the slot(s) where the candidate item would go.
- Unequipping the weapon and equipping the item with an unarmed bonus is overall better.

However, I'm not entirely happy with the solution I came up with. While it functions, I think it has the following deficiencies which I think could use some feedback:

- It uses the String representation of an item's modifiers to determine whether or not it has an unarmed bonus. While effective, this feels icky.
  - I couldn't find any straightforward way to iterate over the modifiers to check them directly, and there doesn't seem to be any other form of helper method available.
- It forces all items with unarmed bonuses to always be considered, regardless of whether the bonus from being unarmed will actually be useful.
  - The methods used at this stage of creating the list of candidates all seem to evaluate the modifiers using the player's current status (including being unarmed), so testing this would probably involve expanding the framework to support it. (Or maybe I'm missing a straightforward way to accomplish it.)
  - At present, this is probably not problematic as-is; there are very few items that even grant unarmed bonuses. It could be a problem if TPTB ever put out Master of the Surprising Fist II: Unarmed Equipment Emporium Edition.


jaadams5:
As a sanity check I would like a test where no weapon is equipped, the time halo is equipped and the maximizer equips a weapon, and something besides the halo.  Could complicate things by manipulating Off Hand items that do or do not contribute as well.

If I wanted an alternate implementation I might look at watches because they do not stack.  Perhaps the time halo does not stack with any weapon?  Alternatively I might define a pseudo-outfit that had the time halo and $NONE as the weapon.  

That said things might get too fiddly because special casing items in code is almost never the best solution.  

Sine we allow conditional modifiers on attributes that the Maximizer knows about perhaps the best way it is to allow UNARMED as a conditional modifier in the data and change code to support it?

Just ideas.  Like you I am not especially happy with the list of negative side effects.

VeeArrKoL:
I didn't track down specifically why, but the current maximizer implementation seems to handle cases with an open slot correctly. (e.g. If you have a weapon equipped and an open accessory slot, it will correctly recommend the time halo, even if it had other options.) I'll see if I can come up with a test case where it gets it wrong on an existing empty slot.

Ironically, unarmed used to be a modifier on the item but was changed to an expression variable  in #3275 to fix another issue, but it would be nice to have it back as a non-functional marker. Perhaps rather than specifically adding a redundant Unarmed modifier in the data, whatever parses the data files into modifiers could detect the use of unarmed in an expression and add the Unarmed marker modifier?

VeeArrKoL:
So the current implementation seems to correctly handle equipping a weapon if it's better than an existing unarmed bonus, but the way I had originally implemented the change was stopping the now-useless unarmed bonus item from being removed and replaced by something worse-if-unarmed-but-better-if-armed (because it was filling up the candidate list with unarmed items, I guess). Pushed a change that fixes that and a test to cover it.

VeeArrKoL:
Another update here. Made two major changes:

First, added a `BooleanModifier` that indicates that an item has an unarmed bonus. This modifier is automatically applied when parsing the modifiers from the data file if any of the expressions use the "unarmed" expression variable. This lets the maximizer code check whether an item has an unarmed bonus without resorting to just converting the `Modifiers` object to a string.

Second, rather than always considering every single item with an unarmed bonus, there's code to evaluate whether the item actually has a relevant unarmed bonus. The code to do this is a bit shakier than I'd like, but it seems to work. Because the expression evaluator uses the current player state to calculate the value of the expression and doesn't seem to have any mechanism to specify an override, I accomplish this by temporarily setting `Modifiers.unarmed` to true. Perhaps there's a better way to do this speculative calculation (surely the maximizer itself does it somehow?), but I couldn't figure out a cleaner way.

jaadams5:
> Because the expression evaluator uses the current player state to calculate the value of the expression and doesn't seem to have any mechanism to specify an override, 

This is a feature of the current implementation .  I, at least, have been overwhelmed by the effort needed to change it.

VeeArrKoL:
> This is a feature of the current implementation . I, at least, have been overwhelmed by the effort needed to change it.

YOLO

#3462

src/net/sourceforge/kolmafia/persistence/ModifierDatabase.java
```diff
@@ -735,6 +735,9 @@ public static final Modifiers parseModifiers(final Lookup lookup, final Modifier
         } else {
           newMods.addExpression(
               new Indexed<>(mod, ModifierExpression.getInstance(matcher.group(2), lookup)));
+          if (matcher.group(2).contains("unarmed")) {
+            newMods.setBoolean(BooleanModifier.HAS_UNARMED_BONUS, true);
```
midgleyc: I don't think this should be a modifier. As a modifier, it'll show up on e.g. StringModifier.MODIFIERS, will be written when writing file overrides...

I think it just wants to be a property on the Modifiers object that you can check with `mods.hasUnarmedBonus()` or something.

midgleyc (approved):
Thanks!

Converting to a Map does make a difference in that previously, you could have had multiple ModifierExpressions for the same Modifier, but code-wise we always tried to combine them, so I don't think this happened in practice. But if it did, this would be a behavioural change.

PR #3444, opened 2026-05-09
refactor: merge parallel entities in FamiliarDatabase into single Map

midgleyc:
Small behavioural changes: setting image now updates the image -> familiar map, modifiers now uses the effective id when determining if a familiar is a fairy or not. I think these are the right moves.


jaadams5 (approved):

midgleyc:
Last up is ItemDatabase, but I also want to make more methods in SkillProxy and FamiliarProxy use the stored data class instead of going via Skill/FamiliarDatabase, which means moving the contents of the non-trivial methods into the data classes.

PR #3449, opened 2026-05-13
Integrate JLine into KolMafia as a GUI replacement

libraryaddict:
This integrates [https://jline.org/](https://jline.org/) into KoLMafia for use when the GUI is not enabled. It also adds timestamps when JLine is in use.

JLine allows for better CLI interaction, such as navigating the command buffer with the arrow keys or scrolling through history. It prints console output above the input, very useful while you are typing something.

image

JLine in this PR starts up relatively late during boot, allowing other text to be printed to console, which is the reason behind the first line of the image not having a timestamp. 

This PR accomplishes the following when JLine is used, which is only when the GUI is disabled.

* JLine support
* Timestamps in the CLI
* Brightens dark true colors, which are otherwise hard to read against a dark background
* Intercepts Ctrl+C to first abort scripts, then shut down KoLMafia cleanly, then kill the process.
* Intercepts all System.out and System.err usage to pipe into JLine, which avoids any large refactoring
* JLine history files, which is not shared with KoLMafia’s gCLI command buffer
* Exceptional commands no longer kills the command retriever thread
* Adds the startup command to JLine's command history

A few notes:

1. I am well aware that the history implementation in this PR may be seen as undesirable. I do not believe the preferences file is a suitable place to store commands, maintainers are welcome to make their changes.
2. True colors can be quite dark and unreadable, which is the reason behind the color mutation. CLI colors are mutated even if not using JLine, but can be disabled via preference.
3. The JLine implemented here does not turn every line into its own command, this was not by design, but I have found useful. This could probably be addressed by splitting the string being read, but this also suits my use case, since I can test multiline scripts by pasting them in from an editor. That said, it will break other things, such as aliases, which do not expect newlines and will never save properly. To restate, this allows multiple lines to be pasted in and executed as a single command, this does not override the behavior when pressing `Enter`.
4. Commands that rely on html formatting for rendering will continue to be broken, such as `prefref`. I don't have a solution for this.

I am not interested in personally working on this further. For my builds of KoLMafia, this already does what I want, and I am happy for someone else to take it over. I made this PR for that purpose, and do not expect it to be accepted in it's current state.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -110,6 +110,7 @@ dependencies {
   implementation("com.jgoodies:jgoodies-binding:2.13.0")
   implementation("org.eclipse.jgit:org.eclipse.jgit:7.4.0.202509020913-r")
   implementation("org.eclipse.jgit:org.eclipse.jgit.ssh.apache:7.4.0.202509020913-r")
+  implementation("org.jline:jline:4.1.0")
 
   checkstyle("com.puppycrawl.tools:checkstyle:${checkstyle.toolVersion}")
 }
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -17,6 +17,7 @@ global	allowNonMoodBurning	true
 global	allowSummonBurning	true
 global	autoLogin
 global	autoHighlightOnFocus	true
+global	brightenDarkCLIColors	true
 global	broadcastEvents	true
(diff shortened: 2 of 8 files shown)


libraryaddict:
Closing this PR as there's a few issues that I need to iron out, and, I don't think the interest for others is really there.

PR #3450, opened 2026-05-13
fix: parse the legendary noodles eating result.

Opassa:
This fixes the parsing of eating legendary noodles dishes.  I have tested it with the relay browser, the CLI, and an ash script.  I ran my full daily scripting with this version, and after fixing a bug I introduced, my scripts run correctly (it had broken behavior of other "use item -> choice" items under ash).  I don't know if there's additional testing that needs to be done, but I have run this version for 2 days now, and it all seems to be working correctly for me.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
+++ b/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
@@ -320,14 +320,6 @@ public class ResponseTextParser {
           SummoningChamberRequest.parseResponse(location, responseText);
         } else if (location.contains("whichchoice=1278")) {
           ClanFortuneRequest.parseResponse(location, responseText);
-        } else if (responseText.contains("How do you want to digest the legendary noodles?")) {
-          AdventureResult item = ChoiceManager.lastItemUsed;
-          if (item != null) {
-            UseItemRequest.setLastItemUsed(item);
-            UseItemRequest.parseConsumption(responseText, false);
-            SpadingManager.processConsumeItem(item, responseText);
-            UseItemRequest.clearLastItemUsed();
-          }
         }
       }
       case "clan_hall.php" -> {
--- a/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
+++ b/src/net/sourceforge/kolmafia/session/ResponseTextParser.java
@@ -320,6 +320,14 @@ public class ResponseTextParser {
           SummoningChamberRequest.parseResponse(location, responseText);
         } else if (location.contains("whichchoice=1278")) {
           ClanFortuneRequest.parseResponse(location, responseText);
+        } else if (responseText.contains("How do you want to digest the legendary noodles?")) {
+          AdventureResult item = ChoiceManager.lastItemUsed;
+          if (item != null) {
+            UseItemRequest.setLastItemUsed(item);
+            UseItemRequest.parseConsumption(responseText, false);
+            SpadingManager.processConsumeItem(item, responseText);
+            UseItemRequest.clearLastItemUsed();
+          }
         }
       }
       case "clan_hall.php" -> {
--- a/src/net/sourceforge/kolmafia/request/UseItemRequest.java
+++ b/src/net/sourceforge/kolmafia/request/UseItemRequest.java
@@ -1416,7 +1416,7 @@ public class UseItemRequest extends GenericRequest {
       // which GenericRequest automates in FightRequest.INSTANCE, we
       // automate this ourself in runOneIteration. Item consumption
       // was handled in GenericRequest.checkItemRedirection, so punt.
-      if (!this.responseText.contains("How do you want to digest the legendary noodles?")) return;
+      return;
     }
 
     switch (this.consumptionType) {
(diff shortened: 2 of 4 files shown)


midgleyc:
I think we want to put the logic in `ChoiceControl.visitChoice`, not in `ResponseTextParser`.

Also, the second `if` in UseItemRequest can be merged with the first (with `&&`) and a comment should be added explaining.

jaadams5:
I fixed and ran spotless because it seemed like the friendly thing to do.  Also as part of *my* re-education commented on the code.

src/net/sourceforge/kolmafia/session/ResponseTextParser.java
```diff
@@ -320,6 +320,14 @@ public static final void externalUpdate(final GenericRequest request) {
           SummoningChamberRequest.parseResponse(location, responseText);
         } else if (location.contains("whichchoice=1278")) {
           ClanFortuneRequest.parseResponse(location, responseText);
+        } else if (responseText.contains("How do you want to digest the legendary noodles?")) {
+          AdventureResult item = ChoiceManager.lastItemUsed;
+          if (item != null) {
```
jaadams5: I am not certain how to combine the ifs.  item is needed after the null check so it needs to be declared and set somewhere.   

null != (AdventureResult item = ChoiceManager.lastItemUsed) doesn't compile for me.

Admittedly I tend to avoid similar syntax because it is harder to comprehend since setting and declaring the variable seems like a side effect in the context of a control statement but that is definitely a me problem :-)

midgleyc: I meant the `if`s in the other place that got changed ;)

```
if (UseItemRequest.currentURL.startsWith("choice.php") && !this.responseText.contains("How do you want to digest the legendary noodles?"))
```

jaadams5: Where's my embarrassed emoji?

Thank you.

Opassa:
I'll work on moving the the logic into ChoiceControl.visitChoice instead, but I'll want to give it a couple days of testing once I've made the change to make sure I don't break anything.

Opassa:
I have moved the logic into ChoiceControl.  It made the implementation much simpler, only 1 file modified from the main branch.  It eliminated the ugly check in UseItemRequest.java.  I tested it with relay browser, CLI, and ash scripting, and it works correctly in all 3 cases.  I am currently running my daily scripting routine against it, and it doesn't seem to be causing any issues.

jaadams5 (dismissed):
@midgleyc - figured I'd let you make the final decision and commit.

midgleyc:
I don't see how this can work properly -- isn't `UseItemRequest.getLastItemUsed()` null? You could move the setting of this and the id above where we return due to detecting a choice.

Also I expected `visitChoice` instead of `postChoice0` -- `postChoice0` is after you make a decision, so seems too late?

Opassa:
UseItemRequest.getLastItemUsed() is set to null later in the process, during ChoiceManager.visitChoice() right before it calls ChoiceControl.visitChoice(), it's why my initial code had to grab it from ChoiceManager.lastItemUsed and assign it to UseItemRequest.setLastItemUsed().  visitChoice() is called from postChoice1() and postChoice2().  I think the code is cleaner if it's called in postChoice0() than in visitChoice(), because I don't have to muck around with UseItemRequest.setLastItemUsed() and .clearLastItemUsed().  In postChoice0() it's in the ideal state it needs to be in to handle the parsing.  Anywhere else requires messing around with the current state.  But if you really want it in visitChoice(), I can move it there.

midgleyc (approved):
Thanks!

midgleyc:
This looks to be adding fullness twice -- exactly what wasn't working before this PR?

Opassa:
I can't reproduce the adding fullness twice.  I would be happy to debug it if you can provide instructions on how to reproduce it.  When I run it, I see fullness get incremented exactly once, in ResultProcessor.java method tallyResult() line 1313.

What wasn't working is the stuff that happens in EatItemRequest.parseConsumption():
1. Doesn't subtract food eaten from session results.
2. Doesn't subtract Special Seasoning consumed from session results.
3. Doesn't mark mayo as consumed via mayoInMouth property set to blank.
4. If you consume mayo from mayo minder, doesn't subtract mayo consumed from session results.
5. Doesn't record milk of magnesium charge consumed (although this is a bug in KoL itself, it's not sending the proper message, I have bug reported this).
6. I don't know if legendary noodles trigger carboloading (it probably does), but if so, that should get reset to 0.
7. I didn't test this, but mini kiwi aloi charge consumption.
8. Can only test this once a day, so I didn't test it thoroughly: Bowl of Infinite Jelly consumption message to set _infiniteJellyUsed flag.

There is a comment in EatItemRequest.java line 1109:

    // If the user has fullness display turned on ( "You gain x
    // Fullness" ) DON'T touch fullness here.  It is handled in
    // ResultProcessor.
    if (!responseText.contains(" Fullness")) {

It always skips past the if statement for me because responseText always contains "Fullness".  I'm wondering if your result has something to do with that, since I don't understand what the comment is referring to.  I don't know of any way to turn off the fullness display in the consumption message.  I tried turning off "Display Fullness Counter" in KoL settings, but that resulted in fullness not getting incremented at all.

midgleyc:
Okay, so the double parsing is because it parses the choice once per time your browser (or anything else) hits the page, and you can't walk away because you're stuck in the choice so all requests send you back there.

Also looks like I was wrong about postChoice0/1/2: I thought those were only for when we actually had a nonzero decision, but looks like they get hit on a zero decision too.

I guess the easiest move is to just hit api.php after the choice completes and get the right results back from that.

PR #3453, opened 2026-05-18
refactor: use try-with-resources to automatically close streams

midgleyc:


jaadams5 (approved):
Motivation?

midgleyc:
In general: ensures the stream is closed even if an exception happens in the middle. Also cleanliness, less explicit code.

In specific: I changed a bunch of these in previous refactors which put it at the forefront of my mind as a cleanup I wanted done.

midgleyc:
Also while testing session_logs on a real Mafia instance: it takes a crazy long time to print the session logs to the screen. They're about 6kb each. Admittedly they're on a network drive, but opening the file in Notepad takes a few seconds, whereas posting the logs took many minutes.

PR #3455, opened 2026-05-18
refactor: merge parallel entities in ItemDatabase into single Map

midgleyc:


midgleyc:
Test shows an error: the entrysets are no longer sorted like they were, which is my mistake, and also affects some other classes I refactored. I'll fix tomorrow.

jaadams5 (approved):
Assuming tests pass after I spotlessed :-)

midgleyc:
This project comprised 
* EffectDatabase: #3435 
* BountyDatabase: #3437
* RestoresDatabase: #3439
* EquipmentDatabase: #3440
* SkillDatabase: #3441 
* FamiliarDatabase: #3444, #3456 
* ItemDatabase: here, #3455 

I lost steam towards the end. This was inspired by every time I would look at all the hashmaps on FamiliarDatabase or SkillDatabase or EffectDatabase and lose motivation to work on whatever I was trying to do in there.

As expected, I also lost motivation towards this in general, but at least the first step was done.

PR #3459, opened 2026-05-20
feat: add a connection timeout of non-infinity

midgleyc:
This should prevent scripts (e.g. ChIT) hanging when zachbardon.com does not respond.

There is a potential issue with `pull all` not responding in time, but I think the connection itself completes within time -- just not the whole request. The timeout on the request itself is still infinity.


jaadams5 (approved):
Trigger?  I've had scripts that failed because zachbardon.com did not respond but those were fixed (elsewhere) in mafia.  Just curious about the different execution path.

midgleyc:
There was a discussion on Discord near the start of the month at https://discord.com/channels/466605739838930955/534709227260870656/1500365570787512471.

There were reports that Mafia stopped adventuring because it was unable to parse the [ChIT](https://github.com/loathers/ChIT) charpane, or that attempts to run ChIT were preventing other scripts from running.

Scripts using https://kolmafia.us/threads/make-your-scripts-automatically-update-their-data-files.1515/ apparently hung.

PR #3462, opened 2026-05-21
Implement framework for Expression evaluator overrides

VeeArrKoL:
Currently, the expression evaluator always uses the current character/mafia state when evaluating expressions. This can make it difficult to perform a simple speculative evaluation of an expression under conditions that aren't currently true. This pull request implements a framework for overriding aspects of the current state while evaluating an expression. Two types of override are provided as a proof-of-concept; it is probably prudent to expand the available overrides only as they are actually needed.

For context, the motivation for this is to be able to quickly evaluate potential efficacy of items with unarmed bonuses within the maximizer, in #3443.


src/net/sourceforge/kolmafia/ExpressionOverrides.java
```diff
@@ -0,0 +1,63 @@
+package net.sourceforge.kolmafia;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+/**
```
jaadams5: There are tags in the comment that are not rendered in my IDE.  I would prefer an unformatted comment.

src/net/sourceforge/kolmafia/ExpressionOverrides.java
```diff
@@ -0,0 +1,63 @@
...
+ * </ol>
+ */
+public class ExpressionOverrides {
+  // Ensure that nothing mutates this.
+  public static final ExpressionOverrides NONE = new ExpressionOverrides();
+
+  // region Overrides for user preferences
+  private Map<String, String> prefOverrides = null;
+
+  public void setPrefOverride(String pref, String value) {
```
midgleyc: naming: this class is already all about overrides, I think `setPref`, `pref`, `setUnarmed`, `unarmed` are clear enough and simpler. In particular, I think the shorter getters are clearer than the longer ones, e.g. `overrides.unarmed().orElse[...]` vs `overrides.overrideUnarmed().orElse`

midgleyc:
Could you extend 3443 to show how you'd use this, if it were merged?

VeeArrKoL:
> Could you extend 3443 to show how you'd use this, if it were merged?

Ugh, I thought I had a plan for this, but there's actually a gap in it now that I'm looking more closely at how I implemented this. I'll need to think on it a bit.

VeeArrKoL:
> Ugh, I thought I had a plan for this, but there's actually a gap in it now that I'm looking more closely at how I implemented this. I'll need to think on it a bit.

On reflection, I suppose it would look something like expanding `Modifiers.override(Lookup)` to also take a `ExpressionOverrides` to pass to its `eval()` calls for expression evaluation, but since the result of the evaluations gets baked into the `Modifiers` you need to make sure to use it right (i.e. make a copy first) to avoid poisoning the values in `ModifierDatabase`. On the other hand, I guess that that's *already* a risk if someone calls `override()` in a dumb way, so this doesn't really make that *worse*.

Having done that, rather than doing the janky thing to temporarily mutate the value of `Modifiers.unarmed`, do:

```
ExpressionOverrides overrides = new ExpressionOverrides();
overrides.setUnarmed(true);
Modifiers unarmedMods = new Modifiers(ModifiersDatabase.getItemModifiers(id)); // Make a copy to avoid poisoning the DB data
unarmedMods.override(null, overrides); // Recalculate the modifiers as if we were unarmed, Lookup=null will make it *only* update expressions
double score = this.getScore(unarmedMods, Map.of(Slot.NONE, item));
```

Thoughts?

midgleyc:
> unarmedMods.override(null, overrides); // Recalculate the modifiers as if we were unarmed, Lookup=null will make it *only* update expressions

Rather than this, make a dedicated method for it, imo. Like `recalculate(overrides)`.

The other method you make can call the first one with null, but I think when we come back to this in several years the first one will be confusing. Especially because what we currently use "override" to mean is nothing to do with the new "overrides" you've implemented.

But other than that, if it works this is really good! The other static we're awkwardly storing on Modifiers that I recall is familiar data.

VeeArrKoL:
Updated the modifier class to have the recalculate method. 

Also updated the copy constructor to maintain expressions since it didn't previously. It's unclear why `expressions` is a `List<Indexed<Modifier,ModifierExpression>>` rather than a `Map<Modifier,ModifierExpression>`, but I left it as-is for now despite it being awkward to deal with.

midgleyc:
> It's unclear why expressions is a List<Indexed<Modifier,ModifierExpression>> rather than a Map<Modifier,ModifierExpression>, but I left it as-is for now despite it being awkward to deal with.

I suspect historical reasons: it started out as a `ModifierExpression[]`, #1389 converted it to `ArrayList<Indexed<ModifierExpression>>` where `Indexed<T>` was a `Pair` of `DoubleModifier` and `T`, and then #1463 converted it to `ArrayList<Indexed<DoubleModifier, ModifierExpression>>` genericising the `Indexed`. So nobody who touched this thought "hmm, maybe this should be a `LinkedHashMap`".

midgleyc (approved):
Thanks!

PR #3463, opened 2026-05-22
Fix choice decoration for NC number 208

VeeArrKoL:
Fixes issue from https://kolmafia.us/threads/ancient-hobo-burial-ground-nc-choice-208-description-is-wrong.32498/


midgleyc (approved):
Thanks!

PR #3464, opened 2026-05-23
Migrated from Java 5 

jaadams5:
Ran Intellij"s anaylzer.  Addressed things that were flagged as Java 5 features that could be migrated so something newer.  29 files are a lot but should be to scan.


jaadams5:
Didn't see any places where the unchanged code followed a coding pattern that was broken when changed.

midgleyc (approved):

PR #3466, opened 2026-05-24
Correctly handle situations where tiny stillsuit is equipped to multiple familiars

VeeArrKoL:
Fixes issue from https://kolmafia.us/threads/tiny-stillsuit-not-tracking-distillate-accurately-when-equipped.32570/


midgleyc (approved):
Thanks!

wow I want multiple stillsuits

PR #3465, opened 2026-05-24
feat: support legendary noodles consumption choice

midgleyc:

PR #3467, opened 2026-05-26
fix: don't double parse the legendary noodle digestion messages.

Opassa:
This fixes the multiple parsing of the legendary noodle digestion messages by parsing it once when it first parses the choice, and then excluding all of the normal parsing when it contains "Legendary Digestion".  This prevents any page refreshes from re-parsing the fullness gain, adventure gain, stat gain, etc.

Also it adds a handler for if you digest with your spleen, since that -fullness/+spleen modification wasn't being noticed by KoLmafia.


gausie:
Apply the automated formatting to this to pass tests. On the subject of tests, please add some unit tests to this. Ideally construct a unit test that fails before your change and passes afterwards.

midgleyc:
Preferably edit the existing EatItemRequestTest that covers noodles.

Dealing with spleen was covered in #3465.

Opassa:
I removed the spleen handling, and added to the legendary noodles test to prove that it doesn't double parse.

Opassa:
I do not know how to apply auto-formatting.  I'm using IntelliJ, and it has its own ideas about formatting, some of which I've had to go through line by line and manually revert when it changed formatting in other code accidentally.

midgleyc:
Run `./gradlew :spotlessApply` from the terminal.

In IntelliJ, this is Gradle -> Tasks -> verification -> spotlessApply.

midgleyc (approved):
Thanks!

PR #3470, opened 2026-05-27
Add tests for unknown preferences in data files

libraryaddict:
The first commit will be a failing test. Unless things break, then the second commit will be a failing test, repeat ad infinitum until the test successfully fails.

This PR aims to try prevent mistakes from occurring when creating preferences in data files, where a preference is referred, but does not exist in defaults.txt


midgleyc (approved):
Thanks!

PR #3474, opened 2026-05-28
More syntax upgrades from Java 5

jaadams5:
Ran Intellij analyze.  Just changed things that could be migrated from Java 5.  Lots of files but at least the changed lines are obvious and adjacent :-)


src/net/sourceforge/kolmafia/textui/command/CleanupJunkRequest.java
```diff
@@ -68,8 +67,8 @@ public static void cleanup() {
     do {
       madeUntinkerRequest = false;
 
-      for (int i = 0; i < items.length; ++i) {
-        currentItem = items[i];
+      for (AdventureResult item : items) {
+        currentItem = item;
```
midgleyc, line 71: just to note here: so `currentItem` is redefined here because it's declared outside the loop. But it's never used outside a loop and always reset inside a loop, so it could just not be declared outside the loop and purely used as a loop variable.

But this is also fine.

Reusing variables is something the code does a lot that modern code would not. It probably had more severe performance concerns back in the day.

midgleyc (approved):

src/net/sourceforge/kolmafia/textui/command/CleanupJunkRequest.java
jaadams5: Noted.  The IDE correctly identified this as a case of a variable being reused but the inspection I am using does not flag it for possible automated action.  Thanks.

PR #3475, opened 2026-05-29
Fix astral energy drink missing from the item manager's Spleen panel

VeeArrKoL:
Fixes https://kolmafia.us/threads/astral-energy-drinks-dont-appear-in-item-manager.29865/

A few notes:
* This changes which consumable name is used by the `ConsumablesDatabase` code that automatically calculates the stats for the old astral energy drinks. I doubt this breaks anything; are these things even obtainable anymore?
* It changes the `UseItemEnqueuePanel` to use the disambiguated name of an item to calculate the characteristics of its consumable. This didn't appear to break anything or make anything display strangely, but I'm not 100% sure where exactly that name gets used.

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/Concoction.java
+++ b/src/net/sourceforge/kolmafia/objectpool/Concoction.java
@@ -197,20 +197,15 @@ public class Concoction implements Comparable<Concoction> {
 
   public ConcoctionType computeType() {
     int itemId = this.getItemId();
-    if (ConsumablesDatabase.getRawFullness(name) != null) {
+    if (forceFood.contains(itemId)) {
       return ConcoctionType.FOOD;
-    } else if (ConsumablesDatabase.getRawInebriety(name) != null) {
+    } else if (forceBooze.contains(itemId)) {
       return ConcoctionType.BOOZE;
-    } else if (ConsumablesDatabase.getRawSpleenHit(name) != null) {
-      return ConcoctionType.SPLEEN;
     } else
       return switch (ItemDatabase.getConsumptionType(itemId)) {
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/objectpool/Concoction.java
```diff
@@ -197,20 +197,15 @@ public Concoction(final String name, final String property) {
...
-      return ConcoctionType.SPLEEN;
     } else
       return switch (ItemDatabase.getConsumptionType(itemId)) {
-        case FOOD_HELPER -> ConcoctionType.FOOD;
-        case DRINK_HELPER -> ConcoctionType.BOOZE;
-        case USE, USE_MULTIPLE ->
-            forceFood.contains(itemId)
-                ? ConcoctionType.FOOD
-                : forceBooze.contains(itemId) ? ConcoctionType.BOOZE : ConcoctionType.NONE;
+        case EAT, FOOD_HELPER -> ConcoctionType.FOOD;
```
midgleyc, line 204: with this change you should be able to remove magical sausage + quantum taco from the `forceFood` set because they're both 0-fullness food

`[glitch season reward name]` should also be fine because we consider it to have primary type food even if on KoL it's usable.

The drunki-bears should both be food, same as before, because we preferred categorising something as food to booze.

so I think this is fine despite the logic change.

VeeArrKoL: Removed magical sausage + quantum taco from `forceFood` and added a test to verify that they still get `ConcoctionType.FOOD`.

midgleyc (approved):
Thanks!

PR #3479, opened 2026-05-30
More Java 5 migration

jaadams5:
A couple of unelated nits plus some unboxing and and contains.

Somehow this is including previously committed and approved changes.  Will try and fix that.  And did.


midgleyc (approved):

src/net/sourceforge/kolmafia/persistence/ConsumablesDatabase.java
```diff
@@ -522,7 +522,7 @@ private static int adventureFlagsToKey(
         | (gainEffect1 ? 1 : 0) << 3
         | (gainEffect2 ? 1 : 0) << 2
         | (gainEffect3 ? 1 : 0) << 1
-        | (gainEffect4 ? 1 : 0) << 0;
+        | (gainEffect4 ? 1 : 0);
```
midgleyc, line 525: while this is useless, it's symmetric with the lines above.

but also I think doing nothing for binary units is also common, so this is fine

jaadams5:
Thank you.  I'm sorry this was such a slog to get fixed.

PR #3490, opened 2026-06-05
feat: retry upon receiving a 502

midgleyc:
From choice adventures, if we receive a 502, the server has not handled our request (unlike a GOAWAY, where it has handled it). So we're safe to retry.

This is particularly seen during the Grimace map choice adventures, where you look for distention pills or synthetic dog hair pills.

We might want to limit this to only choice adventures (as the Grimace map ones are the only ones where I'm sure it's safe), but let's go with this for now and limit it if we see issues.


jaadams5 (dismissed):

jaadams5:
Before commit, check with @Veracity0 .  They have seen this often enough with the Monorail Breakfast Counter in their meat farming script that they have changed the script, at least to the point of telling users that it happened and what to do manually to resolve it.

midgleyc:
copy, I will leave this open for a few days for comments.

VeeArrKoL:
This feels like something that maybe wants a setting/preference to activate (enabled by default) so that if we run into a situation where the behavior is not wanted, people aren't stuck either reverting to a who-knows-how-old mafia version or waiting for a code update.

jaadams5:
> This feels like something that maybe wants a setting/preference to activate (enabled by default) so that if we run into a situation where the behavior is not wanted, people aren't stuck either reverting to a who-knows-how-old mafia version or waiting for a code update.

I don't understand your concern.  How do you think it would break things?  Thanks.

VeeArrKoL:
> I don't understand your concern. How do you think it would break things?

Currently all of the cases where we receive a 502 involve the server not processing the request at all, but imagine a scenario where we receive a 502 but the request *was* processed by the server. In that case, you may not want to automatically retry (if it was a request that uses up some kind of resource, for example).

jaadams5:
> > I don't understand your concern. How do you think it would break things?
> 
> Currently all of the cases where we receive a 502 involve the server not processing the request at all, but imagine a scenario where we receive a 502 but the request _was_ processed by the server. In that case, you may not want to automatically retry (if it was a request that uses up some kind of resource, for example).

OK.  I think you are afraid of things that I think will never happen.  But we may agree to disagree :-)

midgleyc:
Minimally, I can see worth in adding a property to disable the retries because I've set it up to do exponential backoff, and if e.g. a skill description 502s it could consume half a minute retrying a page which is always going to fail. I think the old bookshelf skills 503 sometimes.

Also it is possible the server could 502 after doing something. I think a 500 or 503 is more likely, but I don't know why the choices are 502ing to begin with.

jaadams5 (approved):
Ok to commit.  veracity0's opinion will be give later if necessary.

PR #3491, opened 2026-06-05
Maximizer: Properly consider left-hand man (and other familiars) for item requirements and bonuses

libraryaddict:
The first test is a failing test. The second commit is fixing spotless, whoops. The third commit is the changes to evaluator. 

Currently for maximizer, left-hand man is not considered for items beyond the normal modifiers. Such as `equip bread basket` and `1000 bonus bread basket` will not be considered.
This makes the maximizer fail when it requires bread basket to be worn, but the offhand slot is unavailable. Maximizer will only give left-hand man items that the modifier resolved with say `item drop, meat drop, 10000 bonus bread basket`, it will never be given a bread basket.

This PR aims to fix that, but, it seems to have an issue as seen in the commented out test, where the maximizer is now prioritizing the left-hand man for the equipment needs. Maybe this is a mistake, or maybe it's actually normal. I'm not sure. But the test can't pass if that comment is removed, and I don't know how to resolve it. I'm not sure if it needs to be resolved either.


midgleyc:
For the commented-out code: if you add "-familiar" does it work as expected?

I don't see any reason to prefer the actual offhand to the familiar offhand, given they both contribute enchantments.

Does it also work with "switch left-hand man"? Is there something else elsewhere in the codebase that makes it not try to equip items that can't be equipped?

Preferring familiar items to standard slots will be more of a problem for hatrack / pantsrack / disembodied hand, where the enchantments don't work on the familiar slot (but things like "letter r" for PvP *do* work). If this happens this is an issue.

libraryaddict:
> For the commented-out code: if you add "-familiar" does it work as expected?

Modified the test and yes

> Does it also work with "switch left-hand man"?

Added a test.

> Is there something else elsewhere in the codebase that makes it not try to equip items that can't be equipped?
 
https://github.com/libraryaddict/kolmafia/blob/lefthand-man/src/net/sourceforge/kolmafia/maximizer/Evaluator.java#L1150

It checks if the current fam can equip this item if it's not owned by the familiar slot. If the modifiers in the `getScore` are not optimal, it doesn't bother adding it to `ranked`.
Same thing for `switch fam`

It needs to be in `ranked` to be considered for the familiar slot.

If this change isn't implemented, the normal modifiers for the item are still considered, but the bonus score, which is passed via item reference, is not considered, and the item being required, is not considered. It has to be added to `ranked` if it wants to go in the familiar slot.

I don't really want to look at bigger changes than this, as it would be out of scope for the bug.

midgleyc (approved):
Thanks!

PR #3495, opened 2026-06-06
Fix cafe items not appearing in the Item Manager organ panels

VeeArrKoL:
Changes in #3475 caused the cafe items in Chez Snootee and Gnomish Microbrewery to no longer appear on the organ panels of the Item Manager. This corrects that case.


midgleyc (approved):
Thanks!

PR #3499, opened 2026-06-07
feat: warn when trying to free the king in meat path

midgleyc:
I've lost out a few times, so add a warning.

Also remove the plumber warning to overeat as that now prevents you from adventuring.


jaadams5 (dismissed):

Rinn:
if we're capped at 40 shouldn't we have a warning that you'd lose adventures if over?

midgleyc:
Huh. Maybe.

As you can buy adventures at any time, I've always had as few as possible, hence the addition of a warning for when you're below.

midgleyc:
Added extra warning, tested, works as expected.

PR #3504, opened 2026-06-09
Display Sword of S Words kills on char pane

VeeArrKoL:
Display progress of Sword of S Words kills on the char pane. I considered instead adding this as an entry in the list of familiar drops, but that felt like it would have been janky in a way that the existing "fake" drops (e.g. cookbookbat recipes) aren't.


midgleyc (approved):
Thanks!

PR #3505, opened 2026-06-10
feat: change LASS to loathers in script manager where relevant

midgleyc:
Philter is still on LASS.

### DIFF
--- a/data/SVN/svnrepo.json
+++ b/data/SVN/svnrepo.json
@@ -246,7 +246,7 @@
    "longDesc": "The purpose of this script is to automatically go to the Bounty Hunter Hunter, accept bounties, hunt them and turn in the completed bounties for lucre."
  },
  {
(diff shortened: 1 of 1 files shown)


jaadams5:
What is the effect of this change on people who have already installed the script from the old location?  I am no longer certain I remember how updates work but at one point when the script author moved the repo location, the user got no updates and eventually had to figure that out and manually uninstall and reinstall before things worked as expected.  I don't think dependencies are so specific that moving a repo breaks them but it is worth asking.

These questions are not a reason to block the commit.  Rather they suggest additional support might be needed when this is rolled out.

jaadams5 (approved):

midgleyc:
Everything will continue to work for the users with the script installed from the old location -- it redirects to the new one anyway.

The only difference is that the Script Manager will show the script as uninstalled, or the new location's script as installed.

PR #3507, opened 2026-06-10
Fix mall purchase parse failure for per-day limit

eegeeZA:
When a mall store has a per-day purchase limit and the server returns the
limit message, `parseResponse` fires "Mall purchase parse failure" because
the response's inner `<table>` contains no item row. This is a false
positive -- `processResults` already handles the per-day limit correctly
via `YIELD_PATTERN`.

The fix adds an early-return guard in `parseResponse` for this response,
consistent with the existing guards for other known non-purchase responses
(ignore list, frozen inventory, can't afford). A regression test is
included using the exact HTML from the debug log that surfaced the issue.


test/net/sourceforge/kolmafia/request/MallPurchaseRequestTest.java
```diff
@@ -147,4 +150,16 @@ public void canParsePrice(long price, String whichitem) {
       assertEquals(price, MallPurchaseRequest.priceFromStoreString(whichitem));
     }
   }
+
+  @Nested
+  class ParseResponse {
+    @Test
+    public void perDayLimitDoesNotSetError() {
+      StaticEntity.setContinuationState(MafiaState.CONTINUE);
```
midgleyc: nicer to use Cleanups for this -- `var cleanups = withContinuationState()`, and then `try (cleanups) {}` wrapping the rest of the setup / asserts

midgleyc (changes requested):
one request, other than that this is great, thank you!

midgleyc (approved):

PR #3508, opened 2026-06-11
Track some skills that expend charges even if they fail

VeeArrKoL:
Resolves some of the issues in #3412 

Specifically, tracks spending charges during failure of: 
* Heartstone: GONE
* Assert your Authority
* Transcendent Olfaction (sometimes)
* Feel Hatred


midgleyc (approved):
Thanks!

PR #3509, opened 2026-06-11
Fix maximiser ignoring equip keywords with -tie

eegeeZA:
## Problem

`maximize("-tie, club")` (and similarly for `sword`, `knife`, `utensil`, `accordion`, `shield`) does not recommend any qualifying item even when the character has one. Forum thread: https://kolmafia.us/threads/issues-with-maximiser-command.28691/

With no stat weights, every item scores 0.0. The simplicity comparison in `MaximizerSpeculation.compareTo` is not guarded by `isUsingTiebreaker()`, so it runs even with `-tie` active. The synthetic `(none)`/UNEQUIP item gets +2 simplicity (currently equipped) while a qualifying club scores 0, so UNEQUIP sorts first, its `Integer.MAX_VALUE` count fills the slot quota, and the club never enters `automaticEntry`.

The same mechanism breaks `shield` in the offhand slot: `maximize("-tie, shield")` with an equippable shield recommends nothing.

## Fix

**`MaximizerSpeculation.java`** -- in `getTiebreaker()`, add +3 simplicity for a non-UNEQUIP weapon when a weapon-type keyword is active, and for a non-UNEQUIP offhand when `shield` is required. The qualifying item then scores 3 vs UNEQUIP's 2 and enters the shortlist under the normal `total < useful` cutoff. No `automaticFlag` change is needed since qualifying items with `delta == 0` already reach the ranked list via the normal fall-through path.

**`Evaluator.java`** -- add `isWeaponTypeRequired()` and `isShieldRequired()` helpers used by `MaximizerSpeculation`.

## Tests

Added to `MaximizerTest.WeaponModifiers`:
- `clubModifierWorksWithoutTieBreaker`: `maximize("-tie, club")` recommends a lawn dart (a club) in the weapon slot.
- `shieldModifierWorksWithoutTieBreaker`: `maximize("-tie, shield")` recommends a vinyl shield in the offhand slot. This test fails without the fix.

This relates to PR #1576.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -980,6 +980,18 @@ public class Evaluator {
     return !this.noTiebreaker;
   }
 
+  boolean isWeaponTypeRequired() {
+    return this.requireClub
+        || this.requireUtensil
+        || this.requireSword
+        || this.requireKnife
+        || this.requireAccordion;
+  }
+
+  boolean isShieldRequired() {
+    return this.requireShield;
+  }
+
   enum Constraint {
     /** Item violates a constraint, don't use it */
     VIOLATES,
--- a/src/net/sourceforge/kolmafia/maximizer/MaximizerSpeculation.java
+++ b/src/net/sourceforge/kolmafia/maximizer/MaximizerSpeculation.java
@@ -99,6 +99,20 @@ public class MaximizerSpeculation extends Speculation
         this.simplicity += slot == Slot.WEAPON ? -1 : 1;
       }
     }
+    // When an equipment-type keyword (club, sword, shield, etc.) is active, prefer any
+    // qualifying item over leaving the slot empty: give it a nudge above UNEQUIP's +2.
+    if (Maximizer.eval.isWeaponTypeRequired()) {
+      AdventureResult weapon = this.equipment.get(Slot.WEAPON);
+      if (weapon != null && !weapon.equals(EquipmentRequest.UNEQUIP)) {
+        this.simplicity += 3;
+      }
+    }
+    if (Maximizer.eval.isShieldRequired()) {
+      AdventureResult offhand = this.equipment.get(Slot.OFFHAND);
+      if (offhand != null && !offhand.equals(EquipmentRequest.UNEQUIP)) {
+        this.simplicity += 3;
+      }
+    }
     return this.tiebreaker;
   }
 
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
@@ -804,6 +804,26 @@ public class MaximizerTest {
     }
 
     @Test
+    public void clubModifierWorksWithoutTieBreaker() {
+      final var cleanups = new Cleanups(withEquippableItem("lawn dart"));
+
+      try (cleanups) {
+        assertTrue(maximize("-tie, club"));
+        assertThat(getBoosts(), hasItem(recommendsSlot(Slot.WEAPON, "lawn dart")));
+      }
+    }
+
+    @Test
+    public void shieldModifierWorksWithoutTieBreaker() {
+      final var cleanups = new Cleanups(withEquippableItem("vinyl shield"));
+
+      try (cleanups) {
+        assertTrue(maximize("-tie, shield"));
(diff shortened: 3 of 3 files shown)


src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -1567,6 +1575,12 @@ void enumerateEquipment(EquipScope equipScope, int maxPrice, PriceLevel priceLev
           if (KoLCharacter.hasEquipped(item) && this.current) break gotItem;
           if (item.initial == 0) continue;
           if (item.automaticFlag) continue;
+          // Explicit weapon-type keywords (club, sword, etc.) are requirements, not score
+          // weights. Ensure qualifying weapons rank above UNEQUIP even with zero score.
+          if ((slot == Evaluator.WEAPON_1H || slot == Slot.WEAPON) && this.isWeaponTypeRequired()) {
```
midgleyc: should this not also check that the weapon is of the desired type?

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -1567,6 +1575,12 @@ void enumerateEquipment(EquipScope equipScope, int maxPrice, PriceLevel priceLev
...
+            item.automaticFlag = true;
```
VeeArrKoL: In cases where you have a lot of weapons available (e.g. in aftercore), this change significantly increases the number of combinations checked and running time of the maximizer any time a weapon-type keyword is used in order to properly handle a very narrow corner case. I think the fix should be more-narrowly targeted.

For a point of data, I tried `Maximizer: -tie, club, mus, 0.1 item, 0.01 meat` on a character with a bunch of stuff in aftercore, and it checked 17,664 combinations before this change and 244,224 combinations after this change.

A cursory inspection indicates that simply not setting the `automaticFlag` but still using `break gotItem` will correctly ensure a weapon of the appropriate type gets recommended without exploding the number of options considered, but you should probably test whether it actually fully solves the problem.

VeeArrKoL:
In addition to my concerns above about performance, I'm a little confused about the discussion of the tiebreaker. Does the modification to the tiebreaker stuff actually do anything? The maximizer-with-specific-weapon-type already seems to work as expected with the tiebreaker active and I would be surprised if modifying the tiebreaker code has an effect on cases where the tiebreaker is disabled. Also, the test case should probably be named `clubModifierWorksWithoutTieBreaker`, not `clubModifierWorksWithTieBreaker`, since it's testing the case where the tiebreaker is disabled.

eegeeZA:
Thanks for the review. Updated the branch.

**Performance (@VeeArrKoL):** The `automaticFlag` change has been removed entirely. Clubs with `delta == 0` already reach the ranked list via the normal fall-through path in `enumerateEquipment` (the three conditions on lines 1571-1573 are all false for an unequipped, in-inventory item with no prior `automaticFlag`), so no Evaluator change is needed and combination counts return to baseline.

**Tiebreaker change does nothing (@VeeArrKoL):** It does do something. `simplicity` is compared at line 176 of `MaximizerSpeculation.compareTo`, which is not guarded by `isUsingTiebreaker()` -- it runs even when `-tie` is active. Without the +3 nudge, the synthetic `(none)` item gets +2 simplicity (currently equipped = UNEQUIP in an empty weapon slot) while the club gets 0, so `(none)` sorts first in the per-slot speculation list. Its `Integer.MAX_VALUE` count then fills the `useful` quota in shortlist construction, and the club never enters `automaticEntry`. The test passing with the Evaluator change reverted confirms this is the actual fix path.

**Test name (@VeeArrKoL):** Renamed to `clubModifierWorksWithoutTieBreaker`.

**Type check (@midgleyc):** With the Evaluator change gone the question is moot. Non-clubs are already excluded from `WEAPON_1H` by the existing `requireClub && !isClub -> slot = auxSlot` reassignment earlier in the switch, so only qualifying clubs ever reach the ranked list for that slot.

See also the original forum report: https://kolmafia.us/threads/issues-with-maximiser-command.28691/ and related PR #1576.

eegeeZA:
Updated: the same bug applies to `shield` in the offhand slot (`maximize("-tie, shield")` with an equippable shield recommends nothing; same UNEQUIP ranking mechanism). Extended the fix to cover it: `isShieldRequired()` helper plus the same +3 nudge for a non-UNEQUIP offhand, and added `shieldModifierWorksWithoutTieBreaker`, which fails without the fix. Commit message and PR description updated to match.

midgleyc (approved):

PR #3512, opened 2026-06-14
Fix fold command for familiar-hatchling fold group members

eegeeZA:
## Problem

Two bugs with fold group items whose current form has `ConsumptionType.FAMILIAR_HATCHLING` (e.g. the makeshift crane, which is a member of the makeshift fold group). Forum: https://kolmafia.us/threads/issues-with-maximiser-command.28691/post-173296

1. **`fold` command:** `UseItemRequest.getInstance(item)` for a `FAMILIAR_HATCHLING` item constructs a familiar-hatch request rather than a fold request. Executing it produces "You already have that familiar." and the fold does not happen.

2. **Maximizer:** `+equip makeshift cape` with only a makeshift crane in inventory recommends nothing.

## Fix

**`FoldItemCommand`:** detect `FAMILIAR_HATCHLING` consumption type before dispatching and route through a direct `inv_use.php?which=3` request, bypassing `UseItemRequest`. All other fold group types continue to use `UseItemRequest` as before.

**`MaximizerTest`:** add `singleFoldSourceProducesRequestedForm` — verifies that maximizing `+equip makeshift cape` with only a makeshift crane in inventory recommends the cape in the container slot.

## Testing

- Verified live: `fold makeshift cape` with a makeshift crane now folds correctly (crane → turban → cape via two-step path).
- `MaximizerTest` passes in full.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/command/FoldItemCommand.java
+++ b/src/net/sourceforge/kolmafia/textui/command/FoldItemCommand.java
@@ -351,7 +351,15 @@ public class FoldItemCommand extends AbstractCommand {
         RecoveryManager.recoverHP(damage);
       }
 
-      RequestThread.postRequest(UseItemRequest.getInstance(item));
+      if (UseItemRequest.getConsumptionType(item)
+          == KoLConstants.ConsumptionType.FAMILIAR_HATCHLING) {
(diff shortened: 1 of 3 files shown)


midgleyc (approved):
Thanks!

PR #3518, opened 2026-06-20
Improve valhalla detection

VeeArrKoL:
Improve detection for when we have made it to Valhalla and don't actually perform `onAscension()` until we actually make it there. This should fix #3409. 

Due to the weird mechanism mafia uses to detect the on-ascension redirect from ascend.php to afterlife.php, the `lastBreakfast` preference still ends up getting clobbered if one tries to ascend but fails, but this is probably not that big of a deal on its own.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/AfterLifeRequest.java
+++ b/src/net/sourceforge/kolmafia/request/AfterLifeRequest.java
@@ -4,7 +4,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import net.sourceforge.kolmafia.AscensionPath;
 import net.sourceforge.kolmafia.AscensionPath.Path;
-import net.sourceforge.kolmafia.KoLConstants;
+import net.sourceforge.kolmafia.KoLConstants.MafiaState;
 import net.sourceforge.kolmafia.KoLmafia;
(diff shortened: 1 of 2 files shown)


src/net/sourceforge/kolmafia/request/AfterLifeRequest.java
```diff
@@ -34,6 +36,21 @@ public static boolean parseResponse(final String urlString, final String respons
       return false;
     }
 
+    if (responseText.isBlank()) {
+      KoLmafia.updateDisplay(
+          KoLConstants.MafiaState.ERROR,
+          "Received an empty response from afterlife.php. You are probably not in Valhalla.");
```
midgleyc: nit: extract the identical message into a variable

midgleyc (approved):
Thanks!

Have you confirmed this works both with ascending and failing to ascend?

VeeArrKoL:
I tested it with:
* Failing to ascend via relay browser
* Successfully ascending and navigating Valhalla via relay browser
* Failing to ascend using a personal script
* Successfully ascending and navigating Valhalla using a personal script 

I can't rule out some ascension script happening to visit some URLs in an order I didn't account for and running into an issue, but it works for every case I tested.

PR #3522, opened 2026-06-23
Gate possession-granted noncombat equipped-conditional skills by their unlock state

eegeeZA:
## Problem

`InventoryManager.checkSkillGrantingEquipment()` grants an item's NONCOMBAT
`Conditional Skill (Equipped)` skills from mere inventory possession, with no
check of the per-skill unlock state. With `heartstone*Unlocked=false`, a
refresh flips `have($skill("Heartstone: %pals"))` (and `%luck`/`%buff`) to
true; garbo then casts and KoL rejects it as "That skill is unavailable."

## Root cause

The terminal filter in `checkSkillGrantingEquipment` keeps an
equipped-conditional skill whenever it is NONCOMBAT-tagged, never consulting
`EquipmentManager.shouldApplySkill()` — the function that encodes each skill's
real gate (its `heartstone*Unlocked` pref; default `true` otherwise).

Live repro: with `heartstoneKillUnlocked` / `heartstoneLuckUnlocked` /
`heartstonePalsUnlocked` / `heartstoneBuffUnlocked` all `false` and a Heartstone
only in inventory (not equipped), `refresh all` flips `haveSkill()` to `true`
for `Heartstone: %pals`, `%luck`, `%buff` (all NONCOMBAT) but **not** `%kill`
(combat) — the add/skip-by-tag split fingerprints the NONCOMBAT-only filter.

## Fix

Gate the NONCOMBAT branch by `shouldApplySkill`. Locked Heartstone skills are
no longer granted; default-true skills (Sip Some Sweat, Cincho Fiesta Exit,
Powerful Glove cheat codes, codpiece, etc.) still grant on possession, so the
deliberate auto-equip-on-cast behaviour is preserved.

## Why not the unequip path (supersedes the prior version of this PR)

The earlier version of this branch changed `applyNoncombatSkills` to drop
noncombat skills on unequip. That is the wrong layer: skills are granted by
possession and re-granted on every refresh, so unequip-removal neither
explains nor fixes the bug, and it would have broken the deliberate
auto-equip-on-cast design that @midgleyc and @jaadams5 noted. That change is
reverted here.

## Tests

`InventoryManagerTest.NoncombatConditionalSkillGrant`: locked skill not granted
on possession, unlocked skill still granted, combat skill never granted, and
an inventory-conditional (designer sweatpants) control.

### DIFF
--- a/test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
@@ -30,6 +30,7 @@ import net.sourceforge.kolmafia.StaticEntity;
 import net.sourceforge.kolmafia.equipment.Slot;
 import net.sourceforge.kolmafia.modifiers.DoubleModifier;
 import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.objectpool.SkillPool;
 import net.sourceforge.kolmafia.persistence.ConsumablesDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
@@ -425,6 +426,63 @@ public class InventoryManagerTest {
 
       assertThat(InventoryManager.getCount(ItemPool.WORTHLESS_TRINKET), is(25252));
       assertThat(InventoryManager.getCount(ItemPool.WORTHLESS_GEWGAW), is(10622));
+    }
+  }
+
+  @Nested
+  class NoncombatConditionalSkillGrant {
+    // Live repro: an item with a NONCOMBAT Conditional Skill (Equipped) whose own
+    // unlock gate is false, merely possessed (not equipped), must NOT grant the
+    // skill. Before the fix, checkSkillGrantingEquipment granted it on possession.
+    @Test
+    public void lockedHeartstoneSkillNotGrantedByPossession() {
+      KoLCharacter.resetSkills();
+      var cleanups =
+          new Cleanups(
+              withProperty("heartstonePalsUnlocked", false), withItem(ItemPool.HEARTSTONE));
+      try (cleanups) {
+        InventoryManager.checkSkillGrantingEquipment();
+        assertFalse(KoLCharacter.hasSkill(SkillPool.HEARTSTONE_PALS));
+      }
+    }
+
+    // Positive control: when the gate is unlocked, possession still grants
+    // (preserves the deliberate auto-equip-on-cast behaviour).
+    @Test
+    public void unlockedHeartstoneSkillGrantedByPossession() {
+      KoLCharacter.resetSkills();
+      var cleanups =
+          new Cleanups(withProperty("heartstonePalsUnlocked", true), withItem(ItemPool.HEARTSTONE));
+      try (cleanups) {
+        InventoryManager.checkSkillGrantingEquipment();
+        assertTrue(KoLCharacter.hasSkill(SkillPool.HEARTSTONE_PALS));
+      }
+    }
+
+    // Combat-tagged conditional skill is never granted by possession (NONCOMBAT
+    // branch short-circuits before shouldApplySkill) — unchanged by the fix.
+    @Test
+    public void combatHeartstoneSkillNotGrantedByPossession() {
+      KoLCharacter.resetSkills();
+      var cleanups =
+          new Cleanups(withProperty("heartstoneKillUnlocked", true), withItem(ItemPool.HEARTSTONE));
+      try (cleanups) {
+        InventoryManager.checkSkillGrantingEquipment();
+        assertFalse(KoLCharacter.hasSkill(SkillPool.HEARTSTONE_KILL));
+      }
+    }
+
+    // Covers the e.getKey()==true branch (CONDITIONAL_SKILL_INVENTORY): designer
+    // sweatpants grant "Sweat Out Some Booze" on mere possession, ungated.
+    @Test
+    public void inventoryConditionalSkillGrantedByPossession() {
+      KoLCharacter.resetSkills();
+      var cleanups = new Cleanups(withItem(ItemPool.DESIGNER_SWEATPANTS));
+      try (cleanups) {
+        assertFalse(KoLCharacter.hasSkill(SkillPool.SWEAT_OUT_BOOZE));
+        InventoryManager.checkSkillGrantingEquipment();
+        assertTrue(KoLCharacter.hasSkill(SkillPool.SWEAT_OUT_BOOZE));
+      }
     }
   }
 }
--- a/test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
+++ b/test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
@@ -36,6 +36,7 @@ import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
 import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.EquipmentRequest;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
@@ -431,12 +432,16 @@ public class InventoryManagerTest {
 
   @Nested
   class NoncombatConditionalSkillGrant {
+    @AfterEach
+    void afterEach() {
+      KoLCharacter.resetSkills();
+    }
+
     // Live repro: an item with a NONCOMBAT Conditional Skill (Equipped) whose own
     // unlock gate is false, merely possessed (not equipped), must NOT grant the
     // skill. Before the fix, checkSkillGrantingEquipment granted it on possession.
     @Test
     public void lockedHeartstoneSkillNotGrantedByPossession() {
-      KoLCharacter.resetSkills();
       var cleanups =
           new Cleanups(
               withProperty("heartstonePalsUnlocked", false), withItem(ItemPool.HEARTSTONE));
@@ -450,7 +455,6 @@ public class InventoryManagerTest {
     // (preserves the deliberate auto-equip-on-cast behaviour).
     @Test
     public void unlockedHeartstoneSkillGrantedByPossession() {
-      KoLCharacter.resetSkills();
       var cleanups =
           new Cleanups(withProperty("heartstonePalsUnlocked", true), withItem(ItemPool.HEARTSTONE));
       try (cleanups) {
@@ -463,7 +467,6 @@ public class InventoryManagerTest {
     // branch short-circuits before shouldApplySkill) — unchanged by the fix.
     @Test
     public void combatHeartstoneSkillNotGrantedByPossession() {
-      KoLCharacter.resetSkills();
       var cleanups =
           new Cleanups(withProperty("heartstoneKillUnlocked", true), withItem(ItemPool.HEARTSTONE));
       try (cleanups) {
@@ -476,10 +479,8 @@ public class InventoryManagerTest {
     // sweatpants grant "Sweat Out Some Booze" on mere possession, ungated.
     @Test
     public void inventoryConditionalSkillGrantedByPossession() {
-      KoLCharacter.resetSkills();
       var cleanups = new Cleanups(withItem(ItemPool.DESIGNER_SWEATPANTS));
       try (cleanups) {
-        assertFalse(KoLCharacter.hasSkill(SkillPool.SWEAT_OUT_BOOZE));
         InventoryManager.checkSkillGrantingEquipment();
         assertTrue(KoLCharacter.hasSkill(SkillPool.SWEAT_OUT_BOOZE));
       }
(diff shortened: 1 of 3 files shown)


midgleyc:
This is deliberate because "cast fiesta" should work without the item equipped, and should automatically equip the item.

If the latter isn't working, that's the bug.

jaadams5:
Just so I am sure I am following...

A script decides a skill is available, tries to cast it and it fails.

This proposed fix is to make equipment based skills unavailable when the associated item is unequipped.

This is wrong because there are cases where the availability of a skill depends upon possession of an item and mafia will equip the item as necessary.  This fix breaks that.

My observation is that whatever the script is using to determine availability is not properly understood.  Is it saying it can be cast immediately with no changes or is it saying a skill is available if you equip an item first?  The fact that it fails suggests that mafia does not automatically equip all items when deeded for a case.  This is deliberate because equipment changes sometimes have side effects that players need to manage themselves.

One solution would be to have a function that helps the script writer understand whether casting a skill would require the script to make an equipment change or whether maifa would change equipment automatically.

midgleyc:
> Is it saying it can be cast immediately with no changes or is it saying a skill is available if you equip an item first?

The intent is that if you tell Mafia to cast the skill, Mafia will equip any items it needs to before casting.

In `InventoryManager`, there is a function `checkSkillGrantingEquipment` which adds skills from items we have equipped or in inventory. It is also called in `ResultProcessor.gainItem` for items we obtain during the session. It adds the non-combat skills from equipment to the available skills list because casting them should succeed.

Equipping happens in `UseSkillRequest`'s `optimizeEquipment` in `equipForSkill` (or `unequipForSkill`, if we need an item *not* equipped).

> This is deliberate because equipment changes sometimes have side effects that players need to manage themselves.

We fail with an error message explicitly in the case where replacing the current item would lead to a reduction in HP or MP.

Scripts should check the output of `use_skill` and manually replace equipment in this case. See e.g. https://github.com/loathers/autoscend/blob/9b4bab99470f5f567a1b67ae26d20e9ff901f6b5/RELEASE/scripts/autoscend/auto_restore.ash#L1717

jaadams5:
Seems to me this is fundamentally a problem that could/should be fixed by the script writer.  I'm Ok with closing without commit.

eegeeZA:
I picked this up from https://github.com/loathers/garbage-collector/blob/9eecbf9aa63a7e068e8d39687e1451896cef8b1c/packages/garbo/src/mood.ts#L198 which does internally call `haveSkill` via libram. But I'm rechecking now, and I think I've missed some analysis.

**Update:** yip, picked up a different bug. I'll amend tomorrow and reopen PR. Will revert this change, but reuse PR.

eegeeZA:
Correction / retraction of my earlier comment: I had the mechanism wrong, and this branch has been rewritten to carry the right fix.

The bug is not in the unequip path. `checkSkillGrantingEquipment` (`InventoryManager`) grants an item's NONCOMBAT `Conditional Skill (Equipped)` skills from mere inventory possession, never consulting `shouldApplySkill`. Live repro today: with `heartstoneKillUnlocked` / `heartstoneLuckUnlocked` / `heartstonePalsUnlocked` / `heartstoneBuffUnlocked` all `false` and the Heartstone only in inventory (not equipped), a `refresh all` flips `haveSkill()` to `true` for `Heartstone: %pals`, `%luck` and `%buff` (all NONCOMBAT) but NOT `%kill` (combat) — the add/skip-by-tag split that fingerprints the NONCOMBAT-only filter. garbo then casts `%pals` and KoL rejects it.

@midgleyc @jaadams5 — you were right that auto-equip-on-cast is deliberate and that removing skills on unequip is the wrong fix. I've reverted that change. The new fix gates the NONCOMBAT grant in `checkSkillGrantingEquipment` by `shouldApplySkill`, so a locked Heartstone skill is never granted in the first place, while default-true skills (Sip Some Sweat, Cincho Fiesta Exit, Powerful Glove cheat codes, etc.) still grant on possession and auto-equip on cast as before. Title/body updated; tests added in `InventoryManagerTest`.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -427,4 +428,61 @@ void shouldReadInventoryJson() {
...
   }
+
+  @Nested
+  class NoncombatConditionalSkillGrant {
+    // Live repro: an item with a NONCOMBAT Conditional Skill (Equipped) whose own
+    // unlock gate is false, merely possessed (not equipped), must NOT grant the
+    // skill. Before the fix, checkSkillGrantingEquipment granted it on possession.
+    @Test
+    public void lockedHeartstoneSkillNotGrantedByPossession() {
+      KoLCharacter.resetSkills();
```
midgleyc: instead of being at the start of every function, this should be a Cleanups or an `@AfterEach`.

You could make `withCheckSkillGrantingEquipment` (function only needs to be in this file, not in Player) that calls the relevant function and resets skills afterward, or you could just make it an after each because this pattern is only used here.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
```diff
@@ -427,4 +428,61 @@ void shouldReadInventoryJson() {
...
+    }
+
+    // Covers the e.getKey()==true branch (CONDITIONAL_SKILL_INVENTORY): designer
+    // sweatpants grant "Sweat Out Some Booze" on mere possession, ungated.
+    @Test
+    public void inventoryConditionalSkillGrantedByPossession() {
+      KoLCharacter.resetSkills();
+      var cleanups = new Cleanups(withItem(ItemPool.DESIGNER_SWEATPANTS));
+      try (cleanups) {
+        assertFalse(KoLCharacter.hasSkill(SkillPool.SWEAT_OUT_BOOZE));
```
midgleyc: test shouldn't assert false here imo. That's just how it currently works, but it's not a requirement it works this way.

midgleyc (changes requested):
some test changes, ideally you would also grab fight HTML for learning the Heartstone skills to confirm that works

There is code in `FightRequest` for acquiring skills, but no guarantee the heartstone skill gain HTML matches it.

not a blocker as closing / reopening mafia will fix that.

jaadams5:
Let me think out loud which might help someone tell me if or where I am wrong.

An Item can grant a non-combat skill when the item is merely possessed or when it is equipped.  In addition if the skill has a limited number of uses then at least one use must be available for it to be granted.

If casting a granted skill requires that the item be equipped then KoLmafia will automatically equip the item as part of casting the skill unless equipping the item will change HP or MP, in which case the cast will fail.

The original bug was that a script thought a skill had been granted, attempted to cast it and had the cast fail.  

If the failure was because the skill was incorrectly marked as granted because limited uses was not being taken into account then this PR seems to me to be a proper fix for that bug.

But if the failure was because mafia did not automatically equip an item in order to cast the skill, then this does not fix the bug.

An overcomplicated fix would be to expand the code that determines whether something is granted to check MP/HP before and after equipping and include that in the decision.  This will reduce the chances of a failure but there is no guarantee that the equipment remains fixed between the grant computation and the cast.  There would still be a possibility that a granted skill could not be cast.

A more practical fix would be to document the possibility that a granted skill might not be cast because mafia did not equip the item.  Then a script writer cold be expected to either check for success and respond appropriately or manage the equipment swap themselves.

I would be curious to see whether garbo's behavior is different before and after this change or whether the script needs to changed anyway.

If I am basically correct then I concur with @midgleyc 's Approval and will let them commit when they are satisfied.

Otherwise I welcome to chance to fix my misunderstandings :-)

midgleyc:
> But if the failure was because mafia did not automatically equip an item in order to cast the skill, then this does not fix the bug.

I wasn't clear originally -- we only do not automatically switch items if the skill restores HP or MP, so the only one affected at the moment is Rest upside down. There are also some combat skills like Loofah leglifts that restore MP, but these don't matter because they're combat only.

> An overcomplicated fix would be to expand the code that determines whether something is granted to check MP/HP before and after equipping and include that in the decision.

This is already how it works.

> A more practical fix would be to document the possibility that a granted skill might not be cast because mafia did not equip the item.

`use_skill` is already documented to return false in case the skill is not cast (e.g. the user does not have enough MP). Not equipping the item is just a special case of failure.

jaadams5:
> I wasn't clear originally -- we only do not automatically switch items if the skill restores HP or MP, so the only one affected at the moment is Rest upside down. There are also some combat skills like Loofah leglifts that restore MP, but these don't matter because they're combat only.
> 

OK.  I remember some discussion about automated switching that caused a player to lose HP/MP and how that was not something to be automated.  Example, I have Max HP over 1000 and am at full health.   The Gore Bucket is equipped and I lose 800 hp.  I was probably conflating that discussion with this one.

> > An overcomplicated fix would be to expand the code that determines whether something is granted to check MP/HP before and after equipping and include that in the decision.
> 
> This is already how it works.
> 

OK.  I may have been conflating two cases or may have been expecting the code that "grants" a skill was not being executed under the same conditions as the code that casts.

> > A more practical fix would be to document the possibility that a granted skill might not be cast because mafia did not equip the item.
> 
> `use_skill` is already documented to return false in case the skill is not cast (e.g. the user does not have enough MP). Not equipping the item is just a special case of failure.

In which the original bug is the garbo code that calls use_skill and does not check the return :-)

eegeeZA:
Okay, let me spend some more time going through this in detail. I'll mark it as a draft for now.

midgleyc:
I think the fix is good, just correct the test issues I mentioned.

jaadams5:
> Okay, let me spend some more time going through this in detail. I'll mark it as a draft for now.

I talk a lot and I forget things so there is no significance to the quantity of my comments.  I would like two things and then I would be glad to commit this.  One would be to change the current test, as suggested.  The other would be to add a new test with an item that grants a non-combat skill with a limited number of uses.  The test would show that the skill is granted when there are uses available and not granted when the uses are exhausted.

The rest of my comments concern motivation.  I believe the following assertions.

This work found and fixed a previously unreported bug where KoLmafia would report a skill was available when it had already used the available casts.

The originally reported bug where garbo failed to cast a skill is not a KoLmafia bug but a scripting error caused by the script's failure to check the return value of use_skill.

midgleyc:
> This work found and fixed a previously unreported bug where KoLmafia would report a skill was available when it had already used the available casts.

> The originally reported bug where garbo failed to cast a skill is not a KoLmafia bug but a scripting error caused by the script's failure to check the return value of use_skill.

No, neither of these are true. 

This work found and fixed a previously unreported bug where KoLmafia would erroneously report a skill as available when the character had not unlocked it.

Garbo indeed doesn't check the output of the used skill, but we shouldn't be reporting that a user has a skill when they don't have it unlocked.

Also, we still report `have_skill($skill[whatever])` as true if you can no longer cast the skill. The uncastability is stored in `$skill[whatever].dailylimit`. Or it should be, some skills aren't wired up correctly.

jaadams5:
> > This work found and fixed a previously unreported bug where KoLmafia would report a skill was available when it had already used the available casts.
> 
> > The originally reported bug where garbo failed to cast a skill is not a KoLmafia bug but a scripting error caused by the script's failure to check the return value of use_skill.
> 
> No, neither of these are true.
> 
> This work found and fixed a previously unreported bug where KoLmafia would erroneously report a skill as available when the character had not unlocked it.
> 
> Garbo indeed doesn't check the output of the used skill, but we shouldn't be reporting that a user has a skill when they don't have it unlocked.
> 
> Also, we still report `have_skill($skill[whatever])` as true if you can no longer cast the skill. The uncastability is stored in `$skill[whatever].dailylimit`. Or it should be, some skills aren't wired up correctly.

One of the skills taught to improve communications is to paraphrase what you heard so there is feedback.  The good news is it is helping.  The bad news is I need to spruce up my reading comprehension skills.

While it may be my comprehension and memory are shot I do think I would have been better off if the original PR had been closed and this created as a new one, rather than just editing and deleting.  That would have helped me understand why I got something wrong.

Thanks for your patience.

> Also, we still report `have_skill($skill[whatever])` as true if you can no longer cast the skill. The uncastability is stored in `$skill[whatever].dailylimit`. Or it should be, some skills aren't wired up correctly.

Is this worth addressing separately?

midgleyc:
> Is this worth addressing separately?

Yes, if any skills are not in `dailylimits.txt` but should be, that's a bug and a PR should be filed. Out of scope for this one, though. I noticed "Brandish Moss Medal" while I was writing the original paragraph. Almost surely there are more.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
eegeeZA: Done — moved resetSkills() to an @AfterEach in the nested class.

test/net/sourceforge/kolmafia/session/InventoryManagerTest.java
eegeeZA: Done — removed.

eegeeZA:
Apologies for the confusion in this PR. I initially misidentified the bug as being in the unequip path and pushed a fix for that, then realised I had the mechanism wrong and rewrote the branch for the correct fix (the grant path, gated by shouldApplySkill). I should have closed the old PR and opened a new one rather than silently rewriting this one.. sorry for the context-switching that caused.

Both review comments addressed; marked as ready.

jaadams5:
https://kolmafia.us/threads/have_skill-is-sometimes-incorrect.32632/

Apology for reusing the PR noted but the confusion is ultimately my fault :-)

midgleyc (approved):
Thanks!

PR #3529, opened 2026-06-27
feat: update flatlaf to 3.7.1

midgleyc:
Some visible changes: current tab has a highlighted background, subtabs on the same page have a different underline depending on which area is selected, some more corners are rounded.

One change averted: non-editable panels received the inactive background. For our HTML panels, manually set the background to the active one, which I think we want.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -80,9 +80,9 @@ dependencies {
   }
   testImplementation("org.junit-pioneer:junit-pioneer:2.3.0")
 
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
Easier to approve and see it in action than to ask a bunch of questions :-)  Jumping from 1.x to 3.x says something...

midgleyc:
All it says in practice is that we haven't updated in a long time ;)

PR #3533, opened 2026-06-30
Implement since/script/notify directives for JS scripts

VeeArrKoL:
Implements functionality similar to Ash's `since`, `script`, and `notify` directives for use in JS scripts. This is done by reading from an exported objected named `__metadata` which can contain the necessary values. For example:

```
module.exports.__metadata={script:"test script", notify:"VeeArr", since:"r29081"};
```

These should function more-or-less the same as the corresponding Ash directives. I'm not really familiar with all the ways typescript is used to produce JS scripts, but I tested with a small TS script based on [kol-ts-starter](https://github.com/docrostov/kol-ts-starter), and the following in `main.ts` had the desired effect:

```
export const __metadata={script:"test script", notify:"VeeArr", since:"r29081"};
```

Notes:

* `__metadata` was chosen as a name that was almost certainly not already being used by someone. If someone has a better idea for a name, I'm open to suggestions.
* One file was expanded and renamed (`MainWarningScript` => `JsPostLoadScript`). The rename is in its own commit, so the actual code changes to the file are more easily seen in commit 9be2b13.
* In most cases, the `since` directive will behave in a similar manner to Ash. However, if `require()` is used within a code block (rather than at the top-level scope), the imported file doesn't get processed (and thus any since directive doesn't get checked) until that portion of code executes.
* In my test cases, the directives get processed on a per-file basis. So if you have a bunch of TS that all gets compiled to a single JS script, only `__metadata` exported from the "top" file get read. Again, I'm not super familiar with the TS toolchain so this may be an artifact of how I tested it. Regardless, I don't imagine this being a problem in practice.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/AbstractRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/AbstractRuntime.java
@@ -165,7 +165,7 @@ public abstract class AbstractRuntime implements ScriptRuntime {
         // case you should be able to run anything.
         if (currentRevision != 0 && currentRevision < targetRevision) {
           String template =
-              "'%s' requires revision r%s of kolmafia or higher (current: r%s). Up-to-date builds can be found at https://github.com/kolmafia/kolmafia/releases/.";
+              "'%s' requires revision r%s of kolmafia or higher (current: r%s).  Up-to-date builds can be found at https://github.com/kolmafia/kolmafia/releases/.";
           return new SinceStatus(
               "SINCE", String.format(template, filename, targetRevision, currentRevision));
         }
--- a/src/net/sourceforge/kolmafia/textui/AbstractRuntime.java
+++ b/src/net/sourceforge/kolmafia/textui/AbstractRuntime.java
@@ -3,7 +3,11 @@ package net.sourceforge.kolmafia.textui;
 import java.io.PrintStream;
 import java.util.LinkedHashMap;
 import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.RequestThread;
+import net.sourceforge.kolmafia.StaticEntity;
+import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.request.RelayRequest;
+import net.sourceforge.kolmafia.request.SendMailRequest;
 import net.sourceforge.kolmafia.textui.parsetree.Value;
 import net.sourceforge.kolmafia.utilities.NullStream;
 
@@ -131,5 +135,58 @@ public abstract class AbstractRuntime implements ScriptRuntime {
       this.indentLine(this.traceIndentation);
       traceStream.println(string);
     }
+  }
+
+  public static void handleNotify(
+      String scriptFilename, String scriptName, String notifyRecipient) {
+    String notifyList = Preferences.getString("previousNotifyList");
+    String notifyKey = (scriptFilename == null) ? "<>" : "<" + scriptFilename + ">";
+    if (notifyRecipient != null && !notifyRecipient.isBlank() && !notifyList.contains(notifyKey)) {
+      Preferences.setString("previousNotifyList", notifyList + notifyKey);
+
+      String message =
+          "I have opted to let you know that I have chosen to run <"
+              + scriptName
+              + ">.  Thanks for writing this script!";
+      SendMailRequest notifier = new SendMailRequest(notifyRecipient, message);
+      RequestThread.postRequest(notifier);
+    }
+  }
+
+  public record SinceStatus(String status, String message) {}
+
+  public static SinceStatus handleSince(String revision, String filename) {
+    try {
+      if (revision.startsWith("r")) { // revision
+        revision = revision.substring(1);
+        int targetRevision = Integer.parseInt(revision);
+        int currentRevision = StaticEntity.getRevision();
+        // A revision of zero means you're probably running in a debugger, in which
+        // case you should be able to run anything.
+        if (currentRevision != 0 && currentRevision < targetRevision) {
+          String template =
+              "'%s' requires revision r%s of kolmafia or higher (current: r%s). Up-to-date builds can be found at https://github.com/kolmafia/kolmafia/releases/.";
+          return new SinceStatus(
+              "SINCE", String.format(template, filename, targetRevision, currentRevision));
+        }
+      } else { // version (or syntax error)
+        String[] target = revision.split("\\.");
+        if (target.length != 2) {
+          return new SinceStatus("SYNTAX", "invalid 'since' format");
+        }
+
+        int targetMajor = Integer.parseInt(target[0]);
+        int targetMinor = Integer.parseInt(target[1]);
+
+        if (targetMajor > 21 || targetMajor == 21 && targetMinor > 9) {
+          return new SinceStatus(
+              "SYNTAX", "invalid 'since' format (21.09 was the final point release)");
+        }
+      }
+    } catch (NumberFormatException e) {
+      return new SinceStatus("SYNTAX", "invalid 'since' format");
+    }
+
+    return new SinceStatus("OK", null);
   }
 }
--- a/src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
@@ -3,6 +3,7 @@ package net.sourceforge.kolmafia.textui.javascript;
 import java.util.Map;
 import net.sourceforge.kolmafia.KoLConstants.MafiaState;
 import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.RequestLogger;
 import net.sourceforge.kolmafia.textui.AbstractRuntime;
 import net.sourceforge.kolmafia.textui.ScriptException;
 import org.mozilla.javascript.Context;
@@ -65,7 +66,8 @@ public class JsPostLoadScript implements Script {
           AbstractRuntime.SinceStatus response =
               AbstractRuntime.handleSince(since, scriptShortFilename);
           if (!response.status().equals("OK")) {
-            throw new ScriptException(response.message());
+            throw new ScriptException(
+                String.format("In <%s>: %s", scriptShortFilename, response.message()));
           }
         }
 
@@ -78,7 +80,9 @@ public class JsPostLoadScript implements Script {
     } catch (Exception e) {
       // Apart from the ScriptException we intend to throw, never ever let this handling propagate
       // an unexpected exception.
-      KoLmafia.updateDisplay("Unexpected error while handling metadata: " + e.getMessage());
+      String errMsg = "Unexpected error while handling metadata: " + e.getMessage();
+      KoLmafia.updateDisplay(errMsg);
+      RequestLogger.updateSessionLog(errMsg);
     }
   }
 }
--- a/src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
+++ b/src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
@@ -11,7 +11,7 @@ import org.mozilla.javascript.Scriptable;
 import org.mozilla.javascript.ScriptableObject;
 import org.mozilla.javascript.commonjs.module.ModuleScope;
 
-public class MainWarningScript implements Script {
+public class JsPostLoadScript implements Script {
   // This is a slight hack to perform checks after loading a JS file. At this time, we warn folks
   // who have failed to export main() and check for script/notify/since directives.
   @Override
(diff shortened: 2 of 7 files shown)


jaadams5:
At one time the notify directive was controversial.  What scripts a user is running was considered private by some users and notify violated that privacy.  The compromise, since notify does have benefits for script authors, was to include notify as a user accessible and editable part of the script.  It was also a problem for KoL because messages were stored, but the KoL change to delete messages older than two weeks fixed that :-)

Since JS and TS ae both areas of my ignorance, how does a user opt out of notify?  Does __metadata appear anywhere that is human readable and editable with a text editor or is a user's only option to choose not to use the script?  If a user does make an edit is the code robust enough to work if notify is absent or some other typo is introduced?

The pros and cons of notify are beyond the scope of this but I think the ability of a user to opt out with editing remains important.

Thanks.

jaadams5:
I'd like to the editability question answered but otherwise it looks good.

src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+      }
+    } catch (ScriptException e) {
+      throw e;
+    } catch (Exception e) {
+      // Apart from the ScriptException we intend to throw, never ever let this handling propagate
+      // an unexpected exception.
+      String errMsg = "Unexpected error while handling metadata: " + e.getMessage();
+      KoLmafia.updateDisplay(errMsg);
+      RequestLogger.updateSessionLog(errMsg);
```
midgleyc, line 85: probably want `RequestLogger.printLine` too, in addition to session log.

midgleyc (changes requested):
I will look in more detail later (and have more of a think about the design, which seems reasonable), but minimally I want some tests on this.

e.g.
* since: old version works, r9999999 breaks
* notify: fake HTML shows appropriate request, running script again doesn't make request again
* __metadata: not required (implicit in existing tests), unexpected key doesn't break anything, empty object doesn't break anything

Look at how CustomScriptTest uses CallScriptCommand to call scripts, but do it elsewhere so you can add some setup.

src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
VeeArrKoL: I'll admit some ignorance here: what are the differences between `KoLmafia.updateDisplay()` and `RequestLogger.printLine()`?

midgleyc: From what I recall:
* KoLmafia.updateDisplay(): appears in the black bar on the "Adventure" screen, or at the bottom of the login frame, or various other places
* RequestLogger.printLint(): appears in the "Graphical CLI" tab
* RequestLogger.updateSessionLog(): appears in the log in the `sessions/` folder.

VeeArrKoL: `KoLmafia.updateDisplay()` does print to the gCLI (that's why I put it there), but from your description it sounds like it's probably also doing other stuff that's undesirable. I'll check and probably replace that with a call to `RequestLogger.printLine()`.

src/net/sourceforge/kolmafia/textui/AbstractRuntime.java
```diff
@@ -132,4 +136,57 @@ public final void trace(final String string) {
...
+      String message =
+          "I have opted to let you know that I have chosen to run <"
+              + scriptName
+              + ">.  Thanks for writing this script!";
+      SendMailRequest notifier = new SendMailRequest(notifyRecipient, message);
+      RequestThread.postRequest(notifier);
+    }
+  }
+
+  public record SinceStatus(String status, String message) {}
```
midgleyc, line 160: rather than a String, make `status` an enum (three values: SINCE, SYNTAX, OK).

src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
```diff
@@ -40,4 +48,37 @@ public Object exec(Context cx, Scriptable scope, Scriptable thisObj) {
     }
     return null;
   }
+
+  private void handleMetadata(Scriptable exports, Scriptable scope) {
+    try {
+      Object metadataObj = exports.get("__metadata", exports);
+      if (metadataObj instanceof Map metadata && scope instanceof ModuleScope moduleScope) {
+        String scriptFilename = moduleScope.getUri().getPath().substring(1);
+        String scriptShortFilename = scriptFilename.substring(scriptFilename.lastIndexOf("/") + 1);
```
midgleyc, line 58: `File.separator` instead of "/"

VeeArrKoL:
@jaadams5 

> Does __metadata appear anywhere that is human readable and editable with a text editor

Short answer: *Technically*, yes. Practically, no.

Obviously the directive must actually make its way into the script. However, since mafia doesn't control the parsing of JS scripts, we can't enforce the same requirement that `notify` be at the very start of the script like Ash requires. As a result, the `__metadata` entry could be declared anywhere in the file, even buried down among other pieces of code. Worse still, the TS toolchain results in such a jumble of transpiled JS, it is probably not feasible for a user to disable it manually.

If this is a concern, the better solution is probably to create a preference to opt out of `notify` altogether. (Is a user really going to want to opt out of `notify` in one specific script rather than all scripts?) For visibility, it would probably make sense for this to be in the actual preferences panel.

Thoughts?

gausie:
I don't like this change from a design perspective.

VeeArrKoL:
> I don't like this change from a design perspective.

Can you elaborate? Do you mean that (1) you don't like the idea of having `script`/`since`/`notify` available in JS scripts (and if so, which directive(s) specifically?); or (2) you are okay with the idea in principle but do not like the implementation; or (3) something else that I've entirely misunderstood?

jaadams5:
> @jaadams5
> 
> > Does __metadata appear anywhere that is human readable and editable with a text editor
> 
> Short answer: _Technically_, yes. Practically, no.
> 
> Obviously the directive must actually make its way into the script. However, since mafia doesn't control the parsing of JS scripts, we can't enforce the same requirement that `notify` be at the very start of the script like Ash requires. As a result, the `__metadata` entry could be declared anywhere in the file, even buried down among other pieces of code. Worse still, the TS toolchain results in such a jumble of transpiled JS, it is probably not feasible for a user to disable it manually.
> 
> If this is a true concern, the better solution is probably to create a preference to opt out of `notify` altogether. (Is a user really going to want to opt out of `notify` in one specific script rather than all scripts?) For visibility, it would probably make sense for this to be in the actual preferences panel.
> 
> Thoughts?

My personal concern with the whole JS/TS infrastructure is the loss of control (by the user) because the resulting script is harder to read, harder to understand and harder to modify.  So unconditional NOTIFY at the whim of the scriptwriter is an issue.  I certainly don't have to use a script, unless it is unconditionally a dependency for a script I do want to use, but it is one more small step away from the Good Old Days when the scripting communities that I experienced were much more interested in helping and sharing.

However this is not an insistence that the world conform to me :-)

A preference concerning NOTIFY would address my issue.  I would first consider making it a user preference of scripts that are not allowed to be notified.  In Ye Olden Days I had a feud with someone.  I wanted to use and modify their script but I didn't want them to know.  If I still felt the somewhat childish need to do that with a JS/TS script I would set the preference instead of edit the script to accomplish the same thing.

Note that this is a preference and not something I would use to block approval :-)
 
I am more concerned with the design perspective concerns that have been raised but not elaborated on.

gausie:
I just think the way that it is being proposed is kind of janky and I'm not entirely convinced that the behaviour can't just be replicated (for all intents and purposes) in code. The fact they exist historically in ASH doesn't mean they need to make it to JS too, and they represent a maintenance burden that outweighs their utility in my mind.

src/net/sourceforge/kolmafia/textui/javascript/JsPostLoadScript.java
VeeArrKoL: No, file URIs use `/` as the separator regardless of the platform-specific path separator.

VeeArrKoL:
@midgleyc 
> I will look in more detail later (and have more of a think about the design, which seems reasonable), but minimally I want some tests on this.

I did test the listed cases while writing the change, and they all worked as expected. (Specifically, if whatever it finds for the values of the directives isn't a String, it's just ignored.) I can produce a unit test suite for it, but as some prevailing opinions seem to be that this change isn't good as-is, I'll hold off on spending time doing so until discussions have run their course.

@jaadams5 
> I would first consider making it a user preference of scripts that are not allowed to be notified.

Would a preference like this really be meaningfully different than the already-available solution of manually adding the script to `previousNotifyList` (which has the same effect, by making mafia believe it has already done the notify for that script)?

midgleyc:
> Short answer: Technically, yes. Practically, no.

Practically, yes. You open the file, ctrl+f `__metadata`, find "notify" and delete to the next comma or `}` (or remove the whole `__metadata` object).

I think the way this is being proposed is sensible: one benefit of the ASH approach is that since / notify get called on script load, instead of whenever the script interprets the function. But equally you could have them be functions which error / send a message, and the user has to call them in the appropriate places. If `script` is more than a parameter to `notify`, I think it has to be done in a manner like this.

On the preference: I would like, even without this, a preference to never notify any authors. The message says "I have opted to let you know" which implies the user has taken some positive action, instead of it being impossible to opt out.

jaadams5:
> @jaadams5
> 
> > I would first consider making it a user preference of scripts that are not allowed to be notified.
> 
> Would a preference like this really be meaningfully different than the already-available solution of manually adding the script to `previousNotifyList` (which has the same effect, by making mafia believe it has already done the notify for that script)?

Good catch.  If I have to edit something manually then editing perviousNotifyList instead of a new parameter would have the same effect.  That would handle the case where I never want MrBozo to know that I am using their script Clownface but don't care about notifications to other authors or any other of MrBozo's scripts.  But maybe I do want a less specific preference..  neverNotify could be a Boolean that just disables (or enables) NOTIFY.  Extending that to a list of prohibited or permitted authors could be considered but might make things so niche that very few people use it.

VeeArrKoL:
Leaving a note of something I noticed today: The existing processing for `notify` in Ash is a little weird. It gets processed on the top-level file but not on any imported files. However, `notify` directives of any Ash scripts directly imported into a JS script *do* get processed (but not the `notify` directives of any further Ash scripts imported by those scripts). I'm not sure what the actual desired behavior would even be in all of these cases, but the JS versions of the `notify` directive in this PR get processed on both the top-level script as well as any imported JS scripts.

jaadams5:
> Leaving a note of something I noticed today: The existing processing for `notify` in Ash is a little weird. It gets processed on the top-level file but not on any imported files. However, `notify` directives of any Ash scripts directly imported into a JS script _do_ get processed (but not the `notify` directives of any further Ash scripts imported by those scripts). I'm not sure what the actual desired behavior would even be in all of these cases, but the JS versions of the `notify` directive in this PR get processed on both the top-level script as well as any imported JS scripts.

I believe, and while I was an eyewitness my reliability is suspect, the intent of ash notify was to notify the author of a script that was initiated directly by a user.  You started the script from the Scripts menu or typed something on the command line.  importing the script was not expected to trigger notify.  I don't recall there was much thought put into nesting and calling and cli_execute and I think the code to track notifications and limit them to one per character per script was a way to keep notify manageable.  It should be noted that script names and file names are somewhat coupled and the script name parameter was a convivence.  

So one notify for an ash script chain called by JS seems to be the expected behavior.

The script community has changed and some of the original assumptions about ash scripts are no longer valid and some that are don't apply to JS.

It occurs to me as I write this that the problem could be simplified by wrapping a JS script in an ash script.  The wrapper would have the since, notify and script name directives, execute them and then transfer control to a JS script.  I think we could generate a template for use by script authors but I don't know how mafia could populate it.  Maybe a mafia command that accepts the parameters (and JS file name containing the top level) and spits out the wrapper file.  Or the script author would craft it..  A JS script author who didn't care just would not create the wrapper.

VeeArrKoL:
> It occurs to me as I write this that the problem could be simplified by wrapping a JS script in an ash script.

Technically it's already possible to do this but the other way around: You can create an Ash script with nothing but the directives and import it into the JS script. 

The motivation for this PR is that that felt to me like a bit of a hack compared to having the directives natively in the JS script.

jaadams5:
> > It occurs to me as I write this that the problem could be simplified by wrapping a JS script in an ash script.
> 
> Technically it's already possible to do this but the other way around: You can create an Ash script with nothing but the directives and import it into the JS script.
> 
> The motivation for this PR is that that felt to me like a bit of a hack compared to having the directives natively in the JS script.

But the more I think of it the more I am inclined to suggest that JS writers who want this capability just have to edit and reference an ash file with at most three directives.  One man's hack is another man's taxi? :-)

VeeArrKoL:
It seems to me that the current vibe is that this isn't worth doing (or at least, not in this somewhat-janky manner). Unless someone corrects that understanding in the next day or so, I'll close this PR down.

PR #3534, opened 2026-07-01
Update zapgroups.txt to contain sucker samurai zap group

BlortHorc:
Added  Sucker Samurai Suit zap group


midgleyc (approved):
Thanks!

I see you added this to the wiki in March 2026. That certainly went a long time without being checked.

PR #3536, opened 2026-07-01
Add preference to enable/disable Ash notify directives

VeeArrKoL:
Adds a preference to disable processing of `notify` directives for users who do not wish to notify script authors, per discussion in #3533


midgleyc (approved):
Thanks!

PR #3537, opened 2026-07-02
Suppress abort due to ascension error for 1-day CS

VeeArrKoL:
Follow-up to #3518

Suppresses the error state normally generated when a redirect to `main.php?nope=asc` happens under the specific circumstance where we are body-donation-cancelling at the end of a 1-day Community Service run. The message still gets printed (with an indicator that we *expected* to arrive at this error message), but it should no longer cause scripts to abort in this case.


jaadams5 (approved):
I think I understand this (have not paid attention to CS in years) and it looks correct so I will approve and commit.  We can sort out later if I am wrong.

PR #3540, opened 2026-07-03
Track daily dungeon room order

VeeArrKoL:
The order of daily dungeon rooms is set per-player and does not change from day to day until ascension. This PR creates a preference to track the order of the rooms in the dungeon.

There is some discussion at https://kolmafia.us/threads/request-record-order-of-daily-dungeon-rooms.32648/ regarding whether this is too niche to even bother tracking within KoLmafia. By the time that discussion was in full swing, I'd already sorted out an approach for implementing it, so I figured I may as well toss it into a PR even if it ends up being rejected.

Notes:
* `updateDailyDungeonRoom()` is sort of awkwardly crashing on `QuestManager`'s couch. If there's a better place to put this, I'm open to suggestions.
* The `????_????_????` structure was chosen to make it clear where the fixed treasure rooms are, but I could be convinced that a different representation is better.


jaadams5:
Off the top of my head...

????_????_???? is compact.  It would be nice to have some documentation of what values could replace a question mark and what they mean.  I think M is a room with a monster but I keep remembering all of the trap rooms I needed muscle to defeat and my first though is Muscle trap.  

That is harder to parse than other options.  Imagine what the code to find out what the 7th room was might look like.  

A delimited list would be easier to manage although the builder would have to insert the fixed rooms and possibly be self correcting if there was not a fixed room where one was expected.  If there was a future question about which monsters were encountered in which rooms then something more verbose than a single character per room might be desired and if that was done a delimited list would be a better choice, IMO.

I'd be interested in a test where there was a lost fight and the test showed the preference did not update.  Similarly tests where a trapped room test is both passed and failed would be useful.

I'll circle back to this but like you I think there is something better as a home then Quest Manager.  My first thought is a new class might be a better solution.  If there were a new class I can see it providing utilities that hid the implementation details of the preference which would would futureproof this.  A new class offers opportunities for easy tests as well.

VeeArrKoL:
> ????_????_???? is compact. It would be nice to have some documentation of what values could replace a question mark and what they mean. I think M is a room with a monster but I keep remembering all of the trap rooms I needed muscle to defeat and my first though is Muscle trap.
> 
> That is harder to parse than other options. Imagine what the code to find out what the 7th room was might look like.

It's intended to be compact both for information density and because it makes it easy to programatically parse (e.g. the n-th room is `pref.charAt(n-1)`). I'm not necessarily opposed to making it a comma-separated list of "monster", "trap", "door" or whatever, but that seemed like it would add complexity for little benefit.

> If there was a future question about which monsters were encountered in which rooms then something more verbose than a single character per room might be desired and if that was done a delimited list would be a better choice, IMO.

The intent of the PR was to capture the information that does not change across days of an ascension. The (e.g.) specific monsters encountered change from day to day, but the type of room (monster/trap/door) does not.

> I'd be interested in a test where there was a lost fight and the test showed the preference did not update. Similarly tests where a trapped room test is both passed and failed would be useful.

The preference *should* update in those instances--we know what type of room is it regardless of whether we successfully pass it or not. (There's actually separate checks that happen when encountering the room and when "finishing" the room--winning the combat or selecting a non-retreat choice--and I'd expect that most of the time it will get updated on the first visiting check.)

> I'll circle back to this but like you I think there is something better as a home then Quest Manager. My first thought is a new class might be a better solution. If there were a new class I can see it providing utilities that hid the implementation details of the preference which would would futureproof this. A new class offers opportunities for easy tests as well.

This feels like overkill IMO.

midgleyc:
We have precedence for compact prefs, see e.g. `_trickOrTreatBlock`.

I think the chamber -> type mapping is clear.

midgleyc:
On location: it's not a quest. EncounterManager / SpadingManager are possibilities; I would probably just make a new class for it. I remember making ShrunkenHeadDatabase, so this isn't uncommon.

VeeArrKoL:
Alright, both of you are in agreement so I'll extract `updateDailyDungeonRoom()` into a new `DailyDungeonManager` class. Do you want it to just house this new method, or do you want the existing daily dungeon preference handling refactored into there as well?

jaadams5:
> > ????_????_???? is compact. It would be nice to have some documentation of what values could replace a question mark and what they mean. I think M is a room with a monster but I keep remembering all of the trap rooms I needed muscle to defeat and my first though is Muscle trap.
> > That is harder to parse than other options. Imagine what the code to find out what the 7th room was might look like.
> 
> It's intended to be compact both for information density and because it makes it easy to programatically parse (e.g. the n-th room is `pref.charAt(n-1)`). I'm not necessarily opposed to making it a comma-separated list of "monster", "trap", "door" or whatever, but that seemed like it would add complexity for little benefit.
> 

OK.  I like things I can eyeball and understand immediately.  I also have spent enough time in other languages that I always have to ask or confirm what the start index is for, in this case, a character string.  But this is fine.

> > If there was a future question about which monsters were encountered in which rooms then something more verbose than a single character per room might be desired and if that was done a delimited list would be a better choice, IMO.
> 
> The intent of the PR was to capture the information that does not change across days of an ascension. The (e.g.) specific monsters encountered change from day to day, but the type of room (monster/trap/door) does not.
> 

OK.  I was thinking there might be a way to predict monsters but you are absolutely correct - that is way down the road if it ever becomes useful.

> > I'd be interested in a test where there was a lost fight and the test showed the preference did not update. Similarly tests where a trapped room test is both passed and failed would be useful.
> 
> The preference _should_ update in those instances--we know what type of room is it regardless of whether we successfully pass it or not. (There's actually separate checks that happen when encountering the room and when "finishing" the room--winning the combat or selecting a non-retreat choice--and I'd expect that most of the time it will get updated on the first visiting check.)
> 

I was thinking there were different code paths for success and failure.  But that, if true, is not relevant because the room is known, doesn't change and presumably the preference is recorded.  Underthinking on my part.

> > I'll circle back to this but like you I think there is something better as a home then Quest Manager. My first thought is a new class might be a better solution. If there were a new class I can see it providing utilities that hid the implementation details of the preference which would would futureproof this. A new class offers opportunities for easy tests as well.
> 
> This feels like overkill IMO.

It is good design and implementation in a object oriented world as well as anyplace that uses separation of concerns.  statics are used with care in well formed Java so there is a standard to aspire to.  Mafia is a hybrid implementation that uses a lot or procedural constructs in an object oriented language.  The worst of both worlds. 

A new class handles a separation of concerns even though a class with one static function could considered overkill.  

That said you are not the first person who chose not to make a class that just had one static function and I'm pretty sure there is some class that is nothing but unrelated static functions.  If I can find/remember what it is then this could go there.

jaadams5:
> Alright, both of you are in agreement so I'll extract `updateDailyDungeonRoom()` into a new `DailyDungeonManager` class. Do you want it to just house this new method, or do you want the existing daily dungeon preference handling refactored into there as well?

I'd probably just have the one static function and it reads and writes dailyDungeonRooms.  The only place mafia uses _lastDailyDungeonRoom is when it adds a boring door spoiler in the relay browser.  For this PR I would leave that alone.  If or when we trust out test cases I might consider eliminating that preference and having the relay spoiler use the new preference but that certainly can be another PR.

New class is fine.

jaadams5 (approved):

PR #3543, opened 2026-07-04
Refactor daily dungeon

VeeArrKoL:
I was already elbow deep in the daily dungeon code for #3540 so I figured I may as well refactor some more stuff into the new `DailyDungeonManager`. Now with a shiny new test suite!


jaadams5 (approved):

jaadams5:
Looks good.  Thanks.  I'll let it sit for a bit before committing.

I will probably explore the "self healing" aspects of setting the preference but that will either be for my own amusement or a new PR.

VeeArrKoL:
> I will probably explore the "self healing" aspects of setting the preference but that will either be for my own amusement or a new PR.

Yea, in retrospect it maybe makes more sense to just completely reset it to default if it's not valid.

PR #3544, opened 2026-07-04
Log after-battle message for Sword of S Words

VeeArrKoL:
Log the after-battle message when the Sword of S Words replaces the monster's drop table.


midgleyc (changes requested):
update the test for this

midgleyc (approved):
Thanks!

PR #3545, opened 2026-07-04
Tweak quality control for Daily Dungeon Layout preference

jaadams5:
Added some validation when setting the preference.  If something fails validation just set it to the default.  Guessing what was intended doesn't seem fruitful when the data could be programmatically rebuild.  Fixed one test that broke with the reset to default change and added several tests so that coverage is almost 100%.  Didn't seem worth adding a constructor just to test it.

@VeeArrKoL - comments are welcome.


VeeArrKoL:
I agree that resetting to default is better than whatever jank I was originally trying to do.

That said, might I recommend something like this instead:

```
  private static final Pattern VALID_PREF_PATTERN = Pattern.compile("[MDT?]{4}_[MDT?]{4}_[MDT?]{4}");

  public static boolean validPref(String ddData){
    return VALID_PREF_PATTERN.matcher(ddData).matches();
  }
```

VeeArrKoL:
Also, a test that doesn't really test anything but just makes code coverage bot happier doesn't seem very productive to me, but that's not a hill I'm looking to die on.

src/net/sourceforge/kolmafia/session/DailyDungeonManager.java
```diff
@@ -35,20 +35,44 @@ public static void handleRoomCompletion(int chamber, RoomType roomType) {
...
-      // broken, but oh well.
-      ddData = ddData + "????_????_????".substring(ddData.length());
+    if (!validPref(ddData)) {
+      ddData = Preferences.getDefault("dailyDungeonRooms");
     }
     ddData = ddData.substring(0, chamber - 1) + roomType.code + ddData.substring(chamber);
     Preferences.setString("dailyDungeonRooms", ddData);
   }
 
+  // Visible for testing
```
midgleyc: things don't have to be `public static` for testing, they can just be `static` (package-private)

test/net/sourceforge/kolmafia/session/DailyDungeonManagerTest.java
```diff
@@ -136,15 +139,37 @@ public void dailyDungeonMonsterLoss() {
...
+    "????_??m?_????, false",
+    "????M_???_????, false",
+    "????_??M?_????, true",
+  })
+  public void validityCheckerTest(String beforeLayout, boolean expected) {
+    assertEquals(expected, DailyDungeonManager.validPref(beforeLayout));
+  }
+
+  @Test
+  public void goForTotalCoverage() {
```
midgleyc: tests which call functions and don't test anything are useless. just remove this entirely. maybe the final assertion is worth putting into its own test.

src/net/sourceforge/kolmafia/session/DailyDungeonManager.java
midgleyc, line 8: a lot of classes like this will have a private constructor
```
  private DailyDungeonManager() {}
```
to show they're not supposed to be instantiated. This wouldn't be covered but there's no point to covering it because the whole point is to be unable to instantiate the class.

src/net/sourceforge/kolmafia/session/DailyDungeonManager.java
jaadams5: Fixed

jaadams5:
I left a comment that I cannot find.  If this appears to have duplicate content it is my browser's fault :-)

For me the code as written is easier to read, debug and maintain than a ReGEx and pattern matcher.  We can discuss.

We disagree on the benefits of test code that superficially does nothing that increases coverage,  But the code I was increasing coverage for did have some side effects so I could test whether the side effect happened or not.  Done.

VeeArrKoL:
> For me the code as written is easier to read, debug and maintain than a ReGEx and pattern matcher. We can discuss.

Reasonable minds can differ, but for me the `Pattern` version is much easier to read. Even with essentially no knowledge of regular expressions, it would be clear from reading a single line that "a pref is valid if and only if it matches the format represented by this regex", and the regular expression itself is elementary enough that determining what it does should not be difficult. Further, confusion about what the pattern represents could be resolved by accompanying it with a comment along the lines of:

```
// A valid value for the dailyDungeonRooms preference is of the form "XXXX_XXXX_XXXX", 
// where each X is one of 'M', 'D', 'T', or '?'.
```

jaadams5:
RegEx it is.

midgleyc (approved):
Thanks!

PR #3546, opened 2026-07-05
Fix javascript combat macros by wrapping MonsterData & fix javascript claiming to exit exceptionally

libraryaddict:
MonsterData was not wrapped when it should be, meaning javascript combat macros, passed as a function, via parameter, were erroring out when trying to access the `Monster` that they were given. `JavaScript evaluator exception: Invalid JavaScript value of type net.sourceforge.kolmafia.MonsterData`

This change wraps the `MonsterData` in a javascript wrapper, and fixes the error. Allowing javascript functions to be passed safely as a parameter to `adv1(location, turns, func)` and for the functions to handle monsters where previously they would just error.

The code has overlap with another area, so I extracted it into its own method.

As for tests, there are none. I can't figure out how to do tests without making this a whole ordeal because none of the tests seem to be testable this way, and the closest I got was atrocious, especially for how simple this is.

Regardless, pretty easy to verify at a glance.

I would have preferred to reassign `parameters` in the Macrofier, but the parameters variable needs to be `final`, and reassignment kind of invalidates that... I could have mutated the array, but that would kind of defeat the point of the method existing. And its easier to overcover this than to do the bare minimum.

Regardless, tested this and it seems to be working great, haven't noticed any issues.

# Edit:

Spotted another issue, it seems like the javascript function always sets the EXIT state when it executes, which is wrong in this case. It seems like EXIT is meant to say that the script is aborting or exiting, which seem to be about the same thing for the EXIT?

I can't really find much to indicate "This indicates the script is closing gracefully", it seems more... exceptional.

I only found it an issue because adv1 was returning false on every single invocation.

### DIFF
--- a/src/net/sourceforge/kolmafia/combat/Macrofier.java
+++ b/src/net/sourceforge/kolmafia/combat/Macrofier.java
@@ -133,13 +133,14 @@ public class Macrofier {
         var macroScope = macroOverrideRec.macroScope;
         var macroFunction = macroOverrideRec.macroFunction;
         var macroThisArg = macroOverrideRec.macroThisArg;
+        Object[] jsParameters = JavascriptRuntime.wrapMonsterArguments(macroScope, parameters);
         // Execute a function from the JavaScript runtime maintaining the scope, thisObj etc
         returnValue =
             interpreter.executeFunction(
                 macroScope,
                 () -> {
                   Context cx = Context.getCurrentContext();
-                  return macroFunction.call(cx, macroScope, macroThisArg, parameters);
+                  return macroFunction.call(cx, macroScope, macroThisArg, jsParameters);
                 });
       } else {
         // Execute a single function in the scope of the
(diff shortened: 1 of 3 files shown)


midgleyc:
> I can't really find much to indicate "This indicates the script is closing gracefully", it seems more... exceptional.

I'm not sure whether this works properly at all. There's a comment on `declareWorldPeace`:
```java
  /**
   * Declare world peace. This clears all pending requests and queued commands and notifies all
   * currently running requests that they should stop as soon as possible.
   */
  public static final void declareWorldPeace() {
```

but from my experience smashing escape on JavaScript scripts declaring world peace actually does nothing at all. Or the re-entrant script just immediately re-enters itself.

midgleyc (approved):
Thanks!

libraryaddict:
> > I can't really find much to indicate "This indicates the script is closing gracefully", it seems more... exceptional.
> 
> I'm not sure whether this works properly at all. There's a comment on `declareWorldPeace`:
> 
> ```java
>   /**
>    * Declare world peace. This clears all pending requests and queued commands and notifies all
>    * currently running requests that they should stop as soon as possible.
>    */
>   public static final void declareWorldPeace() {
> ```
> 
> but from my experience smashing escape on JavaScript scripts declaring world peace actually does nothing at all. Or the re-entrant script just immediately re-enters itself.

Pretty sure its because they're wrapping themselves in try {} catch blocks? Or something. It does work, but not reliably yeah.

PR #3548, opened 2026-07-09
feat: update to node 24 for publishing types

midgleyc:
Installing the latest version of NPM installs version 12 which requires node 22. Upgrade to latest LTS.

### DIFF
--- a/.github/npm/package.json
+++ b/.github/npm/package.json
@@ -4,7 +4,10 @@
   "description": "Typings for the KoLmafia JS standard library",
   "main": "index.js",
   "types": "index.d.ts",
-  "repository": "https://github.com/kolmafia/kolmafia",
+  "repository": {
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
This amuses me greatly.  NPM, Node 22 and LTS are not things I deal with normally although I did remember long term stable without help from Google.  I now know the others would be a part of my vocabulary if I did anything with JavaScript besides try and read it.  I also am reminded that they are relevant to KoLmafia because it provides a JavaScript runtime.  But my ignorance amused me.

midgleyc:
Yeah if this doesn't fix it I may have to call in gausie who actually works with this professionally.

midgleyc:
We have hit https://github.com/npm/cli/issues/9722

Tempted to just wait and see if it gets fixed.

PR #3549, opened 2026-07-09
feat: add cup of 13s helper functions

midgleyc:
heartstone_string_length is used to calculate which effect is associated with an item. The 'strip spaces, use utf-8' strategy.

cup_of_13s_tier is probably based on mall price at the time of generation. On discord it's called 'value' but we call so many things 'value' I picked a synonym.

### DIFF
--- a/src/net/sourceforge/kolmafia/persistence/CupOf13sDatabase.java
+++ b/src/net/sourceforge/kolmafia/persistence/CupOf13sDatabase.java
@@ -1,0 +1,50 @@
+package net.sourceforge.kolmafia.persistence;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import net.sourceforge.kolmafia.KoLConstants;
+import net.sourceforge.kolmafia.KoLConstants.MafiaState;
+import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.StaticEntity;
+import net.sourceforge.kolmafia.utilities.FileUtilities;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
+
+public class CupOf13sDatabase {
+  private CupOf13sDatabase() {}
+
+  private static final Map<Integer, Integer> cupMap = new HashMap<>();
+
+  static {
+    CupOf13sDatabase.reset();
+  }
+
+  public static int getTier(int itemId) {
+    return cupMap.getOrDefault(itemId, -1);
+  }
+
+  private static void reset() {
+    boolean error = false;
+    try (BufferedReader reader =
+        FileUtilities.getVersionedReader("cup_of_13s.txt", KoLConstants.CUP_OF_13S_VERSION)) {
+      String[] data;
+
+      while ((data = FileUtilities.readData(reader)) != null) {
+        if (data.length >= 3) {
+          int itemId = StringUtilities.parseInt(data[0]);
+          int itemTier = StringUtilities.parseInt(data[2]);
+
+          cupMap.put(itemId, itemTier);
+        }
+      }
+    } catch (IOException e) {
+      StaticEntity.printStackTrace(e);
+      error = true;
+    }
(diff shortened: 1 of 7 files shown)


src/net/sourceforge/kolmafia/persistence/CupOf13sDatabase.java
```diff
@@ -0,0 +1,50 @@
...
+      while ((data = FileUtilities.readData(reader)) != null) {
+        if (data.length >= 3) {
+          int itemId = StringUtilities.parseInt(data[0]);
+          int itemTier = StringUtilities.parseInt(data[2]);
+
+          cupMap.put(itemId, itemTier);
+        }
+      }
+    } catch (IOException e) {
+      StaticEntity.printStackTrace(e);
```
jaadams5, line 42: This is just a place to put a question and not a blocker at all.  Is there any 'policy" for what we do when KoLmafia tries to read a datafile and gets an error?  Given that we are building a database I can imagine that we might want to reset or invalidate the database or perhaps even use the specific exception to decide whether a second attempt might work?

src/net/sourceforge/kolmafia/persistence/HeartstoneDatabase.java
```diff
@@ -8,6 +8,12 @@ private HeartstoneDatabase() {}
 
   public record MiddleLetter(String letter, int byteIndex) {}
 
+  public static int spaceStrippedStringLength(String str) {
+    var compact = str.replaceAll("\\s+", "");
+    var bytes = compact.getBytes(StandardCharsets.UTF_8);
+    return bytes.length;
```
jaadams5, line 14: Again, non-blocking question based upon my lack of game knowledge.  Are HTML character entities and encoding/decoding a factor here?  Is it the caller's decision or is there some kind of checking elsewhere?  Maybe I have a chance to learn something about UTF8 that I did not know or remember?

jaadams5 (approved):
Questions are not blocking.

src/net/sourceforge/kolmafia/persistence/HeartstoneDatabase.java
midgleyc: HTML entities are a factor, but I'm relying on the caller to pass in the string in the correct format instead of trying to guess what they're doing and do what they probably wanted.

Same thing we do for heartstone_middle_letter.

src/net/sourceforge/kolmafia/persistence/CupOf13sDatabase.java
midgleyc: I copied this from Cultist Shorts Database, I don't think we have a general policy. It would be nice ;)

src/net/sourceforge/kolmafia/textui/RuntimeLibrary.java
```diff
@@ -4006,12 +4007,18 @@ private static VariableReference namedParam(String name, Type type) {
     params = List.of(namedParam("monster", DataTypes.STRING_TYPE));
     functions.add(new LibraryFunction("heartstone_middle_letter", DataTypes.STRING_TYPE, params));
 
+    params = List.of(namedParam("string", DataTypes.STRING_TYPE));
+    functions.add(new LibraryFunction("heartstone_string_length", DataTypes.INT_TYPE, params));
```
VeeArrKoL, line 4011: Seeing as this is now used for more than just Heartstone, maybe a less specific name would be better?

midgleyc: It first came out with heartstone. If I call this "spaceless_string_length" and then TPTB introduce a slightly different way of calculating the length of a string while stripping spaces (e.g. the HTML entity handling differs), I'd be in trouble.

So yeah I considered this but went with heartstone_ in the end.

VeeArrKoL: Fair enough.

PR #3551, opened 2026-07-10
Remove npm update step from publish workflow

gausie:
https://github.com/npm/cli/issues/9722


jaadams5 (dismissed):

midgleyc:
Honestly might be better to just remove that step entirely, I think it was added for trusted publishing on node 20, because the default npm wasn't good enough? But now it is.

gausie:
@midgleyc great point. How about now

midgleyc (approved):
Thanks!

PR #3552, opened 2026-07-10
Fix Encyclopedia wiki links

jaadams5:
The wiki links for items were not working although the in game links were.  I fixed that and the discovered familiars, skills and effects were also broken.  So I fixed them as well.

I am leaving this as draft for now for two reasons.  One, Vampire Bat as a familiar doesn't work as expected.  Either it is not being disambiguated or disambiguation needs to be tweaked.  Second I can't recall any recent changes in this area.  Breaking this may have been a side effect of some refactoring or structural things and if that is correct then my fix may be wrong.

### DIFF
--- a/src/net/sourceforge/kolmafia/utilities/WikiUtilities.java
+++ b/src/net/sourceforge/kolmafia/utilities/WikiUtilities.java
@@ -10,6 +10,7 @@ import net.sourceforge.kolmafia.modifiers.StringModifier;
 import net.sourceforge.kolmafia.objectpool.Concoction;
 import net.sourceforge.kolmafia.persistence.ConcoctionDatabase.QueuedConcoction;
 import net.sourceforge.kolmafia.persistence.EffectDatabase;
+import net.sourceforge.kolmafia.persistence.FamiliarDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
 import net.sourceforge.kolmafia.persistence.MonsterDatabase;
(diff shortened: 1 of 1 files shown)


VeeArrKoL:
Not sure if it's useful for disambiguation here or if it just causes more problems, but the wiki now has `<thing>:<id>` redirects for most kinds of thing. For example, [Monster:734](https://wiki.kingdomofloathing.com/Monster:734) (grouper groupie monster) vs. [Familiar:116](https://wiki.kingdomofloathing.com/Familiar:116) (grouper groupie familiar).

midgleyc:
Your fix looks right to me.

On the redirects: I would prefer not, because I use `to_wiki_url` to get an edit link, and `https://wiki.kingdomofloathing.com/index.php?title=Item:11788&action=edit` doesn't work as I want it to.

jaadams5:
Thank you.  I have undrafted this.

There is a routine to disambiguate types (in the same file) and my first thought when I saw "vampire bat" was that my fix had bypassed code.  Since disambiguation  code deals with types and not names my first observation is wrong.  I just checked against a previous version of mafia and the game description for vampire bat has the same problem with and without this fix.  So this is not a new problem that I introduced.

To the extent that I understand the redirect point I have no intention of doing that with this PR.

I may investigate ambiguous cases and ponder possible fixes.  But that will be another PR :-)

midgleyc (approved):
Thanks!

There are quite a lot of incorrect links, e.g. many items with names identical to other items go to the wrong place. But this is a pre-existing problem.

gausie:
> Your fix looks right to me.
> 
> 
> 
> On the redirects: I would prefer not, because I use `to_wiki_url` to get an edit link, and `https://wiki.kingdomofloathing.com/index.php?title=Item:11788&action=edit` doesn't work as I want it to.

I can probably make that work cor you

PR #3553, opened 2026-07-12
Fix interact() modifier expression parsing

VeeArrKoL:
Because the `interact()` modifier expression function is formatted like a function but does not take any parameters, it is not being correctly parsed by the expression engine and can corrupt the expression stack and produce incorrect results for complex expressions. This fix changes it to be parsed like non-parameter-taking expression "functions" like `paradoxicity` and `unarmed`. 

I opted to leave it expressed as `interact()` rather than `interact`, but if anyone feels strongly about it, we can change it to `interact` in the very few places it's used.


midgleyc (approved):
Thanks!

PR #3556, opened 2026-07-13
Daily limit for Calculate the Universe

VeeArrKoL:
https://kolmafia.us/threads/all-skills-with-daily-limits-should-appear-in-dailylimits-txt.32632/
Another missing dailylimits.txt entry, this time for Calculate the Universe. This one is a bit weird due to all of the weird behavior of Calculate the Universe, which I *think* is now all properly tracked:

* Using the skill only sends you to the Doing the Maths choice, so avoid incrementing the counter when the skill is used and manually increment it if the choice results in something other than a Try Again.
* Some choices result in a combat. This seems to Just Work™️so I assume there's some redirect shenanigans going on, but I won't look a gift horse in the mouth.
* The daily limit depends on whether or not the user is in Ronin/HC, so we have to do weird gymnastics to set the daily limit.
  * Additionally there's a weird interaction where reaching the limit in Ronin/HC blocks further use of the skill for the rest of the day even if the prism is later broken.

### DIFF
--- a/src/data/dailylimits.txt
+++ b/src/data/dailylimits.txt
@@ -60,6 +60,7 @@ Cast	CHEAT CODE: Invisible Avatar	_powerfulGloveBatteryPowerUsed	100
 Cast	CHEAT CODE: Replace Enemy	_powerfulGloveBatteryPowerUsed	100
 Cast	CHEAT CODE: Shrink Enemy	_powerfulGloveBatteryPowerUsed	100
 Cast	CHEAT CODE: Triple Size	_powerfulGloveBatteryPowerUsed	100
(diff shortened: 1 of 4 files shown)


midgleyc (approved):
Thanks!

At first I thought there was an issue with in-run casting setting to max, in that the daily limit is lower and any use of `increment` will set it back to 3 instead of increasing.

But you can't successfully cast it in that condition, so I don't think we'll ever call `increment`.

Malibu-Stacey:
> Additionally there's a weird interaction where reaching the limit in Ronin/HC blocks further use of the skill for the rest of the day even if the prism is later broken.

I assume this has been bug reported in game?

VeeArrKoL:
> I assume this has been bug reported in game?

You know what happens when you assume! (I don't actually have this much numberology--I'm going based off of what the wiki says and butts corroborated in Discord.)

PR #3557, opened 2026-07-13
Fix Javascript swallowing aborts

libraryaddict:
# What's broken

Using `abort()` in a Javascript script does not actually abort. [`JavascriptRuntime.checkInterrupted()`](https://github.com/kolmafia/kolmafia/blob/2977765f95fd5001b5347b7b6649456c9dc65adc/src/net/sourceforge/kolmafia/textui/javascript/JavascriptRuntime.java#L419) treats every stopped state the same, with `!permitsContinue()` clearing the state and turning it into a `JavascriptException`.

1. The new exception is catchable, so scripts can swallow the abort and keep going. Even a user abort doesn't get through.
2. The state is cleared before the throw, so when a nested script aborts (e.g. `cliExecute("ashq abort('...')")` or a `betweenBattleScript` inside `adv1()`), the caller sees `CONTINUE` and carries on.

# How it should work

KoLmafia has three failure modes, I'm using ASH as defining the intended behavior.

1. *Thrown exceptions.* Library functions, bad scripts, or `throw`. These are language exceptions, the state remains `CONTINUE`, and scripts may catch them.
2. *Soft stops.* Something sets `ERROR`, but `refusesContinue()` is false. ASH halts unless the script handles the error. Javascript treats every ASH call as handled, so the call returns `false`, the state is reset, and the script decides what to do. This is unchanged.
3. *Aborts (hard stops).* These stop every script on the stack, regardless of nesting, and cannot be caught. The state remains `ABORT` until the stack finishes unwinding.

Javascript handled the first two, but treated `3. aborts` as soft stops.

# The fix

`checkInterrupted()` now checks `refusesContinue()` first and throws a new `AbortException` extending `RuntimeException`. Rhino does not let Javascript catch those, but still runs `finally` blocks while unwinding, matching ASH.

The exception temporarily suspends the global abort state so `finally` blocks can still make library calls. At the script boundary, `executeFunction()` restores the abort state for the caller. ASH's `Try` does the same: save the abort, `forceContinue()`, run `finally`, then restore.

With `printStackOnAbort` enabled, each unwound script also prints its stack trace, matching ASH. That preference applies to `ABORT`, not `ERROR`, so the failing test is correct.

* `catch`/`capture` cannot swallow an abort.
* `finally` still runs cleanup (including ASH calls).
* Stack traces only print when the preference is enabled.
* The abort clears once the outermost request sequence ends.

The tests should cover most cases, let me know if I missed something.

This will probably break a few scripts, but those were relying on buggy behavior by treating `abort()` like `throw`.

The main change people will see is that a single abort can now reliably stop Javascript scripts instead of requiring repeated aborts. Operations that hang (eg kolmafia code issues or networking) are a separate issue.

Relevant:

* https://kolmafia.us/threads/cant-stop-gash-jumping-in-preascensionscript-from-js-js-abort-and-ash-abort-behave-differently.29618/
* https://github.com/kolmafia/kolmafia/pull/3546#issuecomment-4887310854

# Closing thoughts

I note that when I abort() in my pre-adv script, it works as intended, but the stack trace is slightly confusing. The abort *message* only shows at the original abort, with the following stack traces only being a stack trace, no messages. This, might not be the best behavior.

But as a counterpoint, this is only visible when `printStackOnAbort` is set.

The other perspective is that the script should not be sending another stacktrace, I think it's already captured in the first stacktrace.

This didn't occur to me beforehand, so I'm not sure if it should change or if its best as it is now.

Eg
image

### DIFF
--- a/test/net/sourceforge/kolmafia/textui/AbortPropagationTest.java
+++ b/test/net/sourceforge/kolmafia/textui/AbortPropagationTest.java
@@ -438,15 +438,16 @@ public class AbortPropagationTest {
 
     /**
      * As in ASH, an abort prints its message once (from {@link RuntimeLibrary#abort(ScriptRuntime)}
-     * itself) and unwinds quietly. With 'printStackOnAbort', each script it unwinds also prints its
-     * stack trace, the Javascript version of {@link AshRuntime#setState(ScriptRuntime.State)}'s
-     * stack printing. Two scripts unwind here, the nested one and the caller, so two traces print,
-     * each a single "at command line" frame. Plus one for the test framework.
+     * itself) and unwinds quietly. With 'printStackOnAbort', the script the abort originated in
+     * also prints its stack trace, the Javascript version of {@link
+     * AshRuntime#setState(ScriptRuntime.State)}'s stack printing. Nested scripts share the context,
+     * so only the first trace gets printed, the remaining traces are suppressed for the current
+     * thread.
      *
      * <p>Note that the preference does not control if non-abort error stacks are printed.
      */
     @Test
-    void abortPrintsAStackTracePerUnwoundScriptWhenPreferenceSet() {
+    void abortPrintsOneStackTraceWhenPreferenceSet() {
       try (var cleanups = withProperty("printStackOnAbort", true)) {
         var run =
             runJs(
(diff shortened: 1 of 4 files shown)


libraryaddict:
Hmm, it's only now that I realize that maybe this isn't exactly what I want.
I wanted to have autoscend.js properly abort() in its pre-adventure scripts, but the issue is that it's killing my parent script as well, which is fully intended, but uh, not what I wanted.

So I guess I'll mark this as a draft, and think on it. Ideally, we'd have some way to say "call this script, it can abort, but don't abort the parent script"
But maybe the issue is just that abort() should be left as that, and we should never be able to suppress it. Instead if we're going to use it like that, think about redesigning the script to stop using hard kills.

And, maybe we should be propagating script exceptions as well, though that's probably going to cause some grief? Don't know.

midgleyc:
I think this is better than what we have, and I look forward to being able to stop scripts properly without e.g. escape only canceling the current excavator call.

libraryaddict:
Yeah true, I think my realized issues may be valid, but that's more of a feature.

It's just the stack trace I suppose, simple enough change to add `Script aborted: <Reason>\n`.  - er, maybe not. Looks confusing, and it's still the question of if I should even be printing a stacktrace multiple times or just the initial abort.

libraryaddict:
Yeah a stacktrace only on the initial abort is the ash behavior? And makes more sense.

test/net/sourceforge/kolmafia/textui/AbortPropagationTest.java
```diff
@@ -455,8 +456,27 @@ void abortPrintsAStackTracePerUnwoundScriptWhenPreferenceSet() {
                     print("After abort");
                     """);
 
-        assertThat(run.output().split("Nested abort", -1).length - 1, is(1));
-        assertThat(run.output().split("at command line", -1).length - 1, is(3));
+        // We expect only one "Script aborted: Nested abort\n<trace>"
+        assertThat(run.output().split("Script aborted", -1).length - 1, is(1));
```
midgleyc: Would you mind making a custom matcher for this (in `internal.matchers`)?

I think the intent of this is that the number of times the substring you're searching for appears in the text is the number at the end, so just for clarity it would be nicer to have it called as 
```
assertThat(run.output(), containsStringTimes("Script aborted", 1));
```

midgleyc (approved):
Thanks!

PR #3558, opened 2026-07-15
Fix maximizer changing modes of items in ignored slots

libraryaddict:
Maximizer currently considers items in slots we ignore, to see if it can mess with their modes.

This generally isn't desirable or expected from a user perspective.

The user expects `maximize meat -acc1` not to mess with the backup camera in acc1. And yet...

This PR aims to fix that by skipping items already equipped, in a slot that we're ignoring.


midgleyc (changes requested):
Thanks for the existing tests, can you add one more test that confirms that for an item equipped in a *non* excluded slot, we *do* change the mode (we do, I just want a test for regression).

midgleyc (approved):
Thanks!

PR #3559, opened 2026-07-15
Allow maximizer to define what modes to set modeable equipment to

libraryaddict:
Maximizer currently does not let users specify modes on equipment, such as a backup camera.

This is kind of a pain when you're trying to maximize and you want the parka to go into a mode to use a skill, except the parka's modifiers are not optimal for the maximizer evaluation. It requires the user to execute a command to further adjust the parka, after maximizer has adjusted it.

This PR aims to fix that by allowing users to specify a mode for their equipment, which forces that mode to be used.

Eg, `maximize +equip backup camera (ml), -ml` will now put the backup camera into +ml.

It doesn't require "mode" to be in the parentheses because it'd be inconsistent UI, given that it's being displayed in some modes.

There's a chance this PR will have a merge conflict with the other PR I have open, because of the tests being on the same line.
I seperated the two PR as while they basically touch the same topic, they are doing different things.


test/net/sourceforge/kolmafia/persistence/ItemFinderTest.java
```diff
@@ -855,4 +856,63 @@ public void itShouldReturnGemsWithEquip() {
       assertEquals(1, item.getCount());
     }
   }
+
+  @Test
```
midgleyc, line 860: add a test for parsing an item without "mode" (e.g. "Snow Suit (nose)")

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -1553,6 +1556,179 @@ public void shouldNotSuggestUsingLedCandleWithoutJill() {
...
+
+    @Test
+    public void softRequestShouldGiveWayWhenIndirectConflict() {
+      final var cleanups =
+          new Cleanups(
+              withEquippableItem(ItemPool.CROWN_OF_ED),
+              withEquippableItem(ItemPool.OLD_SCUBA_TANK));
+
+      try (cleanups) {
+        // 'sea' cannot override an explict mode. We equip a scuba tank to sastify 'sea'
```
midgleyc, line 1714: add to maximizer-help.html that "equip" needs to be declared before other modeable-changing keywords or it might error.

Preferably I guess we would have it work either way, but that is a more significant behavioural change.

libraryaddict:
Just as well, I had forgotten to handle the existing normalize() methods, and majority of modes were not being normalized. I'd expect `snow suit (Nose)` to work after all.

libraryaddict:
Not too content with the wording of maximizer help, but idk. My main intent there is just that we don't add punctuation next to the `<b>` to avoid it being confused as part of the command.

Also, just wanted to say thank you @midgleyc for your continued support for kolmafia.

libraryaddict:
Actually, my change does impact the way we use modeables on the commands, I'll update the tests

libraryaddict:
Trying to figure out if I should be changing modeable commands, eg `backupcamera meat` to allow case insensitivity, `backupcamera MEAT`
And if not, if the same rule should be applied to maximizer `backup camera (meat)` not `backup camera (MEAT)` or `backup camera (meat Mode)` but yes `backup camera (meat mode)`

libraryaddict:
Minor thought is that the maximizer help doesn't say you can use (mode) on `bonus`, but that should be easy to figure out.

src/data/maximizer-help.html
```diff
@@ -44,7 +44,7 @@ <h3>Equipment</h3>
 <br><b>knife</b> - With positive weight, only knives will be considered for your main-hand.
 <br><b>accordion</b> - With positive weight, only accordions will be considered for your main-hand.
 <br><b>current</b> - By default Maximizer considers your current equipment in Hardcore and Ronin. With -current keyword it does not, with +current keyword in aftercore it forces it to be considered. Not considering current equipment will reduce the number of combinations considered, but will sometimes mean changes of equipment are suggested even though they don't improve what you are wearing.
-<br><b>equip <i>item</i></b> - The specified item is required (positive weight) or forbidden (negative weight).  Multiple uses of <b>+equip</b> require all of the items to be equipped.
+<br><b>equip <i>item</i></b> - The specified item is required (positive weight) or forbidden (negative weight).  Multiple uses of <b>+equip</b> require all of the items to be equipped. Modeable items such as <i>backup camera</i> can specify a mode <b>+equip backup camera (ml)</b> in parentheses.
```
midgleyc: Add something like "If you are receiving an error about conflicting modes, try moving this 'equip' attribute to the start"

As it's not obvious that something like "sea" would affect the edpiece from the outside.

Also, yes, mention modes in bonus if it works, and add a test.

midgleyc:
>Also, just wanted to say thank you midgleyc for your continued support for kolmafia.

You're welcome; thanks for the PRs.

libraryaddict:
Yeah, I messed up. Forgot to implement the "bonus" side, knew it was too painless to be true.
Should be good to go now.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -1573,7 +1627,8 @@ void enumerateEquipment(EquipScope equipScope, int maxPrice, PriceLevel priceLev
           newMods.add(ModifierDatabase.getModifiers(ModifierType.EFFECT, intrinsic));
           mods = newMods;
         }
-        double delta = this.getScore(mods, Map.of(Slot.HAT, item)) - nullScore;
+        // Modeable items never reach here (they break gotItem above), so we leave out modes
+        double delta = this.getScore(mods, Map.of(Slot.NONE, item), Map.of()) - nullScore;
```
libraryaddict, line 1631: The enum change from "HAT" to "NONE" on that one line in Evaluator should be harmless, I believe it's dead code. Only looked at it because HAT did not make sense at all to be there.

midgleyc (approved):
Thanks!

For the tests with Jarlsberg's pan / Boris's helm, checking they would be equipped doesn't prove anything because they would be equipped even if the bonus keyword didn't work.

But the `assertTrue` on the maximize does mean that either parsing works correctly or it fails to fail on an invalid bonus, but there are tests elsewhere showing the latter works. So it's all good.

libraryaddict:
Yeah, my brain was getting a bit fried

PR #3560, opened 2026-07-16
Ash functions for selecting a seeded random number from a range

VeeArrKoL:
ASH currently provides the ability to get the raw next int for a seeded RNG, but this is a pain to work with (since the user ends up having to normalize it themselves). I don't want to go too crazy exposing additional RNG functions, but these ones (which allow the user to specify the range to select from) seem likely to be much more useful in most cases.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/Rng.java
+++ b/src/net/sourceforge/kolmafia/textui/Rng.java
@@ -20,7 +20,15 @@ public class Rng {
     return rand.nextInt();
   }
 
+  public int nextRandInt(int min, int max) {
+    return rand.nextInt(min, max);
+  }
+
   public int nextMtRandInt() {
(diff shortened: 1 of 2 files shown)


jaadams5 (approved):
`<snark>yet another niche change that could be done by a script</snark>`

:-)

I'll approve and circle back tomorrow to push if no one else does.

I couldn't really think of a useful test either.

VeeArrKoL:
> `<snark>yet another niche change that could be done by a script</snark>`

Haha, I was expecting this. It certainly can be done by a script, but when I got to the point where I had four separate scripts for different things doing the same little normalization I had to wonder if maybe it just made sense for the standard library to have the versions of the functions a user would actually want to use. (If not for backward-compatibility, I'd actually argue for these to *replace* the existing ones!)

PR #3561, opened 2026-07-17
Redirect empty relay url path to /game.php

Rinn:
Rather than call sendLocalFile() with an empty path which will return a blank page


jaadams5:
How does this relate to https://github.com/kolmafia/kolmafia/pull/3399?

midgleyc:
Don't think it does; that has to do with relay scripts, and there are no scripts involved here.

midgleyc (approved):

PR #3564, opened 2026-07-18
Harden preference file saves against power-loss corruption.

DamianDominoDavis:
Write prefs via temp+fsync+atomic replace (with backup-safe fallback), refresh *.bak only on durable save, and restore truncated prefs from backup without clobbering it on the next launch.


jaadams5:
@BadHorseMonkey 

This reminds me of some previous attempts to do something similar that were ultimately not committed.  I'd like some time to study this to understand how it is different.

midgleyc:
Possibly https://github.com/kolmafia/kolmafia/pull/1811 (reverted in https://github.com/kolmafia/kolmafia/pull/1886) and the subsequent https://github.com/kolmafia/kolmafia/pull/1889.

Rinn:
just some positive reinforcement i work on build systems professionally and this style of write to a temp file then rename into the final location is pretty much the exact solution I've always used to prevent output file corruption

jaadams5:
I now understand the difference between this and the previous work.  This is about files on disk where the previous work was about what to synchronize and when.

That said, I have a nit or two.  First a corrupt file is somewhat arbitrarily defined as half the expected size.  Instead of embedding "2" as a "magic constant" give it a name and use the named constant.  Second, I would love to see a test where the file was corrupt but it fails the size test.  If a good file has 100 bytes and a corrupt file has 50 bytes or less then the test would have a 60 byte file.  If this suggestion indicates some kind of misunderstanding on my part then just roll your eyes and explain to me politely :-)

I'll withhold approval until the merge conflict is resolved but I'm fine with this.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -332,6 +317,36 @@ private static void loadUserPreferences(String username) {
...
+    if (prefsFile == null) {
+      return null;
+    }
+    String name = prefsFile.getName();
+    if (!name.endsWith("_prefs.txt")) {
+      return null;
+    }
+    String bakName = name.substring(0, name.length() - ".txt".length()) + ".bak";
+    File parent = prefsFile.getParentFile();
+    return parent == null ? new File(bakName) : new File(parent, bakName);
```
libraryaddict, line 293: This doesn't make sense to me, if the parent is null then there's bigger issues going on and we shouldn't be silently swallowing it

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -332,6 +317,36 @@ private static void loadUserPreferences(String username) {
...
+  private static boolean prefsFileSuspect(File prefsFile, File backupFile, Properties loaded) {
+    if (loaded.isEmpty()) {
+      return true;
+    }
+    if (backupFile == null || !backupFile.exists()) {
+      return false;
+    }
+    long bakLen = backupFile.length();
+    long prefsLen = prefsFile.length();
+    return bakLen > 0 && prefsLen < bakLen / 2;
```
libraryaddict: As mentioned by other people, the length of the file although it can be indicative, it's not conclusive.

This is also very problematic in the event a user has a script that wrote a lot of properties, suddenly this is wiping their prefs. I could've missed something here though.

midgleyc:
In my experience a corrupted file is significantly *larger* than expected, because it has a lot of null bytes in it. Probably better to just switch that to Irrat's check of containing a null byte, unless anyone has experienced the "smaller than expected" version.

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -898,21 +913,96 @@ private static void saveToFile(File file, Map<String, byte[]> encodedData) {
...
+      try {
+        try (FileOutputStream fos = new FileOutputStream(tempFile);
+            BufferedOutputStream bos = new BufferedOutputStream(fos)) {
+          synchronized (encodedData) {
+            for (Entry<String, byte[]> current : encodedData.entrySet()) {
+              bos.write(current.getValue());
+            }
+          }
+          bos.flush();
+          fos.getFD().sync();
```
libraryaddict, line 962: Was benchmarking performed for this change?

Quick testing showed that I was doing more total bytes written, and each write took 15 times longer.

I would prefer this can be opt-out, as it feels like this is going to increase the amount of unneeded wear on a drive, and slow down the program when doing a lot of preference changes.

Referring to `getFD().sync()`

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -898,21 +913,96 @@ private static void saveToFile(File file, Map<String, byte[]> encodedData) {
...
+      Preferences.commitPrefsTempFileNonAtomic(tempFile, targetFile, e);
+    }
+  }
+
+  /**
+   * Portable commit when {@link StandardCopyOption#ATOMIC_MOVE} is unavailable. Relies on {@code
+   * *.bak} holding the previous good contents (refreshed by the caller). Removes the live file,
+   * then renames the complete temp into place — never opens/truncates the live path for writing.
+   */
+  static void commitPrefsTempFileNonAtomic(
```
libraryaddict, line 1016: Feels like there's a lot of code that should've been collapsed into less methods.

libraryaddict (commented):
I'm unsure if solutions like One Drive, google drive and whatever, can handle this smartly. Some people have weird setups.

Also a bit concerned for getFD().sync(), kolmafia doesn't batch the changes.

I'd prefer if this can be opt-out.

src/net/sourceforge/kolmafia/preferences/Preferences.java
libraryaddict: Yeah, the actual wear and tear is probably neglibable, just 14% more for me. The slowdown is the only real issue, and that's mostly people on old drives, and ascension + rollover.

jaadams5:
Not a big fan of making this opt out.  Not a big fan of claims about timing without any discussion of how the timing measurement was made. 

I agree that some of teh nul checks shuld be treated as disasters if they check fails but the cases i checked aren't going to return null so maybe a comment or rewrite would be appropriated.

libraryaddict:
> Not a big fan of claims about timing without any discussion of how the timing measurement was made.

https://unix.stackexchange.com/questions/475500/how-continuously-spinning-sync-call-can-cause-high-io-wait

midgleyc:
In terms of timing, any delay is surely going to be dwarfed by the next network call: milliseconds compared to hundreds of milliseconds.

libraryaddict:
> In terms of timing, any delay is surely going to be dwarfed by the next network call: milliseconds compared to hundreds of milliseconds.

Agreed, it's when we're chaining preference modifications that it gets noticeable. There's no temporary file saving turned off for example with daily prefs or ascensions.

jaadams5:
I'm not certain closing this was the right move so I am reopening it while I ponder that.  I will mark it as a draft while doing so.

In general if we are going to maintain a preference backup and then update the preference file from the backup then I am pretty convinced this is the correct implementation.

I will deal with the merge, consider or reconsider dealing with some of the code nits and explore whether there is an easy way to get before and after performance metrics.  An 7+ year old analysis on *nix is a place to start but is not convincing, by itself.  

Since we are discussing backup and recovery changes elsewhere it is possible that this will become obsolete or merely inspirational but I suspect it is necessary as long as we try and fix a "live" preferences file on the fly.

I have about 25 hours of non-negotiable, and additional, real life tasks this week so I probably will not get much done on this until next week but at least this is a marker.

gausie:
What's going on here? As the owner of an insanely old draft PR I hate draft PRs stacking up 😉

jaadams5:
> What's going on here? As the owner of an insanely old draft PR I hate draft PRs stacking up 😉

:-)

The original author closed it, I think because there was some vocal resistance.  I thought it was a good idea and the right thing to do so I reopened it as a draft.  It turns out merging this with the other PR that changed Preferences and was been committed is much harder than I expected.  I have a failing test and cannot figure out whether I am picking the wrong pieces and places to merge or whether there is "missing" code that needs to be written.  I've been on travel but I need to solve this or give up.  So I will either close or undraft it in a couple of days.  Stand by.

jaadams5:
Will now merge with a main that includes #3565.  Test error is either because code needs to be changed after #3565 or because I screwed up in what I did to handle the merge conflicts.  My hope/intent is to let this sit for a day or two and either make the broken test work or close it in its current state.

src/net/sourceforge/kolmafia/preferences/Preferences.java
jaadams5: No.  If the parent is null then it means prefsFile does not specify a parent.  It is the difference between "fronobulax_prefs.txt" and "C:\Kolmafia\dist\settings\fronobulax_prefs.txt".  If no parent is specified than the backup file will be created using runtime environmental data or defaults.

src/net/sourceforge/kolmafia/preferences/Preferences.java
jaadams5: A preferences file that contains a zero byte is considered bad but recoverable from the beginning of the file to the 0 byte.  If the backup contains more than twice as much data then the prefs then the backup is considered correct.  You are correct in that if I deleted enough preferences it would be possible for a backup that was made before the deletions to replace the correct file.  I am not sure how likely that is however and someone else made the decision that a size comparison was better than doing nothing.  At some point the correct answer is not to replace a preferences file with a backup but to create a new file created by some kind of merge operation between the prefs file, a backup, and memory contents.  Since the replacement decision predates this PR then perhaps considering a merge is a candidate for future work?

jaadams5:
I resolved the issues with merging with #3565 to my satisfaction, added some more tests and cleaned up some lint.

As I see it the PR changes how and when a backup file replaces a preferences file but (with one exception, see below) but does not change the pre-existing decision to do so.  So opting out, or not, is out of scope as far as I am concerned.  

This code does introduce a file length check to decide when to replace a preferences file.  That could be discussed and if necessary the length check could be removed and reintroduced if appropriate.  

Future work might include a more robust means of deciding a file needed to be replaced.  Future work might also consider some kind of merge operation instead of replacement.  Future work might include determining if any of the tests are redundant and refactoring related to duplicate code.

I have not marked them as resolved yet but I do consider previous conversations as resolved or out of scope.  Please repeat them if you disagree.

gausie:
Ok so before we truncated the prefs file then rewrote it. That meant that power loss in the middle of that process wiped the prefs file. This PR includes a change to making that write atomic. Great!

Having made truncation impossible, the PR then adds a whole apparatus to detect and recover from truncation. That seems useless - am I misunderstanding?

Secondly loadUserPreferences and loadPreferencesWithBackup seem to both disagree with eachother and also duplicate a bunch of code.

jaadams5:
> Ok so before we truncated the prefs file then rewrote it. That meant that power loss in the middle of that process wiped the prefs file. This PR includes a change to making that write atomic. Great!
> 
> Having made truncation impossible, the PR then adds a whole apparatus to detect and recover from truncation. That seems useless - am I misunderstanding?
> 
> Secondly loadUserPreferences and loadPreferencesWithBackup seem to both disagree with eachother and also duplicate a bunch of code.

Thank you.  This was basically abandoned work that I thought was worth keeping.  But I had to merge it with with #3565 and lost track of what was happening.  Your comments clarified what should be going on.  The previously existing code that decided a prefs file was bad and loaded from the backup needs to be restored/preserved.  The definition of "bad" needs to be consistently applied.  I'll focus on that.

jaadams5:
WIP.  Test failure expected.

jaadams5:
@gausie

Let's try again.

I cleaned out a lot of stuff that dealt with when to use a backup.  In hindsight I think I introduced it when trying to merge.

This now writes prefs via temp+fsync+atomic replace (with backup-safe fallback) and refreshes *.bak only on durable save,

Pre-existing code decided when to use a backup and that, and associated tests have not changed.

There was code that used file length to decide if a member of a prefs, bak pair was corrupt.  Since that was somewhat controversial I have left it out of this PR.  If/when this is committed the detection of corrupt files can be revisited and expanded.

There are a few places where lint was cleaned up and containsNullBytes throwing an exception was not needed/helpful, in my opinion.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -779,12 +915,19 @@ public void canToggle() {
   }
 
   @Nested
+  /* This class has tests that deal with corrupt preferences and backups.  Several tests were added
+    after work making backup writing and replacing prefs with the backup more robust.
+    Embedded in these tests and the associated code are the ideas that: If length of one of the
+    preference or backup files is less than half that of the other then one is corrupt; if a null or
+    zero byte is in either file then it is corrupt; and a data before the first null is recoverable.
+  */
```
gausie, line 922: This seems wrong now?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -73,10 +77,142 @@ void TestBackupFileWrite() {
...
+        String backupContents = Files.readString(backupFile().toPath(), StandardCharsets.UTF_8);
+        String prefsContents = Files.readString(userFile().toPath(), StandardCharsets.UTF_8);
+        assertThat(backupContents, containsString("durableMarker=first"));
+        assertThat(prefsContents, containsString("durableMarker=second"));
+        assertThat(prefsContents, not(containsString("durableMarker=first")));
+      }
+    }
+
+    @Test
+    @Disabled("Test depends upon size comparison to determine backup validity")
```
gausie, line 145: Should this ship disabled?

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -903,141 +1066,227 @@ public void emptyIsRestoredFromBackup() throws IOException {
...
     }
 
-    @Test
-    void canSetLongWithFunction() {
-      var cleanups = withProperty("example", 10L);
-      try (cleanups) {
-        Preferences.setLong("example", v -> v / 5L);
-        assertThat("example", isSetTo(2L));
+    @Nested
+    class SetterFunctions {
```
gausie: Should this be nested? Probably not

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
```diff
@@ -903,141 +1066,227 @@ public void emptyIsRestoredFromBackup() throws IOException {
...
-        int pos;
-
-        while ((data = FileUtilities.readData(reader)) != null) {
-          if (data.length < 3) continue;
-
-          assertTrue(
-              Preferences.containsDefault(data[2]),
-              "Unknown preference '" + data[2] + "' in dailylimits.txt");
+    @Nested
+    class TestPropertiesExists {
```
gausie: This too re the nesting

gausie (commented):
Some small comments, looks good now. I think Irrat's thoughts re getFD().sync() probably deserve some proper thought also.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
jaadams5: Supposedly fixed unintended "subnesting".

jaadams5:
> Some small comments, looks good now. I think Irrat's thoughts re getFD().sync() probably deserve some proper thought also.

Fixed the nesting.  1300 line files on a 15" laptop screen are a PITA :-)

Since part of the reason this was such a mess was there was code that managed a backup if nulls were in the prefs file and different code that managed backups if there was a significant deviation in file size, I figured the best way ahead was to get this done and then have a general discussion of how to recognize a prefs or bak file was bad.  One being half the size of the other raised some concerns so a PR that (re)implements that code and allows the disabled test to pass is planned for the future.

I don't completely share Irrat's concerns and don't really have the instrumentation in place to make easy measurements.  I understood his comments to mean his concerns were not a great as initially believed so I figured this could go ahead.  Irrat has some work in progress https://github.com/kolmafia/kolmafia/pull/3567 and has graciously agreed to hold it until the current PR is committed or abandoned.

test/net/sourceforge/kolmafia/preferences/PreferencesTest.java
gausie: @jaadams5  what about this?

jaadams5: My suggestion is to commit with the test disabled and revisit after there is some agreement that a pref, bak pair where one file is twice as large as the other is sufficient reason to assume the larger file is correct and replace or delete as appropriate.

gausie:
I think switching from DataUtilities.getOutputStream (which mkdirs the parent) to `new FileOutputStream(tempFile)` means preference saves throw whenever settings/ doesn't exist yet.

My honest preference would be to strip this down to just the atomic write, and work on the backup file separately if we deem it necessary. Even in that case there are some problems though. Like it looks like the fallback only catches AtomicMoveNotSupportedException so ordinary IO errors will fail silently. Also the non-atomic path deletes the target before moving and the outer catch then deletes the temp.

Oh and GLOBAL_prefs.txt as a name breaks if someone runs multiple concurrent instances.

jaadams5:
The relevant code has been moved to #3623.  This will be closed in a day or two.

jaadams5:
Closing in favor of https://github.com/kolmafia/kolmafia/pull/3623

PR #3565, opened 2026-07-19
Better preference backups + global prefs backups

libraryaddict:
I'm not sure why global backups isn't a thing already, but that's included in this PR.

The reason preference files get filled with nulls is because the filesystem metadata was written to disk and the power was cut before the file was written.

Eg; https://serverfault.com/questions/1067720/what-could-cause-files-to-end-up-zeroed-out

I think `fileStream.getFD().sync()` would fix a majority of that for users, by reducing the time at risk?, but I'm not willing to do that as it comes with the tradeoff of constant writes to disk, instead of hitting the cache. It'd also slow down operations.

This PR aims to

1. If a file contains nulls, we always try to restore from backup
2. If the backup is missing, we try to load from only the good lines, a line is not good if the line has null bytes in it
3. Add global prefs to the backups

As a side effect, it'll stop clobbering the backups when the first file attempted is not empty but contains nulls.

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -2,7 +2,9 @@ package net.sourceforge.kolmafia.preferences;
 
 import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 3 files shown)


libraryaddict:
There was discussions of one sort or another about this in the forums, but I don't really care, they are distracting.

This PR mostly just handles nulls in files, which is what the vast majority of users will encounter after a power loss. This doesn't address all possible corruption, that's a different PR someone can make.

An indicator of corruption would be preferences missing that shouldn't be, heck, hash all the keys and if there's a mismatch the user is forced to confirm that they didn't modify it themselves. Another thought is probably can just check for non-escaped characters. But I'm not touching that.

midgleyc (approved):
Thanks!

I think this is a strict improvement over what we currently have. I am aware of the other PR #3564 that this conflicts with, but this one was easier to review.

PR #3567, opened 2026-07-22
Use journals for writing prefs to reduce IO and increase stability

libraryaddict:
Discussion found here; https://kolmafia.us/threads/reduce-the-amount-of-i-o-wear-and-tear-from-writing-preferences.32681/

This PR aims to add journaling to lower IO usage for prefs.
It has the remarkable downside of still no decent globals handling, but that's complicated.
It moves file handling into a seperate class.
A notable downside is the way it loads from Properties file.

### DIFF
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -1,19 +1,11 @@
 package net.sourceforge.kolmafia.preferences;
 
-import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.StandardCopyOption;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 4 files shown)


libraryaddict:
One thing this PR can improve on, is not flushing the prefs to disk when timing in, aka relogging. Only when the current user is genuinely being switched off / on.

Another thought is potentially allowing the max size of the journal to be controlled by a global pref, which wasn't implemented at the time due to the confusion of when such a pref would be initialized, and re-initialized.

Could also be worth a thought that we start noting when an especially large pref is set, something like 2kb in size? Shouldn't be hit, and can suppress the warning for the remainder, as well as a way to turn that warning off.

libraryaddict:
> One thing this PR can improve on, is not flushing the prefs to disk when timing in, aka relogging. Only when the current user is genuinely being switched off / on.

Existing LoginRequest comments appear to sound like this is intended, leaving it as is.

> Another thought is potentially allowing the max size of the journal to be controlled by a global pref, which wasn't implemented at the time due to the confusion of when such a pref would be initialized, and re-initialized.

Leaving it as is, 10mb is fine.

> Could also be worth a thought that we start noting when an especially large pref is set, something like 2kb in size? 

Not adding, this PR isn't about enforcing database rules. I do think this should be added, but that's unrelated.

gausie:
Same on this one. It might be good to keep WIPs on your own fork and once they're fully cooked open a PR here. Thank you for the work

libraryaddict:
It is ready, I was asked to keep it as a draft until the other pr has resolved

gausie:
Oh nice ok. Still, seems good to wait.

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

libraryaddict:
As a small note for any future reviewers, due to the PR currently being in a closed state, the changes Github shows are stale.

Atomic moves is something to consider, although that's currently being dealt with in PR #3623 and my comment there applies.

PR #3568, opened 2026-07-23
Add "transientPreferences" which controls if a preference is written to file or not

libraryaddict:
This PR builds on https://github.com/kolmafia/kolmafia/pull/3567
It will remain a draft until the above PR is resolved.

A transient preference is a preference that is not written to file, it is useful when you plan to mess around with the user's settings, but do not want your changes to persist. It is useful when you're storing data in their settings, but again do not want those settings to persist.

Examples is choice adventure scripts, or recovery settings, or so on.

The preference itself is always transient, and if it was loaded while modified, will when saved, be an empty string again.

Anyone clever enough to use a transient preference, is clever enough to set it up again. This helps prevent issues where this interferes beyond the lifescope of a script.

This is exposed as a preference, because a preference is easy to read, easy to access and can be modified by any script on the stack, up or downstream, and keep compatibility as the scripts responsibility.

A transient preference when reset, does not regain its old setting, it just means any further changes to the preference will now be saved. This perhaps, could be handled better, by storing in a separate map the transient preferences original values, and restoring them as needed.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -201,6 +201,7 @@ global	_g9Effect	0
 global	_gitUpdated	false
 global	_svnRepoFileFetched	false
 global	_svnUpdated	false
+global	_transientPreferences
 user	8BitBonusTurns	0	roa
 user	8BitScore	0	roa
 user	8BitColor		roa
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -179,6 +179,7 @@ global	swingLookAndFeel
 global	switchEquipmentForBuffs	false
 global	syncAfterSvnUpdate	false
 global	toolbarPosition	1
+global	transientPreferences
 global	useChatToolbar	true
 global	useContactsFrame	true
 global	useDevServer	false
@@ -201,7 +202,6 @@ global	_g9Effect	0
 global	_gitUpdated	false
 global	_svnRepoFileFetched	false
 global	_svnUpdated	false
-global	_transientPreferences
 user	8BitBonusTurns	0	roa
 user	8BitScore	0	roa
 user	8BitColor		roa
--- a/src/net/sourceforge/kolmafia/preferences/Preferences.java
+++ b/src/net/sourceforge/kolmafia/preferences/Preferences.java
@@ -280,11 +280,14 @@ public class Preferences {
   }
 
   /**
-   * A preference in "_transientPreferences" behaves normally in memory, but is never written to
+   * A preference in "transientPreferences" behaves normally in memory, but is never written to
    * disk.
    */
   public static boolean isTransient(final String name) {
-    String transientPrefs = Preferences.getString("_transientPreferences");
+    if (name.equals("transientPreferences")) {
+      return true;
+    }
+    String transientPrefs = Preferences.getString("transientPreferences");
     if (transientPrefs.isEmpty()) {
       return false;
     }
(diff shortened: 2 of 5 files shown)


VeeArrKoL:
I like the idea of some kind of temporary property-setting (as discussed in [this thread](https://kolmafia.us/threads/introduce-temporary-preferences-for-scripting.32530/)), but this I'm having trouble understanding when the behavior described in this PR would be useful. As a user, if a script is doing something temporary, I would expect it to either be reverted when the script is done or for it to persist indefinitely. If I'm understanding correctly, this fits into an awkward middle ground where the change stays active but only until you restart mafia. (For example, if I have a preference for a choice adventure set, and a script wants to "temporarily" set it, I wouldn't want to end up with the "wrong" preference value until mafia exits.) 

Maybe I'm misunderstanding the use case for these transient preferences?

libraryaddict:
You're correct, the behaviour is there, but scripts are expected to be cleaning up after themselves.

I would consider this persisting beyond the runtime to be a potential issue. 

The intent is to avoid situations where the user doesn't understand what's going on.

But I should have reset it when the user logs out perhaps.

But perhaps this could be done a different way, like resetting the preference on startup.

Or just having that preference in there which ispreventing itself from being saved. People can remove it and shoot themselves in the foot on their own terms.

VeeArrKoL:
Sorry, but I'm still not grasping it. Can you give an example of how a script would use this? What properties would it add to the transientProperties property, and how would it then use those properties?

libraryaddict:
> Sorry, but I'm still not grasping it. Can you give an example of how a script would use this? What properties would it add to the transientProperties property, and how would it then use those properties?

So in javascript / typescript

```typescript
const tempPrefs: string[] = getProperty("transientPreferences")
  .split(",")
  .filter(Boolean);

const added: boolean =
  tempPrefs.includes("betweenBattlesScript") === false;

try {
  if (added) {
    tempPrefs.add("betweenBattlesScript");
    setProperty("transientPreferences", trans.join(","));
  }

  setProperty(
    "betweenBattlesScript",
    "js abort('We didn't expect to fight!')"
  );

  runTheRestOfMyCode();
} finally {
  if (added) {
    const tempPrefs: string[] = getProperty("transientPreferences")
      .split(",")
      .filter(Boolean);

    const index = tempPrefs.indexOf("betweenBattlesScript");

    if (index >= 0) {
      tempPrefs.splice(index, 1);
      setProperty("transientPreferences", tempPrefs.join(","));
    }
  }
}
   ```

If you had `betweenbattlesScript` set to say `mySpecialScript.js`, you don't need to worry about a power cut locking in the 'temporary' script, because the temporary script was never saved to file.
It's basically a fancier version of setting a property, then the finally cleaning it up, where the fancier = it will be cleaned up even if the script is killed mid-operation.

Another thought is that we could stop writing the preference change to our session. Avoid hitting this https://github.com/s-k-z/kolfix/blob/main/src/main.ts#L103

VeeArrKoL:
Ah, I think I understand. So the idea is that this is an extra guardrail that a script can use *in addition to* cleaning up prefs it temporarily changes, not *instead of* manually cleaning them up itself. 

From a user point-of-view, I think I'd still rather have the latter than the former. (For example I think this solution still leaves stuff in an unexpected state if the script exits abnormally but without mafia exiting?) But maybe I'm letting perfect be the enemy of "good enough".

gausie:
This looks promising. I want to keep the number of PRs low so please reopen when you're ready to make it non-draft!

PR #3569, opened 2026-07-24
Add can_walk_from_choice() - Can the choice be walked away from

libraryaddict:
Exposes `ChoiceManager.canWalkAway()` as `can_walk_from_choice()`.
https://github.com/kolmafia/kolmafia/blob/057f8a9fc80623723ec580bfe0321bf19091dc3d/src/net/sourceforge/kolmafia/session/ChoiceManager.java#L66

The name of the function itself, I'm not sure if it could be better. "walkawayable" is not a real word.

Did consider allowing callers to provide a choice id as a parameter, however I'm struggling to think of a situation where the caller would care about such a choice that's not the current choice.

So, this just tells the caller if the current choice can be walked from or not.

Useful for checking if we're stuck in a choice or not without actually trying to leave, paired with `handling_choice()`.

Possible improvement of also checking if we're handling a choice before returning true. But would pair with `last_choice()` not being reset when we finish a choice.


gausie (approved):
This seems useful and simple. Thanks!

PR #3570, opened 2026-07-24
Add append_buffer_to_file - Ability to append a buffer on a file

libraryaddict:
This PR is partly due to [this thread](https://kolmafia.us/threads/reduce-the-amount-of-i-o-wear-and-tear-from-writing-preferences.32681/), however this function in itself still has value for when someone is writing stuff to file that doesn't require them to rewrite the entire file, they only want to append.

One example is when you're logging stuff, or tracking stuff (eg, inventory or sales). Rewriting the entire file feels a bit like a waste, especially if you're concerned about losing power mid-write, which could corrupt the entire file. You just wanted to add a new line, like a session or mirror log would.

I'm unsure if this function would be useful to others, I already use something similar.

Main thoughts on this are if the function could be named differently.

Eg, `append_to_file` or `buffer_append_to_file` or even `buffer_to_file(buffer, file, appendAsBoolean)`


gausie (commented):
Concerned that there is no test reading through the DataFileCache - it seems like that could be broken here. Could you add one?

gausie:
Thanks for the extra test but I am worried that the cache issue is still unaddressed. `printBytes` calls `updateCache` where on an append data is only the apended bytes, so the read cache holds partial content against the new mtime and the next `getBytes` return the last chunk. New tests all read with`Files.readString` so don't catch it. Am I totally misunderstanding?

libraryaddict:
Yeah, I overlooked the cache.

Had the test pass, which was confusing, but turns out that the cache was updating the un-sanitized file path with write and reading the sanitized file path with read. So that's fixed in the PR as well.

I think it would mean writing to `mydata.txt` would not update the cache for those reading `mydata.txt`, you'd have to write to `data/mydata.txt`. Or maybe I got the reproduction wrong, doesn't really matter as the bug is obvious enough and ultimately I fixed it just so my test would work properly.

gausie (approved):
Awesome, thanks for the contribution

PR #3571, opened 2026-07-25
Fix maximizer recommending mode swaps when no score is achievable

VeeArrKoL:
https://kolmafia.us/threads/enable-item-modes-on-modifier-maximizer.32672/post-179940

Currently, there's a narrow case where if a modeable is already equipped, and there is no equipment available to get the score of the corresponding slot above zero, the maximizer result will recommend swapping to a random mode (depending on which mode's score it happened to calculate first). Instead, it should prefer keeping the current mode.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -1774,10 +1774,13 @@ public class Evaluator {
                         return forcedModeables.get(modeable);
                       }
 
-                      MaximizerSpeculation best = null;
                       CheckedItem item =
                           new CheckedItem(modeable.getItemId(), equipScope, maxPrice, priceLevel);
                       var bestMode = modeable.getState();
+                      MaximizerSpeculation best = new MaximizerSpeculation();
+                      best.attachment = item;
+                      best.equipment.put(modeable.getSlot(), item);
+                      best.setModeable(modeable, bestMode);
 
                       // Check each mode in modeable to determine the best
                       for (String mode : modeable.getModes()) {
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
@@ -1875,6 +1875,20 @@ public class MaximizerTest {
     }
 
     @Test
+    public void dontSwitchModeForZeroScoreWithoutTiebreaker() {
+      final var cleanups =
+          new Cleanups(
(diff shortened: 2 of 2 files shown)


libraryaddict:
Generally when doing a bugfix, it's a good idea to put the failing tests into one commit, then after the tests fail, the changes that fix it.

Checking the code, it looks good, I just can't say the test would have failed to begin with.

VeeArrKoL:
> Generally when doing a bugfix, it's a good idea to put the failing tests into one commit, then after the tests fail, the changes that fix it.

FWIW I did write the change this way--the test did fail before actually making the fix (well, 80% of the time, anyway).

libraryaddict:
> > Generally when doing a bugfix, it's a good idea to put the failing tests into one commit, then after the tests fail, the changes that fix it.
> 
> FWIW I did write the change this way--the test did fail before actually making the fix (well, 80% of the time, anyway).

Oh yes, sorry, I think this looks great and I do believe that your test would've successfully failed.

It was intended to be advice, as it helps streamline the process of approving changes.

jaadams5:
To clarify, the test failed about 80% of the time before the fix and you attribute that failure to something both unidentified and non-deterministic, correct?

The mode work has focused on the parka.  Are there any existing tests that involve other items with modes?

Neither will be blocking for me.  I am just looking at "good" and striving for "excellent".

VeeArrKoL:
> To clarify, the test failed about 80% of the time before the fix and you attribute that failure to something both unidentified and non-deterministic, correct?

The sub-100% failure rate is due to the maximizer checking the scores for each mode in a non-deterministic order. (I did not hunt down why this is the case.) in the event that all scores are equal (e.g. zero), the "best" mode would be whichever it happened to calculate first. Hence the test would succeed in the 20% of cases where the already-equipped mode was calculated first. 

This fix should apply equally to all modeables, not just parka. I think many of the test cases for the previous set of modeables changes use the backup camera instead.

jaadams5:
> > To clarify, the test failed about 80% of the time before the fix and you attribute that failure to something both unidentified and non-deterministic, correct?
> 
> The sub-100% failure rate is due to the maximizer checking the scores for each mode in a non-deterministic order. (I did not hunt down why this is the case.) in the event that all scores are equal (e.g. zero), the "best" mode would be whichever it happened to calculate first. Hence the test would succeed in the 20% of cases where the already-equipped mode was calculated first.
> 
> This fix should apply equally to all modeables, not just parka. I think many of the test cases for the previous set of modeables changes use the backup camera instead.

Thank you.

jaadams5 (approved):

PR #3574, opened 2026-07-28
Additional Portable Laughing Stock support

VeeArrKoL:
Track fights since last fruit drop, track number of fruit dropped, add crafting recipes, add Taunt Your Foe skill


gausie (approved):
Great PR thank you

PR #3576, opened 2026-07-28
TTT main map detection

VeeArrKoL:
The TTT used to be inside of Seaside Town, so mafia would check for it being present whenever the town was visited and set `timeTowerAvailable` as appropriate. However, with this year's map revamp, the TTT is now located on the main map. As a result, mafia doesn't detect the tower automatically (instead detecting when the user actually clicks into the zone); and worse, if the user visits the town, mafia doesn't see the tower there as expected and sets `timeTowerAvailable` to `false`. This PR fixes this issue by checking for the tower on the main map.

It also ~~includes support for a recently-found TTT 2026 item and~~ fixes a missing shoprow.


jaadams5 (dismissed):

src/net/sourceforge/kolmafia/session/QuestManager.java
```diff
@@ -279,6 +279,10 @@ public static final void handleQuestChange(GenericRequest request) {
         BURTRequest.parseResponse(responseText);
       }
     } else if (location.startsWith("main")) {
+      if (location.equals("main.php")) {
```
gausie: I think the previous line is meant to cover this, remove the if?

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -448,11 +449,13 @@ public void checkMapAccessForNoTwitch() {
...
           // the map, we have access today.
-          client.addResponse(200, html("request/test_visit_town_no_twitch.html"));
+          client.addResponse(200, "temporal ether");
+          // TODO replace with actual example of what it looks like when TTT is unavailable
+          // client.addResponse(200, html("request/test_visit_twitch_no_twitch.html"));
           boolean success = BOHEMIAN_PARTY.preValidateAdventure();
           var requests = client.getRequests();
           assertThat(requests, hasSize(1));
-          assertPostRequest(requests.get(0), "/place.php", "whichplace=town");
+          assertPostRequest(requests.get(0), "/place.php", "whichplace=twitch");
```
gausie: Is this right? Isn't the equivalent just visiting main.php, not whichplace=twitch?

src/net/sourceforge/kolmafia/session/QuestManager.java
VeeArrKoL: No, there are cases where the location starts with `main.php` but the main map is not displayed. For example, `main.php?talktosocp=1` and we don't want to check for the presence of the tower in those cases. 

I wasn't immediately sure what was going on with the island thing so I didn't want to just change the condition, but I think that it would probably be fine to do so.

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
VeeArrKoL: `preValidateAdventure()` is only set up to work with a place, so I changed it to use place=twitch (which should also effectively test whether the tower is available) instead of place=town. (See the changes in KoLAdventure.java)

gausie: Hmmm. It seems like we should fix that.

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -448,11 +449,13 @@ public void checkMapAccessForNoTwitch() {
         try (cleanups) {
           // If we have not verified access, but the Time Twitching Tower is on
           // the map, we have access today.
-          client.addResponse(200, html("request/test_visit_town_no_twitch.html"));
+          client.addResponse(200, "temporal ether");
+          // TODO replace with actual example of what it looks like when TTT is unavailable
+          // client.addResponse(200, html("request/test_visit_twitch_no_twitch.html"));
```
gausie: Is this PR meant to be a draft? we have a TODO here

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
```diff
@@ -430,11 +431,11 @@ public void checkMapAccessForTwitch() {
         try (cleanups) {
           // If we have not verified access, but the Time Twitching Tower is on
           // the map, we have access today.
-          client.addResponse(200, html("request/test_visit_town_twitch.html"));
+          client.addResponse(200, html("request/test_visit_twitch_twitch.html"));
           boolean success = BOHEMIAN_PARTY.preValidateAdventure();
           var requests = client.getRequests();
-          assertThat(requests, hasSize(1));
-          assertPostRequest(requests.get(0), "/place.php", "whichplace=town");
+          assertThat(requests.size(), greaterThanOrEqualTo(1));
```
gausie: Why loosen this from hasSize(1)?

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
gausie: I'll push a commit that does this. Also removing some claude slop comments.

gausie (approved):

test/net/sourceforge/kolmafia/KoLAdventureValidationTest.java
VeeArrKoL: It looks like you already handled this another way, but the TODO was intentional--I didn't have a way to capture the data for the test case (until the event is over) but didn't want that to block a PR for something that was actively causing problems for users during the event.

PR #3577, opened 2026-07-29
feat: hamlet monster stats, parts

midgleyc:
Probably something more complicated is happening with the knights' scaling (i.e. I guess it changes with their winrate).

Flimsy armor isn't heistable; beyond that I don't know what's going on.


jaadams5 (approved):

gausie (approved):

gausie:
According to excavator we know that the traveling space soldier has a torso.

PR #3582, opened 2026-07-31
Fix to_url(location) returning a stale string

libraryaddict:
```
 > set pyramidBombUsed=true

 > js toUrl(Location.get("The Lower Chambers"))

Returned: place.php?whichplace=pyramid&action=pyramid_state1
```

It should be
```
Returned: place.php?whichplace=pyramid&action=pyramid_state1a
```

The first test is a failing test, the second is the fix.

In the fix, we create a new AdventureRequest as the url is changed in two places, and the place that's not in the constructor, will mutate preferences if invoked. So we need to construct a new AdventureRequest to get a non-stale url.


libraryaddict:
Although maybe I need to change updateFields() to have some overload and a boolean parameter "mutate fields" or something.

I don't know.
The current fix isn't correct either, cellar for example.

gausie:
Well spotted. I added another commit, I think this could sort it.

gausie:
@libraryaddict let me know if you like the changes I made and we can merge

libraryaddict:
Ended up removing the `updateFields()` overload, it doesn't `@Override` anything and the default behavior is something I can see being in the air everytime someone might need to call it.

I'm mostly of the mind that modifying preferences really doesn't belong in that method, but that's a far larger change and mental load. Especially as it should belong in a successful response. But, that's more effort.

So I'm happy with the PR as it is now.

jaadams5 (approved):

PR #3588, opened 2026-08-03
Cup of 13s tracking

nasurte:
Track Cup of 13s charges and booze drops that day. A drop occurs if the cup has 10 charges. The first drop of the day happens on your 6th combat win with the cup, subsequent drops take 10 wins each, quality of the received booze decreasing throughout the day.


src/data/defaults.txt
```diff
@@ -2138,6 +2138,8 @@ user	_crimboTraining	false
 user	_crimboTree	false
 user	_crToday	false
 user	_cupidBowFamiliars
+user	_cupOf13sCharges	5
```
midgleyc, line 2141: if the first drop occur on the 6th combat, shouldn't this be 4?

midgleyc (changes requested):
A test would be appreciated. Also see comment.

src/data/defaults.txt
nasurte: You get the drop and reset _cupOf13sCharges to 0 before it increments at the end of combat here, so I raised the default by 1 so that the behaviour of the first drop of the day would be consistent with the rest of them (you win the combat with _cupOf13sCharges already at 10, you get the drop and charges increment to 1. If it started at 4, the first drop would happen with _cupOf13sCharges at 9 at the start of combat)

midgleyc (approved):
Thanks!

PR #3590, opened 2026-08-04
build: remove redundant zipTree unpacking from tasks.jar

Aellinsar:
Build seemed slow during the jar packing step. `tasks.jar` had a `from({ configurations.runtimeClasspath... zipTree(it) })` block unpacking dependency jars, which duplicates what `shadowJar` already does.

Ran 25 builds to check timing:
Warm shadowJar builds went from 4.5s avg down to 1.5s avg.


midgleyc:
I don't think this is right: `jar` builds a fat jar, and this changes it to build a thin jar, which wouldn't be runnable. I don't think shadowJar calls jar in the first place.

A cold shadowJar call takes 21s, but after that it reliably takes only 2s for me.

Aellinsar:
> I don't think this is right: `jar` builds a fat jar, and this changes it to build a thin jar, which wouldn't be runnable. I don't think shadowJar calls jar in the first place.
> 
> A cold shadowJar call takes 21s, but after that it reliably takes only 2s for me.

- The output jar runs completely fine. I've been using it for my daily adventuring without any issues. Feel free to build both and diff the actual .jar contents if you want to double check, but nothing is missing. I've been working on other fixes in my local branch (particularly the shutdown inconsistencies when the main Swing window closes) and this was simply an annoyance that I fixed along the way.
- shadowJar already handles bundling runtimeClasspath into the fat jar. Having zipTree(it) in tasks.jar was just redundant work that forced extra unpacking steps before shadowJar ran.
- Timing varies wildly by system, so citing a pre-change speed on your machine without a before/after baseline on that same machine doesn't say much. The 25-build benchmark was an A/B test on the exact same laptop, going from 4.5s down to 1.5s.

jaadams5 (approved):

jaadams5:
I used this in a build and the jar file with and without has the exact same size.  The files are identical except for Build-Branch,  Build-Build and Build-Dirty in the  manifest.  

I am running with the modified jar right now but don't expect any problems.  Anecdotally the build did not seem any faster but I also have more pressing concerns than to try and make a comparable measurement.

Aellinsar:
> I used this in a build and the jar file with and without has the exact same size. The files are identical except for Build-Branch, Build-Build and Build-Dirty in the manifest.
> 
> I am running with the modified jar right now but don't expect any problems. Anecdotally the build did not seem any faster but I also have more pressing concerns than to try and make a comparable measurement.

Perfect! That was my goal, same output, faster build. I have a habit of mashing "build" after every small change and was noticing the filesystem I/O & RAM spikes due to that 2nd copy step. Was actually causing GCs during builds sometimes, with warm builds spiking to 15-20s when it occurred!

midgleyc:
> The output jar runs completely fine. I've been using it for my daily adventuring without any issues. Feel free to build both and diff the actual .jar contents if you want to double check, but nothing is missing. I've been working on other fixes in my local branch (particularly the shutdown inconsistencies when the main Swing window closes) and this was simply an annoyance that I fixed along the way.

I have done: the jar does not run.
```
Exception in thread "main" java.lang.NoClassDefFoundError: com/alibaba/fastjson2/JSONException
        at net.sourceforge.kolmafia.persistence.FlaggedItems.initializeLists(FlaggedItems.java:274)
        at net.sourceforge.kolmafia.KoLmafia.main(KoLmafia.java:326)
```

To confirm, when you run "./gradlew :jar" with your changes, you get a working jar? I do not.

> shadowJar already handles bundling runtimeClasspath into the fat jar. Having zipTree(it) in tasks.jar was just redundant work that forced extra unpacking steps before shadowJar ran.

shadowJar does not run the jar task as far as I am aware. Only jar runs the jar task.

Also, the 2s timing is on both your branch and the main branch. Exactly what commands are you running for your benchmark? I am just running `./gradlew :shadowJar`.

jaadams5:
Oh.  I have never run "gradlew jar".  I always and only use "gradlew shadowjar".  My recollection for the reasons to do so was that I was told that shadowjar always produced a jar with all of the necessary dependencies but that was not the case  for jar.  

I think the discussion at https://kolmafia.us/threads/r27801-fix-gradle-warnings-by-midgleyc-in-t.29542/#post-174151 might be of interest but maybe we need to revisit the difference between jar and shadowjar and if/why we need both.

midgleyc:
Yes, I also only use `:shadowJar`. But this PR changes the code for `tasks.jar`, not for shadowJar. So it's expected that shadowJar would continue to work as before, as the code relating to it hasn't changed. It does read the manifest from `jar`, but that hasn't been changed, only some of the logic later in the task.

Both jar and shadowJar produce a fat jar with all the dependencies. There are minor differences in the ordering of files and the MANIFEST.MF, but I don't think anything that should have any material effect.

As to why we have two tasks that produce fat jars, I don't know.

Aellinsar:
> Yes, I also only use `:shadowJar`. But this PR changes the code for `tasks.jar`, not for shadowJar. So it's expected that shadowJar would continue to work as before, as the code relating to it hasn't changed. It does read the manifest from `jar`, but that hasn't been changed, only some of the logic later in the task.
> 
> Both jar and shadowJar produce a fat jar with all the dependencies. There are minor differences in the ordering of files and the MANIFEST.MF, but I don't think anything that should have any material effect.
> 
> As to why we have two tasks that produce fat jars, I don't know.

Ah, my mistake! I assumed :shadowJar called :jar underneath, but running :shadowJar bypasses :jar entirely.

Having both tasks setup to build fat JARs (and not having a standard thin JAR) definitely confused me. Since removing zipTree breaks ./gradlew jar without offering any benefit for :shadowJar, I'll close out the PR. Thanks for explaining!

Rinn:
i don't remember why we have both either

PR #3592, opened 2026-08-04
Fix Mer-Kin Colosseum corner case

VeeArrKoL:
Weird miniature crystal ball shenanigans can result in mafia being out-of-sync with regard to how many combats have been completed in the Mer-Kin Colosseum, resulting in being done with the zone and ready to move onto the boss while mafia still thinks you have fights to do in the colosseum. There is existing code to handle a case like this in `SeaMerkinRequest`, but it gets skipped if the Mer-Kin quest is already marked as done. Due to 11,037 Leagues, you now can have already reached a "done" state when you go to do the Colosseum (if you did the scholar path first), so this PR removes the check for the quest being done and parses the colosseum result regardless.


gausie:
Can you add some tests for this? It would help mapping out the different states and expectations

VeeArrKoL:
Tests have been added for the Colosseum parsing. Is there anything else needed?

jaadams5 (approved):
Looks OK to me but I think @gausie should be the one to commit.

PR #3593, opened 2026-08-04
Update TCRS data files

gausie:
Re-derived all 54 class×sign TCRS data files with a current KoLmafia build.

- Each non-cafe file is now complete at 12,064 items
- Modifiers on existing items are enriched - mostly `Familiar Effect` modifiers the older derivation didn't carry over.
- Sizes and qualities are unchanged, so this is the same TCRS randomization, just more complete/current. Cafe files unchanged in size (33 booze / 45 food).

Derived with the ConcurrentSkipListMap fix (#3591) applied, so no partial/corrupt files.


gausie:
LLM really wrote some crazy stuff in the PR here, my bad

midgleyc:
I attempted to make LLM agents identify themselves with https://github.com/kolmafia/kolmafia/blob/main/AGENTS.md#pull-requests.

Did that not work because you're using Claude, and Claude doesn't read AGENTS.md?

midgleyc:
Also the cafe files have changed the names by reordering the modifiers; has the underlying code in KoL changed? I don't see how that could happen otherwise.

For this PR you ran "tcrs derive" on each character / sign combination, then used an LLM to make the PR?

midgleyc (approved):

gausie:
This is all done via tcrs derive yeah. It probably didn't read AGENTS.md because I was using it from the directory where I was creating a rig for concurrently running tcrs derive per multi, and as a final step told it to copy the output to kolmafia and make a PR. 

I can imagine how the order has changed, since CDM switched all of the RNG tooling over to something that will survive a PHP version bump. I noticed it and agree it is weird though. I'll manually log in with a TCRS account and verify visually

gausie:
Yep confirmed they have switched around on live. This will mean some tinkering to the tcrs guessing branch I am sure, but this merge is correct.

Veracity0:
I am glad you came back to this.

PR #3595, opened 2026-08-05
remove floristFriarChecked

Rinn:
Now that we can rely on ownsFloristFriar being set from api.php coolitems, remove floristFriarChecked and don't check if floristFriar is available unless it is owned. This should prevent cases where checked was set incorrectly and florist friar was seen as unavailable when it should have been.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -736,7 +736,6 @@ user	flickeringPixel6	false
 user	flickeringPixel7	false
 user	flickeringPixel8	false
 user	floristFriarAvailable	false	roa
-user	floristFriarChecked	false	roa
 user	flyeredML	0
 user	forbiddenStores
 user	fossilB	0
--- a/src/net/sourceforge/kolmafia/request/FloristRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FloristRequest.java
@@ -145,8 +145,10 @@ public class FloristRequest extends GenericRequest {
   }
 
   public static void reset() {
-    Preferences.setBoolean("floristFriarChecked", false);
     FloristRequest.floristPlants.clear();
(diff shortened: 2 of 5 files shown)


midgleyc (approved):
Thanks!

libraryaddict:
Just wanted to confirm that the change here is intended for a http request to be made whever the player owns the friar, but does not have it available? When calling `florist_available()`

Rinn:
that's what it was doing before i didn't change the request being made i just changed the condition to be the preference set from coolitems rather than the manually set preference

libraryaddict:
Sorry I edited my comment.

Before it was
```java
  public static Value florist_available(ScriptRuntime controller) {
    // Pref is reset when we refresh state
    if (!Preferences.getBoolean("floristFriarChecked")) {
      // We have not checked this session, we must perform a http request
      FloristRequest.checkFloristAvailable();
    }
    return DataTypes.makeBooleanValue(FloristRequest.haveFlorist());
  }
  ```
And now it's
  ```java
    public static Value florist_available(ScriptRuntime controller) {
    
    if (Preferences.getBoolean("ownsFloristFriar") && Preferences.getBoolean("floristFriarAvailable")) { 
      return;
    }

    if (!Preferences.getBoolean("ownsFloristFriar")) {
      return;
    }

   // We own the friar, but the friar is not available, we must perform a http request
    FloristRequest.checkFloristAvailable();
    return DataTypes.makeBooleanValue(FloristRequest.haveFlorist());
  }
  ```
  
  The check was preventing the request from being made every time.

libraryaddict:
The reason I'm asking, is because it's entirely reasonable for the expected usage to be scripts to be checking the property instead of calling the function, to check if they can use florist.

PR #3596, opened 2026-08-05
Fix append_buffer_to_file, buffer is not converted to string

libraryaddict:
So this issue was caused by https://github.com/kolmafia/kolmafia/pull/3570

The original PR had tested against `bufferToFile("buffer", "file", true)` - an extra parameter. That was confusing, and was silently dropped, but wasn't re-tested against the newer function name until a hour ago when converting my scripts over to use `appendFileToBuffer`.

It turns out there's a hardcoded list of function names that need their buffer parameter converted, and my previous integration accidentally fell into that hardcoded list. When not using an existing function name, there is no conversion.

The first test is a failing test.


midgleyc (approved):
Thanks!

PR #3597, opened 2026-08-06
Remove Curse of Vacation daily limit

VeeArrKoL:
It turns out the wiki lied and Curse of Vacation doesn't actually have a daily limit.


midgleyc (approved):
Thanks!

PR #3598, opened 2026-08-06
TCRS adjective order had accidentally become unseeded

gausie:
I mentioned to CDM and he fixed it. Confirmed by rederiving, and rederiving again - stable. I didn't notice before because you need to fully relog to get the reroll.


midgleyc (approved):
sweet

PR #3599, opened 2026-08-07
chore: update actions to run on node 24

midgleyc:
Practically speaking, this should just remove the warning GitHub adds about actions being forced to run on Node 24 despite only supporting Node 20. Looking through the release notes, there don't seem to be any other breaking changes.

`checkout` has changed defaults for insecure modes we don't use, and apparently that change was backported to 6 anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         os: [macos-latest, ubuntu-latest, windows-latest]
     steps:
       - name: Checkout repository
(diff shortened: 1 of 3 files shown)


jaadams5 (approved):
Of course I have no recollection of seeing the warnings...

midgleyc:
If you go to a specific action run and scroll down you can see them.

e.g. https://github.com/kolmafia/kolmafia/actions/runs/31140299660

PR #3600, opened 2026-08-07
feat: add display name for legendaries and Crimbo '23 tiny plastics

### LINKED ISSUE #3482: Item names for legendary seal-clubbing club and legendary pasta wand
Because the legendary seal-clubbing club counts as multiple S-words for the Sword of S Words, it seems likely that its name internally includes all of the formatting for its special rendering, similar to `elven <i>limbos</i> gingerbread`.

However, having the item name be `<span style='padding: 2px; color: #ff8000; background-color: black;'>legendary seal-clubbing club</span>` within kolmafia would probably be unwieldy. Perhaps it would be good if kolmafia was able to have both a "human-readable name" and "true name" for items.

(All of the above applies to the legendary pasta wand as well.)

midgleyc:
Adjust the S Word detection to agree with observed behaviour for the legendary weapons.

Closes #3482.

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -1663,9 +1663,9 @@ Item	leechknife	Maximum HP: +10, Maximum MP: +5, HP Regen Min: 4, HP Regen Max: 
 # legendary pasta wand: Summon Legendary Noodles
 # legendary pasta wand: Cook up to 5 items per day without using an Adventure
 # legendary pasta wand: Pasta Thralls Go Up to 11 (Pastamancer Only)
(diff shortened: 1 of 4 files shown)


VeeArrKoL:
Resolves #3482

midgleyc:
On that, btw: something weird is happening because legendary pasta wand has a score of 2, despite the full name having two spans and a style. What I've assumed is that it splits at "<" in addition to whitespace, but it's possible something else is going on.

VeeArrKoL:
I suspect the way that it works is that all of the special characters get stripped out, so pasta wand has 2 words starting with S (`span` and `stylepadding`, but not `wandspan`). Similarly, LSCC has 3 words starting with S (`span`, `stylepadding` and `sealclubbing`, but not `clubspan`).

midgleyc:
Yeah that seems more likely. Updated.

jaadams5 (approved):

PR #3601, opened 2026-08-08
Add daily limits for Heavy Rains 1/day skills

VeeArrKoL:
Adds daily limits for Lightning Rod, Rain Coat, and Thunder Down Underwear. I verified that these can actually only be cast once per day despite the skill description not explicitly saying so (rather than, say, being castable multiple times but only doing something the first time).

ref: https://kolmafia.us/threads/all-skills-with-daily-limits-should-appear-in-dailylimits-txt.32632/post-179914


midgleyc (approved):
Thanks!

PR #3602, opened 2026-08-09
Fix Eternity Codpiece gem inventory tracking

VeeArrKoL:
The current implementation of inventory tracking when gems are inserted into/removed from the Eternity Codpiece has some issues where gems can be double-counted when being added or removed under certain conditions. Additionally, this behavior differed depending on whether the change was initiated from the gCLI (e.g. `equip codpiece1 unblemished pearl`) or via the relay browser.

This PR aims to address these issues and correctly track gems as they move into and out of the codpiece.

ref: https://kolmafia.us/threads/gems-equipped-to-the-eternity-codpiece-are-deducted-from-the-inventory-twice.32546/


src/net/sourceforge/kolmafia/request/EquipmentRequest.java
```diff
@@ -942,7 +942,8 @@ public void processResults() {
     }
 
     if (urlString.startsWith("choice.php") && urlString.contains("whichchoice=1588")) {
-      parseCodpiece(responseText);
+      // Handling the codpiece inventory changes by calling parseCodpiece() was already handled in
+      // ChoiceControl.
```
jaadams5, line 946: The language in the comment confuses me.  Perhaps "codpiece inventory changes have already been handled in ChoiceControl"?  If I understand EquipmentRequest correctly there is no requirement to check for the choice and then do nothing.  Perhaps deleting 944 through 948 would be a better solution?

jaadams5 (approved):

jaadams5:
Since the lines were added a few months ago I would actually prefer that they be deleted.

As a curiosity do the new tests fail when run against old code and visa versa?

src/net/sourceforge/kolmafia/request/EquipmentRequest.java
VeeArrKoL: We want to avoid falling through to the default handling which prints a bunch of red text to the gCLI, so we have to detect this case and specifically do nothing. This is an artifact of the codpiece slots being treated as pseudo-equipment slots--they don't really have much behavior in common with how other equipment is managed on the backend.

VeeArrKoL:
> As a curiosity do the new tests fail when run against old code and visa versa?

I forget exactly which did, but some did (and some of the old tests succeeded only by coincidence--two errors cancelling out to get the right answer).

src/net/sourceforge/kolmafia/request/EquipmentRequest.java
jaadams5: Got it.  Thanks.  I missed it because the "fall through" wasn't obvious with the viewers I was using.  There is an opportunity for a comment but that can be your judgement call :-)

PR #3605, opened 2026-08-10
Add git url remapping

Rinn:
Intent is to be able to use a fork of a git script and have any urls in dependencies.txt be remapped automatically.

The issue I'm trying to solve is that multiple git repos will be cloned and there is no way to control which one is synced to the scripts directory.

Format of the `gitUrlRemaps` property is `a|b:c|d:...`, the first matched is used. I also print out that a url has been remapped before cloning.

### DIFF
--- a/src/data/defaults.txt
+++ b/src/data/defaults.txt
@@ -60,6 +60,7 @@ global	getBreakfast
 global	gitInstallDependencies	true
 global	gitShowCommitMessages	false
 global	gitUpdateOnLogin	false
+global	gitUrlRemaps
 global	greenScreenProtection	false
 global	guiUsesOneWindow	false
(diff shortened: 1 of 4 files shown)


jaadams5:
I'm a little confused, so please be patient :-)

If, for example, I want to use a fork of excavator, I can do that by setting a preference, updating one of the scripts that depends upon excavator and then every script that uses excavator will use the fork.  That is because we don't support versions in scripts and so once the fork has been copied to scripts, every other script is forced to us it.

This seems to be able to create a potential problem if script A depends upon excavator and script B depends upon excavator-fork.  If script A relies on a feature or interface of excavator that is different or not present in excavator-fork then this breaks script A.  Correct?

If I were developing excavator then this problem might be a feature assuming an intent to "publish" the fork.

I can see this being useful in a niche occupied by a script writer building components that other scripts depend on.  I'm not sure it has a role for a basic script user.

I'm picking on excavator because an upgrade once broke things I was doing locally and this feature would have been useful until something else that used excavator needed the newer version :-)

Can you confirm that the tests will run when GitHub is unavailable?

Also please confirm that this will work if a Source Forge/SVN hosted script uses a GitHub hosted dependency and  it is not expected to work is a script is accessed via git but hosted elsewhere.

Thank you.

Rinn:
> I'm a little confused, so please be patient :-)
> 
> If, for example, I want to use a fork of excavator, I can do that by setting a preference, updating one of the scripts that depends upon excavator and then every script that uses excavator will use the fork. That is because we don't support versions in scripts and so once the fork has been copied to scripts, every other script is forced to us it.

Yes

> This seems to be able to create a potential problem if script A depends upon excavator and script B depends upon excavator-fork. If script A relies on a feature or interface of excavator that is different or not present in excavator-fork then this breaks script A. Correct?

This is a problem *now* because you can checkout multiple repos that would sync to the same file path, so whichever syncs last wins.

> If I were developing excavator then this problem might be a feature assuming an intent to "publish" the fork.

Sorry I don't understand the point you're making here.

> I can see this being useful in a niche occupied by a script writer building components that other scripts depend on. I'm not sure it has a role for a basic script user.

A basic script user doesn't have to use this property.

> I'm picking on excavator because an upgrade once broke things I was doing locally and this feature would have been useful until something else that used excavator needed the newer version :-)
> 
> Can you confirm that the tests will run when GitHub is unavailable?

None of theses tests request network resources they're just checking if the string remapping mutates the value.

> Also please confirm that this will work if a Source Forge/SVN hosted script uses a GitHub hosted dependency and it is not expected to work is a script is accessed via git but hosted elsewhere.

Entries in dependencies.txt are resolved the same regardless of what kind of repo is being processed so that doesn't make any difference, any git repo will go though GitManager to resolve the url to be able to clone it.

I'm not sure I follow, this is just a string replacement it doesn't matter what the source or destination url is.

Rinn:
The real issue here is I suppose that a user can have two script repos cloned that conflict with each other, and mafia doesn't have any way to resolve these conflicts and accept one over the other when it does a sync. We also can't control what scripts other repos set as dependencies which is generally how we get these conflicts to exist in the first place.

I'm not really sure how to best handle that if this isn't acceptable, we'd need to add code to detect that files will be overwritten that were already synced, and provide a way for a user to specify which repo is preferred.

jaadams5:
Sorry I wasn't as clear as I could have been but my concerns are addressed even if not clearly stated.

It is opt in so anyone using it presumably knows what they are doing.  It does not address, nor is intended to address, the case where a dependency called out by one script overwrites a different "version" installed by another.

At some point we need to acknowledge that we know how to manage build and runtime dependencies for apps but applying that knowledge to KoLmafia scripting would require a lot of fundamental changes to mafia that are probably not worth the effort.

jaadams5 (approved):

PR #3609, opened 2026-08-11
feat: use [L] for shields with level-based DR

midgleyc:
Correct OPS not setting this correctly.

Follow-up to #3606.


gausie (approved):
Sweet. Maybe I need to update some TCRS stuff now

Issue #3610, opened 2026-08-11
issue with chat pms


chocofungi:
i was using version ...144 with no problems then today it started not showing me pms, though they would display with /last. i upgraded to ...151 and had the same issue. i tried it in vanilla with no trouble

gausie:
I think this could be fixed - try now!

chocofungi:
it is fixed! thank you! it even fixed the "not giving notification of getting a kmail" thing which i had failed to mention, so definitely a job well done

PR #3611, opened 2026-08-11
Handle prefixed chat message ids for events and private messages

gausie:
As of some point recently KoL sends prefixed string mids for events (ev<timestamp>_0) and private messages (pm<timestamp>_<senderId>_<n>) instead of plain numbers, causing a NumberFormatException in ChatPoller.parseNewChat that broke chat polling. It seems public messages still use a plain numeric mid that matches the response's last field. This adds parseMid, which returns the numeric value for public mids (still compared against serverLastSeen for deduplication) and 0 for prefixed or missing mids. The timestamp in a prefixed mid is not used because last is a message counter rather than a timestamp. Events and private messages are no longer deduplicated by mid, but the poll's lasttime already filters new messages server-side so a duplicate is only possible on an overlapping re-poll.


VeeArrKoL:
Resolves #3610, I think?

midgleyc (approved):
Thanks!

Haven't seen this myself but I trust you.

PR #3612, opened 2026-08-11
Adds a new Type column to the Item Manager.

j13waldman:
## Summary

Adds a new **Type** column to the Item Manager.

This change adds a new Type column to the **Usable**, **General**, and **Creatable** item tables, showing normalized player-facing categories. These categories differ from the raw `ConsumptionType` values used internally by KoLmafia.

Examples include:

- **11-leaf clover** → `usable`
- **4:20 bomb** → `combat item`
- **agua de vida** → `spleen item`

## Testing

- Verified the Type column appears correctly.
- Verified sorting by Type works correctly.
- Verified existing columns continue to sort correctly.


gausie:
image

j13waldman:
rofl, this is my first PR so Ryan has a good point.

The why might just might be my own personal needs. I probably spend more time playing with my inventory than playing KoL. I find it hard to organize things from the relay browser, like moving all of my avatar-changing potions to the display case, or sharing spleen items with my clan, especially when there are a lot of them. Having  the Item Manager show the item type helps me figure out the item type easily so I can sort by type and easily see all of my avatar-changing potions and chuck them into my display case.

gausie:
I simplified the code and got rid of a bit that I think was unnecessary duplication. Will merge once tests pass.

gausie:
Oh I didn't even spot that the price formatting was unrelated. Well done. And opposite of well done to your LLM. Can you fix the tests?

j13waldman:
Sorry for the vibe coding. I'm getting a crash course in git and java while I'm doing this.

For these changes, I added a helper to ensure that items display the correct type as a player would see it in the browser or wiki. `EquipmentDatabase.getItemType()` already returns a type for many items, so the helper normalizes those values into player-facing categories where necessary. For items without a type in `EquipmentDatabase`, the helper falls back to `ItemDatabase` and uses the item's consumption type and attributes to determine and display a normalized type.

jaadams5 (dismissed):
Looks OK to me but I'd prefer that @gausie has the final say before commit.

gausie:
I removed the `getDisplayItemType` and you readded it. I don't believe that function is particularly useful for players and it's annoying to maintain so I'd rather you removed it.

j13waldman:
Heard, chef. I see where you're coming from but I don't think it's a full feature without the helper method. I appreciate the feedback and the time you all took to review the PR, but I'll go ahead and close it.

PR #3614, opened 2026-08-12
Print when a git repo has file conflicts

Rinn:
This adds a conflict print after cloning or syncing scripts, but otherwise does not affect what files are copied. My intention is to do a follow up PR which uses a property where you can set a preferred repo when a conflict is found which prevents the other conflicting repos from syncing

Example:
```
> git checkout Ezandora/gain

Starting
remote: Enumerating objects
remote: Counting objects
remote: Compressing objects
Receiving objects
Resolving deltas
Checking out files
Copying: scripts/gain.ash
Cloned project Ezandora-gain

> git checkout loathers/gain

Starting
remote: Enumerating objects
remote: Counting objects
remote: Compressing objects
Receiving objects
Resolving deltas
Checking out files
Conflict: scripts/gain.ash (Ezandora-gain, loathers-gain)
Copying: scripts/gain.ash
Cloned project loathers-gain

> git sync

Conflict: scripts/gain.ash (Ezandora-gain, loathers-gain)
Copying: scripts/gain.ash
Conflict: scripts/gain.ash (Ezandora-gain, loathers-gain)
Copying: scripts/gain.ash

> git delete Ezandora-gain

Removing project Ezandora-gain
scripts/gain.ash => DELETED
Project Ezandora-gain removed.

> git sync

Copying: scripts/gain.ash

> git checkout Ezandora/gain

Starting
remote: Enumerating objects
remote: Counting objects
remote: Compressing objects
Receiving objects
Resolving deltas
Checking out files
Conflict: scripts/gain.ash (Ezandora-gain, loathers-gain)
Copying: scripts/gain.ash
Cloned project Ezandora-gain
```

A few things of note:

https://github.com/kolmafia/kolmafia/blob/6d69f4283b69013f144f93c8a7e5c1b7e248f00e/src/net/sourceforge/kolmafia/scripts/git/GitManager.java#L159-L162
I believe there is a bug here where if the root changes, but also files have been moved/deleted when the repo's root changed it will leave stale files that were previously synced. I think this is when a root is defined in the manifest and that changes?

https://github.com/kolmafia/kolmafia/blob/6d69f4283b69013f144f93c8a7e5c1b7e248f00e/src/net/sourceforge/kolmafia/scripts/git/GitManager.java#L196-L208
I didn't add conflict checking here, maybe the both of these could be simplified by simply deleting all the files that were deleted or moved, then call sync() afterwards once?

### DIFF
--- a/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
+++ b/src/net/sourceforge/kolmafia/scripts/git/GitManager.java
@@ -16,8 +16,12 @@ import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import net.sourceforge.kolmafia.KoLConstants;
 import net.sourceforge.kolmafia.KoLConstants.MafiaState;
 import net.sourceforge.kolmafia.KoLmafia;
@@ -54,6 +58,13 @@ public class GitManager extends ScriptManager {
   protected static final String MANIFEST = "manifest.json";
   protected static final String MANIFEST_ROOTDIR = "root_directory";
 
+  // Cache of sync path -> projects, used to check for file conflicts
+  private static Map<Path, Set<String>> projectsSyncPaths = null;
+
+  private static void invalidateProjectsSyncPaths() {
+    projectsSyncPaths = null;
+  }
+
   public static void clone(String repoUrl) {
     clone(repoUrl, null);
   }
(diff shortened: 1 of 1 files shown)


midgleyc:
> I believe there is a bug here where if the root changes, but also files have been moved/deleted when the repo's root changed it will leave stale files that were previously synced. I think this is when a root is defined in the manifest and that changes?

True. Also when we go from having no manifest (so root is start of repo) to having a manifest.

This crops up elsewhere because we have no concept of "file X belongs to script Y", just files in folders.

> I didn't add conflict checking here, maybe the both of these could be simplified by simply deleting all the files that were deleted or moved, then call sync() afterwards once?

I would prefer not to sync like this because I modify files locally, outside of the git merge structure. Unsupported I know ;)

midgleyc (approved):

PR #3615, opened 2026-08-12
feat: more repeated modifiers that are counted for TCRS

midgleyc:
Successful hit causes bleeding, chance of poisoning opponent, item drops from warbears.

Item drops from seals / hobos (crystalline seal eye / old soft shoes) don't make their way to TCRS, so looks like warbear item drops is special.


jaadams5 (approved):

gausie:
This is a fascinating opportunity to learn what is RPN and what isn't. I might be tempted to extend Modifier to keep it as a boolean or something, but that gets weird.

midgleyc:
Some things are sometimes and are not other times (i.e. we have combined two enchantments with the same practical effect into one modifier).

e.g. the mime army insignia's Stat Tuning isn't, but the tiki lighter's Stat Tuning is.

From the slightly different enchantment text, poison chance on a weapon and poison chance on an effect are likely different too.

PR #3617, opened 2026-08-13
feat: stop making HTTP requests when checking for florist

midgleyc:
After https://github.com/kolmafia/kolmafia/pull/3595 there were comments that in Standard, Mafia makes a request to check for Friar every combat (because "checked" was removed, we don't have Friar -> so we check -> we still don't have Friar, repeat).

Stop making HTTP requests under any circumstances. Instead, just check game state, like Chateau Mantegna.

Add back the floristFriarChecked preference for backwards compatibility. This removal broke loopstar (https://github.com/Kasekopf/loopstar/pull/53).


jaadams5 (approved):

Rinn:
@midgleyc It appears that kol isn't setting a Last Available date for Order of the Green Thumb Order Form, so this is returning as available in standard right now due to this.

midgleyc:
same thing happened with Burning Leaves, and the solution we went for was to use the Other branch of standard.

And then to get CDM to add a Last Available. I will check.

eta: okay there is "Florist Friar" under "Bookshelf Books". That should do us.

PR #3621, opened 2026-08-13
Express rollover adventures and fights as a modifier

gausie:
The 40 adventures and 10 PvP fights granted at rollover weren't modelled anywhere, so `numeric_modifier("Adventures")` and the sidepane's "Rollover:" lines only reported the bonus over an unstated base. They're now granted as `Generated / Rollover` modifiers.

Slow and Steady grants 100 adventures instead, added as a `Path` modifier, and suppresses every other source of the modifier. The suppression lives in `Modifiers.addDouble` (alongside the existing organ-capacity and `noExperience` gates) so it applies to maximizer speculation too. `maximize adv` in Slow and Steady now recommends nothing. No path interacts with PvP fights.

Note that `numeric_modifier("Adventures")` now includes the base, so scripts that add their own 40 will double-count.


gausie:
If people agree on merging this (I think it's right to), then I'll make a few announcements and merge it after a few days.

VeeArrKoL:
With the exception of the changes to make S&S maximizer behavior sane, isn't this a bit inconsistent with what's included in modifiers for almost every other thing? 

Like modifiers for initiative don't include the base 100%, modifiers for familiar XP don't include the base 1, and the modifiers for crit chance don't include the base `<whatever base crit chance is>`%. Why should the base rollover adventures/fites granted be different?

gausie:
I mean the answer is that they really probably should, but some of those are harder to introduce at this stage.

Sent from [Proton Mail](https://proton.me/mail/home) for iOS.

-------- Original Message --------

midgleyc:
On the side panel, can you remove the + before Adv and PvP? That's what I think the difference is between modifiers that show a bonus (above the normal) and modifiers that show the actual.

Here: 
image
 
Also TES asked about adding the standard +1 to Fam XP a while back so consider doing that too. Base crit chance I think would be useful to add; base item, meat, initiative I think not because I'm too used to thinking of these as a bonus.

gausie:
> On the side panel, can you remove the + before Adv and PvP? That's what I think the difference is between modifiers that show a bonus (above the normal) and modifiers that show the actual.

I am in two minds because I agree, but also it really is an addition to your end-of-day adventure (/fight) count.

midgleyc:
It is an addition to your current adventures, but HP / MP is an addition to your current HP / MP, but we only show the value you would be getting at rollover.

midgleyc (approved):
Thanks!

time to break the ecosystem eh

PR #3623, opened 2026-08-15
A new title so as not to be confused #3564 which this simplifies and replaces.  Safer preference writing.

jaadams5:
See https://github.com/kolmafia/kolmafia/pull/3564

It was getting too confusing to manage the discussion.  For example I could not tell whether I did not understand @gausie or whether their comments did not reflect the most recent changes.

This is, AFAIK, a stripped down version of that PR that does not add any new code to detect and restore corrupt preference files.


libraryaddict:
Is this PR ready for review?

jaadams5:
> Is this PR ready for review?

Yes.  Only lightly editing the title didn't make it clear that this is replacing #3564

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -926,6 +941,54 @@ private static String propertyName(final String user, final String name) {
...
+    String message =
+        "Atomic move not supported when saving "
+            + targetFile
+            + "; using backup-safe replace"
+            + (cause.getMessage() != null ? " (" + cause.getMessage() + ")" : "")
+            + ". Prior contents are in "
+            + Preferences.prefsBackupFileFor(targetFile)
+            + " if recovery is needed.";
+    System.out.println(message);
+    RequestLogger.updateSessionLog(message);
```
libraryaddict, line 977: Wouldn't this mean they are getting log spammed with every single prefs update if they don't have atomic moves?

src/net/sourceforge/kolmafia/preferences/Preferences.java
```diff
@@ -938,17 +1001,44 @@ private static void saveToFile(File file, Map<String, byte[]> encodedData) {
...
+          fos.getFD().sync();
+        }
 
-      try (OutputStream fstream = new BufferedOutputStream(DataUtilities.getOutputStream(file))) {
-        synchronized (encodedData) {
-          for (Entry<String, byte[]> current : encodedData.entrySet()) {
-            fstream.write(current.getValue());
+        // Refresh backup from the previous durable prefs before replace, so a crash
+        // during the upcoming commit cannot leave both files damaged.
+        File backupFile = Preferences.prefsBackupFileFor(file);
```
libraryaddict, line 1019: So we're writing to .tmp above, then writing a file out again, below? Wouldn't this double the amount of file writes?
And if atomic moves are not supported, we write out the file again a third time?

`copy` is being used, not `move`, which may be the mistake here.

libraryaddict (changes requested):
I'm concerned about the doubling of file writes, maybe tripling, which is being introduced in this PR.

Not to sound hostile, but my own PR #3567 was intended to address that in particular.

Rinn got back to me

```
settings/rinnith_prefs.txt, 1,090,401 writes, 8.00 GB
data/mallprices.txt, 8,395 writes, 66.4 MB
everything else is trivial
that was over a 3 hr garbo -> cs -> garbo, so what like 4-5 times a second lol
```

which aligns up with my own usage.

Regardless, my PR would be building off something almost identical here regardless for the actual prefs.txt writes. I just didn't bother adding it in, as it seems a bit odd to submit a PR with the intention of completely invalidate an ongoing PR's work. I planned to adapt the changes into my own PR.

I think part of the reason you had so much trouble here, is because the file management in the class is a bit convoluted. I struggled here too, which is a major reason I split the work off into a new class. And the other reason being that in the end, the user file and the global prefs are being managed in basically the same way, why don't they just get a wrapper.

gausie:
I am inclined to agree with @libraryaddict here. Well, specifically that journaling is probably the right solution. I have not read theirs but I am sure it is good.

jaadams5:
The reason I stripped down the old PR and made this one was because the concerns I was hearing seemed to me to be about things that I considered out of scope.  That still seems to be the case.  It could be because my use of git obscured some history or because I am still having trouble understanding the context of the comments.

However, it seems clear that there is interest in reworking Preferences and there is a possible future where this would not be necessary.  In that spirit I'll close it and not impede progress :-)

Thanks.

PR #3630, opened 2026-08-17
feat: update quest progress on Blue vs Red NC

midgleyc:


VeeArrKoL:
Most of the other stuff in `QuestManager.updateQuestData(String,String)` needs to get parsed and updated as necessary as well, since tracking for all of those things gets broken if it happens in a Blue v. Red NC.

midgleyc:
The main problem with this is that we can't be sure what monster we're fighting in the NC all the time. Ideally we want the monster id stuck in a comment like it is in fights.

VeeArrKoL:
CDM was kind enough to update the pre-choice NC pages with a comment containing the monster ID, so that should help. 🎉

jaadams5 (approved):

PR #3632, opened 2026-08-18
BvR: War Progress from NCs

VeeArrKoL:
Move the tracking for war progress into `QuestManager.updateQuestData(String,String)` so that it gets updated during Blue vs. Red NCs as appropriate.

nb: The change in `IslandManager` is safe because the method only gets called in contexts where the fight was successful anyway.


midgleyc (approved):
Thanks!

on the nb: yes, I ran through doing the same check, looks good.

PR #3633, opened 2026-08-19
Track Interesting Coin charges

VeeArrKoL:
Add tracking for charges of Interesting Coin items: toxic asset, liquid asset, and intangible asset.

Also fixes an issue where tracking of Interesting Coin daily purchases would get reset when visiting the shop if the daily purchase limit had not been reached.


midgleyc (approved):
Thanks!

Checking this myself, the intangible asset charge is only consumed against monsters that drop meat, so we do need to check the message and this is all good.

PR #3634, opened 2026-08-19
Add parsing codpiece slots from the codpiece choice itself

libraryaddict:
Currently, Codpiece will fail to update its parsed slots if `api.php?what=status` is not called. In my case, the request is being intercepted by 8bit Realm. That means all `equip` requests will go through, but mafia will now insist that you do not own the gem, and your equipped gems did not change. Or to rephrase, mafia completely loses track of what gems you have equipped. This is likely an issue in more areas.

The persists between `refresh all` `refresh equip` and so on. It fixes itself when visiting api.php yourself. I'm not sure the issue would persist if you relogged. The reason 8bitrealm is intercepting the api.php request, is to parse the 8bit realm score from charpane.

I’m not interested in addressing the interception of that endpoint, but it does expose a gap.

Currently, codpiece slots cannot be parsed simply by visiting the Codpiece choice.

This PR addresses that by ensuring the slots are also updated from the choice, instead of relying solely on a http request that may not go through.


VeeArrKoL:
Nice change. Is there some case where the parsing doesn't give the correct results, or is the forced call to `api.php` entirely extraneous at this point?

libraryaddict:
I'm pretty sure it would be superfluous to keep doing the check, but didn't care to think about it. Kol does tell us to refresh charpane, which may indicate that we are doing it regardless and it's a wasted round trip. 

I just didn't care to think about it further because it means validating it and possibly making more tests.

I expect that if we're not able to parse it, that there's no expectation of the equipment changing. And at that point it should be caught by api

midgleyc:
We only refresh the charpane when we explicitly make a call to do so. We do it a lot if the user is playing in the relay browser, but barely at all if the user is running a script.

midgleyc (approved):

PR #3635, opened 2026-08-19
Honor Stooper's liver capacity under SOME PIGS

masonc15:
Mafia has suppressed all familiar modifiers under SOME PIGS since 2014, but it
turns out KoL still honors the Stooper's extra liver capacity (the Stooper
postdates that spading by a few years). With SOME PIGS active and a Stooper
equipped, mafia computes an inebriety limit one lower than the game's: at 20
drunkenness with a computed limit of 19, I could still drink normally through
the game UI, and after overdrinking the charpane showed `Tipsiness: 30 / 20`.
In practice this means mafia considers a SOME PIGS player falling-down drunk
one drink early, and refuses a perfectly legal Stooper nightcap with "(usable
quantity ... is limited to 0 by inebriety)".

This PR passes Liver Capacity through from the current familiar's modifiers in
the SOME PIGS early-return; everything else stays suppressed. Tests cover both
directions: familiar modifiers are still zeroed out under SOME PIGS, and
getLiverCapacity() picks up the Stooper's +1 with or without it.


Rinn:
I think this would be better to instead not allow any of the familiar modifiers (from modifiers.txt) to be applied if under the effect rather than adjusting the value here. Are you sure the other modifiers are also correct? Like for example the Combat Rate from Jumpsuited Hound Dog and Disgeist

Rinn:
the check that includes stooper at the bottom of allowBoozeConsumption() probably also needs to account for the effect

gausie:
Yeah I suspect this is per-familiar rather than per-modifier.

masonc15:
Not freshly spaded, no. I dug through the old implementation, and the [2014 White Citadel change](https://github.com/kolmafia/kolmafia/commit/75d9dbcff4b52764e8204309b3369755886b6a18) explicitly treated SOME PIGS as suppressing all familiar effects. The only exception I could verify in KoL is the Stooper's +1 liver capacity.

I think the per-familiar reading makes more sense, so I changed the patch to special-case the Stooper instead of copying Liver Capacity from any familiar. Hound Dog and Disgeist Combat Rate still stay at 0 under SOME PIGS, and I added focused tests for both.

I checked the Stooper branch in `allowBoozeConsumption()` too. It already uses `getLiverCapacity()`, so an equipped Stooper gets the expanded limit through the modifier change. If it isn't equipped, the prompt to switch to it is still valid because the Stooper does raise the legal limit under SOME PIGS. I added a `DrinkItemRequest` test for the legal nightcap case rather than adding another effect check there.

The focused `Modifiers`, `KoLCharacter`, and `DrinkItemRequest` tests all pass locally.

test/net/sourceforge/kolmafia/ModifiersTest.java
```diff
@@ -252,6 +252,77 @@ void fixodeneConsideredInFamiliarModifiersNotExceedingTwenty() {
...
+        var familiar = FamiliarData.registerFamiliar(FamiliarPool.STOOPER, 0);
+
+        familiarMods.applyFamiliarModifiers(familiar, null);
+
+        assertThat(familiarMods.getDouble(DoubleModifier.LIVER_CAPACITY), closeTo(1, 0.001));
+      }
+    }
+
+    @Test
+    void somePigsDoesNotApplyLiverCapacityFromOtherFamiliars() {
```
midgleyc: we don't know whether this is true, because there are no other familiars that provide liver. It's how you've implemented it, but we should be testing the reality, not the implementation. Delete this test.

masonc15: Gotcha, makes sense - sorry bout that. I added it as a defensive test but missed that it uses a game state we can't verify. Removed it 👍

midgleyc (approved):
Happy with this but will allow other comments.

PR #3640, opened 2026-08-21
Express more base game state as modifiers

gausie:
Follow-up to #3621: You, Robot grants no rollover adventures, and the base familiar experience and critical hit chance are modifiers now too.


midgleyc (approved):
Nice!

PR #3646, opened 2026-08-25
Fix bento box handling for peppermint eel sauce vs. eel sauce

VeeArrKoL:
ref: https://kolmafia.us/threads/bento-box-wont-use-peppermint-eel-sauce.32765/

Fixes handling of bento boxes with peppermint eel sauce, and adds some general testing for sushi name construction and parsing.


midgleyc (approved):
Thanks!

PR #3651, opened 2026-08-25
Use journals for writing prefs to reduce IO and increase stability

libraryaddict:
This is a re-opened #3567 

This PR aims to add journaling to lower IO usage for prefs.
Discussion found here; https://kolmafia.us/threads/reduce-the-amount-of-i-o-wear-and-tear-from-writing-preferences.32681/

Marked as a draft because I've recently made an atomic change and although I think there's no issues, its untested and it's easier to leave it as a draft while I do my own thing for a week. I'll mark it as ready then. I currently consider it to be finished, as is traditional before revisiting things.

Tests could be better, but eh. It's probably the crypto miner that people may find offensive.

The .bak is not being treated as a backup for the prefs, it's being treated as the previous version, and that's fine. I can imagine more situations where someone might want it, than the other way around. I'd lean towards removing it altogether, but I'm not having that talk.

So I'm a bit tired, but the way this should work is that we

1. Have a new user start a session
2. A user_prefs,txt is created, this is only modified at several critical points
3. A user_prefs.journal is created, we're appending all property changes here
4. When the user logs out, or the journal grows too big, we finally save our prefs
* First the prefs are written out to .tmp
* Then the data is forced to disk
* The old prefs.txt is atomically moved to .bak
* The .tmp is atomically moved to prefs.txt
* The journal is truncated on a success

Only the .tmp gets the flush to disk. The journal is not flushed to disk, I apologize in advance, but I don't think it's worth it.
I don't believe we should see much issues.

On load, we try to
1. Load from prefs.txt
2. Load from .bak
3. Load from .tmp, we should never get this far
4. Replay the journal over the loaded properties, as that journal should've been emptied if we successfully shutdown gracefully. As it didn't, mean we're in a dirty state.

The one point of criticism I have about my PR is the way I don't fail when we fail to rotate to backup. But, we're already failing anyways. I don't know what to do there. The user won't like it if we discard our data.


jaadams5:
> The .bak is not being treated as a backup for the prefs, it's being treated as the previous version, and that's fine. I can imagine more situations where someone might want it, than the other way around. I'd lean towards removing it altogether, but I'm not having that talk.

My recollection of the origin is that the baks were implemented to solve a problem and months later a synchronization issue was found and fixed.  After that the reports about the bak being needed to replace preferences became rarer.  So I'm open to a discussion about whether we still need the baks or not.

I'll hold off on the code until you undraft it.

gausie:
Looking forward to this leaving draft. Thanks irrat

jaadams5:
Might be out of scope immediately but I wonder if we should delete saveSettingsOnSet once this goes in?  It was added a long time ago because preferences were being lost when mafia crashed.  AFAIK journaling solves the same problem and opting out of journaling seems unwise.

libraryaddict:
I had a writeup on the better solution for .bak, but then I found out that Ctrl + Alt + Backspace will kill my entire linux session. I somehow did the same thing a day later.

So anyways, if memory serves, instead of using a main file and a backup file, write the files to disk, timestamped in their names. So `irrat_prefs-2020-05-30-2133.txt` or something.

Multiple files means that you can switch from having written file1 to file2, and despite the drive not having actually written it to disk as it said for w/e reason, the third file is practically confirmed to exist in a good state. Can easily determine the file to load from their names.

The idea being that atomic writes isn't always supported, and telling the OS to flush to disk isn't always going to be obeyed.

I'm not going to be covering that in this PR however, and I'm not going to be making a PR for it. I myself just make backups.

Regardless, marking PR as ready. I haven't noticed any issues.

src/net/sourceforge/kolmafia/preferences/PreferencesFile.java
```diff
@@ -0,0 +1,369 @@
...
+import java.util.concurrent.TimeUnit;
+import net.java.dev.spellcast.utilities.DataUtilities;
+import net.sourceforge.kolmafia.KoLConstants;
+import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.RequestLogger;
+import net.sourceforge.kolmafia.utilities.FileUtilities;
+
+/**
+ * Controls a preference snapshot/backup/journal file ([user]_prefs.txt/.bak/.journal) and it's IO.
+ * Changes are appended to the journal instead of rewriting the whole snapshot, and the journal is
```
jaadams5, line 29: and its IO

:-)

libraryaddict:
> I had a writeup on the better solution for .bak, but then I found out that Ctrl + Alt + Backspace will kill my entire linux session. I somehow did the same thing a day later.
> 
> So anyways, if memory serves, instead of using a main file and a backup file, write the files to disk, timestamped in their names. So `irrat_prefs-2020-05-30-2133.txt` or something.
> 
> Multiple files means that you can switch from having written file1 to file2, and despite the drive not having actually written it to disk as it said for w/e reason, the third file is practically confirmed to exist in a good state. Can easily determine the file to load from their names.

Actually I'm unsure.

So the breaking change of this PR is that it's replacing the `file.copy`, with a `file.move`, which may break on some niche usecases. I don't think it'll break on modern systems, but I'm not an expert.

Where my thought is that the more redundant system would probably be to have two settings

`backupFileCount` - Default of 3, set to 0 to turn off backups
`backupWithAtomicMoves` - Default of true

When atomic moves is enabled, we will
1. Write the new prefs to .tmp
2. Flush .tmp
3. Atomic move prefs.txt to prefs_timestamp.bak
4. Atomic move .tmp to prefs.txt
5. Truncate the journal - could flush, but it seems pretty harmless to leave
6. Prune old backups

So the files are `main.txt` `[N-1].bak`, `[N-2].bak`, `[N-3].bak` - Backups are the _previous_ version.

When moves are turned off
1. Write prefs to timestamped backup, and flush
2. Write prefs to main file, and flush
3. Journal is truncated - could flush
4. Prune old backups

So the files are `main.txt`, `[N].bak`, `[N-1].bak`, `[N-2].bak` - Backups are the _current_ version at time of save

Then some helper function that can give us the files and their timestamps using their names.

Maybe just flush in all the above cases, because it's pretty rare that we're doing that.

Shouldn't need to worry about OS time being wrong, because that'd be almost certainly user error, given kolmafia is meant to be used when online, and it wouldn't prevent the saves from working, just prevents them from being able to automatically recover their actual latest backup, if they have more than one backup enabled. They could always switch to backups of `1`.


`irrat_prefs_20260830-1844.bak` is 14 extra characters.

Windows has a file path limit of 260, not a real issue unless someone is using an obscenely long name, and that's their own fault. Terrible to type ingame really.

And for those who are familiar with the old `.bak`, can always just have it use `.bak` with no timestamp, when its a backup file count of 1.

The main idea behind this is that we avoid every prefs file being potentially still being mid-IO because something lied to kolmafia that the IO operation is complete, and we give the user options.

Go a step further I suppose and allow a new pref `backupDirectory` which is preset to `settings`. We don't need to special handle this, because as per this https://stackoverflow.com/a/25240805 we don't need to worry about atomic moves silently working when we require an atomic move. And if an atomic move fails, we say as such to the user, that they need to turn the setting off.

Could always go further and add a new pref for atomic moves so that you can have backups always be `[N].bak` instead of `[N-1].bak`. But, that's a user's perspective thing only.

For reference, when I say `[N-1].bak` I mean that it's a backup of your last valid prefs, `[N].bak` is your current prefs. So non-atomic moves, are saving their prefs twice, not once.

libraryaddict:
I get the feeling no one is going to read that lol. I don't know, I just don't want to break peoples setups.

jaadams5:
:-)

I've read the long comment a couple of times but am still digesting it.

I feel like at this time you are either overengineering, trying to futureproof against something that is extremely unlikely to happen or trying to optimize something that already performs well.  I am guilty of all of those at some point in my career ;-)  It does seem as if you are trying to apply the reliability standards of mission critical applications to a game client that runs on a desktop.  It is a noble goal but it has costs and we cannot compel users to configure and manage their systems in order to reduce those costs.  

You have concerns (expressed elsewhere as well) about move vs. copy and I really don't understand those concerns in this context.  Specifically I can't figure out a scenario where changing copy to move breaks something.  If there is a specific niche case then I am curious.   If the niche case is possible but not probable then I would not worry about it.  My opinion might be tainted because I remember that using copy instead of move was not discussed when the code was first written.  Solving the preference loss problem was the focus and copy worked.

I have a strong personal preference against cluttered directories.  I accept bak files as a necessary evil that might be useful in a recovery situation.  I am adamantly against multiple backups in the same directory distinguished only by name.  I can't tell you not to do it but I will insist that it would have to be opt-in and default to one backup file without a timestamp in the name.

:-)

libraryaddict:
Yeah, that's true. My main thought is mostly copy vs move, and cloud storage tagging the moved file. But, the chances of real issues at this point is probably very little already since we're not saving the entire file all the time anymore. They'd have to be pretty unlucky. So it's pretty much just copy vs move and that's probably more an issue with their setup I guess, they shouldn't be housing it in one drive, and should have been using something s little more modern by now, and yet, it still won't outright break even if they do.

midgleyc:
Ran with this today. I have Mafia on OneDrive, which I think qualifies as an unusual file system, but everything seemed to go fine.

Other point of note was that the session timed out, then upon logging back in the preferences were written from the journal file, instead of on log out.

jaadams5:
I'm going to run with this in a couple of hours and approve or report problems.  I'll then give others a chance to chime in and commit it tomorrow if no one beats me to it.

On cloud - all the cloud providers I have tried have no problem with renaming a file and some of them are faster with a rename (as opposed to a copy, delete) because they don't synch file contents.  At one time, at least a decade ago, the preferred set up for KoLmafia was to have \dist in the cloud and every machine that ran KoLmafia had a local \dist that contained local copies and synch'd to the cloud.  Local copies were faster to use and more reliable for I/O.  If synch was an issue the trick was to manually disable synch on \dist until Mafia was shutdown or at list resting.

Bottom line - I don't think move vs. copy will break anything that we can't blame on the cloud provider :-)

jaadams5:
> Other point of note was that the session timed out, then upon logging back in the preferences were written from the journal file, instead of on log out.

@midgleyc 
I'm going to approve but please dismiss it if this is something we need to address now.  

Or commit if you agree :-)

jaadams5 (approved):

libraryaddict:
> Other point of note was that the session timed out, then upon logging back in the preferences were written from the journal file, instead of on log out.

Assuming you meant "I believed the prefs file would only be flushed on logout, or sufficient time has elapsed, or file has grown larger. Not on login"

I hadn't considered people being timed out, given I'm always logged in, but it wasn't something this PR introduced, and it's an explicit intent to save the preferences to file.

https://github.com/kolmafia/kolmafia/blob/91a597340c124d67f028324434df9d0b0522ff8e/src/net/sourceforge/kolmafia/session/LoginManager.java#L279

As for replaying journal and saving the prefs on login, that is [by design](https://github.com/kolmafia/kolmafia/pull/3651/changes#diff-9bc57c1a93ff7f30303294d51222eb183d42e8c3e895c8551a67012b2b75c671R987), because if someone keeps exiting without logging out, it means their journal could keep growing larger without being saved if it never hits the time or size threshold. Perhaps could add a timestamp header, but doesn't seem worthwhile.

The timein thing looks like it's from this change: https://github.com/kolmafia/kolmafia/commit/7b28e791fc09db8df2dca96569f7604a176e2f0a#diff-e927bd43b6ff4b94b70256a178bba54a37c247427aea6665238495ffd127adc9R711
That's 2009, and it looks like it wasn't introduced because of a special reason or another, just to ensure the prefs is saved. Maybe, because of the setting "saveSettings" or whatever it's called.

So a PR that removes the explicit (null) and (name), I think it would be harmless? Unless someone modified something on file and this [is part of their setup.](https://xkcd.com/1172/)

midgleyc:
I was intending to comment that the scenario worked, even if it hadn't explicitly been taken into account, instead of saying there was anything wrong.

PR #3655, opened 2026-08-26
Teach the maximizer about Eternity Codpiece gems

mark-e-duda:
## What

Allow the Modifier Maximizer to consider alternate Eternity Codpiece gem configurations instead of evaluating only the gems currently installed.

## Details

- Considers accessible gems, including gems already installed in the Codpiece
- Evaluates valid five-slot gem combinations when the Codpiece is equipped or available
- Considers an unequipped Codpiece even when its current gems are irrelevant
- Respects owned-copy and five-slot limits
- Prunes irrelevant candidates before combination generation
- Avoids evaluating slot-order permutations of the same gem multiset
- Preserves the current configuration when it ties for the best result

## Testing

`MaximizerTest` covers:

- More than five available gems
- More than five matching gems
- No matching gems
- Already-installed gems
- An unequipped Codpiece containing irrelevant gems
- Candidate pruning
- Owned-copy and slot limits

Spotless and Checkstyle main/test also pass.


libraryaddict:
I'm interested in the following maximizers
`+equip heartstone, hp regen` - I expect the heartstone to be worn in an accessory slot, even if the codpiece is empty
`-equip heartstone, -tie` - Equip the heartstone to a codpiece slot. I'd expect it to either unequip the codpiece, or unslot the heartstone. 
`+equip heartstone, -equip codpiece` - The heartstone is already in the codpiece, will it be removed properly and equipped to an accessory slot.
`+equip heartstone, +equip codpiece, -acc1, -acc2, -codpiece1, -codpiece2, -codpiece3, -codpiece4` Is the heartstone equiped in codpiece 5, and the codpiece equipped in acc3? Does maximizer fail properly when you also exclude codpiece5?

If the maximizer is able to figure out the difference between
* User likes the stats of the gem in a non-codpiece slot more
* User likes the stats of the gem in a codpiece slot over not (eg, heartstone)
* User wants the gem equipped, but does not mind if its in the codpiece or not
* When the gems in the codpiece are optimal and would be perfect, except the maximizer said not to wear the codpiece, so now the gems must be unslotted and worn instead.

Then the other large consideration of if the codpiece is worth wearing. Eg, if we have a incredibly dense meat gem, and a codpiece. And we maximize for meat. The gem should not be in the codpiece. If we have three or four dense meat gems, the codpiece should not be worn. If we have six dense meat gems, now it's optimal to wear the codpiece.

Then we have three dense meat gems, a codpiece and a peridot. We demand the peridot is equipped, we maximize for meat. Is it going to equip the peridot in the codpiece. 

I'm sure the maximizer may work fine, but the tests are only proving that maximizer knows what the codpiece gems give in enchants. The codpiece isn't a free slot, there's interplay in what combinations you might wear.

Sorry I'm a bit tired, but the codpiece is fiddly enough that it does need the tests written.

Failing tests are good too, proving that maximizer knows when the user has asked for something that's not happening. Although I can't think of much here, apart from say maximising for only a codpiece slot, but the codpiece isn't equipped? Not sure the normal thing here

mark-e-duda:
Thank you for the feedback. I wrote tests for your scenarios (and a couple others, to cover some negative tests) and made the following improvements:
- Honor `-codpieceN` and `-equip`
- If a gem is selected as equipment, it is not also able to be slotted into the Codpiece
- If the Codpiece is disallowed, slotted gems will still be considered as potential equipment and unslotted as necessary

mark-e-duda:
@libraryaddict , sorry, I pushed another change after you had approved the workflow. Could I get reapproval? Thank you in advance.

libraryaddict:
I never gave approval sorry, no one did. I am also not a part of the kolmafia team in the repo. You been see the history including commits. You might be confused about the other pr

mark-e-duda:
> I never gave approval sorry, no one did. I am also not a part of the kolmafia team in the repo. You been see the history including commits. You might be confused about the other pr

Odd, I could have sworn I saw the "Run Checks" tasks going. Oh, well. I can wait :)

jaadams5:
> > I never gave approval sorry, no one did. I am also not a part of the kolmafia team in the repo. You been see the history including commits. You might be confused about the other pr
> 
> Odd, I could have sworn I saw the "Run Checks" tasks going. Oh, well. I can wait :)

I will periodically visit GitHub and update from main, review, and start workflows for new contributors.  So that was me :-)

midgleyc:
Stellar work!

I'd appreciate a couple more tests:
* check that the codpiece functions as an accessory with its base enchants: 3 consolation ribbons, 1 codpiece, no gems should equip 2 ribbons and the codpiece
* confirm that gems are removed from the codpiece if the score is better that way, not just if the codpiece (or slots) are forbidden. e.g. codpiece contains heartstone, maximize hp regen, heartstone is equipped directly

How does this function with "pullable / buyable" enabled? Does it recommend purchasing gems from the mall?

libraryaddict:
If the codpiece is in storage or closet (and we are going to acquire it), is it going to process it in a valid order? Eg, will it need to equip it first to decorate it, or does it already acquire before decoration. Or is not needed. Namely closet.

mark-e-duda:
> Stellar work!
> 
> I'd appreciate a couple more tests:
> 
> * check that the codpiece functions as an accessory with its base enchants: 3 consolation ribbons, 1 codpiece, no gems should equip 2 ribbons and the codpiece
> * confirm that gems are removed from the codpiece if the score is better that way, not just if the codpiece (or slots are forbidden). e.g. codpiece contains heartstone, maximize hp regen, heartstone is equipped directly

Thank you for your kind words. More tests coming right up!

> How does this function with "pullable / buyable" enabled? Does it recommend purchasing gems from the mall?

I hadn't checked prior to your question; while it appeared that it wasn't recommending purchases from the mall, the actual issue was that it wasn't considering purchasing multiple of the same gem. Either way, a bug. Thanks for leading me to it.


> If the codpiece is in storage or closet (and we are going to acquire it), is it going to process it in a valid order? Eg, will it need to equip it first to decorate it, or does it already acquire before decoration. Or is not needed. Namely closet.

Decorating does not require it be equipped, but you're right that it can't be done in the closet. Given that, it was _not_ pulling first, which would be an issue. I've fixed it locally (soon to push) and have added a test for the scenario.

libraryaddict:
I think I'm doing something wrong because its telling me to wear only the codpiece, and not slot any gems?
getBoosts() doesn't say anything either. Just the codpiece itself.

```

    @Test
    public void allCodpieceSlotsConsideredByDefault() {

      final var cleanups =
          new Cleanups(
              withEquippableItem(ItemPool.THE_ETERNITY_CODPIECE),
              withItem(ItemPool.HAMETHYST, 5),
              withItem("massive gemstone", 5));

      try (cleanups) {
        assertTrue(maximize("tie"));
        assertThat(getBoosts(), hasItem(recommends(ItemPool.THE_ETERNITY_CODPIECE)));
        assertThat(Maximizer.best.equipment.get(Slot.CODPIECE1).getItemId(), equalTo(ItemPool.HAMETHYST));
        assertThat(Maximizer.best.equipment.get(Slot.CODPIECE2).getItemId(), equalTo(ItemPool.HAMETHYST));
        assertThat(Maximizer.best.equipment.get(Slot.CODPIECE3).getItemId(), equalTo(ItemPool.HAMETHYST));
        assertThat(Maximizer.best.equipment.get(Slot.CODPIECE4).getItemId(), equalTo(ItemPool.HAMETHYST));
        assertThat(Maximizer.best.equipment.get(Slot.CODPIECE5).getItemId(), equalTo(ItemPool.HAMETHYST));
      }
    }```

**Comment by mark-e-duda** [comment]
> I think I'm doing something wrong because its telling me to wear only the codpiece, and not slot any gems? getBoosts() doesn't say anything either. Just the codpiece itself.
> 
> ```
> 
>     @Test
>     public void allCodpieceSlotsConsideredByDefault() {
> 
>       final var cleanups =
>           new Cleanups(
>               withEquippableItem(ItemPool.THE_ETERNITY_CODPIECE),
>               withItem(ItemPool.HAMETHYST, 5),
>               withItem("massive gemstone", 5));
> 
>       try (cleanups) {
>         assertTrue(maximize("tie"));
>         assertThat(getBoosts(), hasItem(recommends(ItemPool.THE_ETERNITY_CODPIECE)));
>         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE1).getItemId(), equalTo(ItemPool.HAMETHYST));
>         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE2).getItemId(), equalTo(ItemPool.HAMETHYST));
>         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE3).getItemId(), equalTo(ItemPool.HAMETHYST));
>         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE4).getItemId(), equalTo(ItemPool.HAMETHYST));
>         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE5).getItemId(), equalTo(ItemPool.HAMETHYST));
>       }
>     }```
> ```

I've been relying on the assertions for testing, so if you mean in a live session, then that's deviant behavior from what I expect. If it is live; do you have gems in your inventory or closet, or are you expecting to have them purchased from the mall?

**Comment by libraryaddict** [comment]
> > I think I'm doing something wrong because its telling me to wear only the codpiece, and not slot any gems? getBoosts() doesn't say anything either. Just the codpiece itself.
> > ```
> > 
> >     @Test
> >     public void allCodpieceSlotsConsideredByDefault() {
> > 
> >       final var cleanups =
> >           new Cleanups(
> >               withEquippableItem(ItemPool.THE_ETERNITY_CODPIECE),
> >               withItem(ItemPool.HAMETHYST, 5),
> >               withItem("massive gemstone", 5));
> > 
> >       try (cleanups) {
> >         assertTrue(maximize("tie"));
> >         assertThat(getBoosts(), hasItem(recommends(ItemPool.THE_ETERNITY_CODPIECE)));
> >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE1).getItemId(), equalTo(ItemPool.HAMETHYST));
> >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE2).getItemId(), equalTo(ItemPool.HAMETHYST));
> >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE3).getItemId(), equalTo(ItemPool.HAMETHYST));
> >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE4).getItemId(), equalTo(ItemPool.HAMETHYST));
> >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE5).getItemId(), equalTo(ItemPool.HAMETHYST));
> >       }
> >     }```
> > ```
> 
> I've been relying on the assertions for testing, so if you mean in a live session, then that's deviant behavior from what I expect. If it is live; do you have gems in your inventory or closet, or are you expecting to have them purchased from the mall?

Sorry but, is there a reason you can't test your changes in live?

**Comment by libraryaddict** [comment]
To be more clear, what I posted was a failing test, where apparently the PR isn't working properly.

"Tie" is a keyword that combines a bunch of different modifiers together, it's why maximizing will equip a basic outfit that's not perfect, but is decent for a baseline adventure.

Internally, tiebreaker is currently set to

```
1 familiar weight, 1 familiar experience, 1 initiative, 5 exp, 1 item, 1 meat, 0.1 DA 1000 max, 1 DR, 0.5 all res, -10 mana cost, 1.0 mus, 0.5 mys, 1.0 mox, 1.5 mainstat, 1 HP, 1 MP, 1 weapon damage, 1 ranged damage, 1 spell damage, 1 cold damage, 1 hot damage, 1 sleaze damage, 1 spooky damage, 1 stench damage, 1 cold spell damage, 1 hot spell damage, 1 sleaze spell damage, 1 spooky spell damage, 1 stench spell damage, -1 fumble, 1 HP regen max, 3 MP regen max, 1 critical hit percent, 0.1 food drop, 0.1 booze drop, 0.1 hat drop, 0.1 weapon drop, 0.1 offhand drop, 0.1 shirt drop, 0.1 pants drop, 0.1 accessory drop, 1 DB combat damage, 0.1 sixgun damage
```

If `tie` isn't working, but the other modifiers are, then it's weird. `tie` is one of the most used maximizer expressions, even if people don't realize it.

mark-e-duda:
> > > I think I'm doing something wrong because its telling me to wear only the codpiece, and not slot any gems? getBoosts() doesn't say anything either. Just the codpiece itself.
> > > ```
> > > 
> > >     @Test
> > >     public void allCodpieceSlotsConsideredByDefault() {
> > > 
> > >       final var cleanups =
> > >           new Cleanups(
> > >               withEquippableItem(ItemPool.THE_ETERNITY_CODPIECE),
> > >               withItem(ItemPool.HAMETHYST, 5),
> > >               withItem("massive gemstone", 5));
> > > 
> > >       try (cleanups) {
> > >         assertTrue(maximize("tie"));
> > >         assertThat(getBoosts(), hasItem(recommends(ItemPool.THE_ETERNITY_CODPIECE)));
> > >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE1).getItemId(), equalTo(ItemPool.HAMETHYST));
> > >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE2).getItemId(), equalTo(ItemPool.HAMETHYST));
> > >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE3).getItemId(), equalTo(ItemPool.HAMETHYST));
> > >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE4).getItemId(), equalTo(ItemPool.HAMETHYST));
> > >         assertThat(Maximizer.best.equipment.get(Slot.CODPIECE5).getItemId(), equalTo(ItemPool.HAMETHYST));
> > >       }
> > >     }```
> > > ```
> > 
> > 
> > I've been relying on the assertions for testing, so if you mean in a live session, then that's deviant behavior from what I expect. If it is live; do you have gems in your inventory or closet, or are you expecting to have them purchased from the mall?
> 
> Sorry but, is there a reason you can't test your changes in live?

I've been spot checking, but for many scenarios, I don't actually own the required items, e.g. the Heartstone.

mark-e-duda:
> To be more clear, what I posted was a failing test, where apparently the PR isn't working properly.
> 
> "Tie" is a keyword that combines a bunch of different modifiers together, it's why maximizing will equip a basic outfit that's not perfect, but is decent for a baseline adventure.
> 
> Internally, tiebreaker is currently set to
> 
> ```
> 1 familiar weight, 1 familiar experience, 1 initiative, 5 exp, 1 item, 1 meat, 0.1 DA 1000 max, 1 DR, 0.5 all res, -10 mana cost, 1.0 mus, 0.5 mys, 1.0 mox, 1.5 mainstat, 1 HP, 1 MP, 1 weapon damage, 1 ranged damage, 1 spell damage, 1 cold damage, 1 hot damage, 1 sleaze damage, 1 spooky damage, 1 stench damage, 1 cold spell damage, 1 hot spell damage, 1 sleaze spell damage, 1 spooky spell damage, 1 stench spell damage, -1 fumble, 1 HP regen max, 3 MP regen max, 1 critical hit percent, 0.1 food drop, 0.1 booze drop, 0.1 hat drop, 0.1 weapon drop, 0.1 offhand drop, 0.1 shirt drop, 0.1 pants drop, 0.1 accessory drop, 1 DB combat damage, 0.1 sixgun damage
> ```
> 
> If `tie` isn't working, but the other modifiers are, then it's weird. `tie` is one of the most used maximizer expressions, even if people don't realize it.

Okay, it's a new test. Come 3AM, I'm a bit less observant. Let me replicate it and investigate.

Edit:
It looks like my gem-list pruning (to reduce number of combinations to check) was a bit too aggressive and I wasn't accounting for tie-breaking. I'll add some more tests surrounding this.

Why is Hamethyst expected here? Isn't the test character's base muscle 0 (thus the percentage muscle bonus would also be 0)?

mark-e-duda:
@libraryaddict , I've added the test (among others), but changed the expectation to Massive Gemstones. If I'm incorrect regarding my understanding of the test-character's base stats, or in the expected behavior of the maximizer expressions, please correct me.

mark-e-duda:
Test failure due to `mallPriceLimitControlsCodpieceGemAvailability` modifying `MallPriceDatabase`, thus polluting the state for `MallPriceDatabaseTest`. Didn't show up locally due to test ordering. Working on fix.

libraryaddict:
I can't really find any issues, the tests look good to me. I did file a related PR, but nothing that should impact you hopefully.

I'll be testing your code on my character later today, which does some complicated stuff for codpiece programmatically, which is going to be a pain to migrate. Less painful than my workarounds though.

Hopefully I can get back to you in a few hours on the results, though shouldn't be a reason to block this PR. I don't forsee any real issues.

Edit: This PR doesn't work properly, not sure on the details yet.

libraryaddict:
Ok sorry, but it looks like we still overlooked something that's kind of hilarious in retrospect.

`massive gemstone` gives Item Drop.

`massive gemstone` is not Standard.

A `Standard` run means that we are only allowed to use stuff that's either part of the base game, or was released in the last 2/3? years. Which is most iotms. Massive gemstone is from some iotm, forgot which, but it's not standard anymore. Which meant maximizer was correctly ignoring it. Except, codpiece is allowed to wear it.

I'm in a standard run, luckily for this PR, which meant I immediately faceplanted into this bug. It's impossible to know how iotms will interact in the future when they leave standard before the codpiece does, unless a devster says something. So that's going to maybe break in the future, we shouldn't do anything though until then.

I'll submit a PR against your PR in some minutes.

VeeArrKoL:
> It's impossible to know how iotms will interact in the future when they leave standard before the codpiece does, unless a devster says something. So that's going to maybe break in the future, we shouldn't do anything though until then.

When codpiece came out, CF40 said:

> fwiw my understanding of our intent is that you can have non-standard gems in the codpiece, and they'll give their enchantments in standard, but not any of the secondary effects. So you'd get the +10 hp/mp from peridot of peril in 2028, but not the skills/monster choosing effect

Time will tell whether that's how it actually ends up working.

libraryaddict:
Alright, PR for the standard fix here:

https://github.com/mark-e-duda/kolmafia/pull/1

> > It's impossible to know how iotms will interact in the future when they leave standard before the codpiece does, unless a devster says something. So that's going to maybe break in the future, we shouldn't do anything though until then.
> 
> When codpiece came out, CF40 said:
> 
> > fwiw my understanding of our intent is that you can have non-standard gems in the codpiece, and they'll give their enchantments in standard, but not any of the secondary effects. So you'd get the +10 hp/mp from peridot of peril in 2028, but not the skills/monster choosing effect
> 
> Time will tell whether that's how it actually ends up working.

Ah yep, and its in modifiers.txt it seems, conditionals are used for the skill. There might be something for "item is standard" or so on, but I don't think that's actually related to this PR. Seems like it's only a concern in 15 months. And I'm not sure if it'd impact anything in reality.

Edit: Forgot there's another bug, something to do with acquiring slotted gems. Sec

midgleyc:
I fixed `withMallPrice` leaking prices in #3658, so you should be able to revert to using that over the mocking.

libraryaddict:
Alright, made a new PR that fixes an issue that broke this change.
I've also opened a PR against this PR, that PR can either be accepted or not. Either way, will need those tests to be copied over.

The PR fixes standard items, and another issue where the codpiece was not being properly considered when we had already filled our slots.

libraryaddict:
My last commit to [my PR to your PR](https://github.com/mark-e-duda/kolmafia/pull/1) has some issues, could be in the earlier commits as well.

I'm too tired to continue with this, but the root problem is that I want to be able to maximize and require gems to be equipped, because I just want the skill. I want maximizer to figure out if its optimal to hold it in my codpiece or in my accessory slots. Because maybe it'll give me more score in my accessory slots than it'd otherwise give in my codpiece. Or maybe the codpiece is optimal. That's maximizer's job.

The codpiece is practically telling maximizer that we have 6 accessory slots, or something like that. And it's blowing up the combinations. I tried a fix, and it seemed to work, but now I have a sudden maximizer failure for no good reason. So clearly I introduced some issue.

To rephrase, my PR branch does have one concrete fix of considering items in standard. The rest of the stuff is trying to address the combination limit explosion. As well as a change that was merged into main here.

I don't know what the good solution is here. Maximizer itself, it's not being performant. It probably requires a large rewrite or something, but that's assuming someone even has a performant idea in mind that'd work just as well as the current maximizer. I don't have that idea. I don't know if maximizer can be improved. We have too many synergies, and kol just wants to add more and more.

I'm not too surprised you didn't encounter this issue, but I don't think the official guidance for kolmafia should be that people keep items out of their inventory.

mark-e-duda:
> My last commit to [my PR to your PR](https://github.com/mark-e-duda/kolmafia/pull/1) has some issues, could be in the earlier commits as well.
> 
> I'm too tired to continue with this, but the root problem is that I want to be able to maximize and require gems to be equipped, because I just want the skill. I want maximizer to figure out if its optimal to hold it in my codpiece or in my accessory slots. Because maybe it'll give me more score in my accessory slots than it'd otherwise give in my codpiece. Or maybe the codpiece is optimal. That's maximizer's job.
> 
> The codpiece is practically telling maximizer that we have 6 accessory slots, or something like that. And it's blowing up the combinations. I tried a fix, and it seemed to work, but now I have a sudden maximizer failure for no good reason. So clearly I introduced some issue.
> 
> To rephrase, my PR branch does have one concrete fix of considering items in standard. The rest of the stuff is trying to address the combination limit explosion. As well as a change that was merged into main here.
> 
> I don't know what the good solution is here. Maximizer itself, it's not being performant. It probably requires a large rewrite or something, but that's assuming someone even has a performant idea in mind that'd work just as well as the current maximizer. I don't have that idea. I don't know if maximizer can be improved. We have too many synergies, and kol just wants to add more and more.
> 
> I'm not too surprised you didn't encounter this issue, but I don't think the official guidance for kolmafia should be that people keep items out of their inventory.

Thank you for your support. You're thinking of things that I definitely didn't. I can take a crack at further fixes.

I agree that the combinatorics on this aren't great, which is why I originally had aggressive list pruning, but both that and (as you stated) your optimization made for sub-optimal results. I'm going to add a few more tests that show failure points with the greedy method. 

An illustrative example: `maximize("cold damage 9 max, cold spell damage 9 max, codpiece1, codpiece2, -tie"));`

- 18-picohertz resonator crystal: +5 cold damage, +5 cold spell damage
- shard of double-ice: +10 cold damage
- kumquartz: +10 cold spell damage
- two enabled sockets
- both modifiers capped at 9

The resonator scores 10 alone, so greedy selects it first and finishes at 14. The two specialized gems score 18 together. The current implementation produces 14 and the test expects 18.

Once I have the maximizer maximizing again, I'll see if I can figure out the bonus skill factor.

libraryaddict:
One regression btw, when we maximize we can use keywords such as `-weapon`
If we have something equipped, like say `broken champagne bottle`, we might use `-equip broken champagne bottle` to exclude that item from our result.
Then we use `-weapon` to exclude the weapon from being considered.

So consider it being equipped in our weapon slot at the time of maximize. Then we ran `-equip broken champagne bottle, -weapon`
You might believe that it shouldn't be allowed, it should fail. And yeah, I did. But it looks like it's not against the maximizer scope, which means at some point, probably my changes, it broke. Because now it's not allowed, when before it was permitted. The `-weapon` told maximizer `do not consider the weapon slot`, it did not mean `do not modify the weapon slot`

So long story short, garbo is maximizing with the broken champagne bottle equipped, then explicitly telling maximizer not to use the broken champagne bottle. And now I have to revert if I want to run my day :)

mark-e-duda:
Converting to draft since there is likely a few more commits that'll need to happen.

mark-e-duda:
Fixes so far have moved us to a combination calculation instead of permutation, and restored correctness; so not O(n^5), but not O(5n) either. Current worst case is having 5 copies of all gems (73 gems + empty) C'(74, 5) = 21,111,090 with all slots open for consideration.

mark-e-duda:
The latest optimization commit touches more files than the Codpiece-specific behavior might initially suggest, so I wanted to explain the rationale and boundaries.

The original exact search recalculated the character’s complete modifier state for every Codpiece gem multiset. With all 73 known gems, that means checking C'(74, 5) = 21,111,090 states. Profiling showed that over 90% of the runtime was spent in `KoLCharacter.recalculateAdjustments`, rather than in enumeration or score comparison.

The changes fall into four groups:

### Split modifier calculation into an invariant prefix and dependent suffix

`KoLCharacter` still exposes the original complete recalculation method, but internally separates calculation immediately before the final experience- and accumulator-dependent transformations. During Codpiece enumeration, the invariant prefix is calculated once. Each candidate copies that baseline, adds its gems, and runs the original dependent suffix.

The suffix is still required because gems can affect Monster Level, experience, Initiative, Item Drop, elemental damage, HP/MP, and other modifiers involved in nonlinear or accumulator-based adjustments.

### Make copying Modifiers inexpensive and complete

The previous copy path iterated over every possible modifier enum value. Copying 324,632 realistic modifier states took approximately **3.96 seconds**—enough to erase most of the benefit of caching.

The collection classes now copy their existing sparse/dense backing collections directly. The same workload takes approximately **0.19 seconds**. The copy constructor also preserves accumulators, which is necessary before rerunning effects such as _Bow-Legged Swagger_, _Bendin’ Hell_, _Dirty Pear_, _Squint_, and _champagne doubling_. The existing plain `set()` behavior remains unchanged with respect to accumulators.

### Limit the optimization conservatively

The late cache is scoped to a single Codpiece search and cleared in `finally`, so it cannot leak between outer equipment candidates.

It is enabled only when both of the following are true:

- There is no active familiar, since familiar calculations can depend on gem-provided familiar weight
- Every possible or fixed gem contains only modifiers known to be safe at the selected cache boundary.

Unknown future modifier types automatically use the original full recalculation path. This preserves exact search behavior rather than assuming new gems are additive.

### Reduce score-loop overhead

`Evaluator` now records the numeric modifiers actually referenced by the maximize expression once, rather than scanning every `DoubleModifier` for every candidate. Derived stat prediction is also skipped unless the expression references a derived stat.

### Performance
Using the full 73-gem case:

**Before: 629.171 seconds
After: 93.048 seconds**
States checked remain exactly 21,111,090
The retained 30-gem benchmark checks exactly 324,632 combinations and completes in approximately 1.3–1.6 seconds.

mark-e-duda:
I don't know if we can really escape the math on the number of combinations to test without entering into the realm of heuristics, which may not always achieve optimal results. I'm trying to squeeze more performance out of the path, so maybe the cost can be absorbed without being too noticeable.

Still need to look into bonus skills too.

mark-e-duda:
A follow-up profile found two shared sources of maximizer overhead outside the Codpiece-specific cache.

`Modifiers.add()` previously scanned every bitmap and boolean modifier enum whenever combining modifier sets, even when the source contained none. It also populated bitmap maps with zero-valued entries. The new implementation iterates only the source collection’s populated entries.

Scoring and comparison also called `getStrings(...).isEmpty()` to check for rollover effects. Since `getStrings()` returns a defensive copy, this allocated a list for every check. A non-allocating presence check now handles that case.

These changes are shared by general modifier aggregation and maximizer scoring. They reduced the 324,632-state benchmark from approximately **1.44s to 0.53s**, and the full 21,111,090-state search from **93.048s to 38.961s**.

VeeArrKoL:
I'm a bit confused as to why we need to be checking against all possible combinations of gems. Surely we can create a reasonably-sized shortlist of top candidates like we do for other slots?

mark-e-duda:
We do already filter out unavailable, excluded, and irrelevant gems. The problem with taking only the top N remaining gems is that they’re ranked individually, while their value can change depending on the other four gems.

For example, suppose Initiative is capped at 100:

Gem A: +100 Initiative — standalone score 100
Gem B: +60 Initiative and +30 Meat Drop — standalone score 90
A top-five shortlist could retain five copies of A and discard B. But five As still score only 100 because of the cap, whereas one A plus four Bs scores 220. Minimum requirements, derived modifiers, and tiebreakers can create similar cases.

So an arbitrary shortlist would be faster, but could silently miss the optimum.

The case I'm optimizing  is specifically the worst (and unrealistic) case where a player own 5 of every gem and has put no restrictions on the maximizer.

mark-e-duda:
Since this PR ended up going in an unexpected direction, I could make a separate PR with the shared-path Maximizer optimizations to clean it up. I think those would be a good improvement standalone.

midgleyc:
General performance improvements would be very welcome.

How the "max" keyword works is incredibly confusing and I'm pretty sure it doesn't work as expected in most cases.

mark-e-duda:
Closing since this has been split.

PR #3654, opened 2026-08-26
Fix javascript stacktraces inside promises being swallowed

libraryaddict:
### Description

JavaScript exceptions inside Promises are losing their stack traces.

Before:

```
[00:20:54] JavaScript exception: TypeError: Cannot find function find. (file:KolMafia/scripts/autoscend.js#31548)                                                                                                                             
[00:20:54]  at file:KolMafia/scripts/autoscend.js:31548 (main2)   
```

With the fix:

```
[00:23:00] JavaScript exception: TypeError: Cannot find function find. (file:KolMafia/scripts/autoscend.js#31548)
[00:23:00]  at file:KolMafia/scripts/autoscend.js:20084 (auto_pre_adventure)                                    
[00:23:00]  at file:KolMafia/scripts/autoscend.js:20516 (auto_runPreAdventure)                                  
[00:23:00]  at file:KolMafia/scripts/autoscend.js:20544 (auto_triggerPreAdventure)                              
[00:23:00]  at file:KolMafia/scripts/autoscend.js:20571 (autoAdv)                                                                                                                                                                             
[00:23:00]  at file:KolMafia/scripts/autoscend.js:49061 (L11_hiddenCityZonesDo)                                 
[00:23:00]  at file:KolMafia/scripts/autoscend.js:19664 (_do)                                                                                                                                                                                 
[00:23:00]  at file:KolMafia/scripts/autoscend.js:18054 (execute)                                               
[00:23:00]  at file:KolMafia/scripts/autoscend.js:19730 (runQuestTask)                                          
[00:23:00]  at file:KolMafia/scripts/autoscend.js:49207 (L11_hiddenCityZones)                                   
[00:23:00]  at file:KolMafia/scripts/autoscend.js:64147 (auto_earlyRoutingHandlingDo)                           
[00:23:00]  at file:KolMafia/scripts/autoscend.js:19664 (_do)                                                   
[00:23:00]  at file:KolMafia/scripts/autoscend.js:18054 (execute)                                                                                                                                                                             
[00:23:00]  at file:KolMafia/scripts/autoscend.js:27358 (runNextTask)                                           
[00:23:00]  at file:KolMafia/scripts/autoscend.js:30219 (doTasks)                                               
[00:23:00]  at file:KolMafia/scripts/autoscend.js:30350 (auto_begin)                                            
[00:23:00]  at file:KolMafia/scripts/autoscend.js:30380 (safe_preference_reset_wrapper)                         
[...]

PR #3656, opened 2026-08-26
Save Eternity Codpiece gems with custom outfits

mark-e-duda:
## What

Save the Eternity Codpiece's five gem slots as part of a custom outfit and restore them when that outfit is equipped.

Configurations are stored locally by custom outfit ID because KoL does not include Codpiece gem assignments in its outfit data.

## Details

- Records all five gem slots when saving an outfit containing the Eternity Codpiece
- Restores the saved configuration after equipping the outfit
- Restores gems even when the outfit's regular equipment is already worn
- Verifies required gems are available before removing the current configuration
- Removes saved configurations for deleted outfits
- Safely ignores malformed saved configuration data
- Avoids unnecessary requests when the correct gems are already installed

## Testing

- `EquipmentManagerTest`
- `EquipmentRequestTest`
- Spotless
- Checkstyle main and test

This PR is independent of the Eternity Codpiece maximizer PR.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
+++ b/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
@@ -33,6 +33,7 @@ import net.sourceforge.kolmafia.persistence.DebugDatabase;
 import net.sourceforge.kolmafia.persistence.EquipmentDatabase;
 import net.sourceforge.kolmafia.persistence.ItemDatabase;
 import net.sourceforge.kolmafia.persistence.ModifierDatabase;
+import net.sourceforge.kolmafia.preferences.Preferences;
 import net.sourceforge.kolmafia.session.EquipmentManager;
 import net.sourceforge.kolmafia.session.InventoryManager;
 import net.sourceforge.kolmafia.session.QuestManager;
@@ -181,18 +182,18 @@ public class EquipmentRequest extends PasswordHashRequest {
 
   public EquipmentRequest(final String changeName) {
     this(EquipmentRequestType.SAVE_OUTFIT);
-    String savedName = EquipmentRequest.addCodpieceConfiguration(changeName);
+    String savedName =
+        Preferences.getBoolean("includeCodpieceGemsInOutfits")
+                && KoLCharacter.hasEquipped(ItemPool.THE_ETERNITY_CODPIECE)
+            ? EquipmentRequest.outfitNameWithCodpieceGems(changeName)
+            : changeName;
     this.addFormField("action", "customoutfit");
     this.addFormField("outfitname", savedName);
     this.addFormField("ajax", "1");
     this.outfitName = changeName;
   }
 
-  private static String addCodpieceConfiguration(final String outfitName) {
-    if (!KoLCharacter.hasEquipped(ItemPool.THE_ETERNITY_CODPIECE)) {
-      return outfitName;
-    }
-
+  public static String outfitNameWithCodpieceGems(final String outfitName) {
     String suffix = " c=" + EquipmentRequest.encodeCodpieceConfiguration();
     int nameLength = Math.min(outfitName.length(), 50 - suffix.length());
     return outfitName.substring(0, nameLength).stripTrailing() + suffix;
--- a/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
+++ b/src/net/sourceforge/kolmafia/request/EquipmentRequest.java
@@ -1,6 +1,9 @@
 package net.sourceforge.kolmafia.request;
 
+import java.io.ByteArrayOutputStream;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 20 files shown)


midgleyc:
My instinct is that I don't like this because this isn't how anything else works. We don't store stickers, card sleeve contents, folders, modeables, bjorn / throne contents: we just pass the outfit through to KoL and let it handle it.

Historically, we allow extra stuff to come through in the outfit name, as in `EquipmentRequest.wearCustomOutfit`. I wouldn't be averse to having c1 through c5 with item ids for gems. Though you do only have 50 characters to work with that way, which doesn't leave you much for a name.

Alternatively, perhaps a pref for including codpiece gems in custom outfits? I'll wait for other comments.

Also, your "AshInteropTest" changes are unrelated?

mark-e-duda:
> My instinct is that I don't like this because this isn't how anything else works. We don't store stickers, card sleeve contents, folders, modeables, bjorn / throne contents: we just pass the outfit through to KoL and let it handle it.
> 
> Historically, we allow extra stuff to come through in the outfit name, as in `EquipmentRequest.wearCustomOutfit`. I wouldn't be averse to having c1 through c5 with item ids for gems. Though you do only have 50 characters to work with that way, which doesn't leave you much for a name.
> 
> Alternatively, perhaps a pref for including codpiece gems in custom outfits? I'll wait for other comments.
> 
> Also, your "AshInteropTest" changes are unrelated?

I'm not familiar with the history/conventions, so if this doesn't fit, I understand.

As for the `AshInteropTest` changes: I added tests to specifically cover missed lines from my patch as reported by Codecov bot. I'm perfectly willing to remove them.

midgleyc:
It's okay to have them, I'm just confused: they're testing `JavascriptRuntime`, but your PR doesn't touch that file? Is there some confusion with #3654, which was merged most recently?

mark-e-duda:
> It's okay to have them, I'm just confused: they're testing `JavascriptRuntime`, but your PR doesn't touch that file? Is there some confusion with #3654, which was merged most recently?

Looking at it, that seems likely. I'm not familiar with Codecov's interface and just took it at face value.

mark-e-duda:
To address your idea regarding slipping metadata into the name; since there aren't many gems that socket in, I could encode them with a lookup table and probably only need to take a couple characters.

VeeArrKoL:
Maybe item ID would be better than a bespoke lookup table?

midgleyc:
`c=AAAAA,BBBBB,CCCCC,DDDDD,EEEEE` would give you at least 20 characters to work with, which seems enough.

mark-e-duda:
> `c=AAAAA,BBBBB,CCCCC,DDDDD,EEEEE` would give you at least 20 characters to work with, which seems enough.

Yeah, I think this is a reasonable approach. Should I convert this to a draft while I change the methodology?

mark-e-duda:
> `c=AAAAA,BBBBB,CCCCC,DDDDD,EEEEE` would give you at least 20 characters to work with, which seems enough.

Just to be conservative with character count, I did a base64 encoding of the itemIDs.

src/net/sourceforge/kolmafia/request/EquipmentRequest.java
```diff
@@ -173,12 +181,38 @@ private static String choosePage(final EquipmentRequestType requestType) {
...
     this.addFormField("ajax", "1");
     this.outfitName = changeName;
   }
 
+  private static String addCodpieceConfiguration(final String outfitName) {
+    if (!KoLCharacter.hasEquipped(ItemPool.THE_ETERNITY_CODPIECE)) {
+      return outfitName;
+    }
+
+    String suffix = " c=" + EquipmentRequest.encodeCodpieceConfiguration();
```
midgleyc, line 197: We shouldn't be changing the names of people's outfits without their input, by default.

Instead, I would:
* add a preference, defaulting off, that when switched on does this 
* add a function in RuntimeLibrary called something like "outfit_name_with_codpiece_gems(String originalOutfitName)" taking an outfit name and returning the truncated name + "c=" + base64 encoded string of gems, for manual use of this feature

mark-e-duda: Is the following the behavior you'd expect?

- Add a new preference (defaulted to off)
  - When off:
    - When saving, make no modification to outfit names
    - When loading, if the outfit name already contains the control string, respect it (or do we want the feature entirely disabled?)
  - When on:
    - When saving, create duplicate outfit with a name generated by new `outfit_name_with_codpiece_gems` function
      - This function will truncate the name and append the control characters + encoded gem layout
    - On loading, if an outfit contains the control string, parse it and equip gems

mark-e-duda: @midgleyc , I've implemented the change as described in my response. If my assumptions are incorrect, I can make further changes.

midgleyc: Yes, that's correct.

The reasoning for saving is that if a user saves an outfit with a particular name, and then attempts to load the outfit they just saved, it should work. If we truncate the name, it won't work. However, if they've set the preference, it's on their own head ;)

Loading with the control string should always work.

midgleyc (approved):

PR #3657, opened 2026-08-27
feat: add kill more skeletons, db combat weaken, super accurate

midgleyc:
Modifiers from #1767 I thought were worth adding.

* Kill More Skeletons: 2 items
* Super Accurate!: 2 items
* DB Combat Weaken: 5 items


midgleyc:
Also fix First Hit DR from #3606 the way we do everything else ambiguous instead of relying on modifier ordering.

gausie (approved):

PR #3659, opened 2026-08-27
tests: skip headful tests on a headless environment

midgleyc:
Convenient when running locally.


jaadams5:
What happens locally for you before this?  Do they fail because you choose to run headless locally?

jaadams5 (approved):

midgleyc:
Yeah, when I run inside WSL (with a straight `./gradlew :test`), they fail. It meant that I had to look through to see which ones failed when testing #3658 instead of just seeing green.

The workaround we use in GitHub actions is to install xvfb and run it with `xvfb-run ./gradlew :test`.

PR #3660, opened 2026-08-28
Add tests for when we should consider container slots, and when we shouldn't

libraryaddict:
This adds four tests for the [Card Sleeve](https://wiki.kingdomofloathing.com/Card_sleeve) and [Shoulder Folder](https://wiki.kingdomofloathing.com/Over-the-shoulder_Folder_Holder), along with stickers and cowboy boots.

The card sleeve can have its items switched out at any time, the folder holder will destroy the replaced item.
Currently maximizer suggests items for the card sleeve, but not the folder holder, stickers or cowboy boots. This seems desirable.

The tests are just to establish what's the normal behavior.

Could be an idea to see if maximizer will allow us to +equip the stuff, but that seems debatable, assuming its not a thing.


midgleyc (approved):
Thanks!

On the non-destructive replacement: we might want tests to confirm that we replace items if there are items slotted, not only if empty. But as behaviour doesn't differ in these cases it's fine as is.

PR #3661, opened 2026-08-28
Count equipment container slots properly in codpiece and card sleeve

libraryaddict:
Correctly kolmafia does not track the amount of slotted items properly.
Eg, codpiece will have mafia report that you don't own the slotted gem.
This also applies to card sleeve.

I probably misremember this somewhat. Regardless, it resulted in the case where kolmafia wasn't tracking slotted items properly, and wasn't handling them properly when acquiring items.

It broke #3655 by maximizer failing when it needed to unslot a gem that was going to be slotted elsewhere.

### DIFF
--- a/src/net/sourceforge/kolmafia/session/InventoryManager.java
+++ b/src/net/sourceforge/kolmafia/session/InventoryManager.java
@@ -260,6 +260,20 @@ public abstract class InventoryManager {
         ++count;
       }
     }
+    if (InventoryManager.equippedOrInInventory(ItemPool.THE_ETERNITY_CODPIECE)) {
+      for (var slot : SlotSet.CODPIECE_SLOTS) {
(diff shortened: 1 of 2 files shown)


libraryaddict:
I believe there's still an "issue" where we can silently fail at unequipping items that shouldn't be unequippable (no access to holder), but that's not really a concern. Just mafia reporting that it should've worked yet failed.

libraryaddict:
Ah, forgot to mention that this also doesn't fix the situation where we have closeted the codpiece, and thus it's not available.
I don't think anyways.
Regardless, that's an error on the user's behalf I'd expect.

midgleyc:
In that case we can't unequip a gem, so I think the behaviour is right.

midgleyc (approved):

PR #3665, opened 2026-08-29
Maximizer: optimize shared modifier hot paths

mark-e-duda:
## Summary

This improves several shared hot paths used while evaluating Maximizer
candidates:

- Precompute the numeric modifiers active in an `Evaluator` expression rather
  than scanning every `DoubleModifier` for every candidate.
- Skip derived-modifier prediction unless the expression uses Muscle,
  Mysticality, Moxie, HP, or MP.
- Copy modifier collection backing data directly instead of scanning every
  modifier enum.
- Iterate only stored bitmap and boolean modifiers when combining sparse
  collections.
- Check for string modifier presence without allocating a defensive list.

## Benchmark

I ran the same temporary JUnit benchmark against the PR base and this patch on
the same machine and JVM. Each operation received one warmup followed by five
measured runs; the table reports the median.

| Shared operation | Iterations | Before | After | Reduction |
| --- | ---: | ---: | ---: | ---: |
| `Evaluator.getScore` | 2,000,000 | 6,232.817 ms | 440.492 ms | 92.9% |
| `Modifiers` copy construction | 300,000 | 4,204.954 ms | 153.195 ms | 96.4% |
| Sparse `Modifiers.add` | 300,000 | 389.419 ms | 214.439 ms | 44.9% |

This is a focused microbenchmark rather than JMH, so the exact numbers should
not be treated as production latency estimates. Its purpose is to compare
identical shared operations under identical conditions. The benchmark harness
is not part of the patch.


mark-e-duda:
@midgleyc , This is an extraction from #3655 that improves shared-path performance in the maximizer.

mark-e-duda:
Converting to draft as I benchmark further improvements.

mark-e-duda:
Benchmarking the Maximizer’s per-combination recalculation and scoring hot path shows a **~43% reduction** in elapsed time. Across nine 100,000-iteration samples after warmup, additive scoring improved from **1,362.650 ms to 755.199 ms**, while derived-modifier scoring improved from **1,377.903 ms to 779.162 ms.** Both revisions produced identical score checksums.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -66,6 +66,8 @@ public class Evaluator {
   private final DoubleModifierCollection weight = new DoubleModifierCollection();
   private Map<DoubleModifier, Double> min;
   private Map<DoubleModifier, Double> max;
+  private List<ScoreModifier> activeScoreModifiers = List.of();
+  private boolean predictsDerivedModifiers;
```
midgleyc: more like `shouldPredictDerivedModifiers`?

src/net/sourceforge/kolmafia/Modifiers.java
```diff
@@ -548,10 +535,9 @@ public boolean set(final Modifiers mods) {
       // them to avoid stacking up multiple copies of the same expression. This isn't technically
       // quite right, but it's probably close enough to right.
       this.expressions.putAll(mods.expressions);
-      changed = true;
     }
 
-    return changed;
+    return true;
```
midgleyc: rather than always returning true, just make this void, and make the single caller that checks the value assume it's true (in ChoiceControl).

midgleyc:
Thanks, performance improvements are always very welcome, couple of minor comments.

src/net/sourceforge/kolmafia/Modifiers.java
mark-e-duda: Technically, it does return false if `mods == null`, but you're right that this is no longer doing change checking (checking if something changed cost more than it saved). Making void and performing the null check outside.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
mark-e-duda: Yup, changed.

src/net/sourceforge/kolmafia/Modifiers.java
midgleyc: hah, I missed that... nice spot.

midgleyc (approved):
Thanks!

PR #3666, opened 2026-08-29
feat: Maximizer: consider Eternity Codpiece gem configurations

mark-e-duda:
## Summary

Teach the Modifier Maximizer to consider obtainable Eternity Codpiece gem configurations instead of scoring only the gems that are currently installed.

The Codpiece search runs as part of the ordinary equipment search, so the Maximizer can choose whether a dual-use gem belongs in an equipment slot or in the Codpiece and can account for copies shared between those uses. It honors the active equipment scope, required and forbidden equipment, excluded slots, Standard restrictions, and the usual acquisition sources. The selected result includes the steps needed to acquire, move, insert, or remove gems; running the search itself does not change the installed configuration.

## Motivation

The Eternity Codpiece is effectively a configurable piece of equipment, but the Maximizer previously treated its installed gems as fixed. This meant that it could recommend an otherwise optimal outfit while overlooking a better configuration made from gems the player already had or could acquire. Users had to choose and move those gems manually, and the best Codpiece configuration could also depend on what occupied the ordinary equipment and familiar slots.

The goal of this change is to make Codpiece gems part of the same optimization problem as the rest of the outfit, so a maximize result represents the best complete configuration KoLmafia can reach under the user's constraints.

## Correctness and performance

Codpiece configurations are scored through the existing Maximizer evaluator rather than a separate approximation. The search avoids equivalent slot permutations and uses conservative bounds when the expression can be bounded safely. If an expression or modifier interaction is not supported by those bounds, it falls back to exact enumeration instead of ignoring that behavior.

This keeps ordinary searches practical while preserving the existing Maximizer semantics for caps, derived values, familiar interactions, bitmap modifiers, and other nonlinear cases.

## Compatibility

The new `maximizerConsiderCodpieceGems` preference defaults to `true`. Setting it to `false` restores the previous behavior: the Maximizer leaves the five installed gem slots unchanged and scores the Codpiece as it is currently configured.

## Testing

The end-to-end tests cover gem availability and acquisition, shared-copy accounting, required and forbidden gems, excluded slots, Standard restrictions, nonlinear modifier interactions, recommendation ordering, and the compatibility preference. They also retain regressions for the Maximizer behavior reported during review.

An opt-in benchmark exercises every default Maximizer expression against a deliberately large gem pool and verifies that bounded searches stay well below exhaustive enumeration.

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Evaluator.java
@@ -92,6 +92,8 @@ public class Evaluator {
   /** if slots[i] >= 0 then equipment of type i can be considered for maximization */
   private final EnumMap<Slot, Integer> slots = new EnumMap<>(Slot.class);
 
+  private final CodpieceMaximizer codpiece = new CodpieceMaximizer(this);
+
   private String weaponType = null;
   private int hands = 0;
   int melee = 0; // +/-2 or higher: require, +/-1: disallow other type
@@ -116,7 +118,76 @@ public class Evaluator {
 
   record ItemBonus(double base, Map<String, Double> modes) {}
 
-  private record ScoreModifier(Modifier modifier, double weight, double min, double max) {}
+  record ScoreModifier(Modifier modifier, double weight, double min, double max) {}
+
+  private static final EnumSet<Slot> SEARCHABLE_SLOTS = EnumSet.copyOf(SlotSet.SLOTS);
+
+  static {
+    SEARCHABLE_SLOTS.addAll(SlotSet.CODPIECE_SLOTS);
+  }
+
+  CodpieceMaximizer codpiece() {
+    return this.codpiece;
+  }
+
+  List<ScoreModifier> getActiveScoreModifiers() {
+    return this.activeScoreModifiers;
+  }
+
+  /**
+   * Whether {@link #getScore} is exactly the weighted, capped sum of its active modifier terms plus
+   * the constant rollover-effect fudge factor, with no per-item bonus or bitmap total mixed in.
+   */
+  boolean scoreIsWeightedSumOfTerms() {
+    return this.stinkycheese <= 0 && this.bonuses.isEmpty() && this.bonusFunc.isEmpty();
+  }
+
+  boolean isRequired(AdventureResult item) {
+    return this.posEquip.contains(item);
+  }
+
+  boolean isForbidden(AdventureResult item) {
+    return this.negEquip.contains(item);
+  }
+
+  boolean contributesToMinimum(Modifiers mods) {
+    var empty = new Modifiers();
+    var predicted = this.shouldPredictDerivedModifiers ? mods.predict() : null;
+    var emptyPredicted = this.shouldPredictDerivedModifiers ? empty.predict() : null;
+    return this.activeScoreModifiers.stream()
+        .filter(term -> term.min() != Double.NEGATIVE_INFINITY)
+        .anyMatch(
+            term ->
+                scoreValue(term.modifier(), mods, predicted)
+                    > scoreValue(term.modifier(), empty, emptyPredicted));
+  }
+
+  boolean contributesToCappedScore(Modifiers mods) {
+    var empty = new Modifiers();
+    var predicted = this.shouldPredictDerivedModifiers ? mods.predict() : null;
+    var emptyPredicted = this.shouldPredictDerivedModifiers ? empty.predict() : null;
+    return this.activeScoreModifiers.stream()
+        .filter(term -> term.max() != Double.POSITIVE_INFINITY)
+        .anyMatch(
+            term ->
+                scoreValue(term.modifier(), mods, predicted)
+                    != scoreValue(term.modifier(), empty, emptyPredicted));
+  }
+
+  boolean contributesToNonlinearScore(Modifiers mods, Set<DoubleModifier> summableModifiers) {
+    var empty = new Modifiers();
+    var predicted = this.shouldPredictDerivedModifiers ? mods.predict() : null;
+    var emptyPredicted = this.shouldPredictDerivedModifiers ? empty.predict() : null;
+    return this.activeScoreModifiers.stream()
+        .filter(
+            term ->
+                !(term.modifier() instanceof DoubleModifier modifier)
+                    || !summableModifiers.contains(modifier))
+        .anyMatch(
+            term ->
+                scoreValue(term.modifier(), mods, predicted)
+                    != scoreValue(term.modifier(), empty, emptyPredicted));
+  }
 
   private static final Pattern MUS_EXP_PERC_PATTERN =
       Pattern.compile("^mus(cle)? exp(erience)? perc(ent(age)?)?");
@@ -859,119 +930,8 @@ public class Evaluator {
       var mod = scoreModifier.modifier();
       double weight = scoreModifier.weight();
       double min = scoreModifier.min();
-      double val = 0.0;
+      double val = scoreValue(mod, mods, predicted);
       double max = scoreModifier.max();
-      if (mod instanceof BitmapModifier bitmapModifier) {
-        val = mods.getBitmap(bitmapModifier);
-      } else if (mod instanceof DoubleModifier) {
-        var doubleModifier = (DoubleModifier) mod;
-        val = mods.getDouble(doubleModifier);
-        switch (doubleModifier) {
-          case MUS:
-            val = predicted.get(DerivedModifier.BUFFED_MUS);
-            break;
-          case MYS:
-            val = predicted.get(DerivedModifier.BUFFED_MYS);
-            break;
-          case MOX:
-            val = predicted.get(DerivedModifier.BUFFED_MOX);
-            break;
-          case FAMILIAR_WEIGHT:
-            val += mods.getDouble(DoubleModifier.HIDDEN_FAMILIAR_WEIGHT);
-            if (mods.getDouble(DoubleModifier.FAMILIAR_WEIGHT_PCT) < 0.0) {
-              val *= 0.5f;
-            }
-            break;
-          case MANA_COST:
-            val += mods.getDouble(DoubleModifier.STACKABLE_MANA_COST);
-            break;
-          case INITIATIVE:
-            val += Math.min(0.0, mods.getDouble(DoubleModifier.INITIATIVE_PENALTY));
-            break;
-          case MEATDROP:
-            val +=
-                100.0
-                    + Math.min(0.0, mods.getDouble(DoubleModifier.MEATDROP_PENALTY))
-                    + mods.getDouble(DoubleModifier.SPORADIC_MEATDROP)
-                    + mods.getDouble(DoubleModifier.MEAT_BONUS) / 10000.0;
-            break;
-          case ITEMDROP:
-            val +=
-                100.0
-                    + Math.min(0.0, mods.getDouble(DoubleModifier.ITEMDROP_PENALTY))
-                    + mods.getDouble(DoubleModifier.SPORADIC_ITEMDROP);
-            break;
-          case HP:
-            val = predicted.get(DerivedModifier.BUFFED_HP);
-            break;
-          case MP:
-            val = predicted.get(DerivedModifier.BUFFED_MP);
-            break;
-          case WEAPON_DAMAGE:
-            // Incorrect - needs to estimate base damage
-            val += mods.getDouble(DoubleModifier.WEAPON_DAMAGE_PCT);
-            break;
-          case RANGED_DAMAGE:
-            // Incorrect - needs to estimate base damage
-            val += mods.getDouble(DoubleModifier.RANGED_DAMAGE_PCT);
-            break;
-          case SPELL_DAMAGE:
-            // Incorrect - base damage depends on spell used
-            val += mods.getDouble(DoubleModifier.SPELL_DAMAGE_PCT);
-            break;
-          case COLD_RESISTANCE:
-            if (mods.getBoolean(BooleanModifier.COLD_IMMUNITY)) {
-              val = 100.0;
-            } else if (mods.getBoolean(BooleanModifier.COLD_VULNERABILITY)) {
-              val -= 100.0;
-            }
-            break;
-          case HOT_RESISTANCE:
-            if (mods.getBoolean(BooleanModifier.HOT_IMMUNITY)) {
-              val = 100.0;
-            } else if (mods.getBoolean(BooleanModifier.HOT_VULNERABILITY)) {
-              val -= 100.0;
-            }
-            break;
-          case SLEAZE_RESISTANCE:
-            if (mods.getBoolean(BooleanModifier.SLEAZE_IMMUNITY)) {
-              val = 100.0;
-            } else if (mods.getBoolean(BooleanModifier.SLEAZE_VULNERABILITY)) {
-              val -= 100.0;
-            }
-            break;
-          case SPOOKY_RESISTANCE:
-            if (mods.getBoolean(BooleanModifier.SPOOKY_IMMUNITY)) {
-              val = 100.0;
-            } else if (mods.getBoolean(BooleanModifier.SPOOKY_VULNERABILITY)) {
-              val -= 100.0;
-            }
-            break;
-          case STENCH_RESISTANCE:
-            if (mods.getBoolean(BooleanModifier.STENCH_IMMUNITY)) {
-              val = 100.0;
-            } else if (mods.getBoolean(BooleanModifier.STENCH_VULNERABILITY)) {
-              val -= 100.0;
-            }
-            break;
-          case EXPERIENCE:
-            double baseExp =
-                KoLCharacter.estimatedBaseExp(
-                    mods.getDouble(DoubleModifier.MONSTER_LEVEL)
-                        * (1 + mods.getDouble(DoubleModifier.MONSTER_LEVEL_PERCENT) / 100));
-            double expPct = mods.getDouble(DoubleModifier.primeStatExpPercent()) / 100.0f;
-            double exp = mods.getDouble(DoubleModifier.primeStatExp());
-
-            val = ((baseExp + exp) * (1 + expPct)) / 2.0f;
-            break;
-          case DAMAGE_AURA:
-            val += mods.getDouble(DoubleModifier.SPORADIC_DAMAGE_AURA);
-            break;
-          case THORNS:
-            val += mods.getDouble(DoubleModifier.SPORADIC_THORNS);
-            break;
-        }
-      }
       if (val < min) this.failed = true;
       score += weight * Math.min(val, max);
     }
@@ -1015,6 +975,123 @@ public class Evaluator {
     return score;
   }
 
+  private static double scoreValue(
+      Modifier scoreModifier, Modifiers mods, Map<DerivedModifier, Integer> predictedModifiers) {
+    if (scoreModifier instanceof BitmapModifier modifier) {
+      return mods.getBitmap(modifier);
+    }
+    if (!(scoreModifier instanceof DoubleModifier modifier)) {
+      return 0.0;
+    }
+    double val = mods.getDouble(modifier);
+    switch (modifier) {
+      case MUS:
+        val = predictedModifiers.get(DerivedModifier.BUFFED_MUS);
+        break;
+      case MYS:
+        val = predictedModifiers.get(DerivedModifier.BUFFED_MYS);
+        break;
+      case MOX:
+        val = predictedModifiers.get(DerivedModifier.BUFFED_MOX);
+        break;
+      case FAMILIAR_WEIGHT:
+        val += mods.getDouble(DoubleModifier.HIDDEN_FAMILIAR_WEIGHT);
+        if (mods.getDouble(DoubleModifier.FAMILIAR_WEIGHT_PCT) < 0.0) {
+          val *= 0.5f;
+        }
+        break;
+      case MANA_COST:
+        val += mods.getDouble(DoubleModifier.STACKABLE_MANA_COST);
+        break;
+      case INITIATIVE:
+        val += Math.min(0.0, mods.getDouble(DoubleModifier.INITIATIVE_PENALTY));
+        break;
+      case MEATDROP:
+        val +=
+            100.0
+                + Math.min(0.0, mods.getDouble(DoubleModifier.MEATDROP_PENALTY))
+                + mods.getDouble(DoubleModifier.SPORADIC_MEATDROP)
+                + mods.getDouble(DoubleModifier.MEAT_BONUS) / 10000.0;
+        break;
+      case ITEMDROP:
+        val +=
+            100.0
+                + Math.min(0.0, mods.getDouble(DoubleModifier.ITEMDROP_PENALTY))
+                + mods.getDouble(DoubleModifier.SPORADIC_ITEMDROP);
+        break;
+      case HP:
+        val = predictedModifiers.get(DerivedModifier.BUFFED_HP);
+        break;
+      case MP:
+        val = predictedModifiers.get(DerivedModifier.BUFFED_MP);
+        break;
+      case WEAPON_DAMAGE:
+        // Incorrect - needs to estimate base damage
+        val += mods.getDouble(DoubleModifier.WEAPON_DAMAGE_PCT);
+        break;
+      case RANGED_DAMAGE:
+        // Incorrect - needs to estimate base damage
+        val += mods.getDouble(DoubleModifier.RANGED_DAMAGE_PCT);
+        break;
+      case SPELL_DAMAGE:
+        // Incorrect - base damage depends on spell used
+        val += mods.getDouble(DoubleModifier.SPELL_DAMAGE_PCT);
+        break;
+      case COLD_RESISTANCE:
+        if (mods.getBoolean(BooleanModifier.COLD_IMMUNITY)) {
+          val = 100.0;
+        } else if (mods.getBoolean(BooleanModifier.COLD_VULNERABILITY)) {
+          val -= 100.0;
+        }
+        break;
+      case HOT_RESISTANCE:
+        if (mods.getBoolean(BooleanModifier.HOT_IMMUNITY)) {
+          val = 100.0;
+        } else if (mods.getBoolean(BooleanModifier.HOT_VULNERABILITY)) {
+          val -= 100.0;
+        }
+        break;
+      case SLEAZE_RESISTANCE:
+        if (mods.getBoolean(BooleanModifier.SLEAZE_IMMUNITY)) {
+          val = 100.0;
+        } else if (mods.getBoolean(BooleanModifier.SLEAZE_VULNERABILITY)) {
+          val -= 100.0;
+        }
+        break;
+      case SPOOKY_RESISTANCE:
+        if (mods.getBoolean(BooleanModifier.SPOOKY_IMMUNITY)) {
+          val = 100.0;
+        } else if (mods.getBoolean(BooleanModifier.SPOOKY_VULNERABILITY)) {
+          val -= 100.0;
+        }
+        break;
+      case STENCH_RESISTANCE:
+        if (mods.getBoolean(BooleanModifier.STENCH_IMMUNITY)) {
+          val = 100.0;
+        } else if (mods.getBoolean(BooleanModifier.STENCH_VULNERABILITY)) {
+          val -= 100.0;
+        }
+        break;
+      case EXPERIENCE:
+        double baseExp =
+            KoLCharacter.estimatedBaseExp(
+                mods.getDouble(DoubleModifier.MONSTER_LEVEL)
+                    * (1 + mods.getDouble(DoubleModifier.MONSTER_LEVEL_PERCENT) / 100));
+        double expPct = mods.getDouble(DoubleModifier.primeStatExpPercent()) / 100.0f;
+        double exp = mods.getDouble(DoubleModifier.primeStatExp());
+
+        val = ((baseExp + exp) * (1 + expPct)) / 2.0f;
+        break;
+      case DAMAGE_AURA:
+        val += mods.getDouble(DoubleModifier.SPORADIC_DAMAGE_AURA);
+        break;
+      case THORNS:
+        val += mods.getDouble(DoubleModifier.SPORADIC_THORNS);
+        break;
+    }
+    return val;
+  }
+
   public double getScore(Modifiers mods) {
     return this.getScore(mods, Map.of(), Map.of());
   }
@@ -1025,7 +1102,8 @@ public class Evaluator {
     if (!this.failed && !this.posEquip.isEmpty()) {
       equipSatisfied = true;
       for (AdventureResult item : this.posEquip) {
-        if (!KoLCharacter.hasEquipped(equipment, item)) {
+        if (!KoLCharacter.hasEquipped(equipment, item)
+            && !CodpieceMaximizer.hasEquippedGem(equipment, item)) {
           equipSatisfied = false;
           break;
         }
@@ -1172,6 +1250,16 @@ public class Evaluator {
     SlotList<CheckedItem> ranked = new SlotList<>(this.familiars.size());
 
     double nullScore = this.getScore(new Modifiers());
+    boolean considerCodpieceGems = Preferences.getBoolean("maximizerConsiderCodpieceGems");
+
+    // Codpiece gems are inserted into the codpiece rather than equipped, so they have no ordinary
+    // slot and never reach the per-slot candidate loop below. Discover them here instead,
+    // unless the codpiece itself is forbidden and no gem could ever be worn.
+    if (considerCodpieceGems
+        && (!this.negEquip.contains(EquipmentManager.ETERNITY_CODPIECE)
+            || KoLCharacter.hasEquipped(ItemPool.THE_ETERNITY_CODPIECE))) {
+      this.codpiece.discoverGems(equipScope, maxPrice, priceLevel, nullScore);
+    }
 
     Map<Integer, Boolean> usefulOutfits = new HashMap<>();
     Map<AdventureResult, AdventureResult> outfitPieces = new HashMap<>();
@@ -1647,6 +1735,14 @@ public class Evaluator {
         }
 
         if (id == ItemPool.CARD_SLEEVE) {
+          break gotItem;
+        }
+
+        // The codpiece is worth what its gems are worth, and those are chosen at the leaf of
+        // the search rather than here, so carry it through on its own rather than judging it
+        // by its bare enchantments.
+        if (id == ItemPool.THE_ETERNITY_CODPIECE && this.codpiece.hasCandidates()) {
+          item.automaticFlag = true;
           break gotItem;
         }
 
@@ -2390,7 +2486,10 @@ public class Evaluator {
     for (int thresh = 1; ; --thresh) {
       if (thresh < 0) return; // no slots enabled
       boolean anySlots = false;
-      for (var slot : SlotSet.SLOTS) {
+      // When Codpiece search is enabled, -codpiece2 excludes a gem slot exactly as -hat excludes
+      // the hat. Otherwise, leaving every gem slot seeded preserves the current configuration.
+      for (var slot : SEARCHABLE_SLOTS) {
+        if (!considerCodpieceGems && SlotSet.CODPIECE_SLOTS.contains(slot)) continue;
         if (this.slots.getOrDefault(slot, 0) >= thresh) {
           spec.equipment.put(slot, null);
           anySlots = true;
--- a/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
@@ -1827,13 +1827,20 @@ public class Maximizer {
       }
       text = text + " (";
 
-      CheckedItem checkedItem = new CheckedItem(itemId, equipScope, maxPrice, priceLevel);
+      boolean codpieceSlot = SlotSet.CODPIECE_SLOTS.contains(slot);
+      CheckedItem checkedItem =
+          new CheckedItem(itemId, equipScope, maxPrice, priceLevel, codpieceSlot);
+      boolean preserveCodpieceGem =
+          EquipmentRequest.isCodpieceGem(itemId)
+              && CodpieceMaximizer.copiesInGemSlots(Maximizer.best.equipment, item) > 0;
 
       long price = 0L;
 
       // How many have been needed so far to make this maximization set?
       // We need 1 + that number to equip this item, not just 1
       int count = 0;
+      int consumed = 0;
+      int released = 0;
 
       // If we're running from command line then execute them straight away,
       // so we have to count how much we've used in 'earlier' items
@@ -1849,8 +1856,11 @@ public class Maximizer {
       } else {
         // Otherwise we iterate through the maximization set so far
         for (Boost boost : Maximizer.boosts) {
-          if (item.equals(boost.getItem())) {
+          if (codpieceSlot && Maximizer.releasesItem(boost, item)) {
+            released++;
+          } else if (item.equals(boost.getItem())) {
             count++;
+            if (!boost.getCmd().isEmpty()) consumed++;
           }
         }
       }
@@ -1880,22 +1890,42 @@ public class Maximizer {
       } else if (checkedItem.initial > count) {
         // This may look odd, but we need an item, not a checked item
         // The count of a checked item includes creatable, buyable, pullable etc.
+        boolean keepFamiliarCopy =
+            codpieceSlot && item.equals(Maximizer.best.equipment.get(Slot.FAMILIAR));
         String method =
-            InventoryManager.simRetrieveItem(
-                ItemPool.get(item.getItemId(), count + 1),
-                equipScope == EquipScope.EQUIP_NOW,
-                false);
+            codpieceSlot && checkedItem.inventory + released > consumed
+                ? "have"
+                : codpieceSlot && equipScope != EquipScope.EQUIP_NOW
+                    ? InventoryManager.simRetrieveItemFromAccessibleSources(
+                        ItemPool.get(item.getItemId(), consumed + 1 - released), !keepFamiliarCopy)
+                    : InventoryManager.simRetrieveItem(
+                        ItemPool.get(item.getItemId(), count + 1),
+                        equipScope == EquipScope.EQUIP_NOW,
+                        false);
         if (!method.equals("have")) {
           text = method + " & " + text;
         }
-        cmd =
-            switch (method) {
-              case "uncloset" -> "closet take 1 \u00B6" + item.getItemId() + ";" + cmd;
-              case "unstash" -> "stash take 1 \u00B6" + item.getItemId() + ";" + cmd;
-              // Should be only hitting this after Ronin I think
-              case "pull" -> "pull 1 \u00B6" + item.getItemId() + ";" + cmd;
-              default -> cmd;
-            };
+        Slot movableCodpieceSlot =
+            preserveCodpieceGem ? Maximizer.movableCodpieceSlot(item) : Slot.NONE;
+        if (movableCodpieceSlot != Slot.NONE) {
+          text = "unequip & " + text;
+          cmd = "unequip " + movableCodpieceSlot.name + ";" + cmd;
+        } else {
+          cmd =
+              switch (method) {
+                case "uncloset" -> "closet take 1 \u00B6" + item.getItemId() + ";" + cmd;
+                case "unstash" -> "stash take 1 \u00B6" + item.getItemId() + ";" + cmd;
+                // Should be only hitting this after Ronin I think
+                case "pull" -> "pull 1 \u00B6" + item.getItemId() + ";" + cmd;
+                case "free pull" ->
+                    preserveCodpieceGem ? "pull 1 \u00B6" + item.getItemId() + ";" + cmd : cmd;
+                case "create", "create or buy" ->
+                    preserveCodpieceGem ? "make \u00B6" + item.getItemId() + ";" + cmd : cmd;
+                case "buy" ->
+                    preserveCodpieceGem ? "buy 1 \u00B6" + item.getItemId() + ";" + cmd : cmd;
+                default -> cmd;
+              };
+        }
       } else if (checkedItem.creatable + checkedItem.initial > count) {
         text = "make & " + text;
         cmd = "make \u00B6" + item.getItemId() + ";" + cmd;
@@ -1944,6 +1974,9 @@ public class Maximizer {
         }
       } else { // Mall buyable
         text = "acquire & " + text;
+        if (preserveCodpieceGem) {
+          cmd = "buy 1 \u00B6" + itemId + ";" + cmd;
+        }
         if (priceLevel != PriceLevel.DONT_CHECK) {
           price = MallPriceManager.getMallPrice(itemId);
         }
@@ -1966,6 +1999,27 @@ public class Maximizer {
       Maximizer.boosts.add(boost);
     }
     return equipScope;
+  }
+
+  private static boolean releasesItem(Boost boost, AdventureResult item) {
+    Slot slot = boost.getSlot();
+    return slot != null
+        && slot != Slot.NONE
+        && EquipmentManager.getEquipment(slot).equals(item)
+        && !Objects.equals(Maximizer.best.equipment.get(slot), item);
+  }
+
+  private static Slot movableCodpieceSlot(AdventureResult item) {
+    for (Slot slot : SlotSet.CODPIECE_SLOTS) {
+      if (EquipmentManager.getEquipment(slot).equals(item)
+          && !Objects.equals(Maximizer.best.equipment.get(slot), item)
+          && Maximizer.boosts.stream()
+              .map(Boost::getCmd)
+              .noneMatch(command -> command.contains("unequip " + slot.name))) {
+        return slot;
+      }
+    }
+    return Slot.NONE;
   }
 
   private static boolean excludedTCRSItem(int itemId) {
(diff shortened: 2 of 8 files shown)


mark-e-duda:
@libraryaddict , to your comment [regarding accessory vs codpiece placement](https://github.com/kolmafia/kolmafia/pull/3655#issuecomment-5429744063); Yes, this is supported. `+equip <gem>` requires a dual-use gem to be equipped, but does not prescribe whether it occupies an ordinary equipment slot or a Codpiece slot. Maximizer evaluates both placements and chooses whichever produces the better score. I added a test covering both outcomes with the same required gem.

libraryaddict:
> @libraryaddict , to your comment [regarding accessory vs codpiece placement](https://github.com/kolmafia/kolmafia/pull/3655#issuecomment-5429744063); Yes, this is supported. `+equip <gem>` requires a dual-use gem to be equipped, but does not prescribe whether it occupies an ordinary equipment slot or a Codpiece slot. Maximizer evaluates both placements and chooses whichever produces the better score. I added a test covering both outcomes with the same required gem.

To prefix, I'm currently not using your latest changes. So don't expect any of these issues to be reproducible.

I've forgotten if this was already addressed in a test, but I believe part of the trouble I had was when I am using `+equip` on at least 3 gems, which naturally takes all the accessory slots.

I did notice that maximizer will explode the combinations if I don't have anything equipped, which makes sense as it's scoring against 'nothing'.

But regardless, for the future I think when this is ready again, will need to validate that the maximizer computed score is consistent with the main branch beforehand.

I did a quick hack on my own changes because I do love the codpiece support, but, I think my last change had regressed somehow in that it's generating a final score that's lower than what I would normally get. Only by something like 5% though. Which is why I think it'd be a good idea for a more shiny player to run some maximizes when this PR is ready, determine that the scores are just as good as before. Probably need to grep all our maximizes and setup a script to run through it all or something I suppose. Handling for the expiring items too.

Not really something I think you can do yourself, given you're not a very shiny player, and though junit tests should be able to catch it, that'd require knowing what's wrong.

I think one of the bigger weaknesses of maximizer is that there's not sufficient junit tests to ensure we don't regress, which isn't this PR's fault, and could be addressed in a future PR. And by that I don't mean "We have no coverage" but "There's so many niche cases". This isn't me saying you need to add those, because that's a little unreasonable, I'm saying that I'm happy to add them myself when testing.

Regardless, let me know when you're confident in the current changes and I'll give it another try.

libraryaddict:
Wait, what's the difference between this PR and the other PR?

Why does this PR exist?

mark-e-duda:
This PR has the standalone Maximizer optimizations separated out, so this one is dependent on that new one. I think I need to abandon the original Codpiece PR.

mark-e-duda:
@libraryaddict , I think this is in good enough shape to test now, if you're willing. While still not fantastic, the optimizations + pruning yield results in seconds instead of minutes. Of course, when the Codpiece isn't available or explicitly disallowed, the benchmarks improves to their prior sub-second values. I tried to add equivalence checks to keep in line with prior behavior, but like you said; there are a lot of edge cases.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -63,11 +63,14 @@ public class Evaluator {
   public boolean failed;
   boolean exceeded;
   private Evaluator tiebreaker;
+  private final CodpieceEvaluator codpieceEvaluator = new CodpieceEvaluator(this);
   private final DoubleModifierCollection weight = new DoubleModifierCollection();
+  private final EnumSet<DoubleModifier> explicitScoreModifiers =
```
libraryaddict, line 68: This looks like dead code

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
```diff
@@ -183,8 +183,8 @@ public static void maximize(
     Maximizer.boosts.clear();
     if (filter.contains(KoLConstants.filterType.EQUIP)) {
       Maximizer.best.getScore();
-      // In case the current outfit scores better than any tried combination,
-      // due to some newly-added constraint (such as +melee):
+      MaximizerSpeculation currentEquipment = Maximizer.best.clone();
```
libraryaddict, line 186: This isn't mentioned in the PR description? What is this doing and what happened before? I can see tests, but it'd be helpful to show you understood what happened before and why this is an improvement.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -503,11 +503,28 @@ public void surgeonosityTriesSurgeonEquipment() {
     }
 
     @Test
-    public void surgeonosityItemsDontStack() {
+    public void defaultSurgeonosityAccountsForShirtEligibility() {
```
libraryaddict, line 506: Feels a bit weird to add a surgeon bugfix into a codpiece PR without mentioning it

libraryaddict (commented):
This is legal in current kolmafia, but appears to fail for your branch
```
    @Test
    public void forbiddenEquipInExcludedSlotIsLeftAlone() {
      final var cleanups = new Cleanups(withEquipped(Slot.WEAPON, ItemPool.BROKEN_CHAMPAGNE));

      try (cleanups) {
        assertTrue(maximize("-equip broken champagne bottle -weapon"));
        assertThat(
            Maximizer.best.equipment.get(Slot.WEAPON).getItemId(),
            equalTo(ItemPool.BROKEN_CHAMPAGNE));
      }
    }
```
https://github.com/loathers/garbage-collector in particular, likes to do that kind of thing.

I also note that you're hardcoding a bunch of modifiers, which seems a bit difficult for future modifications.

It could be an idea to add tests that would catch that kind of regression, eg, a gem gives a modifier that isn't being handled currently, and silently gets ignored.

Personally I'd prefer something where maximizer constructs that kind of data on runtime, but I don't understand enough about what's going on to make that kind of recommendation. Esp if it may be expensive.

Separating codpiece into a new class, to me feels a bit fiddly. It's practically another edition of maximizer's evaluator and it makes me unsure how well this can be maintained in the future.

I can see some of how this works together, but it feels odd to me. The description feels a bit obtuse and I don't understand how it directly maps into the chosen code design. Others may disagree with me though.

If you do end up resubmitting this, separating the PR into smaller PRs that can be considered and merged in individually may be easier for others to review, as currently this seems like something that's hard to take in at a glance.

midgleyc:
> While still not fantastic, the optimizations + pruning yield results in seconds instead of minutes. 

Can we have a pref to switch this whole thing off? I think you could do it with "-codpiece1 -codpiece2 -codpiece3 -codpiece4 -codpiece5", but existing scripts can't necessarily do this easily.

Also as irrat says any non-codpiece changes should be split out.

mark-e-duda:
> > While still not fantastic, the optimizations + pruning yield results in seconds instead of minutes.
> 
> Can we have a pref to switch this whole thing off? I think you could do it with "-codpiece1 -codpiece2 -codpiece3 -codpiece4 -codpiece5", but existing scripts can't necessarily do this easily.

Yes, I can add that. I had considered adding the switch but hadn't simply because live testing with anything other than enabling buying along with "don't check" had about the same response time with or without the Codpiece. The time really only becomes noticeable when approaching the worst case scenario.


> Also as irrat says any non-codpiece changes should be split out.

Can do.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
mark-e-duda: It seemed like a small enough incidental change that it wasn't worth calling out, but I'm more than happy to split it off.

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
mark-e-duda: Maybe this should be split out too, because it is a separate bug that I had noticed. If you already have the optimal outfit equipped, the maximizer sometimes recommends a worse result. In my case it was recommending to replace the Grandfather Watch with Boots of Twilight Whispers when maximizing for Adventures. (Watch provides 6, Boots provide 3).

I added the fix as I was churning since I thought it was due to the Codpiece changes, but I was able to replicate it on main.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
mark-e-duda: I'll check this. I went through a lot of iterations trying to tune performance (a lot of commit history was lost because I was dumb and really messed up my work area at one point), so it's entirely possible.

mark-e-duda:
> This is legal in current kolmafia, but appears to fail for your branch
> 
> ```
>     @Test
>     public void forbiddenEquipInExcludedSlotIsLeftAlone() {
>       final var cleanups = new Cleanups(withEquipped(Slot.WEAPON, ItemPool.BROKEN_CHAMPAGNE));
> 
>       try (cleanups) {
>         assertTrue(maximize("-equip broken champagne bottle -weapon"));
>         assertThat(
>             Maximizer.best.equipment.get(Slot.WEAPON).getItemId(),
>             equalTo(ItemPool.BROKEN_CHAMPAGNE));
>       }
>     }
> ```
> 
> https://github.com/loathers/garbage-collector in particular, likes to do that kind of thing.

Thank you for the find. I'll fix this.

> I also note that you're hardcoding a bunch of modifiers, which seems a bit difficult for future modifications.

I'm not sure what you mean. I don't believe I've hardcoded anything more than what already existed. I personally agree that the Maximizer is not very extendable due to exactly that reason. Trying to generalize further would either be forced abstraction for the Codpiece or a major redesign of the Maximizer.

> It could be an idea to add tests that would catch that kind of regression, eg, a gem gives a modifier that isn't being handled currently, and silently gets ignored.

While silent, I don't believe it's ignored; it just falls back to the (very expensive) exhaustive search path. Unless you mean an entirely new modifier, then yes, but that's an issue for the Maximizer as a whole.

> Personally I'd prefer something where maximizer constructs that kind of data on runtime, but I don't understand enough about what's going on to make that kind of recommendation. Esp if it may be expensive.

A lot of the odd things that I did were a result of needing to squeeze performance. But maybe there could be some things I could do on startup instead of having them static? The value of that would be dependent on how often new gems become available.

> Separating codpiece into a new class, to me feels a bit fiddly. It's practically another edition of maximizer's evaluator and it makes me unsure how well this can be maintained in the future.

I feel the opposite. I originally made the changes inline, but adding thousands of lines to an existing class to support one item seemed like a bad idea. A new class makes the seam clear and any future changes to the Codpiece can be localized.

> I can see some of how this works together, but it feels odd to me. The description feels a bit obtuse and I don't understand how it directly maps into the chosen code design. Others may disagree with me though.

A number of things changed since the original description was written. I need to rewrite it, and more clearly. That's definitely on me.

> If you do end up resubmitting this, separating the PR into smaller PRs that can be considered and merged in individually may be easier for others to review, as currently this seems like something that's hard to take in at a glance.

I may be able to extract a couple things, but the issue is that the reason it got so big was because of what it took to make gem selection feasible. To split it up would be to have intermediate stages where it isn't really a usable feature.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
mark-e-duda: Moved to #3688

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
midgleyc: yes, please extract any non-codpiece changes into their own PRs

There also seems to be a bunch of code on volleyball tuning / sombrero tuning in modifiers that again would be better split out.

midgleyc:
I appreciate the codpiece changes being in their own classes -- maximizer is large and a big ball of mud at this point -- but I would appreciate all the tests being at the end-to-end MaximizerTest level. Unit tests prohibit refactoring the architecture with confidence nothing breaks, as the tests are tied to that architecture. And I'm not sure I *like* the architecture.

As-is, CodpiecePruning alone being 1.3k lines (most of which seems not strictly codpiece related) and having its own 700 line test file, and the whole PR adding 7000 lines... this might be too large to review. Some of this can definitely be split out.

Also, I'm concerned that essentially every stopping point has had irrat find a bug (thanks, irrat!). Where there's one bug, more often lurk, and I'm not confident that there won't be more issues.

midgleyc:
(ref #3655, original implementation)

src/net/sourceforge/kolmafia/maximizer/Maximizer.java
mark-e-duda: Moved to #3704

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
mark-e-duda: Removed in latest iteration.

mark-e-duda:
@libraryaddict @midgleyc, thanks for the detailed reviews. The branch has changed substantially since those comments, so I went back through the concerns against the current implementation.

I agree that correctness and performance need to be reviewed separately, but I do not think they should land in separate PRs. Exact enumeration is too slow for the feature to be usable, while the optimization is unsafe without the exact-search behavior and tests in the same change. I have instead reduced the implementation and organized the remaining review around explicit seams.

`CodpiecePruning` and `CodpieceEvaluator` are gone. Codpiece-specific discovery, copy accounting, enumeration, and bounds now live in one `CodpieceMaximizer`; final scoring and comparison still use the existing `Evaluator` and `MaximizerSpeculation.checkBest()` paths. At current head `ee3cbc7`, the PR changes 9 files with 2,319 additions and 242 deletions. Of those additions, 1,015 are in tests; the production and data changes account for 1,304 additions and all 242 deletions. The unrelated surgeonosity fix moved to #3688, and the pre-existing current-equipment bug moved to #3704.

Gem data is discovered at runtime from the `ETERNITY_CODPIECE` modifier records; there is no fixed list of gem ids or supported combinations. Exact enumeration remains the reference behavior. Pruning is enabled only for scoring semantics represented by an admissible bound; unsupported or nonlinear scoring retains candidates and uses exact enumeration. The recently added linear `modbonus` term is included directly in the bound and has an end-to-end pruning regression.

The dedicated tests use the normal `Maximizer.maximize` entry point and assert selected equipment, scores, and recommendations. They cover the reported forbidden-item regression, three required dual-use gems sharing accessory and Codpiece slots, copy reuse, unavailable sources, Standard restrictions, nonlinear and bitmap interactions, recommendation ordering, and retaining a better currently equipped item. There is no separate unit-test API tied to the removed pruning classes.

The `maximizerConsiderCodpieceGems` preference defaults to `true`. Setting it to `false` preserves the installed gem configuration and bypasses alternative-gem discovery, acquisition, and search. It is an operational fallback and isolates the new gem search for live comparisons; it is not a claim that every surrounding Maximizer code path is identical to pre-PR main.

For a new review, I've organized changes into four commits that should be independently reviewable; the first two commits are behavior-preserving preparation, the third is the exact feature, and the fourth is the performance improvement. Live comparison with normal maximize expressions and expiring items would also be especially helpful since my character is very shiny.

1. **Shared combination evaluation (`140bad6`):** The main points of interest are `Evaluator`, `MaximizerSpeculation.checkBest()`, and the accompanying `MaximizerTest` regressions. This is a behavior-preserving refactor that extracts score calculation and completed-combination handling so a specialized search can use the same path. Its intended contract is that ordinary equipment maximization remains unchanged. The commit changes 3 files with 222 additions and 170 deletions; 40 of the additions are tests.
2. **Generic availability support (`eb95b22`):** This commit is mostly contained in `CheckedItem` and `InventoryManager.simRetrieveItemFromAccessibleSources()`. `CheckedItem` gains generic policies for maximum useful quantity and for retaining owned or equipped copies despite Standard, while the existing constructor preserves the old one-copy acquisition behavior. `InventoryManager` gains side-effect-free simulation across accessible sources. This commit contains no Codpiece or gem policy. It changes 2 production files with 103 additions and 37 deletions, with no test-file changes.
3. **Exact Codpiece search (`68814e7`):** This is the central feature with the naive, but provably correct, exhaustive search. The central path runs through `CodpieceMaximizer.initialize()`, `discoverGems()`, `selectCandidates()`, and `searchGems()`, with copy accounting in `MaximizerSpeculation.countAfterCodpieceSlots()` and recommendation generation in `Maximizer.emitSlot()`. The main behaviors covered here are activation and lifecycle, required/forbidden/excluded slots, canonical multiset enumeration, physical-copy accounting, and recommendation order. Codpiece-specific quantity and Standard rules are supplied to the generic availability APIs here. There is deliberately no pruning in this commit, so it serves as the reference behavior for the next commit. It changes 7 files with 1,432 additions and 38 deletions; 811 of the additions are tests, leaving 621 production additions.
4. **Pruning (`ee3cbc7`):** The optimization is concentrated in `CodpieceMaximizer.buildBound()` and `ScoreBound`. The relevant safety questions are whether each contribution used by `Evaluator.getScore()` is represented in the upper bound or causes an exact fallback, whether directed rounding keeps the bound optimistic, and whether pruning follows `MaximizerSpeculation.compareTo()` ordering. The differential cases, `modbonus` regression, and opt-in benchmark belong to this layer. It changes 3 files with 570 additions and 5 deletions; 164 of the additions are tests, leaving 406 production additions.

libraryaddict:
Your comment looks like a wall of text that's not really meant to be read. (to me)

mark-e-duda:
> Your comment looks like a wall of text that's not really meant to be read.

Sorry, I thought the spacing and formatting made it rather readable. The short of it is that I reworked a lot of this feature to directly address the comments received, and given that it's still quite large, I organized the commits to be independently reviewable.

libraryaddict:
Yeah, my last comment was unnecessarily hostile. Sorry about that.

I'm currently testing the changes, but will know more after rollover.

mark-e-duda:
> Yeah, my last comment was unnecessarily hostile. Sorry about that.

No worries, I sometimes have a hard time reading tone.

> I'm currently testing the changes, but will know more after rollover.

Excellent, thank you. I appreciate the help.

jaadams5:
>I agree that correctness and performance need to be reviewed separately, but I do not think they should land in separate PRs.

I think they should be in separate PRs because it reduces the intellectual overload of reviewing.

A principle that guided my coding career is "It is easier to make working code fast than to make fast code work." 

A second principle is that you should always make measurements before optimization.  That will show that the effort was applied to the real hot spot and that the optimization works.

So I would suggest two PRs and the second one should have before and after timing measurements.  If you are concerned about the user experience we can certainly delay committing the first until the second is approved.

mark-e-duda:
> A second principle is that you should always make measurements before optimization. That will show that the effort was applied to the real hot spot and that the optimization works.
> 
> So I would suggest two PRs and the second one should have before and after timing measurements. If you are concerned about the user experience we can certainly delay committing the first until the second is approved.

As explained in my comment above, SHA 68814e7 is the feature before optimization, which I benchmarked prior to applying optimization using:

 `$env:KOLMAFIA_CODPIECE_DEFAULTS_BENCHMARK = "true"; .\gradlew.bat :test --tests "net.sourceforge.kolmafia.maximizer.EternityCodpieceMaximizerTest.benchmarksEveryDefaultExpressionWithEveryCodpieceGem"`

Before optimization >45 minutes, after is ~4.2 seconds.

However, the break is pretty clean. If it's easier to review as a separate PR, I'd just want to have the feature disabled by default until the performance improvement landed.

VeeArrKoL:
> ~4.2 seconds

This is against a benchmark in which the maximizer string and gem inventory are chosen to be pathological, right? Is a seconds-long runtime also likely for a "typical" use case for a player in aftercore? If so, while it's an obviously dramatic improvement over 45 minutes, I feel like this is still too slow for something enabled by default.

libraryaddict:
`equipment.values()` CodpieceMaximizer iterates over this, but doesn't check for nulls, which throws an error.

mark-e-duda:
> > ~4.2 seconds
> 
> This is against a benchmark in which the maximizer string and gem inventory are chosen to be pathological, right? Is a seconds-long runtime also likely for a "typical" use case for a player in aftercore? If so, while it's an obviously dramatic improvement over 45 minutes, I feel like this is still too slow for something enabled by default.

I was looking at the wrong column in my table, the 4.2 seconds was actually the time for the first string to complete which includes initialization overhead. The more meaningful numbers are total benchmark and median times since the benchmark contains all 36 default maximizer strings.

Total time was 14.8 seconds, Median was 258 milliseconds. Omitting the first string, the next highest were `+four songs` at 996ms and `familiar weight` at 866ms.

These, as you noted, are absolute worst case where you have 5 of every gem available, which isn't possible for IOTM gems like Heartstone or Baseball Diamond.

From a subjective point of view, it hasn't been noticeably slower when I've used it interactively.

mark-e-duda:
> `equipment.values()` CodpieceMaximizer iterates over this, but doesn't check for nulls, which throws an error.

Good find, thanks. I've added a check + regression test.

libraryaddict:
So, the codpiece isn't very performant.

I did not test what the performance was like before this PR.

I turned off auto sastify with closet.

33,408 combinations checked when codpiece is in closet
1,924,416 combinations checked when codpiece is not in closet. Takes just over a minute.

You can effectively assume I have most items in the game. I'm in aftercore.

I think maximizer used to be slower before your improvements though, which is nice.

```

> maximize 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Brimstone Bu
nker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken champagne
 bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip ¶5869", -"
equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus rake", 12
2.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit", 253.25 "b
onus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"                                                                         
[16:24:28] Maximizer: 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Br
imstone Bunker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken
 champagne bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip 
¶5869", -"equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus
 rake", 122.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit",
 253.25 "bonus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"                                                               
[16:24:28] Loading character status...               
[16:24:28] Maximizing...            
[16:24:30] 33408 combinations checked, best score 8,806.39
 > closet take 1 the eternity codpiece                                                                                                                                                                                                                      
[16:24:38] Removing items from closet...                                                                                                                                                                                                                    
[16:24:38] You acquire an item: The Eternity Codpiece                                                                                                                                                                                                       
[16:24:39] Requests complete.                                                                                                                                                                                                                               
[16:24:39]                                                                                                                                                                                                                                                  
 > maximize 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Brimstone Bu
nker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken champagne
 bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip ¶5869", -"
equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus rake", 12
2.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit", 253.25 "b
onus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"
[16:24:40] Maximizer: 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Br
imstone Bunker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken
 champagne bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip 
¶5869", -"equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus
 rake", 122.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit",
 253.25 "bonus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"
[16:24:40] Loading character status...
[16:24:40] Maximizing...
[16:24:45] 76800 combinations checked, best score 8,806.39
[16:24:50] 157696 combinations checked, best score 8,806.39
[16:24:56] 236544 combinations checked, best score 8,806.39
[16:25:01] 316416 combinations checked, best score 8,806.39
[16:25:06] 397312 combinations checked, best score 8,806.39
[16:25:11] 478208 combinations checked, best score 8,806.39
[16:25:16] 559104 combinations checked, best score 8,806.39
[16:25:21] 641024 combinations checked, best score 8,806.39
[16:25:26] 721920 combinations checked, best score 8,806.39
[16:25:31] 802816 combinations checked, best score 8,806.39
[16:25:36] 884736 combinations checked, best score 8,806.39
[16:25:41] 966656 combinations checked, best score 8,806.39
[16:25:46] 1047552 combinations checked, best score 8,806.39
[16:25:51] 1129472 combinations checked, best score 8,806.39
[16:25:56] 1210368 combinations checked, best score 8,806.39
[16:26:01] 1292288 combinations checked, best score 8,806.39
[16:26:06] 1374208 combinations checked, best score 8,806.39
[16:26:11] 1456128 combinations checked, best score 8,806.39
[16:26:16] 1537024 combinations checked, best score 8,806.39
[16:26:21] 1618944 combinations checked, best score 8,806.39
[16:26:26] 1700864 combinations checked, best score 8,806.39
[16:26:31] 1782784 combinations checked, best score 8,806.39
[16:26:36] 1864704 combinations checked, best score 8,806.39
[16:26:40] 1924416 combinations checked, best score 8,806.39
 > js Slot.all().map(s => s + " " + equippedItem(s))
[16:26:50] Returned: aggregate string [31]
[16:26:50] 0 => hat eldritch hat
[16:26:50] 1 => weapon broken champagne bottle
[16:26:50] 2 => holster none
[16:26:50] 3 => off-hand KoL Con 13 snowglobe
[16:26:50] 4 => back Camp Scout backpack
[16:26:50] 5 => shirt BGE 'cuddly critter' shirt
[16:26:50] 6 => pants eldritch pants
[16:26:50] 7 => acc1 Mayor Ghost's sash
[16:26:50] 8 => acc2 lucky gold ring
[16:26:50] 9 => acc3 Jekyllin hide belt
[16:26:50] 10 => familiar toy Cupid bow
[16:26:50] 11 => crown-of-thrones none
[16:26:50] 12 => sticker1 none
[16:26:50] 13 => sticker2 none
[16:26:50] 14 => sticker3 none
[16:26:50] 15 => card-sleeve Alice's Army Foil Coward
[16:26:50] 16 => folder1 folder (holographic fractal)
[16:26:50] 17 => folder2 folder (rainbow unicorn)
[16:26:50] 18 => folder3 folder (owl)
[16:26:50] 19 => folder4 none
[16:26:50] 20 => folder5 none
[16:26:50] 21 => buddy-bjorn none
[16:26:50] 22 => bootskin none
[16:26:50] 23 => bootspur none
[16:26:50] 24 => codpiece1 hamethyst
[16:26:50] 25 => codpiece2 hamethyst
[16:26:50] 26 => codpiece3 hamethyst
[16:26:50] 27 => codpiece4 hamethyst
[16:26:50] 28 => codpiece5 hamethyst
[16:26:50] 29 => fakehand none
[16:26:50] 30 => hats none

**Comment by mark-e-duda** [comment]
> You can effectively assume I have most items in the game. I'm in aftercore.
```
maximize 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Brimstone Bu
nker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken champagne
 bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip ¶5869", -"
equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus rake", 12
2.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit", 253.25 "b
onus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"                                                                         
[16:24:28] Maximizer: 2.21 Item Drop 10000 max, Familiar Weight, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Br
imstone Bunker", -"equip Buddy Bjorn", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken
 champagne bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip 
¶5869", -"equip ¶5872", -buddy-bjorn, -crown-of-thrones, -familiar, -weapon, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 1087.2 "bonus lucky gold ring", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus
 rake", 122.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Crown of Thrones", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 214.29 "bonus tiny stillsuit",
 253.25 "bonus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 71.05 "bonus Pantsgiving", 75 "bonus toy Cupid bow"
 ```

This is exactly the real-world perspective that I'm lacking, thank you again. How do you even come up with an expression like this? Just trial and error?

> So, the codpiece isn't very performant.
> 
> 33,408 combinations checked when codpiece is in closet 1,924,416 combinations checked when codpiece is not in closet. Takes just over a minute.

Yeah, I don't think a minute wait is acceptable. In theory, I know why the combination count ballooned; it has to do with what is considered prune-able by individual contribution; there are still several modifiers that fallback to pure combinatorial searches. I've wracked my brain but haven't found good, safe, and effective pruning conditions.

What's interesting is that the best combination was found very early; I wonder if I'm still being too liberal with my upper bound...

Since this is a matter of performance vs. correctness, is the solution for now to just disable the feature by default?

libraryaddict:
> Since this is a matter of performance vs. correctness, is the solution for now to just disable the feature by default?

If the performance isn't an issue currently, this doesn't seem reasonable. By which I mean the tradeoff doesn't look like anyone would use this unless they have little equipment available.

The bigger issue is that I don't see a reason we need to explode that high. Codpiece has effectively only two gems I think for that maximizer. Massive gemstone and heartstone.

mark-e-duda:
> If the performance isn't an issue currently, this doesn't seem reasonable. By which I mean the tradeoff doesn't look like anyone would use this unless they have little equipment available.

I guess I still don't have a handle on what the typical use case looks like. I can only speak from my personal testing which isn't nearly as complex. But I have a few ascensions, am in aftercore, enable mall purchasing without budget limit, and can't say I've noticed any particular lag beyond what I see with the codpiece disabled. So for a player like me (however common or uncommon that is), this would be useful.

> 
> The bigger issue is that I don't see a reason we need to explode that high. Codpiece has effectively only two gems I think for that maximizer. Massive gemstone and heartstone.

The reason is the implicit tiebreaker which prevents us from pruning gems that don't contribute to the primary score.

libraryaddict:
Yeah, I don't really know.
```
Maximizer: -tie, 0 Item Drop, 2.75 Meat Drop, -"equip Amulet of Perpetual Darkness", -"equip Brimstone Beret", -"equip Brimstone Bludgeon", -"equip Brimstone Boxers", -"equip Brimstone Bracelet", -"equip Brimstone Brooch", -"equip Brimstone Bunker", -"equip Crown of Thrones", -"equip Spooky Putty ball", -"equip Spooky Putty leotard", -"equip Spooky Putty mitre", -"equip Spooky Putty snake", -"equip angelbone chopsticks", -"equip angelbone dice", -"equip angelbone totem", -"equip broken champagne bottle", -"equip devilbone corset", -"equip devilbone greaves", -"equip devilbone rosary", -"equip dice-shaped backpack", -"equip papier-masque", -"equip papier-mitre", -"equip smoke ball", -"equip stinky fannypack", -"equip ¶5868", -"equip ¶5869", -"equip ¶5872", -acc1, -buddy-bjorn, -crown-of-thrones, -off-hand, 100 "bonus pantogram pants", 100000 "bonus small peppermint-flavored sugar walking crook", 120.14 "bonus Mr. Cheeng's spectacles", 122.9 "bonus rake", 122.9 "bonus tiny rake", 126.62 "bonus can of mixed everything", 137.71 "bonus June cleaver", 139.5 "bonus Snow Suit", 144 "bonus Buddy Bjorn", 153.85 "bonus cursed magnifying glass", 174 "bonus lucky gold ring", 180 "bonus Mr. Screege's spectacles", 192.89 "bonus Cup of 13s", 250 "bonus mafia thumb ring", 253.23 "bonus KoL Con 13 snowglobe", 317.03 "bonus aviator goggles", 35.17 "bonus Portable Laughing Stock", 6.25 "bonus Powerful Glove", 9000 "bonus ¶11942"
Loading character status...
Maximizing...
5888 combinations checked, best score 15,718.27
```
That didn't need many combinations at all, and the codpiece has baconstone gems closeted. Some combination somewhere bouncing off each other I suppose.

Edit: Sorry, just realized I misread it. So, nothing really changed.

libraryaddict:
Incase you didn't see my edit, I was wrong. I hadn't accounted for the `tie` or combination limit. So my last message is effectively meaningless sorry.

VeeArrKoL:
I checked out this branch and tried it and am also seeing poor results. Even a simpler string like `2.21 Item Drop, Familiar Weight` is checking millions of combinations and taking over a minute. (92,442 combinations and ~instantaneous results with the codpiece closeted; 4,962,049 combinations and ~2 minutes with it available)

Is the issue that it decides there aren't *enough* relevant gems, so it's considering *everything* for tiebreaker value or something? Unfortunately, the codpiece portion of the maximizer doesn't seem to respond to the `dump` keyword, so there's no way to tell exactly what's going on.

VeeArrKoL:
Whoops, this thread moved a bit while I was doing my testing. If the issue is that guaranteeing we get the tiebreaker right requires checking orders of magnitude more combinations: personally, I'd sooner see it give potentially not-quite-tiebreaker-optimal results and finish quickly rather than give-optimal-results-including-tiebreakers but take so long that it's never worth using.

That said, `2.21 Item Drop, Familiar Weight -tie` actually made it test *more* combinations for me than the version permitting the tiebreaker, which is very confusing.

mark-e-duda:
> Whoops, this thread moved a bit while I was doing my testing. If the issue is that guaranteeing we get the tiebreaker right requires checking orders of magnitude more combinations: personally, I'd sooner see it give potentially not-quite-tiebreaker-optimal results and finish quickly rather than give-optimal-results-including-tiebreakers but take so long that it's never worth using.

That's possibly a workable angle that could be user selectable.

> That said, `2.21 Item Drop, Familiar Weight -tie` actually made it test _more_ combinations for me than the version permitting the tiebreaker, which is very confusing.

Yeah, that's weird. That points to me just having done something wrong.

I don't think it's quite "back to the drawing board," but, evidently, I have more work to do

midgleyc:
From your long comment, your first two commits should be reviewable separately, and able to go in as refactoring improvements without any of the codpiece stuff, right? Can you file them as separate PRs?

For the two codpiece commits, I agree that there's no point merging an incredibly unperformant solution, so keep that as a PR with two (or however many) commits.

Also, as VR mentions, "dump" (and presumably "2dump") should work for codpiece as well.

jaadams5:
Observations - while I noticed some timing comments I did not realize the were before and after for the latest version of the code.  Sorry for my confusion.

My take away from three weeks, multiple pages of discussion and no resolution is that too much has been dumped into this.  I very much agree with @midgleyc.  Let's deal with the codpiece separately.  If, indeed the codpiece is too slow to be practical then I'm fine with one codpiece PR that covers everything.  I think our goals are the same and we don't have to follow my path to get there :-)

I note that none of my characters has a codpiece so unless I was doing a speculative run to decide what to buy I would never run this code,

Would a two stage "maximization" work?  Stage one "maximizes" just the gem configuration.  Stage two treats the codpiece as an item that cannot be configured.  Either it is equipped or it isn't.  If my recollection of my days doing operations research are correct, the codpiece configuration could be expressed as an optimization problem solved by linear programming.  That gives us tools to determine if there is a feasible solution and, if so, prove a configuration achieves it.  We have choices for LP solvers that are guaranteed to avoid the exponential time costs.

I can construct all sorts of hypotheticals and edge cases where a two stage formulation might not have the best solution but since the maximizer is a heuristic algorithm that does not produce a provable maximum, just a number that is accepted as being "close enough" it probably doesn't make a difference.

mark-e-duda:
I've been able to replicate the performance issues in testing and agree that with the current number of "fallback" paths, this isn't a viable feature. I think any further optimization is going to have to be more general, rather than specific to this feature. And before optimization, I think cleanup is necessary for better separation of responsibilities, which I think will make future Maximizer improvements easier to write and review. That would be a multistage process to avoid the burden of review huge line counts.

So, I think I have two paths forward with respect to this;
1) Go back to draft mode (or closed if that is the preference) until other base work is done, or
2) Add a simple heuristic in place of the current "exact" optimization, so it can be a somewhat useful feature in the interim

In any case, I'm grateful for all of your time, effort and patience. Thank you.

mark-e-duda:
> My take away from three weeks, multiple pages of discussion and no resolution is that too much has been dumped into this. I very much agree with @midgleyc. Let's deal with the codpiece separately. If, indeed the codpiece is too slow to be practical then I'm fine with one codpiece PR that covers everything. I think our goals are the same and we don't have to follow my path to get there :-)

This is a new way of working for me, so I appreciate the feedback; I'm working on getting better breaking up work into chunks. 
And I think yo and @midgleyc are correct that this should be split further.

> Would a two stage "maximization" work? Stage one "maximizes" just the gem configuration. Stage two treats the codpiece as an item that cannot be configured. Either it is equipped or it isn't.

I agree, as a heuristic, that'd probably be a good way to handle it. As you point out there are cases where this doesn't yield an optimal result; familiar weight bonus from heartstone being a prominent example.

> If my recollection of my days doing operations research are correct, the codpiece configuration could be expressed as an optimization problem solved by linear programming. That gives us tools to determine if there is a feasible solution and, if so, prove a configuration achieves it. We have choices for LP solvers that are guaranteed to avoid the exponential time costs.

Based on this, I imagine you're more qualified than me to make such a determination, but from what I've looked into, this seems to fit a Mixed-Integer Programming model rather than LP. Though your point still stands that we shouldn't need to reinvent the wheel. I was looking into what it would take, in terms of redefining the problem, to utilize something like ojAlgo.

libraryaddict:
```Returned: aggregate string [31]
0 => hat - Apriling band helmet
1 => weapon - legendary seal-clubbing club
2 => holster - none
3 => off-hand - oversized monocle on a stick
4 => back - bat wings
5 => shirt - surgical apron
6 => pants - tearaway pants
7 => acc1 - Heartstone
8 => acc2 - The Eternity Codpiece
9 => acc3 - Möbius ring
10 => familiar - toy Cupid bow
11 => crown-of-thrones - none
12 => sticker1 - none
13 => sticker2 - none
14 => sticker3 - none
15 => card-sleeve - Alice's Army Foil Coward
16 => folder1 - folder (holographic fractal)
17 => folder2 - folder (rainbow unicorn)
18 => folder3 - folder (owl)
19 => folder4 - none
20 => folder5 - none
21 => buddy-bjorn - none
22 => bootskin - none
23 => bootspur - none
24 => codpiece1 - massive gemstone
25 => codpiece2 - massive gemstone
26 => codpiece3 - massive gemstone
27 => codpiece4 - massive gemstone
28 => codpiece5 - massive gemstone
29 => fakehand - none
30 => hats - none
Maximizer: 5 item drop, 0.05 meat drop, 0.5 init 440 max, 0.1 da 1000 max, dr, 0.5 all res, 1.5 Mainstat, 0.4 hp, 0.2 mp 1000 max, 3 HP Regen, 1.5 Weapon Damage, 0.75 Weapon Damage Percent, 1.5 Elemental Damage, 2 Familiar Weight, 5 Familiar Experience, 5 Experience, 2.5 Moxie Experience Percent, -200 Combat Rate 32 max, -fumble, Effective, +250"bonus Baseball Diamond", +200"bonus Monodent of the Sea", +400"bonus blood cubic zirconia", +200"bonus Möbius ring", +40"bonus Cup of 13s", +40"bonus Portable Laughing Stock", +50"bonus spring shoes", +200"bonus bat wings"
Loading character status...
Maximizing...
74176 combinations checked, best score 11,157.97
Putting on fake arrow-through-the-head...
Equipment changed.
Holding Monodent of the Sea...
Equipment changed.
Putting on silky pirate drawers...
Equipment changed.
Putting on McHugeLarge left ski...
Equipment changed.
Putting on blood cubic zirconia...
You acquire an item: massive gemstone
Loading character status...
Equipment changed.
Putting on Baseball Diamond...
You acquire an item: massive gemstone
Loading character status...
Equipment changed.
Maximizer: 5 item drop, 0.05 meat drop, 0.5 init 440 max, 0.1 da 1000 max, dr, 0.5 all res, 1.5 Mainstat, 0.4 hp, 0.2 mp 1000 max, 3 HP Regen, 1.5 Weapon Damage, 0.75 Weapon Damage Percent, 1.5 Elemental Damage, 2 Familiar Weight, 5 Familiar Experience, 5 Experience, 2.5 Moxie Experience Percent, -200 Combat Rate 32 max, -fumble, Effective, +250"bonus Baseball Diamond", +200"bonus Monodent of the Sea", +400"bonus blood cubic zirconia", +200"bonus Möbius ring", +40"bonus Cup of 13s", +40"bonus Portable Laughing Stock", +50"bonus spring shoes", +200"bonus bat wings"
Loading character status...
Maximizing...
10580 combinations checked, best score 11,179.43
Putting on duonoculars...
Equipment changed.
Maximizer: 5 item drop, 0.05 meat drop, 0.5 init 440 max, 0.1 da 1000 max, dr, 0.5 all res, 1.5 Mainstat, 0.4 hp, 0.2 mp 1000 max, 3 HP Regen, 1.5 Weapon Damage, 0.75 Weapon Damage Percent, 1.5 Elemental Damage, 2 Familiar Weight, 5 Familiar Experience, 5 Experience, 2.5 Moxie Experience Percent, -200 Combat Rate 32 max, -fumble, Effective, +250"bonus Baseball Diamond", +200"bonus Monodent of the Sea", +400"bonus blood cubic zirconia", +200"bonus Möbius ring", +40"bonus Cup of 13s", +40"bonus Portable Laughing Stock", +50"bonus spring shoes", +200"bonus bat wings"
Loading character status...
Maximizing...
10580 combinations checked, best score 11,179.43
Returned: aggregate string [31]
0 => hat - fake arrow-through-the-head
1 => weapon - legendary seal-clubbing club
2 => holster - none
3 => off-hand - Monodent of the Sea
4 => back - bat wings
5 => shirt - surgical apron
6 => pants - silky pirate drawers
7 => acc1 - McHugeLarge left ski
8 => acc2 - duonoculars
9 => acc3 - Möbius ring
10 => familiar - toy Cupid bow
11 => crown-of-thrones - none
12 => sticker1 - none
13 => sticker2 - none
14 => sticker3 - none
15 => card-sleeve - Alice's Army Foil Coward
16 => folder1 - folder (holographic fractal)
17 => folder2 - folder (rainbow unicorn)
18 => folder3 - folder (owl)
19 => folder4 - none
20 => folder5 - none
21 => buddy-bjorn - none
22 => bootskin - none
23 => bootspur - none
24 => codpiece1 - blood cubic zirconia
25 => codpiece2 - Baseball Diamond
26 => codpiece3 - massive gemstone
27 => codpiece4 - massive gemstone
28 => codpiece5 - massive gemstone
29 => fakehand - none
30 => hats - none

```

Maximizer doesn't seem to be equipping the codpiece sometimes when it should be. And I'm not sure why it switched gear again with the second run.

jaadams5:
> I'm not sure why it switched gear again with the second run.

For what it is worth I see this a lot running from the GUI.  I run, receive equipment recommendations, accept them and then run again.  The recommendation for one slot changes with no reason given and the contribution value of the change is "(+0)".  Sometimes the second recommendation includes something unrelated to the maximizer string that is beneficial.  A couple of times I have seen the recommendation toggle between two pieces of equipment.  

My maximizer strings tend to be very simple - maybe three or four keywords at most.  My hypothesis is that the implicit tie-breaking is a factor.  I also suspect something is not always being cleared or updated between runs.  Since I simultaneously don't trust the maximizer to be "correct" but find it close enough to be useful I have not really researched because I don't expect a fix to be a quality of life improvement for me.  FWIW I don't have a codpiece of many items that are configurable.

PR #3669, opened 2026-08-30
Add a new function to expose maximizer success / fail

libraryaddict:
Expose whether the last maximizer run succeeded or failed to scripts.

This is mainly useful for cases where a script wants to verify that maximizer worked / failed when unable to use a function that returns a boolean.


src/net/sourceforge/kolmafia/modifiers/BooleanModifier.java
```diff
@@ -108,7 +108,8 @@ public enum BooleanModifier implements Modifier {
       Pattern.compile("All Spells Cast Are Sleazy"),
       Pattern.compile("All Spells Cast Are Sleazy")),
   SUPER_ACCURATE(
-      "Super Accurate!", Pattern.compile("Super Accurate!"), Pattern.compile("Super Accurate!"));
+      "Super Accurate!", Pattern.compile("Super Accurate!"), Pattern.compile("Super Accurate!")),
+  RESULT("Result", null);
```
midgleyc: would be nice to exclude this from BOOLEAN_MODIFIERS too (so `$modifier[result]` isn't valid, so it can't be maximized, etc.)

so BOOLEAN_MODIFIERS is only those modifiers with a non-null tagPattern.

src/net/sourceforge/kolmafia/modifiers/DoubleModifier.java
```diff
@@ -624,7 +624,9 @@ public enum DoubleModifier implements Modifier {
   KILL_MORE_SKELETONS(
       "Kill More Skeletons",
       Pattern.compile("Kill (\\d+)% More Skeletons"),
-      Pattern.compile("Kill More Skeletons: " + EXPR));
+      Pattern.compile("Kill More Skeletons: " + EXPR)),
+  COMBOS_CONSIDERED("Combos Considered", null),
```
midgleyc: same as BOOLEAN_MODIFIERS comment for DOUBLE_MODIFIERS

but, drat: the tagPattern comment would also exclude prismatic damage and raw combat rate, which should be allowed. so maybe just exclude these manually

gausie:
I like the idea but don't love the pseudomodifiers. I will think about it

gausie:
Maybe we could have like

```
boolean last_maximizer_succeeded()
float   last_maximizer_score()
int     last_maximizer_combinations_considered()
```

Which would be a lot easier and has the benefit of being typed

libraryaddict:
> Maybe we could have like
> 
> ```
> boolean last_maximizer_succeeded()
> float   last_maximizer_score()
> int     last_maximizer_combinations_considered()
> ```
> 
> Which would be a lot easier and has the benefit of being typed

Yeah, I wasn't too sure about my implementation either. I'll make the change later today

VeeArrKoL:
In an ideal world it feels like a single method with a structured return type would be better, but I guess making a specialized record type just for this would be overkill.

libraryaddict:
> In an ideal world it feels like a single method with a structured return type would be better, but I guess making a specialized record type just for this would be overkill.

Actually, that does seem like an idea. We already have that kind of overload on functions, 
eg,

```
{ display: string; command: string; score: number; effect: Effect; item: Item; skill: Skill; }[] maximize(string maximizerStringValue, number maxPriceValue, number priceLevelValue, boolean isSpeculateOnlyValue, boolean showEquipment)
```

You might be confused with `Monster` or `Location`.

Regardless, I'll wait to see what the others think. It could go either way here.

VeeArrKoL:
Oh, I didn't realize we could just do that already. In that case, is there a particular reason to not just add these into the record that the 5-arg version of `maxmize()` already returns?

libraryaddict:
> Oh, I didn't realize we could just do that already. In that case, is there a particular reason to not just add these into the record that the 5-arg version of `maxmize()` already returns?

That's an array

gausie:
Tried this out: it looks like nothing resets best/bestChecked between runs. I did a valid maximize but then after `maximize nonsense` fails to parse, last_maximizer_succeeded() is still true (and score was still a number also). We could snapshot the outcome or something. I can write a failing test if it's not clear.

libraryaddict:
Kind of a bigger headache than I expected.

So to be clear, the new functions will _not_ report anything about the maximizer outcomes on non-equipment. That's part of maximizer design.

There was also a long old standing bug apparently, where the first maximize of the day if it was not equipment, could throw an error. Because `best` was still null because it wasn't assigned unless it was run with the equip filter.  Not that it really was a concern.

I believe moving the assignments of the two, to be moved up, shouldn't effect anything, unless you're the kind of person who likes to run maximize attempts, and expects stale maximizes to persist afterwards.

This is impacting anyone who maximizes with equipment, then maximizes with effects then tries to use `numericModifier("_spec:something")` on the maximizer to get the equipment speculation.

As for the change not working for `maximize("meat", 0,0,0,"wish")`, simply put, maximizer never hit the code paths for it, it'd be a new change, and I'm not entirely sure it's worthy. Not really the change I wanted to make either.

Honestly the only change I really wanted was to get the maximizer outcome, success or fail. The other stuff was because it felt pretty underwhelming to add container boilerplate on for a simple true/false. So I'm happy to strip the PR down to just true/false, it might be useful for debugging purposes, but otherwise I don't think people really care about the other stuff other than "Did my maximizer pass the test"

As an example of debugging purposes, would be the recent codpiece PRs, when you want to autonomously figure out if codpiece regressed something, ingame, without actually equipping.

gausie:
> Honestly the only change I really wanted was to get the maximizer outcome, success or fail. The other stuff was because it felt pretty underwhelming to add container boilerplate on for a simple true/false. So I'm happy to strip the PR down to just true/false, it might be useful for debugging purposes, but otherwise I don't think people really care about the other stuff other than "Did my maximizer pass the test"

I totally agree with this if it would help you get it there.

gausie (approved):
LGTM

PR #3671, opened 2026-08-30
Add Blue vs. Red team affiliation to monster data

VeeArrKoL:
Adds Blue vs. Red team affiliation to monster data

ref: https://kolmafia.us/threads/path-blue-vs-red.32757/post-180103


src/net/sourceforge/kolmafia/textui/parsetree/ProxyRecordValue.java
```diff
@@ -1692,6 +1692,7 @@ public static class MonsterProxy extends ProxyRecordValue {
             .add("random_modifiers", new PluralValueType(DataTypes.STRING_TYPE))
             .add("manuel_name", DataTypes.STRING_TYPE)
             .add("wiki_name", DataTypes.STRING_TYPE)
+            .add("bvr_team", DataTypes.STRING_TYPE)
```
midgleyc: while "BvR" is good for the monsters.txt data file, I think we should be exposing this as "blue_vs_red_team" for clarity.

Similarly with the enum name, attribute name, etc.

jaadams5 (approved):

midgleyc (approved):
Thanks!

PR #3672, opened 2026-08-31
Fix API status updates with Continuum Transfunctioner

VeeArrKoL:
There's a number of places in mafia where `ApiRequest.updateStatus()` gets explicitly called to update information from api.php. However, if the user happens to have a Continuum Transfunctioner equipped at that time, the call gets dispatched as a char pane update instead. While this is useful if the intent is to update 8-Bit realm points, it would break updating other things. 

For example, if you happen to graft something in Z is for Zootomist while wearing the Continuum Transfunctioner, the call to `ApiRequest.updateStatus()` intended to cause your grafted familiar preferences (e.g. `zootGraftedButtCheekLeftFamiliar`) to get updated gets dispatched only to charpane.php and the desired update never happens.

This PR changes the behavior to always do the actual api.php call (unless we're in a limit mode), and *additionally* refreshes the charpane if the transfunctioner is equipped. I changed the control flow a little to ensure the return value is unchanged from the previous behavior (though I don't know that anything actually cares either way).

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ApiRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ApiRequest.java
@@ -81,15 +81,17 @@ public class ApiRequest extends GenericRequest {
       return ApiRequest.updateStatusFromCharpane();
     }
 
-    // Similarly, if you have the continuum transfunctioner equipped,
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks!

PR #3676, opened 2026-08-31
Retain pull order in _roninStoragePulls

libraryaddict:
Currently the pull order of the preference is unsorted, it's using a HashSet which does not promise any order that makes sense to a normal user.

This means when you pull an item, your storage preference of the pulls done so far may be out of order from what actually happened.

This PR switches it to a LinkedHashSet which keeps the order. Even if there was a performance hit, it's only 20 items.

First commit is failing.


gausie (approved):
Assuming these tests fail before the fix, lgtm

PR #3677, opened 2026-08-31
feat: add maximum hp / mp modifier

midgleyc:
Like #3674, but for HP/MP.

Stop displaying collapsible modifiers on Gear Change Panel.

Correctly add subsumed modifiers that are expressions, in addition to non-expressions.

### DIFF
--- a/src/data/modifiers.txt
+++ b/src/data/modifiers.txt
@@ -2459,7 +2459,7 @@ Item	KoL Con IV Pole	Maximum HP / MP: +4, Muscle: +4, Mysticality: +4, Moxie: +4
 Item	KoL Con Six Pack	Muscle: +5, Mysticality: +5, Moxie: +5, Booze Drop: +25
 Item	KoL Con X Treasure Map	Item Drop: +5, Meat Drop: +10
 # Kramco Sausage-o-Matic&trade;: Shows you how the sausage is made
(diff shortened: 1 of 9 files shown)


gausie (approved):
Nice

midgleyc:
Some more context on the TCRSDatabaseTest change:

There were three failing items. One was Vicar's Tutu in Seal Clubber Packrat. For `classify`, it had:
* expStr as Maximum HP Percent: +20, Maximum HP: +50, Maximum MP: +20, Item Drop: +5, Familiar Effect: "1.2xFairy, cap 25", Last Available: "2013-12"
* gotStr as Last Available: "2013-12", Maximum HP: [2*K], Familiar Effect: "1.2xFairy, cap 25", Maximum HP Percent: +20, Maximum MP: +20, Item Drop: +5

So the expression for Maximum HP was passed through from the original modifiers and the extra Maximum HP value silently dropped. As the tests didn't check this case to begin with, I just did the same thing to ensure we skip Maximum HP / MP in this case, but it's a bug in derive, I think.

midgleyc:
Also outstanding from this PR, Drunkula's ring of haze is going to be showing Maximum HP / MP of -200. I think if the two enchants are on opposite sides of zero, we should just show zero for the combination.

eta: fixed.

midgleyc:
Also also I need to go through the non-items and update their modifiers where relevant.

PR #3678, opened 2026-08-31
feat: add All Attributes (Percent) modifiers

midgleyc:
Add two modifiers for "All Attributes" and "All Attributes Percent", removing the collapsible TCRSDatabase code.

Remove 9 explicit enchantment counts. Add 10 more, for items that vary with the moon, like Talisman of Baio.

(ref #3674)


jaadams5 (approved):

gausie:
Why can't the moon-varying modifiers also be expressed with this new modifier?

midgleyc:
If you check the wiki for e.g. "Talisman of Baio" or "The One Mood Ring", you'll see that the Muscle, Mysticality, Moxie attributes appear on different lines, instead of being "All Attributes".

So when we parse the enchantment, we'll think it's separate enchants, instead of All Attributes. As presumably this is one RPN enchant you could bug report it and get CDM to change the output ;)

Jekyllin hide belt has a different way of working entirely where it does display "All Attributes", but that isn't an enchant as far as TCRS is concerned.

PR #3682, opened 2026-09-01
Let JS scripts import the constants we export by default

gausie:
We currently don't allow a top level `const { Item } = require("kolmafia")`, because the original implementation of JS decided to make our constants globals by default so this is as a redeclaration. This was a bad idea, but is also annoying for trying to write JS with `// @ts-check`, so I decided to allow redeclaring them. In my implementation, nothing changes about what is exported by default, shadowing stays local to the script that does it, and `undefined`, `NaN` and `Infinity` are explicity off limits for shadowing.


jaadams5 (approved):
Passes my "pattern recognition" checks but I'm mostly trusting you for correctness :-)

PR #3687, opened 2026-09-02
fix: fixes to checkitems, checkeffects and checkskills

midgleyc:
Also run them and save the updates.

We were interpreting many Hat / Pants enchants as only Hat.


midgleyc:
@gausie Confirm this fix is reasonable or suggest an alternative?

jaadams5 (dismissed):
Approved but since @gausie 's opinion was explicitly solicited I will not commit.

jaadams5 (approved):
Again approved but give @gausie time to chime in.

gausie (approved):

PR #3688, opened 2026-09-02
fix: Remove -osity defaults to prevent unexpected maximizer failure

mark-e-duda:
If a character does not have Torso Awareness, the Maximizer would fail with the default Surgeonosity target. This is due to not being able to equip the Surgical Apron. 

Original proposed solution was to reduce the target to the max achievable.

However, this was just masking the issue of unclear defaults. Instead all -osity defaults were removed, forcing the user to be explicit (and thus responsible for setting an possible target).


jaadams5:
I'm going to say NO.  If someone requests an optimization that has no solution then they should be told and given the chance to change their request.  Since this is caused by a default that is not specified by the user the preferred solution would be to tell them the default is impossible and have them adjust their input.  I'd address this by requiring a maximizer string with surgeonosity to include a weight.

mark-e-duda:
> I'm going to say NO. If someone requests an optimization that has no solution then they should be told and given the chance to change their request. Since this is caused by a default that is not specified by the user the preferred solution would be to tell them the default is impossible and have them adjust their input. 

A fair enough interpretation.

> I'd address this by requiring a maximizer string with surgeonosity to include a weight.

Just remove the default? If so, then I'd propose removing the default for the other "-osity"s to keep consistent.

jaadams5:
> 
> Just remove the default? If so, then I'd propose removing the default for the other "-osity"s to keep consistent.

That works for me.  Might bother a lot of lazy people but IMHO a lot of the problems with maximizer strings come from defaults that people don't realize are being used.

mark-e-duda:
> That works for me. Might bother a lot of lazy people but IMHO a lot of the problems with maximizer strings come from defaults that people don't realize are being used.

Will do. I'll also update the PR title and description.

Edit: @jaadams5 , done.

midgleyc:
It makes sense to fail on clownosity / raveosity because with too low a value you can't fight the clownlord / get into the club.

It doesn't make sense to fail on surgeonosity because any amount of looking like a surgeon is enough.

mark-e-duda:
> It makes sense to fail on clownosity / raveosity because with too low a value you can't fight the clownlord / get into the club.
> 
> It doesn't make sense to fail on surgeonosity because any amount of looking like a surgeon is enough.

I think the distinction is that the -osities aren't really weights like a lot of the other terms; they're targets. So if the user specifies a target and it isn't possible to reach it, then I think it makes sense to call that a failure.

Whether or not they _should_ be targets instead of weights is a different matter

midgleyc:
Yeah, it's weird. And also the user isn't "specifying" a target, and we shouldn't really be using weights for this when we already have "min" (and weights mean something different).

Early instalment weirdness like the beeosity failures.

mark-e-duda:
> Yeah, it's weird. And also the user isn't "specifying" a target, and we shouldn't really be using weights for this when we already have "min" (and weights mean something different).
> 
> Early instalment weirdness like the beeosity failures.

The existence of `min` is a compelling argument to change the way these are handled. I guess the question is how surgical or broad is appropriate.

If we did change them further, would we want them to still be a special case term that's just handled as `min` in the background, or treat them as any other term and allow the full suite of min, max, weights, etc? While I personally like removing special cases (as the latter would do), I can't think of a scenario where you would not want it to behave as a `min`, and it would break compatibility with existing scripts.

midgleyc:
I would prefer surgical for now; gives us more time to consider the design.

What triggered this, anyway? I agree that not failing if the user can't get all 5 items is sensible.

Generally when doing possibly contentious changes like this I:
* have a scan over scripts that might be affected (e.g. search the "loathers" org)
* make sure all features are in before the change goes in so users have a good version to use if they hate it
* merge to main and wait for complaints

I would remove the min for surgeonosity entirely as 1 is enough. For the others, I think we could change the weight to min if none is specified, or we could remove the possibility of failure entirely. I don't think anyone would specify a weight that isn't the default and expect failure on that, that wouldn't be useful.

jaadams5:
As an aside, before we commit this it might be worthwhile to review and update maximizer-help.html.

The documentation says that "clownosity" is the equivalent of "5 clownosity" and "raveosity" is the equivalent of "7 raveosity".

The current, undocumented, behaviour is that "surgeonsity" is equivalent to "5 surgeonosity", correct?

5 surgeonosity is unachievable if the player lacks torso awareness or lacks items.  If the player specifies something unachievable then that should be a failure and reported to the player who would presumably change something.

Given the game mechanics I would think making "surgeonosity" mean "equip as much surgical gear as possible/available" would be useful and acceptable.  Furthermore the surgeonosity clause without a user supplied weight would never trigger a maximizer failure.  A result with surgeonosity of 0 is acceptable.

We could tweak "surgeonosity" to mean "equip as much surgical gear as possible/available and fail if nothing is possible/available".  That is probably my preferred solution in terms of being useful and not breaking much but I don't know how to implement it without iterating over target surgeonosity weights.

midgleyc:
> Furthermore the surgeonosity clause without a user supplied weight would never trigger a maximizer failure.

No user-supplied weight is a weight of 1, so fails if less than 1.

> The documentation says that "clownosity" is the equivalent of "5 clownosity"

This is wrong already, it's either 4 or 100 depending on where we do the multiplication.

jaadams5:
> > Furthermore the surgeonosity clause without a user supplied weight would never trigger a maximizer failure.
> 
> No user-supplied weight is a weight of 1, so fails if less than 1.
> 

Works for me.

> > The documentation says that "clownosity" is the equivalent of "5 clownosity"
> 
> This is wrong already, it's either 4 or 100 depending on where we do the multiplication.

Brain flatulence or typo on my part.  I went to fix it and push a correction and realized it already says 4,  Whoops.

mark-e-duda:
> What triggered this, anyway?

I was adding tests to the Maximizer as part of my Eternity Codpiece work, and this happened to come up as a failure since the default test character doesn't have Torso Awareness. It didn't make sense to me for the default to be a failure.

> No user-supplied weight is a weight of 1, so fails if less than 1.

This is how other weighted terms are handed; no number supplied yields a weight of 1. In this case instead of weight, it'd specify a minimum. Then the existing Maximizer machinery takes over to determine anything beyond that.

My question is whether we completely eliminate the special behavior and make the user specify `min`, and otherwise have it treated as just a score weight. I could make an argument either way, but my instinct is to let the -osities be a little special by treating them mechanically as `min` terms.

midgleyc:
I would say:
* remove the custom weight behavior
* for clowniness, if the user does not specify, set min 100
* for raveosity, if the user does not specify, set min 7
* for surgeonosity, if the user does not specify, set min 1

mark-e-duda:
> I would say:
> 
> * remove the custom weight behavior
> * for clowniness, if the user does not specify, set min 100
> * for raveosity, if the user does not specify, set min 7
> * for surgeonosity, if the user does not specify, set min 1

@midgleyc , Like the following, for example?

| Input              | Equivalent         | Note                                         |
|--------------------|--------------------|----------------------------------------------|
| clownosity         | clownosity 100 min | Bare keyword treated as default min          |
| clownosity 100 min | No Change          | Require 100% clownosity, else fail           |
| 100 clownosity     | No Change          | Clownosity weighted 100x in combined scoring (no special meaning |

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -844,112 +858,118 @@ public double getScore(
...
-        case DAMAGE_AURA:
-          val += mods.getDouble(DoubleModifier.SPORADIC_DAMAGE_AURA);
-          break;
-        case THORNS:
-          val += mods.getDouble(DoubleModifier.SPORADIC_THORNS);
-          break;
+      if (mod instanceof BitmapModifier bitmapModifier) {
+        val = mods.getBitmap(bitmapModifier);
+      } else {
+        var doubleModifier = (DoubleModifier) mod;
```
midgleyc, line 866: do an else if (mod instanceof DoubleModifier doubleModifier) here

it should always be hit but saves a crash if something weird happens

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -393,21 +401,28 @@ private void parse(String expr) {
...
+        }
       }
-
-      if (keyword.equals("surgeonosity")) {
-        // If no weight specified, assume 5
-        this.surgeonosity = (m.end(2) == m.start(2)) ? 5 : (int) weight;
+      if (osityModifier != null) {
+        index = osityModifier;
+        this.weight.put(osityModifier, weight);
+        if (m.group(1).isEmpty() && m.group(2).isEmpty()) {
```
midgleyc, line 423: I think we should break with backwards compat here and add a min if none is specified, including if there is a weight.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -420,9 +420,10 @@ private void parse(String expr) {
       if (osityModifier != null) {
         index = osityModifier;
         this.weight.put(osityModifier, weight);
-        if (m.group(1).isEmpty() && m.group(2).isEmpty()) {
-          this.min.put(osityModifier, defaultMinimum);
-        }
+
+        // Even if the user specified a weight for an -osity, but did not specify a min, then use a
+        // default value.
+        this.min.put(osityModifier, defaultMinimum);
```
midgleyc: we set min up at line 296, does this not need a `containsKey` check to make sure we're not overriding a user-set min?

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -444,8 +444,9 @@ class Clownosity {
     public void clownosityTriesClownEquipment() {
       final var cleanups = new Cleanups(withEquippableItem("clown wig"));
       try (cleanups) {
-        assertTrue(maximize("4 clownosity -tie"));
+        assertTrue(maximize("clownosity 50 min -tie"));
```
midgleyc, line 448: I don't understand how this is passing. We have 1 clownosity, not 50.

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
midgleyc: also, should this be defaultMinimum * weight?

mark-e-duda: The user supplied `min`, by necessity, must come after the index. This default is set in the same pass as the index, so a user set value will supercede it.

To your second point, yes, the default should get multiplied since the min isn't just counting pieces anymore. Wil fix.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
mark-e-duda: Clownosity in `modifiers.txt` is Clowniness and is scored as 0-100, in increments of 25. I didn't want to change that so, the Clown Wig adds 50 Clownosity.

midgleyc: ah right, I forgot we persisted it like that

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
mark-e-duda: Fixed

src/data/defaults.txt
```diff
@@ -1204,7 +1204,7 @@ user	maraisWildlifeUnlock	false
 user	maraisWizardUnlock	false
 user	markYourTerritoryCharges	0	roa
 user	masksUnlocked	0
-user	maximizerList	mainstat | mus | mys | mox | familiar weight | HP | MP | ML | DA | DR | +combat -tie | -combat -tie | initiative | exp | meat drop | item drop | 2.0 meat, 1.0 item | item, sea | weapon dmg | ranged dmg | elemental dmg | spell dmg | adv | pvp fights | hot res | cold res | spooky res | stench res | sleaze res | all res | mp regen | ML, 0.001 slime res | 4 clownosity, -tie | 7 raveosity, -tie | surgeonosity | +four songs
+user	maximizerList	mainstat | mus | mys | mox | familiar weight | HP | MP | ML | DA | DR | +combat -tie | -combat -tie | initiative | exp | meat drop | item drop | 2.0 meat, 1.0 item | item, sea | weapon dmg | ranged dmg | elemental dmg | spell dmg | adv | pvp fights | hot res | cold res | spooky res | stench res | sleaze res | all res | mp regen | ML, 0.001 slime res | 4 clownosity, -tie | 7 raveosity, -tie | 5 surgeonosity | +four songs
```
midgleyc: remove the weights from this, I think

and probably `-tie` to surgeonosity for consistency

src/net/sourceforge/kolmafia/maximizer/Evaluator.java
```diff
@@ -393,21 +401,29 @@ private void parse(String expr) {
...
-      }
-
-      if (keyword.equals("raveosity")) {
-        // If no weight specified, assume 7
-        this.raveosity = (m.end(2) == m.start(2)) ? 7 : (int) weight;
-        continue;
+      BitmapModifier osityModifier = null;
+      double defaultMinimum = 0.0;
+      switch (keyword) {
+        case "clownosity" -> {
```
midgleyc: add "clowniness" here too

mark-e-duda: Done

src/data/defaults.txt
mark-e-duda: Done

midgleyc (approved):
Thanks!

PR #3689, opened 2026-09-02
Update image server URL to use HTTPS

linzinha:
Referencing this report of a fix to mafia's images and js files breaking here:

OK, I figured it out, they moved it from http://images/... to https://images/... KoLmafia is specificially searching for the http:// version.
KoLmafia.java line 173 needs to be changed to https, I assume someone else can push this through a lot faster than I can.

https://discord.com/channels/466605739838930955/581155825385472001/1544834598029234337

PR #3690, opened 2026-09-02
Update KOL image servers

VeeArrKoL:
Adds a new image server that images and other files (e.g. CSS) just started getting served from, and sets the new server as the default. This was breaking the relay browser badly.

Simply adding the new server to the list of image servers fixed a lot of things, but the combat form was still broken for me. Setting it as the default server also fixed that issue.

I'm not sure that this is a total solution, but it seems to at least work for clicking around and doing stuff.

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafia.java
+++ b/src/net/sourceforge/kolmafia/KoLmafia.java
@@ -170,7 +170,8 @@ public abstract class KoLmafia {
       Set.of(
           PREFERRED_IMAGE_SERVER_PATH,
           "https://s3.amazonaws.com/images.kingdomofloathing.com/",
-          "http://images.kingdomofloathing.com/");
+          "http://images.kingdomofloathing.com/",
+          "https://images.kingdomofloathing.com/");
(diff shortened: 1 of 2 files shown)


VeeArrKoL:
This is similar to #3689, but the solution there was not enough to fix everything in my testing.

Opassa:
I just tested these changes, and it fails during combat the same as #3689.  Still trying to track down why it's screwing up the substitutions in the combat UI.

midgleyc (approved):
aight

midgleyc:
We also use it in VolcanoMazeManager and `src/relay/volcanomaze.5.js`. Whether those still work I don't know.

VeeArrKoL:
Yea, I saw that. I'll investigate them post-loop.

VeeArrKoL:
> We also use it in VolcanoMazeManager and `src/relay/volcanomaze.5.js`. Whether those still work I don't know.

I checked both the relay browser solve button and `volcano solve` in the gCLI, and they both seem to still be working fine for now, so I'm inclined to leave them unless it actually breaks at some point.

PR #3692, opened 2026-09-03
feat: add All Resistance modifier

midgleyc:
New method of calculation as items like the perfume-soaked bandana can have both a subsumed and a combined modifier.

Also remove enchantment count for the perfume-soaked bandana.

(ref #3674)


jaadams5 (approved):

VeeArrKoL:
FYI, I noticed today that the constituent modifiers don't display correctly in modtrace. (For example, `modtrace sleaze res` gives confusing and incorrect information if you also have a source of all resistance.)

midgleyc:
I'll look into it, thanks for the report.

PR #3694, opened 2026-09-03
Parse 'Calculate the Universe' from the skills page

libraryaddict:
A recent change by cdmoyer (blessed be) has changed the little known skill [Calculate the Universe](https://wiki.kingdomofloathing.com/Calculate_the_Universe) to include in its description, an indicator of the casts available and the max casts.
image

The first commit is a failing junit test, and the second commit is the solution.

I edited the html of an existing file, because I'm a bit lazy and because it already fits the criteria here.

We do not set the max level when we're either in ronin/hardcore. Or we are under restrictions (avatar path)

It's not spaded if the max level will be 3 or will display just fine as 4+ when we're under restrictions. I cannot test this.

It's unlikely that the max level will continue to show as 3 when out of ronin/hardcore when in an avatar path, I've chosen to do it this way instead of leaving the PR in draft limbo until I can acquire my 4th level.

As part of adding the new "restrict" modifier, as the modifier is not actually a function that accepts parameters, it has no `()`, and because `interact()` sticks out like a sore thumb, that was also corrected to use `interact` instead of `interact()`


VeeArrKoL:
Things that are worth checking: Does this counter behave in some reasonable way while in Ronin/HC? Do we know if the weirdness around using 3 charges and then breaking the prism has been resolved?

libraryaddict:
> Things that are worth checking: Does this counter behave in some reasonable way while in Ronin/HC? Do we know if the weirdness around using 3 charges and then breaking the prism has been resolved?

Yeah I can't check that, I only have level 3.
Worth checking because the html may be different, I don't know.

VeeArrKoL:
As of yesterday, the max charges displayed is limited to 3 while in Ronin/HC. The quirk remains, where using up 3 charges in Ronin/HC and breaking prism does not grant access to the remaining charges. Confusingly, the in-game indicator changes the displayed max charges (so it says e.g. "3/4 casts used today"), despite the fact that it can't be cast again. 

In its current state we probably only want to use it to read the max charges, and only if the user is not in Ronin/HC.

libraryaddict:
image

libraryaddict:
Marked as a draft again because apparently something is bugged on the backend again.

VeeArrKoL:
MalibuStacey reports that as of today this is now working a sane way and that the quirk about 3 in-run uses blocking additional post-prism uses is no longer present. So I think the only caveat now is to not clobber the `skillLevel144` value if 3 max charges is parsed from the skill tile while restricted.

Should also remove the mafia handling for the weird quirk, which was added in #3556.

src/net/sourceforge/kolmafia/request/UseSkillRequest.java
```diff
@@ -1848,6 +1851,15 @@ public static final boolean parseResponse(final String urlString, final String r
           case SkillPool.HEARTSTONE_LUCK -> Preferences.setBoolean("heartstoneLuckUnlocked", true);
           case SkillPool.HEARTSTONE_PALS -> Preferences.setBoolean("heartstonePalsUnlocked", true);
           case SkillPool.HEARTSTONE_BUFF -> Preferences.setBoolean("heartstoneBuffUnlocked", true);
+          case SkillPool.CALCULATE_THE_UNIVERSE -> {
+            // KoL misreports the maximum number of casts on restricted paths
+            if (!KoLCharacter.getRestricted()) {
```
VeeArrKoL: I think this should check `KoLCharacter.canInteract()` instead. The limit applies in Ronin or HC regardless of whether the path is restricted.

gausie:
This is looking good to me

libraryaddict:
I think we could move it into `consequences.txt` but we need to handle restricted as well I believe. Unless someone is in a restricted path and can confirm that they see all their uses (4+) in the skill page.

I did look at setting up the modifier expression, but that's using unicode control markers and I don't know the rules there, especially since it looks vaguely like we "ran out" of control characters. I don't have the mental energy for it atm.

My thoughts are that if we move it into consequences.txt, we can eliminate the code from the .java side, add a "restricted" modifier, then do some math tomfoolery so that if interact() is 1, and restricted is 0, then we can use $2, othewise we fall back to pref(someValue)

VeeArrKoL:
As far as I know there's no difference in behavior for restricted vs. unrestricted, but I don't have enough charges to verify that personally. (I see now I incorrectly said "while restricted" in my previous comment when I actually meant "in Ronin/HC".)

libraryaddict:
Alright, because we have no idea what the maximum displays as when unrestricted, even though I'm 80% it would be fine. I've decided not to think about it, and instead just don't set the skill level when under restrictions. Aka, when in an avatar path.

We add a new expression, because this can be handled in consequences if so.

src/net/sourceforge/kolmafia/ModifierExpression.java
```diff
@@ -134,6 +134,9 @@ protected String function() {
     if (this.optional("interact()")) {
       return "\u0094";
     }
+    if (this.optional("restricted()")) {
```
VeeArrKoL: Recommend `restricted` rather than `restricted()`, since the keyword isn't actually taking any parameters. (`interact()` is inconsistent in this regard, but was left that way the last time I touched it to avoid breaking backward-compatibility.)

libraryaddict: Yeah, fixed `interact()` while I was at it. Also turns out I forgot to add the expression in other places.

src/net/sourceforge/kolmafia/ModifierExpression.java
```diff
@@ -131,9 +131,12 @@ protected String function() {
     if (this.optional("mox")) {
       return "\u0082";
     }
-    if (this.optional("interact()")) {
+    if (this.optional("interact")) {
```
midgleyc:
```suggestion
    if (this.optional("interact") || this.optional("interact()")) {
```

Keep the function around in case some script is using it in `expression_eval(str)`

src/net/sourceforge/kolmafia/ModifierExpression.java
```diff
@@ -131,9 +131,12 @@ protected String function() {
...
       return "\u0094";
     }
+    if (this.optional("restrict")) {
```
gausie: Can we do `restricted` also?

libraryaddict: Alright, `restrict` is renamed to `restricted`

midgleyc (approved):

PR #3681, opened 2026-09-01
Command (checkambiguous) to display encyclopedia types that might be ambiguous.

jaadams5:
Using the Encyclopedia, I discovered that right clicking on some items produced an empty game or wiki display.  (Monster and familiar vampire bat being one).  See #3552.  A general solution to ambiguity was mentioned but not implemented.  Out of curiosity I wrote the command "checkambiguous" to get some idea of how frequent the problem might be.

I chose not to make tests because the associated data files do change and some things might be disambiguated in the future and invalidate the expected results.  I could write a test that cherry picked the output and used things that were unlikely to be changed but that didn't seem especially useful.

I made some implementation choices to keep the focus on Encyclopedia entries and keep my changes somewhat localized.  If this is worth keeping I can imagine exploring making SkillData a class (not a record), not using String for outfit, making an Encyclopedia interface and then collapsing the sections of almost identical code.  I would also explore using canonical names more explicitly.

PR #3695, opened 2026-09-04
REVISED : Command (checkambiguous) to display encyclopedia types that might be ambiguous.

jaadams5:
I screwed up a merge in #3681.  This cherry picks that because my git skills are lacking.

The goal is to identify Encyclopedia "types:" that might be considered ambiguous when looked for on the wiki.

When I switched from #3681 the comment in https://github.com/kolmafia/kolmafia/pull/3681#issuecomment-5532301277 had not been resolved.  This will be a draft until I address it.


gausie:
I don't really understand what the benefit of this is, I barely understood for the last one but was fine since it's a debug command.

jaadams5:
> I don't really understand what the benefit of this is, I barely understood for the last one but was fine since it's a debug command.

This and the previous one are pretty much the same.  I screwed up a merge and ended up with about 16 files being flagged as changed by me when they were already in main.  Restart :-)

I was looking at the Encyclopedia and found entries that did not display a wiki page upon request because the text mafia was passing was ambiguous.  I wondered how many wiki ambiguous pairs there were and this started out as an attempt to list them.

I have since discovered that my brute force case insensitive compare was generating a lot of false positives.  I also discovered there is disambiguateTypes in WikiUtilities.  So my plan is to find ambiguous pairs and flag ones where disambiguateTypes does not produce an unambiguous string for the wiki.  Eventually someone would run this and change disambiguateType appropriately.

That is the benefit/goal above and beyond the fact that it is amusing me.

OK?

jaadams5:
I have learned much but not enough to make this useful to anyone else :-)  Closing

PR #3696, opened 2026-09-04
fix: display combined modifiers in modtrace again

midgleyc:
As reported in #3692, I think this is the fix.


VeeArrKoL:
Confirming that this fixes the issue I was seeing. 👍

jaadams5 (approved):

Issue #3652, opened 2026-08-26
ballpitBonus defaults to 1, under-reporting all three buffed stats by up to 99 percentage points


H0-LawJik:
### Summary

While the `Having a Ball!` effect is active, KoLmafia under-reports all three buffed stats by up to **99 percentage points**, because `ballpitBonus` defaults to `1` and is only corrected by a lazy description scrape that nothing triggers automatically.

Measured on a level 13 Pastamancer, r29189:

| Stat | mafia modifier total | actual | mafia predicted buffed | actual buffed |
|---|---:|---:|---:|---:|
| Muscle | 76% | 175% | 198 | **308** |
| Mysticality | 101% | 200% | 328 | **487** |
| Moxie | 76% | 175% | 228 | **343** |

That is a 36% under-estimate of buffed Muscle, which silently affects the equipment maximizer and any script that reasons about stats.

### Root cause

`modifiers.txt:6791` models the effect as a computed modifier reading a preference:

```
Effect	Having a Ball!	Muscle Percent: [pref(ballpitBonus)], Mysticality Percent: [pref(ballpitBonus)], Moxie Percent: [pref(ballpitBonus)]
```

`defaults.txt:323` sets `user ballpitBonus 1`.

The only thing that ever corrects it is `consequences.txt:133`:

```
DESC_EFFECT	Having a Ball!	\+(\d+)%	ballpitBonus=$1
```

That fires **only when mafia fetches the effect's description**. Nothing does so on login, on gaining the effect, or on `refresh all`. Until the user happens to open that description, mafia believes the ball pit grants 1% when the effect actually reads:

```
Having a Ball!
Muscle +100%
Mysticality +100%
Moxie +100%
```

### Second issue: correcting the pref does not invalidate cached totals

After the description was fetched and `ballpitBonus` became `100`:

- `numeric_modifier($effect[Having a Ball!], "Muscle Percent")` → `100.0` (correct)
- `numeric_modifier("Muscle Percent")` → `76.0` (still stale)

Only an explicit `refresh status` reconciled the character total. Changing `ballpitBonus` should trigger a recalculation.

### Steps to reproduce

1. Have `Having a Ball!` active from a clan VIP ball pit whose bonus exceeds 1%.
2. Without viewing the effect description, run `ash print(numeric_modifier("Muscle Percent"))` — under-reports by (actual − 1) points.
3. `ash visit_url("desc_effect.php?whicheffect=b806c288732675bf60db7b948d748320")`
[...]

PR #3699, opened 2026-09-05
Parse variable effects for Meat Shield Maiden

VeeArrKoL:
When receiving a buff from [%fn, protect me!](https://wiki.kingdomofloathing.com/%25fn,_protect_me!), view the effect description and parse the tier of the buff. Also adds the %ML modifier to the [Singing of your Prowess](https://wiki.kingdomofloathing.com/Singing_of_your_Prowess) effect.

Also provides a partial fix for #3652, where a few similar effects are set up to be parsed when their effect is viewed, but nothing was triggering that view to occur when gaining the effect.


VeeArrKoL:
In retrospect I'm wondering if it would be simpler to just handle these the same way that Heartstone Attunement is handled. It's nice to have the scaling codified in the modifier expressions, but I guess no one's really going to care much what tier of buff they got (vs. what the corresponding modifiers are, which they could get either way).

src/data/consequences.txt
```diff
@@ -160,6 +160,13 @@ DESC_EFFECT	Wine-Dark	\+(\d+) .*? Damage	vintnerWineLevel=[$1/4]
 DESC_EFFECT	Wine-Befouled	\+(\d+) .*? Damage	vintnerWineLevel=[$1/3]
 DESC_EFFECT	Wine-Friendly	\+(\d+) .*? Damage	vintnerWineLevel=[$1/3]
 
+DESC_EFFECT	Blessing of the Heat Shield	\+(\d+)% Combat Initiative	blessingShieldHotTier=[$1/10-2]
+DESC_EFFECT	Blessing of the Riot Shield	Muscle \+(\d+)%	blessingShieldColdTier=[$1/10]
+DESC_EFFECT	Blessing of the Mental Shield	Mysticality \+(\d+)%	blessingShieldSpookyTier=[$1/10]
+# Blessing of the Privacy Shield handled in EffectDatabase.parseBlessingOfThePrivacyShieldEffect()
```
midgleyc, line 166: Can you not parse this from the stench resistance?

midgleyc:
> just handle these the same way that Heartstone Attunement is handled

No, this is right, we use that way only when the modifiers change so much we can't handle it this way.

e.g. even complicated items like Mobius ring or Yearbook Camera display the mods this way if possible

src/data/consequences.txt
VeeArrKoL: No, the stench resistance stops at tier 5, and the combat rate mod goes up to tier 6. (That is, you can have +5 resist and -13% combat rate *or* +5 resist and -15% combat rate.)

midgleyc: ooh

midgleyc (approved):
Thanks!

PR #3700, opened 2026-09-05
Refactor Maximizer for  + Add Eternity Codpiece Support

mark-e-duda:
# Summary

This adds Modifier Maximizer support for alternate Eternity Codpiece gem
configurations and replaces the three-file Maximizer core with named parsing,
candidate-compilation, search, evaluation, and rendering boundaries.

The existing public entry points and recommendation model remain. Search is
unlimited by default; the new deadline remains opt-in.

## Why this change is necessary

The Codpiece has five interchangeable child slots. With 73 gems, exact search
has 21,111,090 canonical configurations even after removing slot-order
duplicates. Correct integration also has to account for:

- gem copy availability and acquisition;
- the Codpiece occupying an accessory slot;
- required and forbidden gems;
- custom-outfit gem state;
- familiar-dependent and derived modifiers;
- modifier caps and non-additive semantics;
- recommendation order and tie behavior.

Adding those branches directly to the previous `Evaluator` /
`MaximizerSpeculation` recursion would make an already coupled core harder to
review and extend.

## Design

The package now has explicit stages:

```text
expression
  -> MaximizerExpressionParser / MaximizerTermRegistry
  -> candidate and availability compilation
  -> specialty candidate selection
  -> EquipmentSearchProblem
  -> AnytimeSearch
  -> MaximizerSession / SolutionQuality
  -> Boost recommendations
```

Key constraints:

- `AnytimeSearch` knows nothing about KoL items, slots, paths, or familiars.
- KoL mechanics remain in candidate/evaluation owners.
- Unsupported Codpiece semantics use full exact modifier evaluation.
- Parent/child slot handling remains local to the Maximizer.

See `src/net/sourceforge/kolmafia/maximizer/README.md` for the class map,
data-flow diagram, invariants, and extension recipes.

## Scope and size

| Metric | Main | This PR |
| --- | ---: | ---: |
| Maximizer package | 5,578 lines / 12 Java files | 8,053 lines / 42 Java files |
| Former three-file core | 5,063 lines | approximately 2,900 lines across focused owners |
| `Maximizer.java` | 1,955 | 990 |
| `Evaluator.java` | 2,269 | 359 |
| `MaximizerSpeculation.java` | 839 | deleted |

The package grows because this includes a new five-slot search mechanic and
substantially more tests. The purpose is not minimizing total lines; it is
making each semantic boundary reviewable and independently testable.


midgleyc:
As you noticed, this is rather too large to review easily.

But we can be pretty sure it mostly works (unless shenanigans are happening) because all the tests pass.

If you're willing to throw tokens at it, expanding our test coverage (of https://app.codecov.io/gh/kolmafia/kolmafia/tree/main/src%2Fnet%2Fsourceforge%2Fkolmafia%2Fmaximizer) to as high as we can get it so we can be more confident an extensive refactor doesn't break anything would be appreciated before embarking on said massive refactor ;)

Doing it one step at a time, e.g. extracting parsing before anything else, would also be appreciated.

Also doing the refactor separately to the feature addition. Normally for a large refactor I would read it commit-by-commit instead of all at once, but there are a hundred commits here.

mark-e-duda:
> As you noticed, this is rather too large to review easily.
> 
> But we can be pretty sure it mostly works (unless shenanigans are happening) because all the tests pass.
> 
> If you're willing to throw tokens at it, expanding our test coverage (of https://app.codecov.io/gh/kolmafia/kolmafia/tree/main/src%2Fnet%2Fsourceforge%2Fkolmafia%2Fmaximizer) to as high as we can get it so we can be more confident an extensive refactor doesn't break anything would be appreciated before embarking on said massive refactor ;)
> 
> Doing it one step at a time, e.g. extracting parsing before anything else, would also be appreciated.
> 
> Also doing the refactor separately to the feature addition. Normally for a large refactor I would read it commit-by-commit instead of all at once, but there are a hundred commits here.

I immediately closed this because I intended to open it against my fork, so I could play with staging/stacking. Asking anyone to review 13k lines over >100 commits is a bit much.

Since I, inadvertently, got your attention, let me ask; is this a desirable change? I've spent a fair amount of time on it just out of my own interest, but before taking it further, I'd like to know if it'd be seen as a worthwhile improvement.

If you didn't see it, I did add a README in the maximizer directory that contains a bit more detail regarding the suggested architecture.

midgleyc:
> is this a desirable change?

as-is, I don't think so: you've got too many classes for it to be easily comprehensible, it seems a bit enterprise-Java-y. "AnytimeSearch" is a "Domain-independent deterministic depth-first anytime branch-and-bound." which by the architecture diagram is used in exactly one place and, as it's part of KoLmafia, could easily and probably should be *not* domain independent. You're not writing a library here.

But the idea of splitting up maximizer seems reasonable. The files are huge and complicated. Although, splitting it up has to trade off against the fact that adding new stuff is currently relatively simple, *because* it's (Evaluator, that is) a single file with a fairly linear chain. For example, consider the "letter" feature I added in https://github.com/kolmafia/kolmafia/pull/491: one addition to the parsing, one to the scoring, store bonus functions on the class itself. It would be possible to make an architecture that was a lot more tightly coupled to modifiers for which something like this would be a lot harder to do.

Also, the more classes you have, and the more they interact, the harder it is to keep the design in your head. Scanning through what you've done, I think your design is harder to reason about than the current design which is mostly linear. Look at Kent Beck's [Rules of Simple Design](https://martinfowler.com/bliki/BeckDesignRules.html) and keep them in mind.

Also, we definitely want more tests before a substantial refactor.

Also, trying to extract parts at a time (e.g. just the scoring) would be preferred to doing the whole thing at once big bang style. Much easier to review that way.

midgleyc:
Some commits are just straight-up beneficial, like https://github.com/kolmafia/kolmafia/pull/3700/changes/7cbc6ce526efd1b2dedd0daa9e643ee767f665c0.

Some commits are just wrong(?) like https://github.com/kolmafia/kolmafia/pull/3700/changes/f1a4c2d8b6d887cfb32bb9981c24b632db883fb0. Why does this one remove part of modifiers.txt?

mark-e-duda:
Addressing out of order:

> Scanning through what you've done, I think your design is harder to reason about than the current design which is mostly linear. Look at Kent Beck's [Rules of Simple Design](https://martinfowler.com/bliki/BeckDesignRules.html) and keep them in mind.
Thank you for the reference. I like this philosophy and I agree with the author that the tricky balance is between the points "Reveals Intentions" and "Fewest Elements". I think my set-point is to prefer many classes, but I've found that many outside my typical domain disagree :).

> as-is, I don't think so: you've got too many classes for it to be easily comprehensible, it seems a bit enterprise-Java-y. "AnytimeSearch" is a "Domain-independent deterministic depth-first anytime branch-and-bound." which by the architecture diagram is used in exactly one place and, as it's part of KoLmafia, could easily and probably should be _not_ domain independent. You're not writing a library here.

I had started with that class as a guiding star. I prefer when top level algorithms are essentially human readable, with domain specific logic hidden, though it has served it's purpose and can be done away with.

> But the idea of splitting up maximizer seems reasonable. The files are huge and complicated. Although, splitting it up has to trade off against the fact that adding new stuff is currently relatively simple, _because_ it's (Evaluator, that is) a single file with a fairly linear chain. For example, consider the "letter" feature I added in #491: one addition to the parsing, one to the scoring, store bonus functions on the class itself. It would be possible to make an architecture that was a lot more tightly coupled to modifiers for which something like this would be a lot harder to do.

You're far more experienced in this domain, so I'll defer to your opinion on how difficult it would be, but in my mind the tradeoff in clarity is well worth it. It may have touched more files and been more literal lines of code, but I don't think that necessarily translates to harder.

> Also, the more classes you have, and the more they interact, the harder it is to keep the design in your head. 
> 
> Also, we definitely want more tests before a substantial refactor.
> 
> Also, trying to extract parts at a time (e.g. just the scoring) would be preferred to doing the whole thing at once big bang style. Much easier to review that way.

Agreed on all fronts. Jumping between numerous classes makes having the whole thing in mind more difficult, though the tradeoff is that it makes the specific responsibility easier to keep in mind.

There's definitely room for improvement in the existing suites, so legacy behavior can be guarded.

I wasn't intending the big bang, thus the immediate closure. Working through the whole thing was just me working through the problem, so I could get a better grasp of the direction I wanted to go with it, and the kind of tendrils it might have. Not the most efficient use of time, but it's a hobby.

> Some commits are just wrong(?) like https://github.com/kolmafia/kolmafia/commit/f1a4c2d8b6d887cfb32bb9981c24b632db883fb0. Why does this one remove part of modifiers.txt?

I saw the asymmetry with Left vs Right Bear Arm, but I didn't mean to keep the change. That was actually reverted later.

@midgleyc , part of my concern with splitting into a bunch of smaller PRs is just a human one; I don't want to annoy people. Despite programming professionally (Hardware Verification is a very different domain, that I found isn't quite compatible with how the Software world does things), I'm very timid as a hobbyist and don't have much experience contributing to projects like this. So, I really appreciate your patience and willingness to help instead of completely shutting it down.

midgleyc:
Responding similarly out of order:

> So, I really appreciate your patience and willingness to help instead of completely shutting it down.

Glad to hear it! I appreciate your contributions in turn. The first I saw was your work on https://github.com/mark-e-duda/goldmine, which I was impressed by, and you managed to turn it into an [oreo](https://github.com/loathers/oreo) PR well, which was also nice. This had me in a more conciliatory mood when you filed your first PR, and I thought you responded well: you reminded me of phulin, from whom we've had many good PRs in the past, and your maximizer optimization work solidified that impression. So I'm very happy to have you contributing :)

> part of my concern with splitting into a bunch of smaller PRs is just a human one; I don't want to annoy people. 

Well, quite the opposite! Many smaller PRs are easier to review, and can be merged piecemeal, each one an improvement to the main codebase, that can be tested by our users with each release and bugs fixed (or a very small amount reverted). A large PR is a large undertaking to review, more likely to have hidden bugs, harder to revert... I also work on [autoscend](https://github.com/loathers/autoscend), and we've had many month-long PRs that I've split up into smaller PRs before merging, to make reviewing more convenient. Large PRs that aren't split tend to drag on.

> Jumping between numerous classes makes having the whole thing in mind more difficult, though the tradeoff is that it makes the specific responsibility easier to keep in mind.

Agreed. After the whole thing is checked for correctness, after it's in the codebase you should only have to look at a single class for a given change (or the classes should be clear, if e.g. you're doing parsing and scoring). But before that, a reviewer has to look at and understand the whole thing. So it's tricky.

> in my mind the tradeoff in clarity is well worth it

I think the issue here is us having completely different intuitions in what is clearer ;)

> I think my set-point is to prefer many classes

I say this with the greatest respect: I can tell ;). Regrettably, I have the opposite preference. I think many classes are definitely useful in a large team where each person can be focused on a small number, but for a smaller team where people have to be aware of the whole context, fewer classes are easier to keep in one's head. And I've mostly worked in the latter setup. Alternatively, that's just a post-hoc justification for what I prefer to see...

> I prefer when top level algorithms are essentially human readable, with domain specific logic hidden

I see, that does make sense. I think if I were more familiar with these algorithms in the abstract, this would make more sense to me as well. Unfortunately, I came to programming not through Computer Science but through Mathematics, so my familiarity with these things comes with a preference for concrete examples, so the algorithm has the business logic in-line and not abstracted.

PR #3702, opened 2026-09-05
Scrap single server session support

gausie:
Sessions are no longer attached to a single server so measuring ping at login and relogging to chase a faster server does nothing. The `ping` command (and `ping()` and thus `pingLatest`) seemed fine to stay. `timein` theoretically is useful otherwise, and `relog` seemed vaguely keepable too, so they also stay.


jaadams5:
> Sessions are no longer attached to a single server so measuring ping at login and relogging to chase a faster server does nothing.

Is that based on logic or experience?

jaadams5 (approved):
I'm fine with the code.

gausie:
> > Sessions are no longer attached to a single server so measuring ping at login and relogging to chase a faster server does nothing.
> 
> Is that based on logic or experience?

Having been reliably informed, so the latter perhaps

jaadams5:
> > > Sessions are no longer attached to a single server so measuring ping at login and relogging to chase a faster server does nothing.
> > 
> > 
> > Is that based on logic or experience?
> 
> Having been reliably informed, so the latter perhaps

OK.

I ran without it this morning and the first check was twice the average.  I'm used to seeing high first pings that were 10x the average 5-7 times a week (out of 21 logins per week).  Today was lower so your source is probably correct.  Thanks.

PR #3704, opened 2026-09-07
Fix: Preserve current equipment after limited Maximizer searches

mark-e-duda:
## Summary

Preserve the character's valid current equipment as a fallback when a
combination-limited or interrupted Maximizer search only evaluates worse
loadouts.

## Problem

Before beginning equipment enumeration, the Maximizer marks its current
speculation as failed so that an equal-scoring candidate can replace it.
This also removes the current equipment as a usable incumbent.

If the search then reaches `maximizerCombinationLimit` before finding an
equally good candidate, the Maximizer can recommend a worse loadout than
the one already equipped.

The regression test reproduces this with an Adventures maximization:

- the current loadout includes a grandfather watch;
- an alternative accessory is available;
- `maximizerCombinationLimit` is 1;
- the limited search finds a 52-adventure loadout while the current
  equipment scores 55.

## Change

Clone the valid current speculation before marking the search incumbent
as failed. After enumeration finishes or is stopped, restore that
speculation if every evaluated candidate scored worse.

This preserves existing behavior when the search finds an equal or
better result. It only changes the fallback when an incomplete search
would otherwise recommend a downgrade.

## Scope

This is a general Maximizer correctness fix. It does not change candidate
enumeration, scoring, ordering, or the meaning of the combination limit.

## Testing

Added `keepsCurrentEquipmentWhenCombinationLimitIsReached`, which verifies
that a limited search:

- does not return a score below the current equipment;
- retains the grandfather watch instead of recommending the inferior
  accessory configuration.

The test fails against the parent behavior with:

- current score: 55
- selected score: 52

### DIFF
--- a/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
+++ b/src/net/sourceforge/kolmafia/maximizer/Maximizer.java
@@ -183,6 +183,7 @@ public class Maximizer {
     Maximizer.boosts.clear();
     if (filter.contains(KoLConstants.filterType.EQUIP)) {
       Maximizer.best.getScore();
+      MaximizerSpeculation currentEquipment = Maximizer.best.clone();
       // In case the current outfit scores better than any tried combination,
       // due to some newly-added constraint (such as +melee):
       Maximizer.best.failed = true;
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks!

This has been a long-standing issue, and we never knew what caused it, but the combination limit makes sense.

PR #3705, opened 2026-09-07
Feat: Handle Codpiece insertion legality in the equip command

mark-e-duda:
## Summary

Treat Eternity Codpiece slots as insertion slots when processing the
`equip` command.

- Allow registered Codpiece gems to be inserted even when the character
  does not meet the gem's ordinary wearable requirements.
- Reject non-gem equipment passed to `codpiece1` through
  `codpiece5`.

## Why

Codpiece gems are inserted into the Eternity Codpiece rather than worn
as ordinary equipment. Applying `EquipmentManager.canEquip` to them can
incorrectly reject a valid insertion because of the gem's normal
equipment requirements.

Conversely, specifying a Codpiece slot previously allowed ordinary
equipment to proceed to the Codpiece choice request. The command now
validates that the selected item is a registered Codpiece gem before
sending any requests.

This change is limited to legality handling in the `equip` command. It
does not change general equipment legality or Maximizer search behavior.

## Testing

- Added an end-to-end command test showing that a valid Codpiece gem can
  be inserted with stats too low to equip it normally.
- Added a regression showing that attempting to insert a hot plate is
  rejected without sending requests.


midgleyc (approved):
Thanks!

PR #3707, opened 2026-09-07
Testing: Add user-facing Maximizer coverage

mark-e-duda:
## Summary

Add a dedicated `MaximizerCharacterizationTest` suite with 97 end-to-end cases that characterize the Maximizer's current user-facing behavior.

The suite covers:

- Modifier aliases and exact expression-parser errors
- Weapon types, hand requirements, and melee/ranged qualifiers
- Positive and negative class-restricted equipment eligibility
- Elemental immunity, vulnerability, and aggregate resistance behavior
- current, -current, empty, and implicit current-outfit handling
- GUI “show all” explanations for unavailable global options, skills, items, familiars, installations, access, and other effect sources
- Preference-controlled suppression of adventure-consuming and exhausted daily effect sources
- Familiar switching and special Hatrack/Scarecrow equipment
- Crown of Thrones and Buddy Bjorn assignment
- Two-item and three-accessory synergies
- Class-, path-, blessing-, and motorbike-dependent effects
- Hardcore path-exclusive equipment
- Closet retrieval commands
- Default tiebreak behavior
- Ranged dual-wielding and Hat Trick slot handling
- Preservation of actual equipment during speculative searches

## Motivation

A planned Maximizer refactor may substantially change how candidate equipment and effects are enumerated, pruned, and compared. Before changing those internals, we need a stronger behavioral baseline so that compatibility can be measured against user-visible results rather than private implementation details.

These tests characterize existing recommendations, commands, errors, and equipment-state behavior. They intentionally preserve current behavior, including potentially surprising behavior, rather than attempting to fix it in the same change.

## Scope

- Adds tests only.
- Tests observable recommendations, commands, errors, and retained equipment.
- Avoids assertions about private candidate lists, search counters, or intermediate speculation state.

### DIFF
--- a/test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
+++ b/test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
@@ -1,0 +1,738 @@
+package net.sourceforge.kolmafia.maximizer;
+
+import static internal.helpers.Maximizer.getBoosts;
+import static internal.helpers.Maximizer.maximize;
+import static internal.helpers.Maximizer.maximizeAny;
+import static internal.helpers.Player.withAdventuresLeft;
+import static internal.helpers.Player.withClass;
+import static internal.helpers.Player.withEffect;
+import static internal.helpers.Player.withEquippableItem;
+import static internal.helpers.Player.withEquipped;
+import static internal.helpers.Player.withFamiliar;
+import static internal.helpers.Player.withFamiliarInTerrarium;
+import static internal.helpers.Player.withHardcore;
+import static internal.helpers.Player.withInteractivity;
+import static internal.helpers.Player.withItem;
+import static internal.helpers.Player.withItemInCloset;
+import static internal.helpers.Player.withOutfit;
+import static internal.helpers.Player.withOverrideModifiers;
+import static internal.helpers.Player.withPath;
+import static internal.helpers.Player.withProperty;
+import static internal.helpers.Player.withSign;
+import static internal.helpers.Player.withSkill;
+import static internal.helpers.Player.withStats;
+import static internal.matchers.Maximizer.recommends;
+import static internal.matchers.Maximizer.recommendsEffect;
+import static internal.matchers.Maximizer.recommendsSlot;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasProperty;
+import static org.hamcrest.Matchers.hasToString;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.startsWith;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import internal.helpers.Cleanups;
+import java.util.EnumSet;
+import net.sourceforge.kolmafia.AscensionClass;
+import net.sourceforge.kolmafia.AscensionPath.Path;
+import net.sourceforge.kolmafia.KoLCharacter;
+import net.sourceforge.kolmafia.KoLConstants;
+import net.sourceforge.kolmafia.KoLmafia;
+import net.sourceforge.kolmafia.ModifierType;
+import net.sourceforge.kolmafia.ZodiacSign;
+import net.sourceforge.kolmafia.equipment.Slot;
+import net.sourceforge.kolmafia.objectpool.FamiliarPool;
+import net.sourceforge.kolmafia.objectpool.ItemPool;
+import net.sourceforge.kolmafia.objectpool.OutfitPool;
+import net.sourceforge.kolmafia.persistence.FamiliarDatabase;
+import net.sourceforge.kolmafia.preferences.Preferences;
+import net.sourceforge.kolmafia.session.EquipmentManager;
+import net.sourceforge.kolmafia.swingui.MaximizerFrame;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class MaximizerCharacterizationTest {
+  @BeforeAll
+  static void resetCharacter() {
+    KoLCharacter.reset("MaximizerCharacterizationTest");
+    Preferences.reset("MaximizerCharacterizationTest");
+  }
+
+  private static void maximizeAll(String expression) {
+    MaximizerFrame.expressionSelect.setSelectedItem(expression);
+    Maximizer.maximize(
+        EquipScope.SPECULATE_INVENTORY,
+        0,
+        PriceLevel.DONT_CHECK,
+        true,
+        EnumSet.allOf(KoLConstants.filterType.class));
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "all resistance | All Resistance: +1",
+        "elemental damage | Cold Damage: +1",
+        "organ capacity | Stomach Capacity: +1",
+        "crit | Critical Hit Percent: +1",
+        "spell crit | Spell Critical Percent: +1",
+        "sprinkle | Sprinkle Drop: +1",
+        "stomach | Stomach Capacity: +1",
+        "liver | Liver Capacity: +1",
+        "spleen | Spleen Capacity: +1",
+        "ocrs | Random Monster Modifiers: +1",
+        "weapon dmg percent | Weapon Damage Percent: +1",
+        "organ | Stomach Capacity: +1",
+        "mys exp perc | Experience Percent (Mysticality): +1",
+        "mys exp | Experience (Mysticality): +1",
+        "mys perc | Mysticality Percent: +1",
+        "mox exp perc | Experience Percent (Moxie): +1",
+        "mox exp | Experience (Moxie): +1",
+        "mox perc | Moxie Percent: +1",
+        "\"item drop\" | Item Drop: +1"
+      })
+  void recognizesModifierAliases(String expression, String modifiers) {
+    try (var cleanups =
+        new Cleanups(
+            withOverrideModifiers(ModifierType.ITEM, ItemPool.HELMET_TURTLE, modifiers),
+            withEquippableItem(ItemPool.HELMET_TURTLE))) {
+      assertTrue(maximize(expression + ", -tie"));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.HAT, "helmet turtle")));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {"utensil | pasta spoon", "knife | asparagus knife", "accordion | aerogel accordion"})
+  void honorsWeaponRequirements(String expression, String itemName) {
+    try (var cleanups = new Cleanups(withStats(100, 100, 100), withEquippableItem(itemName))) {
+      assertTrue(maximize(expression + ", -tie"));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.WEAPON, itemName)));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "type sword | lupine sword | seal-clubbing club",
+        "2 hands | stone banjo | seal-clubbing club",
+        "melee | seal-clubbing club | disco ball",
+        "-melee | disco ball | seal-clubbing club"
+      })
+  void weaponQualifiersRestrictScoredRecommendations(
+      String qualifier, String expected, String alternative) {
+    int expectedId = ItemPool.get(expected).getItemId();
+    int alternativeId = ItemPool.get(alternative).getItemId();
+    try (var cleanups =
+        new Cleanups(
+            withStats(100, 100, 100),
+            withOverrideModifiers(ModifierType.ITEM, expectedId, "Item Drop: +10"),
+            withOverrideModifiers(ModifierType.ITEM, alternativeId, "Item Drop: +10"),
+            withEquippableItem(expected),
+            withEquippableItem(alternative))) {
+      assertTrue(maximize("item, " + qualifier + ", -tie"));
+
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.WEAPON, expected)));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource({"ACCORDION_THIEF, true", "SEAL_CLUBBER, false"})
+  void twoHandedAccordionRequirementRespectsClass(
+      AscensionClass ascensionClass, boolean recommended) {
+    try (var cleanups =
+        new Cleanups(
+            withClass(ascensionClass),
+            withStats(100, 100, 100),
+            withEquippableItem("stolen accordion"))) {
+      assertTrue(maximize("accordion, -tie"));
+
+      assertThat(
+          getBoosts(),
+          recommended
+              ? hasItem(recommendsSlot(Slot.WEAPON, "stolen accordion"))
+              : not(hasItem(recommends("stolen accordion"))));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "Scalp of Gorgolok | SEAL_CLUBBER | TURTLE_TAMER | muscle",
+        "Elder Turtle Shell | TURTLE_TAMER | SEAL_CLUBBER | muscle",
+        "Colander of Em-er'il | PASTAMANCER | SEAL_CLUBBER | mysticality",
+        "Ancient Saucehelm | SAUCEROR | SEAL_CLUBBER | mysticality",
+        "Disco 'Fro Pick | DISCO_BANDIT | SEAL_CLUBBER | moxie",
+        "El Sombrero De Lopez | ACCORDION_THIEF | SEAL_CLUBBER | moxie"
+      })
+  void classRestrictedEquipmentIsRecommendedOnlyToItsClass(
+      String itemName, AscensionClass requiredClass, AscensionClass otherClass, String expression) {
+    try (var cleanups =
+        new Cleanups(
+            withClass(requiredClass), withStats(100, 100, 100), withEquippableItem(itemName))) {
+      assertTrue(maximize(expression + ", -tie"));
+      assertThat(getBoosts(), hasItem(recommends(itemName)));
+    }
+
+    try (var cleanups =
+        new Cleanups(
+            withClass(otherClass), withStats(100, 100, 100), withEquippableItem(itemName))) {
+      assertTrue(maximize(expression + ", -tie"));
+      assertThat(getBoosts(), not(hasItem(recommends(itemName))));
+    }
+  }
+
+  @Test
+  void allResistanceDoesNotScoreIndividualResistanceModifiers() {
+    try (var cleanups =
+        new Cleanups(
+            withOverrideModifiers(ModifierType.ITEM, ItemPool.HELMET_TURTLE, "Cold Resistance: +1"),
+            withEquippableItem(ItemPool.HELMET_TURTLE))) {
+      assertTrue(maximize("all resistance, -tie"));
+
+      assertThat(getBoosts(), contains(hasToString(containsString("nothing useful found"))));
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"Cold", "Hot", "Sleaze", "Spooky", "Stench"})
+  void elementalImmunityAndVulnerabilityAffectResistanceRecommendations(String element) {
+    String expression = element.toLowerCase() + " resistance, -tie";
+    int alternative = ItemPool.get("bounty-hunting helmet").getItemId();
+    try (var cleanups =
+        new Cleanups(
+            withOverrideModifiers(ModifierType.ITEM, ItemPool.HELMET_TURTLE, element + " Immunity"),
+            withOverrideModifiers(ModifierType.ITEM, alternative, element + " Resistance: +99"),
+            withEquippableItem(ItemPool.HELMET_TURTLE),
+            withEquippableItem(alternative))) {
+      assertTrue(maximize(expression));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.HAT, "helmet turtle")));
+      assertThat(getBoosts(), not(hasItem(recommends("bounty-hunting helmet"))));
+    }
+
+    try (var cleanups =
+        new Cleanups(
+            withOverrideModifiers(
+                ModifierType.ITEM,
+                ItemPool.HELMET_TURTLE,
+                element + " Resistance: +200, " + element + " Vulnerability"),
+            withOverrideModifiers(ModifierType.ITEM, alternative, element + " Resistance: +101"),
+            withEquippableItem(ItemPool.HELMET_TURTLE),
+            withEquippableItem(alternative))) {
+      assertTrue(maximize(expression));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.HAT, "bounty-hunting helmet")));
+      assertThat(getBoosts(), not(hasItem(recommends("helmet turtle"))));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "nonsense | Unrecognized keyword: nonsense",
+        "item,, | Unable to interpret: ,",
+        "outfit not an outfit | Unknown or custom outfit: not an outfit",
+        "switch not a familiar | Unknown familiar: not a familiar"
+      })
+  void reportsInvalidExpressions(String expression, String error) {
+    assertFalse(maximize(expression));
+    assertThat(KoLmafia.lastMessage, is(error));
+  }
+
+  @Test
+  void coldPlumberExplainsWhyItCannotRecommendEquipment() {
+    try (var cleanups =
+        new Cleanups(
+            withClass(AscensionClass.PLUMBER),
+            withPath(Path.PATH_OF_THE_PLUMBER),
+            withEquippableItem("work boots"),
+            withEquippableItem("frosty button"))) {
+      assertFalse(maximize("cold plumber"));
+      assertThat(KoLmafia.lastMessage, is("You don't have an appropriate flower to wield"));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource({
+    "SEAL_CLUBBER, Silent Hunting, Nearly Silent Hunting",
+    "TURTLE_TAMER, Nearly Silent Hunting, Silent Hunting"
+  })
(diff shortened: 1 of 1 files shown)


test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+import net.sourceforge.kolmafia.preferences.Preferences;
+import net.sourceforge.kolmafia.session.EquipmentManager;
+import net.sourceforge.kolmafia.swingui.MaximizerFrame;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class MaximizerCharacterizationTest {
```
midgleyc, line 62: after we get to a good state with the review, these tests should likely be folded into the main `MaximizerTest` class

Alternatively this class should have a better name than "MaximizerCharacterizationTest" (i.e. it characterizes... what?) but the test coverage seems so broad I think it should just get folded in.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+        PriceLevel.DONT_CHECK,
+        true,
+        EnumSet.allOf(KoLConstants.filterType.class));
+  }
+
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "all resistance | All Resistance: +1",
```
midgleyc, line 83: all resistance is, after https://github.com/kolmafia/kolmafia/pull/3692, a modifier, not an alias, so this line should be removed

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+      })
+  void weaponQualifiersRestrictScoredRecommendations(
+      String qualifier, String expected, String alternative) {
+    int expectedId = ItemPool.get(expected).getItemId();
+    int alternativeId = ItemPool.get(alternative).getItemId();
+    try (var cleanups =
+        new Cleanups(
+            withStats(100, 100, 100),
+            withOverrideModifiers(ModifierType.ITEM, expectedId, "Item Drop: +10"),
+            withOverrideModifiers(ModifierType.ITEM, alternativeId, "Item Drop: +10"),
```
midgleyc, line 141: give the alternative a higher score to prove we don't consider it?

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquippableItem(alternative))) {
+      assertTrue(maximize("item, " + qualifier + ", -tie"));
+
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.WEAPON, expected)));
+    }
+  }
+
+  @ParameterizedTest
+  @CsvSource({"ACCORDION_THIEF, true", "SEAL_CLUBBER, false"})
+  void twoHandedAccordionRequirementRespectsClass(
```
midgleyc, line 152: two handed? what matters is that it's stolen, e.g. the antique accordion is two-handed.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+    try (var cleanups =
+        new Cleanups(
+            withClass(otherClass), withStats(100, 100, 100), withEquippableItem(itemName))) {
+      assertTrue(maximize(expression + ", -tie"));
+      assertThat(getBoosts(), not(hasItem(recommends(itemName))));
+    }
+  }
+
+  @Test
+  void allResistanceDoesNotScoreIndividualResistanceModifiers() {
```
midgleyc, line 198: similarly to the above comment this tests how the modifier "all resistance" works and should be removed from here

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquippableItem(ItemPool.HELMET_TURTLE))) {
+      assertTrue(maximize("all resistance, -tie"));
+
+      assertThat(getBoosts(), contains(hasToString(containsString("nothing useful found"))));
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"Cold", "Hot", "Sleaze", "Spooky", "Stench"})
+  void elementalImmunityAndVulnerabilityAffectResistanceRecommendations(String element) {
```
midgleyc, line 211: while the implementation values use 100, this feels more like it's testing an implementation detail than the intent of the maximizer, which is to value "vulnerability" extremely low or "immunity" really high if the user is maximizing for resistances, e.g. if you go for "muscle, hot res" and something has +50 muscle, hot vulnerability we score it lower.

using resistances alone the highest item has 5 resistance, so the test doesn't feel reflective of the game

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
VeeArrKoL: Actually, what *is* the expected behavior of `X all resistance` now? Historically in the maximizer it was shorthand for `X cold res, X hot res, X sleaze res, X spooky res, X stench res` (the help still says this), and that's distinctly different from `X All Resistance` (i.e. the new modifier).

Should this be a weird case where `all resistance` retains its old behavior for backward-compatibility (at the cost of not being able to maximize on the actual All Resistance modifier), or should a new shorthand be introduced to replace the old one (maybe `any resistance`?), or should the shorthand be retired altogether?

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+    try (var cleanups = new Cleanups(withClass(ascensionClass), withSkill("Silent Hunter"))) {
+      assertTrue(maximize("initiative, -tie"));
+
+      assertThat(getBoosts(), hasItem(recommendsEffect(expected)));
+      assertThat(getBoosts(), not(hasItem(recommendsEffect(unavailable))));
+    }
+  }
+
+  @Test
+  void showAllExplainsUnavailableGlobalOptions() {
```
midgleyc, line 283: ...in opposition to the top comment, these tests using "maximizeAll" can go in a "MaximizerIncludeAllTest" class, similar to how we extracted MaximizerCreatableTest.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+        getBoosts(), hasItem(hasToString(containsString("no known source of Shadow Waters"))));
+  }
+
+  @Test
+  void noAdventuresPreferenceSuppressesAdventureCostEffects() {
+    try (var cleanups =
+        new Cleanups(
+            withAdventuresLeft(3),
+            withItem(ItemPool.GONG),
+            withProperty("maximizerNoAdventures", false))) {
```
midgleyc, line 347: parameterized test this instead of setting the pref in the middle of the test

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+
+  @Test
+  void dailyUsePreferenceSuppressesExhaustedEffectSource() {
+    try (var cleanups =
+        new Cleanups(
+            withEquipped(Slot.PANTS, ItemPool.GREAT_PANTS), withProperty("_gapBuffs", 4))) {
+      assertTrue(maximize("item drop"));
+      assertThat(getBoosts(), hasItem(hasProperty("cmd", is("gap vision"))));
+
+      Preferences.setInteger("_gapBuffs", 5);
```
midgleyc, line 365: parameterized test this too

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withOverrideModifiers(ModifierType.ITEM, ItemPool.HELMET_TURTLE, element + " Immunity"),
+            withOverrideModifiers(ModifierType.ITEM, alternative, element + " Resistance: +99"),
+            withEquippableItem(ItemPool.HELMET_TURTLE),
+            withEquippableItem(alternative))) {
+      assertTrue(maximize(expression));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.HAT, "helmet turtle")));
+      assertThat(getBoosts(), not(hasItem(recommends("bounty-hunting helmet"))));
+    }
+
+    try (var cleanups =
```
midgleyc, line 225: also feels like this should be two tests, the setup isn't that onerous (and could be in a nested class if you really felt like sharing it)

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+
+      Preferences.setInteger("_gapBuffs", 5);
+      assertTrue(maximize("item drop"));
+      assertThat(getBoosts(), not(hasItem(hasProperty("cmd", is("gap vision")))));
+    }
+  }
+
+  @ParameterizedTest
+  @ValueSource(strings = {"current", "-current"})
+  void explicitCurrentKeywordsKeepEquallyScoredEquippedItem(String currentKeyword) {
```
midgleyc, line 373: given this asserts the same with and without current, I think it's not testing how current actually works

although admittedly I can't get it to do anything different even if I set the combination limit to 1. I thought it was something about items that scored 0 before the tiebreaker, but perhaps not.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: My intent in having these separated out (and what I mean by "characterization") was to differentiate them from correctness tests. While they do display correct behavior, that is secondary to just capturing what the current implementation does, so that we have a basis for a claim of backwards compatibility in future revisions. And if any user facing behavior does change, it will need to be explicitly addressed.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquippableItem(alternative))) {
+      assertTrue(maximize("item drop, -tie, " + currentKeyword));
+      assertThat(getBoosts(), hasItem(hasToString(containsString("keep hat: helmet turtle"))));
+    }
+  }
+
+  @Test
+  void outfitWithoutANameKeepsTheCurrentlyWornOutfit() {
+    try (var cleanups = new Cleanups(withOutfit(OutfitPool.WAR_FRAT_OUTFIT))) {
+      KoLCharacter.recalculateAdjustments();
```
midgleyc, line 389: we have `withAdjustmentsRecalculated`

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquipped(Slot.HAT, ItemPool.HELMET_TURTLE),
+            withEquippableItem(alternative))) {
+      assertTrue(maximize("item drop, -tie, " + currentKeyword));
+      assertThat(getBoosts(), hasItem(hasToString(containsString("keep hat: helmet turtle"))));
+    }
+  }
+
+  @Test
+  void outfitWithoutANameKeepsTheCurrentlyWornOutfit() {
+    try (var cleanups = new Cleanups(withOutfit(OutfitPool.WAR_FRAT_OUTFIT))) {
```
midgleyc, line 388: with no other items that score better it will always keep the outfit anyway?

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+      assertThat(getBoosts(), hasItem(hasToString(containsString("keep hat: beer helmet"))));
+      assertThat(
+          getBoosts(), hasItem(hasToString(containsString("keep pants: distressed denim pants"))));
+      assertThat(
+          getBoosts(), hasItem(hasToString(containsString("keep acc1: bejeweled pledge pin"))));
+    }
+  }
+
+  @Test
+  void negativeFamiliarSwitchAfterPositiveSwitchIsIgnored() {
```
midgleyc, line 401: negative familiar switch after /matching/ positive switch

you can specify a positive switch for a familiar you don't have

and you should add a test for that

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+  @ParameterizedTest
+  @CsvSource(
+      delimiter = '|',
+      value = {
+        "nonsense | Unrecognized keyword: nonsense",
+        "item,, | Unable to interpret: ,",
+        "outfit not an outfit | Unknown or custom outfit: not an outfit",
+        "switch not a familiar | Unknown familiar: not a familiar"
+      })
+  void reportsInvalidExpressions(String expression, String error) {
```
midgleyc, line 249: +outfit while not wearing an outfit should go here too

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquippableItem("snarling wolf shield"))) {
+      assertTrue(maximize("spooky damage, -tie"));
+
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.WEAPON, "lupine sword")));
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.OFFHAND, "snarling wolf shield")));
+    }
+  }
+
+  @Test
+  void recommendsThreeAccessorySynergy() {
```
midgleyc, line 443: all these items have +item drop (5) even without the synergy, so you'll need to add more items with >15 but <45 item drop total to confirm it's actually using the synergy

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withProperty("garbageShirtCharge", 1),
+            withOverrideModifiers(ModifierType.ITEM, alternative, "Experience: +4"),
+            withEquippableItem(alternative),
+            withEquippableItem(ItemPool.MAKESHIFT_GARBAGE_SHIRT))) {
+      assertTrue(maximize("experience, -tie"));
+
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.SHIRT, "makeshift garbage shirt")));
+    }
+
+    try (var cleanups =
```
midgleyc, line 600: similarly split these tests, use a `@Nested` class if you want to associate them

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
```diff
@@ -0,0 +1,738 @@
...
+            withEquippableItem(ItemPool.HELMET_TURTLE),
+            withEquippableItem("bounty-hunting helmet"))) {
+      assertTrue(maximize("item drop"));
+
+      assertThat(getBoosts(), hasItem(recommendsSlot(Slot.HAT, "helmet turtle")));
+    }
+  }
+
+  @Test
+  void defaultTiebreakerPrefersEquipmentWithARolloverEffect() {
```
midgleyc, line 675: ninjammies has higher power and thus higher DA

not 100% sure maximizer counts that as part of the tiebreaker but if it does you'll need to add some more stuff to old sweatpants

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
midgleyc: I honestly can't think of a use-case for "maximize my resistances across all elements", but "any resistance" is fine for that, if we want to keep it

"all resistance" does "highest minimum". So it can pick up something with cold res alone if there are other items offering the other resistances.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: It only matters in that the other class specific weapons are not two handed, so it could theoretically affect the maximizer differently in terms of off-hand processing.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
VeeArrKoL: > "all resistance" does "highest minimum". So it can pick up something with cold res alone if there are other items offering the other resistances.

I'm not sure what you mean by this. Are you saying that's what it *should* do, or that's what the current behavior is (it isn't), or that's what the previous behavior was (it wasn't), or that that's how the new All Resistance modifier behaves (I don't think it does?), or something else?

midgleyc: That's what the current behaviour should be (based on how the modifier works), to my understanding, and that's also how it looks to work where I'm standing.

That's not what the previous behaviour was, which was to sum all the resistances.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
midgleyc: I don't think tests that test implementation details are useful, because if they break that doesn't give us any useful feedback. Rather, these tests are useful because they test the intent of maximizer given various inputs. While we try for backwards compatibility, this is backwards compat of intent. We don't promise bug-for-bug replication.

A test failing is an indicator that your code has broken something. It shouldn't be an indicator that your code still works fine, but the implementation is different.

That is to say, I think these tests *are* correctness tests, and if they're not, they should be. While the intent was to improve coverage, the result is we have a bunch more correctness tests.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
midgleyc: I mean, it's not a two-handed accordion requirement. It's an accordion requirement. There are some accordions which are two-handed (most of them), and some which are one-handed (accord ion, aerogel accordion).

Some accordions are stolen, and can only be equipped by accordion thieves. Some aren't (toy accordion, antique accordion, aerogel accordion). So the reason "stolen accordion" can't be equipped by a Seal Clubber is because it's stolen. That's what I thought this test was testing.

The other weapons aren't class-specific in this way, e.g. seal-clubbing club can be equipped by any class.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
VeeArrKoL: > That's what the current behaviour should be (based on how the modifier works), to my understanding, and that's also how it looks to work where I'm standing.

Perhaps I'm still misunderstanding, but I set one of my resistances significantly lower than the others and maximized on `all resistance` and the only recommendations were equipment and effects that actually have the All Resistance modifier (not anything that would boost only the lowest resistance). However, it did calculate the current score as the value of the lowest current resistance.

midgleyc: Ah, my mistake. I thought it was recommending items without "all resistance", but I was using a character whose mafia version is before the version I added "all resistance" in. My bad.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Apologies for the delay, I've been on vacation and just recently got back.

I agree that implementation details should not cause failures, which is why I'm targeting the user boundary. I'm fine with moving these tests and removing the "characterization" label, since they are correctness tests. (I thought there may be value in the distinction, but it seems it is a distinction without a difference in this case).

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Thank you for the clarification. This will be removed from the test.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Not sure what I was thinking in the moment. You're correct, and I'll rework/rename the test to be "stolen accordion" focused.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Agreed, it'll read easier separated.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Yes, this was vacuous. Will fix.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: I reworked this to show what I think is the intended difference between `current` and `-current`; `current` ensures that the current equipment is explicitly considered a by the maximizer, while `-current` allows it to be pruned. (See currentKeywordControlsWhetherEquippedItemsAreConsidered)

Though this might just be an implementation detail, so I'd be okay removing it if so. Though, that would beg the question; what is the intent?

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Adding as a separate test since the error message depends on character state (rather than being stable like those in this test) and it makes sense as a pair with outfitWithoutANameKeepsTheCurrentlyWornOutfit.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: I've increased the sweatpants DA to overcome the ninjammies' power advantage. Explicit equal bonuses then leave the rollover effect as the behavior being tested by the default tiebreaker.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Understood, the +100 sentinel is a detail. I've reworked the test to reflect realistic values.

mark-e-duda:
@midgleyc I believe I was able to address your concerns. I have two items left open that I want to make sure meet expectation s before marking them resolved.

midgleyc:
Thanks for the updates, I'm ill and brain fogged at the moment but I will try to review this more thoroughly tomorrow.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
mark-e-duda: Post #3744, I've added tests to clearly distinguish the function of "all resistance" vs "any resistance".

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
midgleyc: I'm impressed you managed to find a visible difference!

Yes, I think that's the intent, and also it's an implementation detail. We should keep the test in this PR, which is all about adding tests, and later on we can remove it (or rather, we should change it to be a no-op, so that people's existing maximizer strings work)

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -3666,4 +3676,654 @@ void keepsCurrentEquipmentWhenCombinationLimitIsReached() {
...
+  void negativeFamiliarSwitchAfterPositiveSwitchIsIgnored() {
+    try (var cleanups = withFamiliarInTerrarium(FamiliarPool.BABY_GRAVY_FAIRY)) {
+      assertTrue(maximize("switch Baby Gravy Fairy, -switch Baby Gravy Fairy, item drop"));
+
+      assertThat(getBoosts(), hasItem(hasProperty("cmd", is("familiar Baby Gravy Fairy"))));
+    }
+  }
+
+  @Test
+  void weightedPositiveSwitchCanUseAnUnownedFamiliar() {
```
midgleyc, line 4082: haha, this wasn't what I meant, sorry. Also I forgot we had this functionality.

What I meant is that if you do e.g. `adv, switch Left-Hand Man, -switch Trick-or-Treating Tot`, and you have a Tot but not a Left-Hand Man, it'll pick the Tot. But if you have a LHM and a Tot, it'll always pick the LHM.

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
```diff
@@ -3666,4 +3676,654 @@ void keepsCurrentEquipmentWhenCombinationLimitIsReached() {
...
+            withEquipped(Slot.HAT, "helmet turtle"),
+            withEquipped(Slot.PANTS, "old sweatpants"),
+            withEquipped(Slot.ACCESSORY1, "gold wedding ring"))) {
+      assertThat(KoLCharacter.currentStringModifier(StringModifier.OUTFIT), is(""));
+      assertFalse(maximize("+outfit, -tie"));
+    }
+  }
+
+  @Test
+  void negativeFamiliarSwitchAfterPositiveSwitchIsIgnored() {
```
midgleyc: also this should be matching familiar, and should specify a different familiar (preferably owned) to the one it matches.

test/net/sourceforge/kolmafia/maximizer/MaximizerCharacterizationTest.java
midgleyc: No worries, hope you had a nice time, thanks for the update!

test/net/sourceforge/kolmafia/maximizer/MaximizerTest.java
mark-e-duda: That makes sense, I'll add another test for those scenarios.

midgleyc (approved):
Thanks!

midgleyc:
So from here, I suspect we can / should remove:
* current
* weight > 1 on switch allowing familiars you don't have

I never liked the "includeAll" behaviour (and it's not consistently supported), but I can see how it would be useful.

PR #3709, opened 2026-09-08
Buy interesting coin items one at a time

gausie:
The Spend Your Interesting Coins shop ignores the quantity field, so `buy(coinmaster, n, item)` only ever got one item. Drop the count field, as the star chart already does, and the existing coinmaster code visits once per item.


midgleyc (approved):
Thanks!

PR #3714, opened 2026-09-10
fix: log speakeasy drinks and hot dogs in the session log

ivanlei:
### What

`ClanLoungeRequest.registerRequest` has `case "speakeasydrink"` / `case "eathotdog"` branches that build `drink 1 <name>` / `eat 1 <name>` and write them to the session log — but they were unreachable. The switch dispatches on `GenericRequest.getAction()`, whose pattern is `(?<!pre|sub)action=…`, i.e. it deliberately ignores `preaction=`. The speakeasy, hot dog stand, pool table, shower and swimming pool all submit their real action as `preaction=…`, so `getAction()` returned `null`, `registerRequest` returned early, and nothing was logged.

Net effect: buying a speakeasy drink or eating a fancy hot dog left no trace in the session log beyond the adventures / effect / stats / meat it granted — you couldn't tell *which* drink or dog it was.

This falls back to `getPreaction()` when `getAction()` is `null`. It also:

- fixes the speakeasy case reading capture group `0` (`"drink=5"`) instead of group `1` (`"5"`) — it happened to work via digit-stripping in `parseInt`, but it was wrong;
- decodes HTML entities in the speakeasy drink name, so `glass of "milk"` / `cup of "tea"` / `thermos of "whiskey"` are logged with real quotes instead of `&quot;`.

Side effect: pool games, showers and swim workouts (also `preaction=`) now get their existing `pool …` / `shower …` / `swimming pool …` session-log lines too. Those switch cases were already written; they just couldn't run.

### Testing

New `ClanLoungeRequestTest.RegisterRequest` nested class asserts `drink 1 Bee's Knees`, `drink 1 glass of "milk"`, `eat 1 basic hot dog` and `eat 1 savage macho dog` reach the session log. Full test suite + `spotlessCheck` pass.

### Why

Log-analysis tooling (and anyone reading their own session log) currently has to guess the drink/dog from the effect gained or the stat split. This makes the log self-describing on the same code path that already logs the consumption's results.

### DIFF
--- a/src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
+++ b/src/net/sourceforge/kolmafia/request/ClanLoungeRequest.java
@@ -1999,7 +1999,12 @@ public class ClanLoungeRequest extends GenericRequest {
     String message = ClanLoungeRequest.equipmentVisit(urlString);
 
     if (message == null) {
+      // The speakeasy, hot dog stand, pool table, shower and swimming pool all
+      // submit their real actions as "preaction", which getAction() ignores.
       String action = GenericRequest.getAction(urlString);
+      if (action == null) {
+        action = GenericRequest.getPreaction(urlString);
+      }
       if (action == null) {
         return true;
       }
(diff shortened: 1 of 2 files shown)


midgleyc (approved):
Thanks!

PR #3715, opened 2026-09-10
feat: data-drive monster trackers + expose them to ASH

ivanlei:
### What

Moves `TrackManager.Tracker`'s per-tracker metadata into a new data file, `src/data/monstertrackers.txt`, and adds an ASH function to read it.

**`src/data/monstertrackers.txt`** — one row per tracker:

```
tracker <tab> copies <tab> duration <tab> phylum <tab> reset <tab> ignoreQueue
```

- `copies` — extra encounters the track generates (the max, for the zootomist kicks)
- `duration` — turns the track lasts, or `-1` when it lasts until its reset condition instead of a turn count
- `phylum` — `true` if it tracks a phylum rather than a specific monster
- `reset` — when the track clears (`turn`, `turn-rollover`, `rollover`, `effect`, `avatar`, `avatar-turn`, `avatar-rollover`, `ascension`)
- `ignoreQueue` — `true` if the tracked monster ignores the recent-monster queue

The `Tracker` enum keeps its constants and the behavior that genuinely needs code — the zootomist-kick `getCopies()` override and the `NOSY_NOSE` / `RED_SNAPPER` familiar checks in `isEffective()` — but the plain numbers now load from the file in a static initializer. `Reset.fromString()` maps the tokens back to the enum.

**`monster_trackers()`** — new ASH function returning a `record[int]`:

```
record { string tracker; boolean phylum; int copies; int duration; }
```

`tracker` is the name as it appears in `trackedMonsters` / `trackedPhyla`. Also adds `TrackManager.Tracker.isPhylumTrack()`.

### Why

A script that reads `trackedMonsters` / `trackedPhyla` gets a monster and a tracker name, but no way to ask *how long* that track lasts or *how many* copies it makes — that data lived only in the enum, so external tooling (data-of-loathing, ASH scripts) had to parse the `.java`. This came up in a ChIT PR ([loathers/ChIT#56](https://github.com/loathers/ChIT/pull/56)) adding a "sniffs" brick that wanted to show "23 turns left" for `Offer Latte to Opponent` (30 turns), `Be Superficially interested` (80), etc. The maintainer asked for the data to live in mafia rather than a hardcoded table in ChIT. `$skill[…].dailylimit` already covers cast caps; this covers the rest.

### Testing

- `TrackManagerTest.TrackerData` — durations / copies / queue / phylum load from the file.
- `RuntimeLibraryTest.Tracking` — registry size and `duration` / `phylum` / `copies` for a few trackers.
- `monstertrackers.txt` added to `DataFileMechanicsTest`.
- Full test suite + `spotlessCheck` pass.


VeeArrKoL:
A few notes/questions:
* I don't understand why exposing the information about trackers to ASH requires moving the data into an external file. It seems like the RuntimeLibrary method for it reads the information from the enum, so this shouldn't be necessary.
* Having the fields of the enum be mutable to support this feels like a recipe for problems down the line.
* Why doesn't the ASH representation contain the reset type or queue rejection information?
* It seems like you've condensed TrackType down to a boolean in most representations. Currently there's only two possible values, but what if we new types of tracking are introduced in the future?

midgleyc:
Yeah, I'm not sure we want to do this. Splitting out the trackers into a file would let them be emitted by data-of-loathing, if @gausie wanted to do that, but I think we'd want to swap "ignoreQueue" and "phylum", and also remove the enums in favour of records as we use elsewhere. But also, that isn't necessary for the fundamentals of the feature you want, which is to expose this functionality at all.

On the ASH function, your design is a single function that returns all the trackers as a record. With how records work, we couldn't add any fields to that record without breaking backwards compat (although perhaps we could fix ASH so that works). Instead, you should make many functions, each of which returns a single field of information from a single tracker.

gausie:
I wouldn't put this sort of thing into data-of-loathing in any case. Not to mention, I have tooling to extract data from Java enums which is what powers the Class and Path tables.

midgleyc:
Sweet, that sorts that. Please keep the data in the enum and remove all code related to the external file (and the file itself).

VeeArrKoL:
> With how records work, we couldn't add any fields to that record without breaking backwards compat (although perhaps we could fix ASH so that works).

The issue here is that the return type doesn't have a name, so the caller has to specify the whole definition of the record as the type, right? So if we made some lightweight way to define a named type for library functions, that would avoid the problem (the same way that adding a field to a proxy value type doesn't break backward compatibility)? 

That's probably worth trying to do--having a bunch of separate functions needed to assemble what should be structured data sounds pretty annoying both for mafia maintenance and as a script developer.

midgleyc:
> The issue here is that the return type doesn't have a name, so the caller has to specify the whole definition of the record as the type, right?

Not exactly, but I think that would fix it, at the expense of adding another type-like item. The issue, as I remember it, is that you can't assign a record with fields `a`, `b` and `c` to a record expecting fields `a` and `b` alone. So if more fields were added to the return type of `monster_tracker`, existing scripts would break. But there's no reason why you shouldn't be able to assign a larger record to a subset of its fields, and just drop the ones that aren't present.

ivanlei:
Thanks for the reviews. Sounds like I can regroup and approach my goals a different way.

ivanlei:
Reworked per the feedback — thanks all.

- Dropped `src/data/monstertrackers.txt` and everything that read it. The data stays in the `Tracker` enum; fields are back to `final`, constructors unchanged.
- `TrackManager` gains only two small string accessors (`getResetType()`, `getTrackTypeName()` — the `Reset` / `TrackType` enums are private) plus a `Reset.shortName()` helper.
- Replaced the single record-returning `monster_trackers()` with per-field functions, keyed by the tracker name (the string `tracked_by()` already returns):
  - `string[] monster_trackers()`
  - `int monster_tracker_copies(string)`
  - `int monster_tracker_duration(string)` — `-1` = lasts until a reset condition
  - `string monster_tracker_reset(string)` — `"rollover"`, `"ascension"`, `"turn_rollover"`, …
  - `string monster_tracker_type(string)` — `"monster"` or `"phylum"` (string, not boolean, so a third type stays additive)
  - `boolean monster_tracker_ignores_queue(string)`

Force-pushed. Naming is open to bikeshedding.

midgleyc (approved):
Sure, thanks

PR #3716, opened 2026-09-10
Automatically trigger description reads for unusual events

### LINKED ISSUE #3652: ballpitBonus defaults to 1, under-reporting all three buffed stats by up to 99 percentage points

VeeArrKoL:
Add automatic triggering of description reads to keep mafia prefs (used to calculate certain modifiers) in sync with reality:

* `fossilB/D/N/P/S/W`: When completing combat against one of the creatures that adds a tooth to the necklace, if a tooth is stolen, parse the fossilized necklace's item description to ensure mafia has the prefs set correctly.
* `_disavowed`: In LTA, when adventuring somewhere other than the Villain's Lair before completing the Villain's Lair, parse the Disavowed effect description to get the current penalties.
* `skillLevel227`: When using residual chitin paste, automatically parse the Chitinous Soul skill description to ensure the strength of the bonus is up-to-date. (We already did something similar for the Slime Tube skills.)

Closes #3652

### DIFF
--- a/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
+++ b/src/net/sourceforge/kolmafia/objectpool/ItemPool.java
@@ -1614,6 +1614,7 @@ public class ItemPool {
   public static final int FOSSILIZED_TORSO = 4693;
   public static final int FOSSILIZED_SPINE = 4694;
   public static final int GREAT_PANTS = 4696;
+  public static final int FOSSILIZED_NECKLACE = 4697;
   public static final int IMP_AIR = 4698;
   public static final int BUS_PASS = 4699;
   public static final int FOSSILIZED_SPIKE = 4700;
--- a/src/net/sourceforge/kolmafia/request/FightRequest.java
+++ b/src/net/sourceforge/kolmafia/request/FightRequest.java
@@ -57,6 +57,7 @@ import net.sourceforge.kolmafia.persistence.AdventureSpentDatabase;
 import net.sourceforge.kolmafia.persistence.BountyDatabase;
 import net.sourceforge.kolmafia.persistence.ConsumablesDatabase;
 import net.sourceforge.kolmafia.persistence.DailyLimitDatabase.DailyLimitType;
+import net.sourceforge.kolmafia.persistence.DebugDatabase;
 import net.sourceforge.kolmafia.persistence.EffectData;
 import net.sourceforge.kolmafia.persistence.EffectDatabase;
 import net.sourceforge.kolmafia.persistence.EquipmentDatabase;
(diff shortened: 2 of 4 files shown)


midgleyc (approved):
Thanks!

PR #3719, opened 2026-09-11
fix: Menu item Scripts -> Refresh rebuilds the entire menu

Opassa:
The Scripts menu has an issue where it gets screwed up in various ways.  Sometimes it won't let you scroll at all.  Sometimes the first two menu items are so big that you can't see or click on anything else in the menu.  Sometimes it scrolls a little ways, but the bottom items are unreachable.  I experienced this behavior under Windows 10, and now I experience it every day under Linux Mint, so it doesn't seem to be operating system specific.  It started happening after the Java version upgrade, and has been a plague ever since.

The best way of dealing with it is by hitting Refresh.  This has the effect of re-importing all the scripts listed in the menu.  But this only fixes the menu for a single click.  After refreshing, you get one chance to scroll to the correct script, and then you have to refresh again.

This pull request deals with the issue by completely rebuilding the entire menu, and discarding the screwed up menu, instead of refreshing part of the screwed up menu.  I also moved Refresh to the top of the menu so it will always be accessible.  After working with my changes for a day, it mostly fixes the problem.  Clicking Refresh semi-permanently fixes the Scripts menu, no more "click refresh, click script, click refresh, click script, click refresh, click script".  Of course, the menu will still break eventually, but this change means refreshing every several hours instead of once per script run.

I wish I knew how to fix the menu permanently, but it looks like it's probably a problem in some third party component, darrylbu.util.MenuScroller..  So for now, this is the best I can do.


jaadams5:
Does this look familiar?

GUI

The Scripts Menu started breaking for me back when the Bookmarks functionality, as a menu, was added, so it is probably not a new problem :-)  I would see scripts lists that were too long to view entirely and did not scroll, sometimes lists that had duplicate entries and lists that had more spaces than needed.   It broke enough things that I gave up and started using Bookmarks.  I abandoned the ScriptMRU menu and used the operating system's file chooser when I wanted to select a script.

I never found a solution but always felt there was something called in the build process that was being called more than necessary.  I have no recollection of ever trying to see what would happen if the menu was always rebuild and not just selectively added to.

I am gong to run with this and see whether the ScriptMRU menu is usable again.  (Usable in the sense of the menu display since enabling it has a significant performance cost now and I have toyed with the idea of just ripping it out).

I'll review but not for a day or two.

jaadams5:
I note that the character I enabled scriptsMRU on does not have a Refresh button anywhere in the Scripts dropdown.  That is not what I expected although it could be my expectations are wrong.  It is there on the other characters so this will be something I need to check.

jaadams5:
From today but before a couple of updates that came out in the last couple of hours.

The one with the triangles was after adventuring.  The other was after finishing consumption for the day.  Both from the character with scriptMRU.  

Screenshot 2026-09-11 083852


Screenshot 2026-09-11 084446

jaadams5:
I see that the absence of Refresh on the ScriptMRU menu is deliberate and was a feature before this PR.  Might be worth deciding if that should still be the case or not.

Opassa:
I think it makes sense to always have the refresh, since it's serving a purpose it was not intended for originally.

jaadams5:
> I think it makes sense to always have the refresh, since it's serving a purpose it was not intended for originally.

Fair enough.  It will be useful information to know what happens when I see a "bad" MRU list and then refresh.

What made the scriptsMRU menu less useful was that the GUI always updated when a script was added.  There might be some benefit to considering some kind of model/view where the updates happened to view.  Not sure why I didn't think of that before but I have learned much since then.

jaadams5 (approved):

jaadams5:
This wasn't supposed to fix some of the previously noted issues and I don't think it does, but the Refresh option is a quality of life improvement if nothing else.  I'll push tomorrow if no one beats me to it.

PR #3720, opened 2026-09-11
Named records for RuntimeLibrary types

VeeArrKoL:
In #3715 there has been some discussion of the difficulties involved in using records as the return types of ASH methods in the runtime library. There are several issues:

1) Since the records do not have a name, callers need to define their variable to have the same type as the record signature, which can be cumbersome. For example, 
```
record {string page; int count; int low; int high; int total; int bytes; int average; int bps;} result=ping();
```
This is not only difficult to understand at a glance, but it also has the problem of drifting out of sync if the structure of the return type ever changes (e.g. by adding a field). Having a simple name to use to refer to the record type would both make it easier to comprehend and also prevent type drift between the runtime library and scripts.

2) Because assignment of records requires the source and destination to be structurally identical, any drift between types breaks backward-compatibility, even in cases where the only change is the addition of a field (which an oblivious caller could just ignore if not for the fact that the script now fails to validate). Ideally, it would be possible to assign a record onto a record variable who fields are a subset of the source record's fields.

This PR serves as a proof-of-concept for an idea I had to resolve the first issue and to reduce the number of cases where the second issue is a problem. It's not intended as a full solution in it's current state; instead the intent is to provide something to discuss without diving into too deep of an implementation first.

The main idea is to allow developers using record types as parameter/return values in the runtime library to assign a name to that record type. A `typedef` tying that name to the record type definition would then be included in the top-level scope of an ASH script. For the sake of example, I have given the record used in the `ping()` runtime library function the name `km_ping_result`. With these changes, the calling script can now work with the resulting record type using this name:
```
km_ping_result res=ping();
print(res.page +": "+res.average);
```

Since this isn't actually a new type definition but rather a typedef, the named version of the record and the unnamed version can be used interchangeably (which enables upgrading an existing library function to use a named record without breaking backward-compatibility). These are all valid operations:
```
km_ping_result res=ping();
record {string page; int count; int low; int high; int total; int bytes; int average; int bps;} res2=res;
print(res2.page +": "+res2.average);

record {string page; int count; int low; int high; int total; int bytes; int average; int bps;} res3=ping();
print(res3.page +": "+res3.average);

km_ping_result res4=res3;
print(res4.page +": "+res4.average);
```

I expect that this change would generally drive users to use the typedef'd representation of the record rather than the unnamed representation, which should (going forward) reduce the amount of problems like those from (2) above that will arise. Nonetheless, we'd likely want to implement a real fix for that issue before being too brazen about making changes to the record types. I haven't ruled out the possibility of including that in this PR; I just haven't had a chance to look at it yet.

Finally, as part of this POC I also made a modification to how these named records are represented by the `ashref` command. The function definition uses the named record, but displays the full structure on hover. (I'm open to ideas about how this could be better--this is just a proof-of-concept idea.)

image

Finally, I suspect this could maybe have some impact on how the TS type stuff works, but I really don't know anything about that stuff. Maybe @gausie might have an opinion.

Thoughts? Is this a reasonable idea to pursue further, or are people satisfied with how the current system works?

### DIFF
--- a/src/net/sourceforge/kolmafia/KoLmafiaASH.java
+++ b/src/net/sourceforge/kolmafia/KoLmafiaASH.java
@@ -16,8 +16,10 @@ import net.sourceforge.kolmafia.textui.ScriptRuntime;
 import net.sourceforge.kolmafia.textui.javascript.JavascriptRuntime;
 import net.sourceforge.kolmafia.textui.parsetree.Function;
 import net.sourceforge.kolmafia.textui.parsetree.FunctionList;
+import net.sourceforge.kolmafia.textui.parsetree.Type;
 import net.sourceforge.kolmafia.textui.parsetree.VariableReference;
 import net.sourceforge.kolmafia.utilities.FileUtilities;
+import net.sourceforge.kolmafia.utilities.StringUtilities;
 
 public abstract class KoLmafiaASH {
   private static final HashMap<String, File> relayScriptMap = new HashMap<>();
@@ -249,6 +251,27 @@ public abstract class KoLmafiaASH {
     KoLmafiaASH.showFunctions(RuntimeLibrary.getFunctions(), filter.toLowerCase(), true);
   }
 
+  private static String getTypeString(Type type, boolean addLinks) {
+    String typeName = type.getName();
+
+    int idx = typeName.indexOf("=");
+    if (idx < 0) {
+      return typeName;
+    }
+
+    String displayName = typeName.substring(0, idx).trim();
+    String details = typeName.substring(idx + 1).trim();
+    if (addLinks) {
+      return "<font color='#008000' title='"
+          + StringUtilities.getEntityEncode(details)
+          + "'>"
+          + displayName
+          + "</font>";
+    } else {
+      return displayName;
+    }
+  }
+
   private static void showFunctions(
       final FunctionList functions, final String filter, boolean addLinks) {
     addLinks = addLinks && StaticEntity.isGUIRequired();
@@ -278,7 +301,7 @@ public abstract class KoLmafiaASH {
 
       StringBuilder description = new StringBuilder();
 
-      description.append(func.getType());
+      description.append(getTypeString(func.getType(), addLinks));
       description.append(" ");
       if (addLinks) {
         description.append("<a href='https://wiki.kolmafia.us/index.php?title=");
@@ -296,7 +319,7 @@ public abstract class KoLmafiaASH {
         description.append(sep);
         sep = ", ";
 
-        description.append(var.getRawType());
+        description.append(getTypeString(var.getRawType(), addLinks));
 
         if (var.getName() != null) {
           description.append(" ");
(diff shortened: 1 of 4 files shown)


VeeArrKoL:
To briefly clarify the motivation here (but only briefly; it's late): In my opinion for an interface providing data like the tracker information in #3715, it would be much better to have a single method that returns a structured representation of the tracker's characteristics--this is a perfect use for a record return type. However, the current way that using a record return type works causes problems both in terms of being understandable from a script author POV and in terms of being maintainable/extensible from a mafia developer POV. The aim here is to start moving things in a direction where having structured data returned from runtime library functions is not onerous.

jaadams5:
Is  this OBE if #3755 is committed?  If not then I'm confused about what problem it solves.

VeeArrKoL:
> Is this OBE if #3755 is committed? If not then I'm confused about what problem it solves.

The `var` change definitely covers most of ground that this would have. Personally, I still think it would be helpful to actually have a named "type" to look at (just in terms of making code easier to read/reason about), but the combination of `var` and assign-record-to-similar-record make the usability gap much smaller already. The cases where I think something like this could still be useful include:
* Cases where you want the type to be visible for readability, but don't want the verbosity of the full record definition. (That is, for me, having `km_maximizer_recommendation` or similar as a "type" is a useful readability middle-ground between `var` and `record {string display; string command; float score; effect effect; item item; skill skill;}`.)
* Cases where you might want to pass a record around (because it would simplify the method signature of the parameter/return types).

Of course, any script author can already get around any of these "problems" by using their own typedefs, so this is necessarily only a convenience feature. It may be that given the other changes, the juice isn't worth the squeeze, but I'd still like to hear thoughts on it.

midgleyc:
I'm flip-flopping on this being a good move. Initially I was against because I thought introducing these typedefs would mean new reserved words so that people couldn't define variables named the same as the typedefs, but it turns out that we don't reserve typedefs to begin with so that's consistent at least.

The prefix reduces the chance that someone has named a variable the same as the typedef at the expense of slightly uglier code because the prefix exists.

And, as you say, someone can always define their own typedefs.

But it does make the ashref look nicer. But how does the on-hover behaviour work if someone isn't using the GUI (i.e. they're using the text-only interface)?

VeeArrKoL:
> But how does the on-hover behaviour work if someone isn't using the GUI (i.e. they're using the text-only interface)?

As currently implemented, it just prints the typedef name, but it was pretty easy to tweak `getTypeString()` to display both the typedef name and the record definition instead when in CLI-only mode:

```
 > ashref ping
aggregate get_monster_mapping(  )
aggregate get_monster_mapping( string path )
km_ping_result={string page; int count; int low; int high; int total; int bytes; int average; int bps;} ping(  )
km_ping_result={string page; int count; int low; int high; int total; int bytes; int average; int bps;} ping( int count, string page )
km_ping_result={string page; int count; int low; int high; int total; int bytes; int average; int bps;} ping( string pingTest )
```

(I'll once again stress that everything so far is very proof-of-concept brainstorming. I haven't even checked whether stuffing the name of the typedef in as part of the type name--the way that I have in this sample--is going to break something else somewhere.)

PR #3721, opened 2026-09-11
Tracking preferences for all lucky gold ring drops

Rinn:
- Is it fine _luckyGoldRingVolcoino switches from a bool to int?
- I added tests if I found an existing html that contained a drop
- _luckyGoldRingMeat just tracks activation, not the actual meat amount
- _luckyGoldRingBit combines both 0 and 1, it should probably be split to two prefs but it seems unimportant so I didn't bother


jaadams5 (approved):

midgleyc:
> Is it fine _luckyGoldRingVolcoino switches from a bool to int?

All scripts using it need to be updated to consider that multiple volcoino drops are possible, so I figure it's okay.

"0".to_boolean() is false and "1".to_boolean() is also false, so it does roughly the right thing in practice.

PR #3724, opened 2026-09-11
feat: use modifier names as tags

midgleyc:
Change the modifiers which had separate "tags" (used in data files) to their "name" (used to reference them) to just use the name as the tag.

Look like a Pirate has changed name from "Pirate", which remains as an alias.

Massive amount of changes, mostly TCRS data files.


gausie (approved):

gausie:
A brave change, love it

PR #3725, opened 2026-09-11
feat: allow assigning a record to a record with a subset of its values

midgleyc:
In #3715, there was discussion that record types were somewhat tricky to use if we wanted to change the output of a function in the future.

Allow assigning a record to a type with a subset of its values, in any order. Also allow coercion on e.g. float -> int, which would have been useful when we changed the type of `item_drops_array` a while back ;)

The code was written by GLM and tidied up by me.

Might add some more tests if we're missing coverage.

### DIFF
--- a/src/net/sourceforge/kolmafia/textui/parsetree/ArrayValue.java
+++ b/src/net/sourceforge/kolmafia/textui/parsetree/ArrayValue.java
@@ -61,6 +61,9 @@ public class ArrayValue extends AggregateValue {
 
     if (baseType.equals(valType)) {
       array[index] = val;
+    } else if (baseType instanceof RecordType recordType
+        && val instanceof RecordValue recordValue) {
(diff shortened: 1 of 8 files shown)


VeeArrKoL:
Maybe some tests with typedefs would be good for coverage purposes, but it seems to work as expected with them.

midgleyc:
Found a bug, it says the path <-> int and string -> path conversions are valid but doesn't do them

midgleyc:
On booleans: they don't appear in `validCoercion`. However, they *do* appear in `autoCoerceValue`. So they're in the record coercion just in case this is possible somehow.

jaadams5 (dismissed):
Makes sense but I'm trusting you :-)

midgleyc:
Given that comment, I went through it again and found another bug. I think we're done now.

gausie (approved):

PR #3727, opened 2026-09-12
Track dieting pill charges

VeeArrKoL:
Add tracking for whether any dieting pill charges are active. 

Initially, this PR doesn't aim to add any further support beyond tracking the charges. Perhaps `EatItemRequest.maxmimumUses()` should be updated to reflect the information as well, but beyond that I feel like trying to incorporate the extra fullness more-broadly is going to run into a lot of corner cases or places where it's not obvious what answer is "correct". Any particularly strong opinions?


midgleyc:
A test for using multiple charges at once would probably be a good idea, if it's easy enough?

jaadams5 (dismissed):

VeeArrKoL:
> A test for using multiple charges at once would probably be a good idea, if it's easy enough?

Done

midgleyc (approved):
Thanks!

PR #3728, opened 2026-09-12
Update inflationary coinmaster prices on shop load

VeeArrKoL:
When visiting the Interesting Coin shop, set the internal prices of the inflationary items to reflect the actual current costs. This enables things like `sell_price($coinmaster[Spend your Interesting Coins], $item[homeowner's loam])` to return correct values. Note that when mafia starts (and maybe on ascension?) the prices will be out-of-sync until the shop is visited in the relay browser or `visit($coinmaster[Spend your Interesting Coins])` is called.

Side note: I initially had a more complicated solution that had a pref for each inflationary item that tracked how many had been bought this ascension, but the logic for fixing the pref if it was found to be out-of-sync was a bit annoying, and I couldn't really think of a case where a caller would actually care about how many were purchased (rather than what the cost of the next one is). If anyone has strong feelings that that version would be better for some reason, I can drop it in.


midgleyc (approved):
Thanks!

Yeah this seems good, like we did the wad prices during crimbo.

PR #3729, opened 2026-09-12
feat: many gear drops were changed to hat / pants

midgleyc:


jaadams5 (approved):

midgleyc:
The only outstanding one is Boxing Day Breakfast.

PR #3730, opened 2026-09-13
Start tracking digitized counters properly

libraryaddict:
## Background

This PR was originally intended to correct how Mafia starts the Digitize and Romantic Arrow counters, bringing them more in line with how KoL tracks the monsters.

However, the way KoL was starting those counters was bug reported and has since been fixed. This means the changes KoLmafia was making to account for that behaviour are no longer needed.

## Testing

I checked the behaviour in-game, and it now appears to be consistent with how other wanderers work.

I digitized a monster, did a free fight, and then encountered the digitized monster 8 turns later, which is consistent with normal wanderer behaviour.

Garbo broke, as expected, because the bugfix meant KoLmafia's tracking was now "one turn too soon."

Further testing could be done, but it feels fairly unnecessary at this point. I’ll mark this PR as ready if garbo gets through the rest of the day without complaining about Digitize while running with my changes.

No junit tests added, as this is already covered under existing behavior and tests.

## Notable Changes

* 1 was added to the counters used internally, which is consistent with wanderer counters
* Counter now starts on the same turn as digitize
* Removed the old temp pref, which was only used for fixing the inconsistent tracking

## User Impact

Currently, the only difference between this PR and how KoLmafia already handles things should be noticeable to the end user if they are doing free fights on the Digitize turn or the turn immediately after.

Otherwise, this PR will appear to have no effect.

### Notes

Wrote the PR this way because I'm feeling kind of dull atm, and I'd rather not stress. 

I had no further issues running garbo for the day.

## Historical

<details> 
  <summary>Original PR content</summary> 

So garbo is failing and it's because of digitize monsters counter being wrong.

Going by what was said in discord, as well as what's on the wiki pages for [Digitize](https://wiki.kingdomofloathing.com/Digitize) and [Wandering Monsters](https://wiki.kingdomofloathing.com/Wandering_monsters), the counters should only start when

1. We enter a combat
2. The combat is not against a no-wander monster
3. The monster is not against a superlikely monster

If all of the above pass, then we start the counter.

Currently, kolmafia is starting the counter unless we are fighting a no-wander monster. There is no checks for combat, no checks for superlikely.

Now, I have no idea why kolmafia is doing this, because [it looks deliberate.](https://github.com/kolmafia/kolmafia/commit/dc43c86d26f5e8ff0c24fa6130d8ece8504cac59#diff-0302978e7b60472b10e21b8105f2f66ec4aef6ad5452070e9e57efc45632a8d6R515)

I'm choosing to trust the wiki here. There's a few locations missing in kolmafia that the wiki has for no-wander zones, but no one cares about the barracks or goo fields. I don't anyways.

Marking this as a draft for a bit (probably just a few days at most) unless someone can point out why kolmafia is correct and the wiki is wrong. Because it's really weird as an oversight.

Failing tests first. Fix second.
</details>


libraryaddict:
Actually, I have no clue how this works for superlikely wandering monsters. Eg, club em into next week.

Unless someone wants to test it, I'll be removing it from the PR before marking this as ready, leaving the PR with just a change to ignore non-combat encounters.

Really depends doesn't it, if a normally superlikely will **not** start the counter despite it being setup as a wanderer. I cba going through the encounter priorities.

libraryaddict:
This is apparently fixed by kol, praise be they. Just needs to be spaded, then this PR can either be closed, or refactored into a PR  to use the same counter logic as "club 'em into next week" for example.

midgleyc (approved):
Thanks!

PR #3731, opened 2026-09-13
Fix -osity default minimum incorrectly scaling with assigned weight

VeeArrKoL:
When a keyword expression includes a `min`, that requires the *value* of the corresponding modifier to be at least that much, regardless of the assigned weight. (For example `0.1 hp, 100 min` requires only at least 100 HP, not the 1000 HP that would be required for the score to be at least 100.)

If a weight was specified for an -osity modifier but no `min` was specified, the default minimum was being multiplied by the weight, in some cases creating impossible minimums.

Also patch up some out-of-date things in the maximizer help.


midgleyc (approved):
Drat, I asked for it to be changed because I thought it worked the other way!

Thanks!

midgleyc:
(ref #3688)

PR #3732, opened 2026-09-13
Fix X-in-the-box free crafting priority

VeeArrKoL:
Recently, the free crafting priority changed to move chef-in-the-box *et al* to the bottom of the free crafting priority rather than the top. As a result, mafia was incorrectly tracking chef-in-the-box uses when another free crafting method was actually being consumed.


src/net/sourceforge/kolmafia/request/concoction/CreateItemRequest.java
```diff
@@ -601,15 +601,16 @@ public static int parseCrafting(final String location, final String responseText
...
-          Preferences.increment("bartenderTurnsUsed", qty);
+        if (mode.equals("cook")) {
+          fancyType = 1;
+        } else if (mode.equals("cocktail")) {
+          fancyType = 2;
         }
       }
 
-      craftComments.add(new Integer[] {m.start(), qty});
+      craftComments.add(new Integer[] {m.start(), qty, fancyType});
```
midgleyc: make `craftComments` a list of records, something like `CraftComment(int index, int quantity, FancyCrafting fancy)`, with the last being `FancyCrafting.NONE`, `FancyCrafting.COOK`, `FancyCrafting.COCKTAIL`.

midgleyc (approved):
Thanks!

Those names are much better than what I suggested :)

PR #3733, opened 2026-09-13
Special formatting for legendary IOTM weapons no longer considered for Swordosity

VeeArrKoL:
ref: https://forums.kingdomofloathing.com/vb/showpost.php?p=5118240&postcount=79

As far as I can tell, there's no other display names that change the number of S words, but if there ever is one we'll need to check whether this exception is only for these or what.


midgleyc (approved):
Thanks!

PR #3734, opened 2026-09-14
Sea floor zone availability

VeeArrKoL:
Adjusts the `canAdventure()` for sea floor zones to correctly match when the zones are available.


midgleyc (approved):
Thanks!

PR #3735, opened 2026-09-14
fix: stop at 100 clownosity

midgleyc:
(ref https://github.com/kolmafia/kolmafia/pull/3688)

Keep behaviour same as previous if no specified max.


VeeArrKoL:
Should add the default maxima to the maximizer help file

jaadams5 (approved):
I presume the change to Player is to help in catching errors when writing tests?  After a couple of self-inflicted wounds I started using ITEMPOOL :-)

midgleyc:
correct, I had "clown buckler" at first and the test was unexpectedly passing, which confused me

PR #3738, opened 2026-09-14
Update to use new mallprices api

libraryaddict:
As of earlier today, `api.php` has support for `mallprices`, allowing for bulk mall price data without having to perform as many requests. So, faster for the player, lighter on the backend.

> mallprices
>     The cheapest listings of every item in a Mall search category, in one response: pass 'category' (allitems, food, booze, othercon, weapons, hats, shirts, container, pants, acc, offhand, famequip, combat, potions, hprestore, mprestore, new, mrstore, unlockers, familiars; default allitems), optionally 'tiers' (comma-separated crappy,decent,good,awesome,EPIC, for food and booze) and 'count' (listings per item, default 5, max 20). Each item has id, name, descid and its stores (id, name, price, quantity, limit, bought -- your purchases from that store today). Pass 'fields' to get only the fields you need (comma-separated; item fields are id, name, descid; store fields are store.id, store.name, store.price, store.quantity, store.limit, store.bought; 'stores' means every store field), e.g. fields=id,store.id,store.price,store.limit,store.bought. Results are shared and up to two minutes old; please use this instead of paging through mall.php for bulk prices.

As part of this PR, I have

1. Removed the `json` field from ApiRequest

That field was only used in two places, neither of which gained much from it. It was also incorrect because it assumes every response would be a JSON object, when `events` returns an array.

2. Added basic support for parsing multiple `what`

Supplying multiple `what` parameters `what=storage,inventory` is valid API usage. I plan to support this more fully in a follow-up PR, so I intentionally left that out.

> what
>     What you are requesting, options include: (you can pass multiple at once, too. status,inventory) 

The alternative fix for the PR would have been simpler, but it would introduce code that would immediately need to be replaced for a following PR. I do not see much gain in writing something that would immediately need to be replaced for a followup PR.

Without some implementation, the changes to `what` would look unnecessary here.


src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -692,13 +697,74 @@ public static int getMallPrices(String category, String tiers) {
...
+
+    List<PurchaseRequest> results = new ArrayList<>();
+
+    for (int i = 0; i < items.size(); ++i) {
+      var item = items.getJSONObject(i);
+      int itemId = item.getIntValue("id");
+      String itemName = item.getString("name");
+
+      // Unknown item
+      if (!itemName.equals(ItemDatabase.getItemDataName(itemId))) {
```
libraryaddict: Need to double check this is the correct usage, re-reading it, it feels odd. Why can't we check by item ID?

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -692,13 +697,74 @@ public static int getMallPrices(String category, String tiers) {
...
+        ItemDatabase.registerItem(itemId, itemName, item.getString("descid"));
+      }
+
+      var stores = item.getJSONArray("stores");
+      // Should never be empty
+      if (stores.isEmpty()) {
```
libraryaddict: Comment is wrong, if there was only one item in mall, then yes it'd be empty. But, I doubt the api would ever return an empty array. So, probably harmless? Though may be a better idea to just remove this check altogether, if the game wants to say no stores explicitly for an item id, maybe we should honor that

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -692,13 +697,74 @@ public static int getMallPrices(String category, String tiers) {
...
+        continue;
+      }
+
+      results.addAll(NPCStoreDatabase.getAvailablePurchaseRequests(itemId));
```
libraryaddict, line 714: I personally think this is weird, but, it's inline with how kolmafia does things.
We don't do the coinmaster, as that's filtered out in the following method.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -692,13 +697,74 @@ public static int getMallPrices(String category, String tiers) {
...
+        boolean canPurchase = limit == 0 || store.getIntValue("bought") < limit;
+
+        results.add(
+            new MallPurchaseRequest(
+                itemId,
+                quantity,
+                store.getIntValue("id"),
+                store.getString("name"),
+                store.getLongValue("price"),
+                limit == 0 ? quantity : limit,
```
libraryaddict, line 731: Need to double check this, it'd be "amount purchasable" or some other variant, which makes this feel odd without context. Feels like it should be Math.min

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -726,6 +792,7 @@ public static int getMallPrices(String category, String tiers) {
       MallPriceDatabase.writePrices();
     }
 
+    MallPriceManager.pricesUpdated = count;
```
libraryaddict, line 768: Done this way because there's no standard for request.run().getResult(), its all statics.

test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
```diff
@@ -782,6 +784,86 @@ public void canGetMallPricesByCategory() {
...
+        assertEquals(0, MallPriceManager.getMallPrices("booze", "awesome,EPIC"));
+        assertPostRequest(
+            builder.client.getRequests().getFirst(),
+            "/api.php",
+            "what=mallprices&for=KoLmafia&category=booze&tiers=awesome,EPIC");
+      }
+    }
+
+    @Test
+    public void honoursPerStoreLimits() {
```
libraryaddict, line 785: This test, I'm unsure. It proves it parses items and finds limits. But it feels like a waste of breath.

libraryaddict (commented):
I'm a bit unsure about the allitems.json, it's several mb and we really don't need that many items in the test.
We can probably just test on a smaller subset, likely by pruning it down, or category filter. Pruning means it's not vanilla, category means its natural but not "allitems". I'll likely change it over later, after I finish with the other PR

Edit: Leaving the large json file as is, unless someone else wants to weigh in, I'm inclined to leave it as regression. The unfortunate effect is that it's a larger file for no real gain at this point.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
libraryaddict: Alright, so this is _correct_ but it's also weird.
Leaving it as is as it's inline with normal kolmafia.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
libraryaddict: This was indeed wrong

test/net/sourceforge/kolmafia/session/MallPriceManagerTest.java
libraryaddict: Leaving as is, for regression I guess

libraryaddict:
Yeah for `searchmall` command, if the api allowed us to provide a list of item IDs to check, it'd be in scope. But it doesn't, and it would be slower.

That command has confusing output regardless for multiple items.

midgleyc:
Yeah, keep the full file: we're big on using test fixtures from the actual game; suspect we have several megabytes worth already. None of it should go into the actual jar, it's just test code.

libraryaddict:
I can understand some reluctance about this PR, because although some changes in it are good, other changes, notably the new field, are questionable. But I don't really see what a good alternative would be there.

The `pricesUpdated`, can move it to ApiRequest where it would belong more, though would then be making it public instead of private.
Honestly, it's a case where it is serving as a workaround where there's only two ugly solutions. I'm not really a fan of having statics exposed everywhere. I'll make that change though if anyone else thinks it should be done.

Also possible is this PR doing further work to migrate away from bulk mall requests, but I saw that as extra work that hits a smaller subset of players.

midgleyc:
Don't take my slight delay in reviewing as reticence; I'm ill at the moment and avoiding reviewing PRs that are too complicated.

I think it's a good move and will review in time (or someone else will).

libraryaddict:
> Don't take my slight delay in reviewing as reticence; I'm ill at the moment and avoiding reviewing PRs that are too complicated.
> 
> I think it's a good move and will review in time (or someone else will).

Oh I didn't mean to imply as such, I'm just critical of my own changes

src/net/sourceforge/kolmafia/request/ApiRequest.java
```diff
@@ -129,6 +130,17 @@ public static synchronized String updateStorage(final boolean silent) {
     return ApiRequest.STORAGE.redirectLocation;
   }
 
+  public static void updateMallPrices(final String category, final String tiers) {
+    ApiRequest request = new ApiRequest("mallprices");
+    request.addFormField("category", category);
+    if (!tiers.isEmpty()) {
+      request.addFormField("tiers", tiers);
+    }
+    // Default of 5, max of 20.
```
jaadams5: Perhaps I am missing context but what has a default of five and a max of 20?

I'm doing the review sequentially but if the answer is the second parameter (MallPriceManager.NTH_CHEAPEST_COUNT)) then this is at least superficially confusing because that is a constant that is set elsewhere and the concept of "default" is at odds with being a constant.

jaadams5 (dismissed):

jaadams5:
@midgleyc - Get Well Soon :-)

I will push this tomorrow if no one else beat me to it.

Adding the extra field doesn't bother me although perhaps it should :-)

The size of the test file is not a problem for me.  In general if we are going to write tests on files that the game generates I would prefer that the test file be unmodified from what the game sends.  If, in the future, the game changes what it sends then we can just drop in the new file and test.  If the test fails then we are pretty sure mafia code has to be changed.  But if we edit the file and a test fails then we have to confirm that our edit was not the cause.

I will ask - is the file sent to the user always going to be this big ?

src/net/sourceforge/kolmafia/session/MallPriceManager.java
libraryaddict: I'm wrong, stores can be empty.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
libraryaddict: Not too sure I understand the logic of coinmaster.
So, in the original mall search code we are adding all coinmaster items to the search results.
When it's passed to the following function, we early exit if the list is empty. If a list has a coinmaster item, we don't.
So, this path only does something if the list contains only coinmaster items.

So, if a list only contains coinmaster items, then this pathway will clear the prices from the database because the final result will have no coinmasters and no stores.
bonerang is an empty of no stores.  So, at mall max.

Which means I need to be adding coinmasters.

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: The api returns a default of 5 stores per item listing, up to a max of 20 stores.
It's possible that setting it to higher than 20 may behave unexpectedly, but, that's a modified mafia and they undertake their own risk.

We currently show the price of the 5th cheapest item in mall, as there could be 5 stores that all sell 1 item each, this seems, preferable?
We could increase this higher to have more data, as the current bulk mall search returns 6 stores, not 5. But it does come at the cost of larger requests, and I'm not sure it matters for the majority of usecases that they can't make a second request for a specific item.

libraryaddict:
> Adding the extra field doesn't bother me although perhaps it should :-)

Assuming you are referring to `count`, it's included because although it's listed as optional in the api, we use for `mallPrice` the cheapest X item, which is the constant.

I figure we should probably request exactly what we want.

Though, the original request is effectively a default of 6. But it largely only matters for wellStocked?, and I'm not sure the extra store in the results is worth the overhead.

> I will ask - is the file sent to the user always going to be this big ?

Yes, it's because the store names are included in the response for the most part, followed by the item data.

The API doesn't seem to allow skipping it, and, if we did skip it, it would require some internal changes which as far as I can see, mostly cosmetic. But not too relevant as we can't filter the data out in our request.

The network response is, 950kb or 750kb, which I forget. The data outside of gzip is 4.X mb currently.
So yes, the request is slow, and likely to be worse for users on poor internet.

Could make a request to TPTB to expand the API to allow filtering of the data, I suppose, and maybe to request the cheapest X items instead of cheapest X stores. Not sure, I have a headache atm.

Removing the store name from the request, would need us to

1. Stop doing equality checks on the store name
2. Skip the store name from being written to log, which only matters for those who parse the log, but they can update their parsers.

I don't think it really impacts anything tbh, other than the session log. We already can use the player id. But it also seems like unneeded complexity.

src/net/sourceforge/kolmafia/request/ApiRequest.java
```diff
@@ -129,6 +130,17 @@ public static synchronized String updateStorage(final boolean silent) {
...
 
+  public static void updateMallPrices(final String category, final String tiers) {
+    ApiRequest request = new ApiRequest("mallprices");
+    request.addFormField("category", category);
+    if (!tiers.isEmpty()) {
+      request.addFormField("tiers", tiers);
+    }
+    // The api returns a subset of stores per item, with the default 5 and the max 20
+    request.addFormField("count", String.valueOf(MallPriceManager.NTH_CHEAPEST_COUNT));
+    request.run();
```
gausie: can we RequestThread.postRequest so that errors get logged properly? I think this will currently NPE if the JSON response is invalid (rollover or whatever) which is fine but should be logged here.

src/net/sourceforge/kolmafia/request/ApiRequest.java
```diff
@@ -129,6 +130,17 @@ public static synchronized String updateStorage(final boolean silent) {
...
   }
 
+  public static void updateMallPrices(final String category, final String tiers) {
+    ApiRequest request = new ApiRequest("mallprices");
+    request.addFormField("category", category);
+    if (!tiers.isEmpty()) {
+      request.addFormField("tiers", tiers);
+    }
+    // The api returns a subset of stores per item, with the default 5 and the max 20
+    request.addFormField("count", String.valueOf(MallPriceManager.NTH_CHEAPEST_COUNT));
```
gausie, line 157: This is not what NTH_CHEAPEST_COUNT means right? There's a comment in MallPriceManager explaining this, but we usually select more so we can discount forbidden/ignored/etc stores etc

src/net/sourceforge/kolmafia/session/MallPriceManager.java
```diff
@@ -692,13 +682,65 @@ public static int getMallPrices(String category, String tiers) {
       return 0;
     }
 
-    // Issue the search request
-    MallSearchRequest request = newMallSearchRequest(category, tiers);
-    RequestThread.postRequest(request);
+    // api.php doesn't return the changes directly, so we use a helper field
+    MallPriceManager.pricesUpdated = 0;
+    ApiRequest.updateMallPrices(category, tiers);
```
gausie: Because the old tiers handling used to use contains, you used to be able to do like `mall_prices(..., "EPIC awesome")` and now that would just go straight to live.

gausie:
> allitems, food, booze, othercon, weapons, hats, shirts, container, pants, acc, offhand, famequip, combat, potions, hprestore, mprestore, new, mrstore, familiars; default allitems

Did unlockers get removed?

src/net/sourceforge/kolmafia/request/ApiRequest.java
```diff
@@ -519,6 +533,10 @@ public static final void parseStorage(final String responseText) {
     StorageRequest.parseStorage(ApiRequest.getJSON(responseText, "storage"));
   }
 
+  public static final void parseMallPrices(final String responseText) {
+    MallPriceManager.parseMallPrices(ApiRequest.getJSON(responseText, "mallprices"));
```
gausie: I think we're parsing the JSON twice here, as it also gets parsed into `this.json` later. It's a big blob so maybe we should fix that.

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: My comment isn't wrong, as it's referring to kol's api and not mafia.
But, I was presuming that we're including the 5th cheapest item, and forgetting about stores we're skipping. Eg, presuming that we want to return 5 stores, because each store has a limit / stock of 1.

src/net/sourceforge/kolmafia/session/MallPriceManager.java
libraryaddict: Ah yep, and the backend is doing a case sensitive equality check

libraryaddict:
> > allitems, food, booze, othercon, weapons, hats, shirts, container, pants, acc, offhand, famequip, combat, potions, hprestore, mprestore, new, mrstore, familiars; default allitems
> 
> Did unlockers get removed?

It was fixed, I didn't update my comment.

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: Oddly this isn't the case. `processResults` is called by `GenericRequest.processResponse()` which has an early exit for `api.php`.

`processResults` appears to be dead code only called by junit tests.

I could be wrong!

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: Ah yep, forgot it's not actually a thread..

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: https://github.com/kolmafia/kolmafia/pull/3738/changes#diff-fbd69a5126a319134c4b122ae800f33eb2cb518794cba8cb17e973a07a4d9fc1L95

It looks like the existing implementation is also limiting the results to 5.
I might have misunderstood what you were trying to say sorry.

src/net/sourceforge/kolmafia/request/ApiRequest.java
gausie: Huh how weird

libraryaddict: I think it's for the case we call api.php outside of ApiRequest, such as a relay request and the like.

libraryaddict:
I delayed this a bit because I considered loading unknown items by `descid` when their item ID isn't known.

That seemed unnecessarily complicated. If an item is relevant, the user will likely already own it or encounter it naturally. Requesting every unknown item for outdated clients also seems unwanted from kol's perspective. This isn't an item database.

This PR does support loading unknown items by `descid`, but KoLmafia won't request that field itself. If a response already includes a `descid`, however, KoLmafia can use it to load the unknown item.

src/net/sourceforge/kolmafia/request/ApiRequest.java
jaadams5: How about you expand the comment with portions of what you just said here?

jaadams5:
Would you consider marking comments that you think are resolved as resolved?  I tend to defer reviewing something again if there are unresolved comments from me or others.

src/net/sourceforge/kolmafia/request/ApiRequest.java
libraryaddict: Rewritte to
    // The backend api returns a subset of X cheapest stores per item, defaulting to 5 when 'count'
    // is omitted, with a maximum of 20.
    // KoLMafia requests N, as we resolve mall prices based on the Nth cheapest listing.

libraryaddict:
The PR was breaking when `api.php` requested `events`. This is a valid request, even though KoLmafia itself does not use it. A script can still call it as I did.

As part of this PR, I have

1. Removed the `json` field from ApiRequest

That field was only used in two places, neither of which gained much from it. It was also incorrect because it assumes every response would be a JSON object, when `events` returns an array.

2. Added basic support for parsing multiple `what`

Supplying multiple `what` parameters `what=storage,inventory` is valid API usage. I plan to support this more fully in a follow-up PR, so I intentionally left a more full implemention out.

> what
>     What you are requesting, options include: (you can pass multiple at once, too. status,inventory) 

The alternative fix for the PR would have been simpler, but it would introduce code that would immediately need to be replaced for a following PR. I do not see much gain in writing something that would immediately need to be replaced for a followup PR.

Without some implementation, the changes to `what` would look unnecessary here.

jaadams5 (dismissed):
Thanks for "comment" resolution.

I'll approve but would feel better if someone else pushed - potentially more eyeballs.

jaadams5 (approved):
Let's do this.

PR #3739, opened 2026-09-15
Add modbonus keyword to maximizer

VeeArrKoL:
Adds a `modbonus` keyword to the maximizer. The expression `<weight> modbonus <boolean modifier>` provides a score bonus of *weight* to any item with the specified boolean modifier.

The intent is for this to provide a way to *prefer* (or avoid, if weight is negative) a particular boolean modifier, without placing a hard requirement or restriction. Conceptually, `modbonus <modifier>` has the same relationship to `+<modifier>` and `-<modifier>` that `bonus <item>` has to `+equip <item>` and `-equip <item>`.


midgleyc:
I want to bikeshed the name because it's not clear what "modbonus" should do, but if you think "what's a good name for a bonus given if a boolean modifier exists", I can't think of anything better than "modbonus".

VeeArrKoL:
I did consider just overloading `bonus`, but that felt perilous. Otherwise, yea I couldn't come up with anything better; open to suggestions.

midgleyc:
Go for "boolmodbonus" or "boolbonus" considering it only works for boolean modifiers? Otherwise you might expect to be able to e.g. give a bonus to anything with a rollover effect.

GLM recommends "prefermod" or "ifmod"... don't think I really like either. They're at least clearer than modbonus, though less consistent.

VeeArrKoL:
Personally, `boolmodbonus` and `boolbonus` feel clunky to me. The documentation says that it only works on boolean modifiers, and no one will know about it without seeing it in the documentation anyway. I'll change it if you really want, but meh.

I dislike the GLM recommendations--I chose `...bonus` specifically to mirror the already-existing `bonus` keyword that has similar functionality.

midgleyc:
> no one will know about it without seeing it in the documentation anyway

Unless they see it in a script!

I guess the error message mentions boolean modifiers explicitly.

You say it's similar, but I'm not sure it is:

> Conceptually, modbonus <modifier> has the same relationship to +<modifier> and -<modifier> that bonus <item> has to +equip <item> and -equip <item>.

* "equip item" means "fail unless item is equipped"
* "bonus item" means "add a bonus to item"
* "+modifier" means "add a weighted bonus to items with modifier"
* "modbonus modifier" means "add a weighted bonus to items with modifier, but for boolean modifiers this time"

Actually, could you just remove the keyword entirely, and let people mention either double or boolean modifiers?

VeeArrKoL:
> "+modifier" means "add a weighted bonus to items with modifier"

No, for a boolean modifier, `+modifier` means "require this boolean modifier" and `-modifier` means "prohibit this boolean modifier". (At least according to the maximizer help. The `-modifier` version doesn't seem to work as intended.)

midgleyc:
Urgh, what a pain. So we have an inconsistency with double modifiers there.

VeeArrKoL:
Yea, that's true, but I don't see how to fix that without breaking backward-compatibility.

midgleyc:
Yeah, we could say that if weight > 1 it's a weighted score addition instead of a demand, but that would introduce an inconsistency.

VeeArrKoL:
Yea, I feel like that would produce very surprising results for anyone trying to actually use a weight of 1 as a weight.

midgleyc (approved):
In lieu of a better name, sure.

If we think of one later, it can be aliased.

PR #3740, opened 2026-09-15
Add missing and incorrect modifiers for campground furnishings

libraryaddict:
This adds a bunch of modifiers that are missing from furnishings, and fixes some incorrect (as per wiki) values for restored HP and MP.
Where the restored HP & MP is in a range, the average is used.

https://wiki.kingdomofloathing.com/Resting#Housing

We could be adding modifiers for the effects gained, and stats gained. But, too much effort.

I did quickly check giant pilgrim hat, and it wasn't what kolmafia said it was.


midgleyc (approved):
Thanks!

PR #3741, opened 2026-09-15
Voting booth parsing fix

VeeArrKoL:
The change to the voting booth ballot interface broke the parsing; this fixes it. 

Note: This was developed and tested using the absentee ballot; if there's anything broken in the actual voting booth, someone else will need to check that.


midgleyc (approved):
Thanks!

midgleyc:
Tested with the actual Voting Booth, works fine.

PR #3745, opened 2026-09-16
Add basic support for `graf[]` which is sent by modern KoL chat

libraryaddict:
## What a `graf` is

A `graf` is a single posted message to the chat backend.
`graf[]` is kol's new API where you can now post multiple commands in a single request.
It's used for the `/who` setting, which runs `/who` for every channel when you open the chat.

Currently, kolmafia silently drops that. Nothing looks for `graf[]`, so when the relay browser forwards the request and finds no `graf`, it sends up silently dropping the request.

## Changes

`GenericRequest` gets `getFormFields` which supports multiple values under the same key, rather than just the first. `RelayRequest` uses it to read `graf[]`, falling back to a single `graf`.

`ChatSender.sendMessages` handles `graf[]` via a new ChatRequest constructor. The constructor writes the entries into this.data itself, since `addFormField` doesn't work that way.

The game merges the replies to every graf in the batch into a single output, so `ChatParser. processResponse` returns early when the graf is null.

I feel like the charset is wrong, but, I don't think I can easily determine if kol support utf8 or not.


jaadams5:
What is graf[]?  How/why is if different from the status quo?  Is there anything driving this besides personal preference?

gausie:
This PR is written as if you want it to be closed, which I'm sure is not the case. Lay out what the issue was, what this changes, and how you did it.

libraryaddict:
Alright, trimmed the PR and the PR message down. Removed all the other solutions to just this specific bugfix.

libraryaddict:
Yeah sorry, was tired and the PR was stressful to go through. I'm not a fan of the chat code, and, a good PR would have taken twice the work easily. Eg, standardize how everything is treated and all that.

I've stripped the PR down to just the fix required, and added a helper method to GenericRequest, though, feels like other code could be migrated over to it as well. But, I'm trying to keep it simple.

A clan member had an issue with the /who not firing as it used to, and I just wanted to address that, but the new grafs thing is fired through json, which the current implementation does not use. Thus the larger PR. But, the new grafs thing is fired by kol code only atm, and we don't really need kolmafia to understand how to use it, and my original PR didn't really address that either, as nothing would use that path that we use.

src/net/sourceforge/kolmafia/request/ChatRequest.java
```diff
@@ -40,6 +41,23 @@ public ChatRequest(final long lastSeen, final boolean tabbedChat, final boolean
...
 
+  /** Constructs a new <code>ChatRequest</code> that will send several messages at once */
+  public ChatRequest(final List<String> grafs) {
+    super("submitnewchat.php", true);
+
+    this.addFormField("playerid", String.valueOf(KoLCharacter.getUserId()));
+    this.addFormField("j", "1");
+
+    for (String graf : grafs) {
+      // addFormField would collapse a repeated name, and KoL wants one entry per message.
```
gausie: There is a three arg overload that doesn't collapse (look at AutoSellRequest's whichitem[])

src/net/sourceforge/kolmafia/request/RelayRequest.java
```diff
@@ -3756,12 +3756,16 @@ private void handleChat() {
...
-          ChatSender.sendMessage(
-              new LinkedList<>(), this.getFormField("graf"), true, false, tabbedChat);
+          graf != null
+              ? ChatSender.sendMessage(new LinkedList<>(), graf, true, false, tabbedChat)
+              : ChatSender.sendMessages(grafs);
 
       if (tabbedChat && chatText.startsWith("{")) {
-        ChatPoller.handleNewChat(chatText, this.getFormField("graf"), ChatPoller.localLastSeen);
+        ChatPoller.handleNewChat(chatText, graf, ChatPoller.localLastSeen);
       }
```
gausie, line 3765: It looks like now if we have no graf[] or graf then we would still hit submitnewchat.php, which there's no need to do.

src/net/sourceforge/kolmafia/request/ChatRequest.java
libraryaddict: Missed that, fixed.

src/net/sourceforge/kolmafia/request/RelayRequest.java
libraryaddict: Yes, easy one-line fix, now checks for empty grafs[]. Not 100% correct, but not relevant.

gausie:
Still some issues I think. I am looking at a local editor so will comment here instead of inline as I usually do.

1. sendMessages needs to pipe through executeCommand, so now `/clik blah` would be sent to the chat channel I think.
2. duplicate grafs in one batch will not duplicate (addEncodedFormField doesn't let you add an exactly equal element)
3. this test would fail:

```
  @Test
  void findsBatchedGrafsInEncodedQueryString() {
    var request =
        new GenericRequest(
            "submitnewchat.php?j=1&graf%5B%5D=%2Fwho+clan&graf%5B%5D=%2Fwho+games", false);
    assertThat(request.getFormFields("graf[]"), contains("/who clan", "/who games"));
  }
  ```
  
  I'm not sure if this matters but I saw some encoded strings in this codebase at some point - it probably does based on browser quirks?
  
  4. I don't know if this matters but `boolean tabbedChat = path.contains("j=1");` will not work anymore as the querystring is being moved to data for POSTs, which would make the later if statement dead code.  So this would cause issues when you open chat i the relay browser and send a bunch of messages from there.

gausie (approved):
Thank you for fixing this stuff.

PR #3746, opened 2026-09-16
Add missing wiki names for 2 effects

VeeArrKoL:
Adds wiki names for the Category and Serenity effects, since they were not directing to the correct page.


gausie:
I actually want to get rid of Wiki Name, since the wiki now supports visiting entities by their in game id

midgleyc (approved):
Thanks!

PR #3748, opened 2026-09-16
Fix mushroom plot param name

### LINKED ISSUE #3742: Field command is broken
As I noticed several days ago, Mafia stopped to farm my mushrooms. It reports it does, but field doesn't change.

After some experimentation, I found that "field" command doesn't work anymore. It pretends to work, but as "pick", as "plant" commands are applied to square 1 each time, regardless of supplied number.

VeeArrKoL:
At some point in the recent (or at least, hopefully not-too-distant?) past, the parameter KOL uses to specify which square of the mushroom plot to interact with changed from `pos` to `whichpos`. 

This PR updates `MushroomRequest` to use the correct parameter name, which in turn fixes the `field` gCLI command.

Closes #3742


midgleyc (approved):
Thanks!

PR #3749, opened 2026-09-17
Shrunken head zombie weights

VeeArrKoL:
Adds ASH function for the weights of shrunken head zombie modifiers, thanks to spading by Jeparo.


midgleyc (approved):
Thanks!

midgleyc:
Possibly we want to show these new weights in the interface in the places we show the abilities alone.

PR #3750, opened 2026-09-17
Use the worst case scenario for predicting a monster's elemental attacks

libraryaddict:
### Introduction

In PR https://github.com/kolmafia/kolmafia/pull/799 the ability for monsters to have multiple elemental attacks was added.

It seems like at the time, only one monster was expected to use multiple elemental attacks, so it wasn't too integrated.

Yesterday I noticed that during combat, my script failed to anticipate how much damage it can receive from the hippy camp boss [The Big Wisniewski](https://wiki.kingdomofloathing.com/The_Big_Wisniewski), and took enough damage to lose the fight.

There may be other factors at play, but this PR aims to fix the most obvious one, where the predicted damage is currently based on only one attacking element.

### Changes

1. Changes `expected_damage` to always return the worst case
2. Changes `elemental_resistance` to always return the worst case. Probably less than ideal for monsters with a physical attack, but more honest.
3. Partially removes obsolete internal helper "main attack element"
4. Changes an unrelated test to use a different monster (`Zombo`) as the monster used (`Lord Spookyraven`) effects the test's outcome, due to having multiple attacks, despite it not being the aim of the test.

Before this change, the expected damage from the boss, when you have [2 stench resistance](https://wiki.kingdomofloathing.com/Anti-Odored), is `275`.
After this change, the expected damage from the boss is `344` which reflects its physical attack.

The `elemental_resistance` before is `20`, the after is `0`.

### Closing thoughts

It's possible that the "average" damage may be wanted, but I feel that would be a downgrade as the main reason someone is looking at the expected damage is to try figure out if they're going to survive combat.

Being told that you would survive the hit, but actually, you're going to die half the time because of skewed resistances making the possible range of damage between 10 and 100, doesn't seem that great an idea.

---------

For restoring the old functionality, that can be done by a future PR if there's a need, by providing via parameter the element to test against.

Or alternatively, if requested.


jaadams5:
Since AFAIK this is only visible to scripts how about parameterized versions of expected_damage and elemental_resistance where the parameter selects maximum possible value, minimum possible value or expected (average) value?  

Existing scripts don't have to change and scripts that want this just add the parameter.

midgleyc:
"expected" sort of already assumes the worst, because the monster damage formula includes a part which is "20 to 25% of Monster Attack" and we always take it as 25%, so that is in keeping with this. And in practice it's used by scripts like autoscend that want to check they can survive another turn.

elemental resistance against a monster I'm less convinced by. I'm also not sure when you would even use that.

ckb11:
This is loosely related to https://kolmafia.us/threads/allow-monster-ea-none.29374/
I noticed the same thing - that expected_damage() is often wrong.  I agree that it should always return the worst case.  FWIW, I do my own damage calculations in my script to generate worst case damage and use that - this would fix this problem and I could just use Mafia's internal function.
+1

libraryaddict:
So, either revert the resistance changer, 
or skip it for physical?

midgleyc:
Yes, either or. I don't think the way it works at the moment is good, and I don't think there's a coherent way to call it at present, but if somebody happened to be doing so, it should continue to work similarly. So either of those proposals are fine.

midgleyc (approved):

PR #3751, opened 2026-09-17
Derive Sword of Procedural Generation enchantments from player id

gausie:
Retires jickSwordModifier, which only worked once the sword had been seen in inventory. The enchantments are seeded off the player id, so they can be derived for everyone.


midgleyc (approved):
Thanks!

There's always a concern when removing a preference that someone may have been using it. For this one, I think it's unlikely, and also I searched GitHub and found no non-Mafia matches.

PR #3753, opened 2026-09-17
feat: expose banisher metadata to ASH

ivanlei:
### What

Adds ASH functions to read `BanishManager.Banisher`'s per-banisher data, the banish counterpart of the `monster_tracker_*` functions from #3715. Each takes the banisher name as it appears in `banishedMonsters` / `banishedPhyla`, which is the same string `banished_by()` returns.

```
string[] banishers()
int      banisher_duration(string banisher)      // -1 = lasts until its reset condition
int      banisher_queue_size(string banisher)
string   banisher_reset(string banisher)         // "turn", "turn_rollover", "rollover", "effect",
                                                 // "avatar", "never", "cosmic_bowling_ball"
string   banisher_type(string banisher)          // "monster" or "phylum"
boolean  banisher_is_turn_free(string banisher)  // true if banishing doesn't cost an adventure
```

The data stays in the enum. `BanishManager` gains only a `Reset.shortName()` helper and two string accessors, `getResetTypeName()` and `getBanishTypeName()`, because the `Reset` / `BanishType` enums are private.

`banisher_duration` returns `Banisher.getDuration()`, the value `BanishManager` itself counts down from. For a turn-taking banisher, that's one less than its listed duration (`Patriotic Screech` → 99), matching how `banishMonster` records the banish on the following turn. So `turn banished + banisher_duration() − turns played` gives the same turns-left as the `banishes` command. Durations below zero are normalized to -1.

Unknown banisher names return `0` / `""` / `false`, matching `monster_tracker_*`.

### Why

`is_banished()` and `banished_by()` tell a script *what* is banished and *by what*, but not how long the banish lasts, whether it cost a turn, or when it resets. That data lived only in the enum, so a script showing turns left per active banish, or annotating banish sources as "free, lasts 30 turns", had to hardcode its own copy of the table.

### Testing

- `RuntimeLibraryTest.BanisherData`:
  - `banishers()` includes `snokebomb` and has at least 50 entries.
  - Duration, queue size, reset, type and turn-free checks for `snokebomb`, `Patriotic Screech`, `batter up!`, `beancannon` and `Bowl a Curveball`, including case-insensitive lookup.
  - Unknown names return the defaults.
- Full test suite and `spotlessCheck` pass.


VeeArrKoL:
I am once again asking whether we should instead do things like this as a single function with a structured return type.

ivanlei:
> I am once again asking whether we should instead do things like this as a single function with a structured return type.

I defer to the collective maintainers. I could do it either way.

midgleyc:
#3725 isn't in yet (and as such there are backwards compatibility concerns), and better to be consistent with the existing tracker functions either way.

After records improve, we can add a single function in addition, if still desired.

midgleyc (approved):
not particularly sold on `cosmic_bowling_ball` as a reset type, but whatever I guess. Scripts can deal.

PR #3754, opened 2026-09-18
Share the good effects pool between TCRS, yam batteries and busking

gausie:
Yam batteries don't exclude Floundering after all, so the pool is the same generic one TCRS rolls from. It now lives in EffectDatabase with a cutoff parameter. I also moved beret busking out of RuntimeLibrary into a BeretManager to match others.


midgleyc (approved):
okay!

jaadams5 (approved):

PR #3755, opened 2026-09-18
feat: add a 'var' type to get the compiler to infer the type

midgleyc:
Allow using `var` (like Java's `var` or C++'s `auto`) in ASH, with types where the compiler can infer what the type should be. Like #3725, this is intended to make records returned by KoLMafia functions more convenient to use (records returned by user-defined functions can already have typedefs assigned).

Instead of writing
```kolmafia-ash
int x = 1;
record {string page; int low; int high; int average;} y = ping();
```

you can write
```kolmafia-ash
var x = 1;
var y = ping();
```

Reserve the word "var", preventing it from being used as a variable name.

Don't support it in anonymous aggregate literals: while technically we should be able to infer that `var x = {"strings"}` is an `string[int]`, in practice we use the type information to parse the aggregate.

Support it in some strange circumstances: e.g.
```kolmafia-ash
int x = x + 1
```
currently works and assigns 1 to x, and similarly
```kolmafia-ash
var x = x + 1
```
infers `x` as an int.

Add some more tests for weirder behaviour.


jaadams5 (approved):

gausie (commented):
Remarkably concise solution, well done

midgleyc:
Turns out the reserving broke zlib. I did look around scripts I had downloaded locally, but I didn't have that one.

It was fixed quickly, though. So quickly I didn't even realise it had broken.

PR #3756, opened 2026-09-19
Improve monster stat tracking during switchmonster

VeeArrKoL:
Fixes a few issues with handling of monster switching mid-fight:
* Previously, if Monster Manuel was available, switching to a new monster would result in the tracking of monster stats (e.g. the return value of `monster_attack()` or `monster_hp()`) being inaccurate until the next fight page load (i.e. either a combat round or a twiddle), because the switch was parsed after the current stats and resulted in the Manuel-parsed stats being replaced by an estimate of the monsters stats.
* Previously, the value of `FightRequest.currentEncounter` was not updated when a monster switch occurred. This resulted in problems like stale responses for `heartstone_middle_letter()`.

Also fixes a related problem, where `heartstone_middle_letter()` was not decoding HTML entities in `FightRequest.currentEncounter` before calculating the middle letter.

ref: https://kolmafia.us/threads/heartstone-middle-letter-uses-lastencounter-which-is-not-updated-on-monodent-talk-to-some-fish.32720/


src/net/sourceforge/kolmafia/combat/CombatActionManager.java
```diff
@@ -171,8 +171,11 @@ public static final String encounterKey(String line, final boolean changeCase) {
     } else if (key.startsWith("the ")) {
       // It really is "The Man" or "The Big Wisniewski"
     } else if (key.startsWith("some ")) {
-      key = key.substring(5);
-      line = line.substring(5);
+      // some fish is literally named "some fish"; otherwise, remove the "some"
+      if (!key.equals("some fish")) {
+        key = key.substring(5);
+        line = line.substring(5);
+      }
```
gausie, line 178: We actually know the article of monsters now, and have it in their monster data. Can we make this more reliable?

VeeArrKoL: It could be tricky. We don't seem to actually have a `MonsterData` at most of the call sites, and it seems like this normalization might actually be used to *lookup* the appropriate `MonsterData` in some cases, so that would be circular.

If we did want to do this, it's probably a large enough change with enough potential for regression to warrant its own, separate PR.

jaadams5: Perhaps not relevant but KoL and mafia do not agree on the article for The Maety Bonerdagon.

See https://kolmafia.us/threads/29230-effects-monster-manual-name-checking.32824/

VeeArrKoL:
Marking as draft for the moment; I found a stupid OCRS case that I won't be able to address until later tonight.

VeeArrKoL:
Updates:
* Added `monster_name()` library function to provide access to `FightRequest.currentEncounter`, updating over the course of a fight in a similar vein to `monster_attack()` and `monster_hp()`.
* Re-parse the name the round after a switchmonster, since OCRS monster modifiers don't get added to the name until the next round/page load.
* Update `FightRequest.currentEncounter` after using Steal Monster's Heart.

src/net/sourceforge/kolmafia/combat/CombatActionManager.java
midgleyc: that's probably because we strip whitespace, but KoL ends with whitespace.

PR #3758, opened 2026-09-20
ASH method for predicting Portable Laughing Stock drops

VeeArrKoL:
Adds two ASH methods for predicting drops from the [Portable Laughing Stock](https://wiki.kingdomofloathing.com/Portable_Laughing_Stock):
* `item [int] portable_laughing_stock_drops( class class, path path, int day, int minFights, int maxFights )`: Lists drops for the given class/path/day, starting with the `minFights`<sup>th</sup> fight and ending with the `maxFights`<sup>th</sup> fight.
* `item [int] portable_laughing_stock_drops( int maxFights )`: Lists drops for the player's current class/path/day, from the current value of `_laughingStockCharges` up to the `maxFights`<sup>th</sup> fight.


src/net/sourceforge/kolmafia/session/PortableLaughingStockManager.java
gausie, line 1: I admit the difference is not super clear but this is a manager not a database. See voting booth, beret, etc

VeeArrKoL: :notlikeduck: I modeled it off of `ShrunkenHeadDatabase`, but I can switch it.

gausie: That should also be a manager to be honest. Difference is that databases load and present data from a file as their primary function. TCRSDatabase should be renamed too now that it no longer does that.

gausie (approved):

PR #3760, opened 2026-09-20
Add missing Drops Items tags

VeeArrKoL:
Adds the `Drops Items` modifier to a few items that seem like they should have it.


midgleyc (approved):
cool

PR #3761, opened 2026-09-22
ASH methods for predicting B&W Apron kit contents

VeeArrKoL:
This PR does two things related to [Black and White Apron Meal Kits](https://wiki.kingdomofloathing.com/Black_and_White_Apron_Meal_Kit):
* Removes some one-off special handling for the `bwApronMealsEaten` pref that was intended to be temporary to facilitate spading.
* Adds `RuntimeLibrary` methods to predict the contents of a kit, either based on current character state or for a specified path/class/# of kits eaten:
  * `record black_and_white_apron_kit_contents()`
  * `record black_and_white_apron_kit_contents(path path, class cls, int prevEaten)`

## Return Type Complexity

I have some concerns about the complexity of the return type, which I'm open to having some discussion on. The PR as written does not seem to have any *technical* problem--it all functions as expected--but maybe there are "better" design choices to be made. Please read below and provide any feedback on whether you think any of the alternative choices would be better (or if you think there's another choice I've overlooked that is better still than any of these).

As implemented, the return type has the structure:

```
record {
    string main_ingredient; 
    effect main_effect; 
    aggregate {
        item ingredient; 
        effect effect; 
        int effect_turns; 
        int mus; 
        int mys; 
        int mox; 
        int meat; 
        int familiar_xp; 
        string other;
    }[3, 5] meals;
}
```

The complexity comes from two design choices: (1) the structure representing each ingredient; and (2) the nested records.

### Ingredient Structure

Much of the complexity here comes from wanting to represent the bonus for each ingredient in a structured way. The inner record is essentially a union type, where most of the fields will be unfilled/default for any given ingredient. There are several alternate choices that could reduce this complexity, including:
* Reducing the amount of fields in the structure. For example, only provide structured data for the `ingredient`, `effect` and `effect_turns`, leaving everything else as unstructured (but consistently-formatted) text in `other`.
* Reducing the structure *even further*, so that it is only the `item` representing the ingredient and a `string` representing the bonus.
* De-structure it entirely, representing both the ingredient and bonus in a single `string`.

In general, progressing down the list trades off structural complexity for increased difficulty parsing out and using the information. The mostly-fully structured approach is what I landed on, but maybe callers are unlikely to be interested in whether an ingredient grants 75 muscle substats or whatever, and only the effect information is really relevant?

### Nested record return type

The structure of the return type being a nested record is largely due to that being a natural representation of the meal kit. Conceptually, it isn't too complicated, but actually specifying the data type in ASH becomes a bit cumbersome. This is offset somewhat by the recent updates allowing `var` and imperfect-assignment for records in ASH. There are some alternatives I considered (before ultimately landing on this design):
* Attempt to represent the main ingredient using the same structure as the other ingredients. While this would theoretically allow flattening the structure into a single, multi-keyed map, the main ingredient really isn't very similar to the other ingredients, and the most straightforward way to implement it would be to add it as a "fake" 6th ingredient into each of the three meal options, which doesn't feel great.
* Split the API into two separate methods: one to get the main ingredient effect and one to get the ingredients for one of the three meals. I'm generally against artificially splitting apart related information into separate calls, but the division would be at least somewhat natural to how I would expect a caller to iterate over the data anyway. It would be something like:
  * `effect black_and_white_apron_kit_main_effect()`
  * `record[5] black_and_white_apron_kit_ingredients(int mealIndex)` (where the record type is the "inner" record from the current implementation)

Thoughts?


midgleyc:
Thoughts: with `var` complex datatypes aren't so much an issue, so you could do
```kolmafia-ash
var bw_contents = black_and_white_apron_kit_contents();
for (int i = 0; i < 3; i++) {
  var selections = bw_contents.meals[i];
  foreach selection in selections {
    [...]
  }
}
```
and be happy with that.

Should add a test for the ASH part though.

VeeArrKoL:
> with `var` complex datatypes aren't so much an issue

Yea, this is where I landed too (even if the return type still looks pretty gnarly in `ashref`), but I wasn't sure whether I was off the mark.

> Should add a test for the ASH part though.

It's a little ugly, but I added one. Let me know if you were envisioning something different.

midgleyc:
Rather than testing the full output, I was expecting just a short script (e.g. like in my comment) with some print statements that you'd assert on (to also serve as a small bit of tests-as-documentation on how you'd expect to use the function).

Having the whole thing is okay, I guess.

VeeArrKoL:
> Rather than testing the full output, I was expecting just a short script (e.g. like in my comment) with some print statements that you'd assert on (to also serve as a small bit of tests-as-documentation on how you'd expect to use the function).

Oh, yea, that makes sense. I can update it tomorrow.

src/net/sourceforge/kolmafia/session/BlackAndWhiteApronManager.java
```diff
@@ -0,0 +1,177 @@
...
+          new MainIngredient("beef", "Beefy Heart"),
+          new MainIngredient("chicken", "Winner, Winner, Chicken!"),
+          new MainIngredient("potatoes", "Spudly"),
+          new MainIngredient("lettuce", "Lettuce Go!"),
+          new MainIngredient("pork", "Porked Up"),
+          new MainIngredient("flounder", "Fishing for Meat"));
+
+  private static final List<Ingredient> INGREDIENTS =
+      List.of(
+          otherIngredient("wad of dough", "Restore 100 HP"),
```
gausie: Why are these not ItemPool.WAD_OF_DOUGH etc?

VeeArrKoL: Primarily because I already had the data in tabular form with item/effect names, and looking up and adding ~50 items and ~30 effects to their respective pools was not appealing. 

I can do it if we're concerned about the name of wad of dough changing or whatever, but meh.

jaadams5 (approved):
But someone else should push.

jaadams5:
@VeeArrKoL - This would have been easier for me to review and approve if you had separated the B&W apron and the Laughing Stock into two different PRs.  I understand why you didn't since you were already "in the file" but I would suggest that reducing the scope and the complexity of the PR is also benefical.

On using ItemPool and EffectsPool etc. there was some discussion about this on KoLmafia.  Is using them a requirement or a suggestion?  Should an entry be added for every new item even if it is not used immediately?  Should some kind of automated process detect changes to the underlying data files and generate the Java code?  There was no strong agreement about changing the status quo.  Code is more readable if it uses a named constant instead of a magic number and more robust when using a named constant instead of a string that might (now or in the future) be ambiguous.  New code is encouraged to use them and retrofitting them is not discouraged but there is no hard and fast rule.

midgleyc:
Personally, I'm happy for code to use item names if unambiguous and covered by tests (which would error if the names changed or didn't match an item).

gausie:
I do lean towards exhaustive ItemPool, but I agree with Ryo - a test that they are valid items probably covers most of the problem

> if we're concerned about the name of wad of dough changing or whatever

Yeah in general with mafia when we are producing features we are unlikely to touch again the most important thing is that they work even when weird shit changes, in my opinion.

